PR #2, opened 2016-05-11
Bug 1333172 - differ between route hostname and navigation within the console

jhadvig:
Making difference between route hostname and navigation within the console more obvious, by prepending the `pficon-route` icon before the exposed route link.

Screen prepending the icon(which I like more then the appending one):
screenshot-11

Screen appending the icon:
screenshot-12

Also I've created a standalone class for the icon `font-size` since by default it will take the size of `h2` style which is 19px and that seems to be too big for the icon.

@liggitt suggested we use the `fa-external-link` icon instead the `pficon-route` as original was suggested:
screenshot-17

@jwforres PTAL

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -314,8 +314,11 @@ angular.module('openshiftConsole')
         return url;
     };
   })
-  .filter('routeLabel', function(routeHostFilter) {
+  .filter('routeLabel', function(routeHostFilter, routeWebURLFilter, isWebRouteFilter) {
     return function(route, host) {
+      if (isWebRouteFilter(route)) {
+        return routeWebURLFilter(route, host);
+      }
       var label = (host || routeHostFilter(route));
       if (route.spec.path) {
         label += route.spec.path;
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -7501,10 +7501,11 @@ return function(b, c) {
 var d = b.spec.tls && "" !== b.spec.tls.tlsTerminationType ? "https" :"http", e = d + "://" + (c || a(b));
 return b.spec.path && (e += b.spec.path), e;
 };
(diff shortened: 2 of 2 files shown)


jhadvig:
Will add bindata after review :)

jwforres:
@rhamilto this is something you could help with, is there really no way for us to get an icon on the right of the route without it disappearing in some cases :(

rhamilto:
> @rhamilto this is something you could help with, is there really no way for us to get an icon on the right of the route without it disappearing in some cases :(

@jwforres given the complexity of the layout, I was unable to find a reliable, consistent [solution for the warning icon](https://github.com/openshift/origin/issues/7844#issuecomment-215106617) since there are so many different layout permutations at the various screen resolutions...  :(

spitballing here... what if we included the protocol (e.g., http://) instead of an icon in this context?

jhadvig:
@rhamilto I like the idea of including the protocol. Would either go with that or prefixing with `external-link` icon .

jwforres:
so we have non http(s) route ports coming soon and we wouldn't know what
protocol to show there, we'll only know generically that it is a TCP port.

rhamilto:
> so we have non http(s) route ports coming soon and we wouldn't know what
protocol to show there, we'll only know generically that it is a TCP port.

Will a hyperlink even make sense in those cases?

jwforres:
true it wont be a link anymore

jhadvig:
@jwforres @rhamilto Ive updated the PR based on our chat. PTAL

screenshot-18

rhamilto:
Works for me.

jwforres:
looks good, run grunt build and check in the dist/* changes

and go ahead and backport the change to ose/master (you'll still need bindata.go there)

jhadvig:
@jwforres updated

jwforres:
[merge]

PR #5, opened 2016-05-12
Bump k8s-container-terminal to 0.0.13

spadgett:
Fixes a problem with unicode characters.

@jwforres PTAL


jwforres:
looks like this also picks up the support for the autofocus param on the directive, but we still need to fix the multiple containers thing before we enable that

[merge]

PR #11, opened 2016-05-13
Add hack/run-all.sh

benjaminapetersen:
@spadgett @jwforres @sg00dwin @rhamilto  This is the "I'm getting a cup of coffee, plz run everything, thx" script.  I should probably run a `grunt build` in here as well.
Thoughts?

Perhaps 2 scripts, a 'clean install build' and a 'all that + test'.

### DIFF
--- a/hack/run-all.sh
+++ b/hack/run-all.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# Run all the important scripts in a sensible order,
+# handy for switching branches, etc
+wd="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+$wd/clean-deps.sh &&
+$wd/install-deps.sh &&
+echo '[build-assets]' &&
+grunt build &&
+echo '[test-assets]' &&
+grunt test &&
+grunt test-integration
+# $wd/test-integration-headless.sh &&
+# $wd/verify-dist.sh
--- a/hack/run-all.sh
+++ b/hack/run-all.sh
@@ -1,14 +1,22 @@
 #!/bin/bash
 
-# Run all the important scripts in a sensible order,
-# handy for switching branches, etc
-wd="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-$wd/clean-deps.sh &&
-$wd/install-deps.sh &&
-echo '[build-assets]' &&
-grunt build &&
-echo '[test-assets]' &&
-grunt test &&
-grunt test-integration
+# This script is for local development usage & will run the
+# entire clean/build/install/test flow in order.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+OS_ROOT=$(dirname "${BASH_SOURCE}")/..
+${OS_ROOT}/hack/clean-deps.sh
+${OS_ROOT}/hack/install-deps.sh
+echo '[build-assets]'
+grunt build
+echo '[test-assets]'
+# SKIP_TESTS flag
+#if !-o SKIP_TESTS; then
+  grunt test
+  grunt test-integration
+#fi
 # $wd/test-integration-headless.sh &&
 # $wd/verify-dist.sh
--- a/hack/clean-deps.sh
+++ b/hack/clean-deps.sh
@@ -2,6 +2,7 @@
 
 set -e
 
+echo "[clean-assets]"
 echo "Cleaning up bower_components and node_modules..."
 rm -rf bower_components/*
 rm -rf node_modules/*


hack/run-all.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Run all our scripts in order.
+wd="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+$wd/clean-deps.sh &&
+$wd/install-deps.sh &&
+$wd/test-integration-headless.sh &&
+$wd/verify-dist.sh
```
spadgett: verify-dist will fail if the clean/install did anything ... so I'd add grunt build. But then I'm not sure you need verify-dist at all. I'd want grunt test as well.

benjaminapetersen: Added grunt build & grunt test & grunt test-integration.

benjaminapetersen:
updated

hack/run-all.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# Run all the important scripts in a sensible order,
+# handy for switching branches, etc
+wd="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
spadgett: Are the nested double quotes really correct?

spadgett: How about

```
OS_ROOT=$(dirname "${BASH_SOURCE}")/..
"${OS_ROOT}/hack/clean-deps.sh"
....
```

benjaminapetersen:
@jwforres curious, I think the "install" part of this is substantially longer because of the `bower cache clear`.  I'm wondering if thats essential every time.  Since we ask for an explicit version in our `bower.json`, seems like we don't need the cache clearing.  Thoughts?  Or is it essential for merge queue?

hack/run-all.sh
benjaminapetersen: So the `$( )` is a sub-shell and you can put additional unescaped quotes in it.
My bash lesson for the day. 😄 
Neat it works, not sure if its a good practice.  Will def update to look like our other scripts.

spadgett: Ah, ok so I don't know much about bash :)

I'm fine with it as long as it's correct.

benjaminapetersen: Heh, well, I used it without being able to explain it (bad).  Had to look it up.

jwforres:
cc @stevekuznetsov since he is our resident bash expert now ;-)   we apparently have standards to follow

jwforres:
@benjaminapetersen on your bower cache question, if we are using the cache and a dependency repo misbehaves and actually tags a fix into an existing version, then we are pulling from the cache and not guaranteed that we are pulling latest

i think its only happened once that someone did that.  So now that we are in our own merge queue, and would only be affecting our team if things go awry, i'm ok with commenting that out for awhile and seeing how things go.

hack/run-all.sh
```diff
@@ -0,0 +1,14 @@
...
+$wd/clean-deps.sh &&
+$wd/install-deps.sh &&
+echo '[build-assets]' &&
+grunt build &&
+echo '[test-assets]' &&
+grunt test &&
+grunt test-integration
+# $wd/test-integration-headless.sh &&
```
jwforres: which one of these you use depends on where you want to use this script, if you want it to be able to be run on an ec2 instance / as part of jenkins job then you need to use headless.  If this is just for devs running locally then the grunt version is sufficient.  Comment at the top of the file if thats what you decide so its clear it shouldn't be run in a non-desktop environment.

hack/run-all.sh
```diff
@@ -0,0 +1,14 @@
...
+# $wd/verify-dist.sh
```
jwforres: i would leave this out

jwforres:
if you want to optionally disable the tests, dont duplicate the script, check an env var.  I would lean towards really running all by default and then accepting a SKIP_TESTS env var to not run the tests

hack/run-all.sh
benjaminapetersen: Yeah, thinking this is definitely for local dev usage.

hack/run-all.sh
stevekuznetsov: Ben's approach gives you the absolute path to the directory, Sam's (what we use in Origin) is a relative path. Shouldn't make any difference, really. If you're keeping this approach, keep your quotes the way they are in case `$ BASH_SOURCE` contains spaces.

stevekuznetsov: Should call this var `$OS_ROOT` for consistency

hack/run-all.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# Run all the important scripts in a sensible order,
+# handy for switching branches, etc
+wd="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+$wd/clean-deps.sh &&
```
stevekuznetsov: What are the `&&` accomplishing here? I can't remember off hand if they cause the newline to be ignored on parsing, but of what you want is conditional continue to next statements, use `set -o errexit` (which should be on anyway).

hack/run-all.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
stevekuznetsov: In all of our scripts it's customary to use this in the preamble:

```sh
set -o errexit
set -o nounset
set -o pipefail
```

The first option exits the script on an error, the second errors when a variable is used before it's assigned, the third fails the entire pipe chain if any member fails.

stevekuznetsov:
This seems like a good place for a `Makefile` that references these scripts, as in Origin. That allows for a little more intuitive access to the actions.

If your scripts begin to get complicated, I can bring in the error handler and stack trace from Origin so debugging failures is easier.

hack/run-all.sh
stevekuznetsov: The purpose of these is to set the shell to "sane mode" and make it's execution easier to understand / work with, as it's rare that the behavior from these options isn't what you wanted in the first place.

hack/run-all.sh
benjaminapetersen: Updated to `$OS_ROOT` & the relative path version.

hack/run-all.sh
benjaminapetersen: Meaning leave out `verify-dist.sh`, correct?

hack/run-all.sh
benjaminapetersen: Right,continue only if the prev command returns non-zero.

hack/run-all.sh
benjaminapetersen: Nice, that makes sense.  Will look at a few more scripts as see if I can better line up with any other conventions.

hack/run-all.sh
jwforres: correct, its always going to fail if you haven't committed your dist yet, so if you are running through the clean/install/build/test you may or may not have actually commited dist already.  Plus with us having enabled travis now you'll know pretty quick if you forgot that step

benjaminapetersen:
@stevekuznetsov @jwforres I have no objection to a `Makefile`.  It might be a bit abnormal for the current UI world that seems to stick to raw `grunt`/`gulp`/etc. but if we want a `make clean build` I can set that up here as well.

stevekuznetsov:
I'm not sure how much parity you want with Origin, so I don't know if the `Makefile` makes sense or not. The one benefit is that it would promote small targets whereas "do everything" scripts Hoover up complexity until they've got more special flags than you can count. If the risk of that happening is low due to the UI workflow, then I'd say keep the script, forget `make`.

spadgett:
Probably not needed with #614 since you should be able to run `make clean test`

jwforres:
agreed, @benjaminapetersen ok to close?

benjaminapetersen:
Works for me

PR #14, opened 2016-05-16
Check if hostname is present when appending path to it

jhadvig:
@spadgett PTAL

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -320,10 +320,13 @@ angular.module('openshiftConsole')
         return routeWebURLFilter(route, host);
       }
       var label = (host || routeHostFilter(route));
+      if (!label) {
+        return '<unknown host>';
+      }
       if (route.spec.path) {
         label += route.spec.path;
       }
-      return label || '<unknown host>';
+      return label;
     };
   })
   .filter('parameterPlaceholder', function() {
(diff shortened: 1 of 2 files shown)


app/scripts/filters/resources.js
```diff
@@ -320,7 +320,7 @@ angular.module('openshiftConsole')
         return routeWebURLFilter(route, host);
       }
       var label = (host || routeHostFilter(route));
```
spadgett: If we want the path to show up, we could just use

```js
var label = host || routeHostFilter(route) || '<unknown host>';
```

spadgett: If we don't want the path to show up, I'd just return early here:

```js
var label = (host || routeHostFilter(route));
if (!label) {
  return '<unknown host>';
}
```

but I have a slight preference for showing the path.

jhadvig: true that, now we just figure out if we want it.

jhadvig: I think I won't rather display the path, since the hostname is missing anyway.. 
It could be confusing to show `<unknown host>/path/to/something`

@spadgett @jwforres thoughts ?

jwforres: honestly i don't have a strong preference either way, i'm leaning towards dropping it just so its very obvious your route is not currently in a usable state

jhadvig:

spadgett:
[merge]

PR #15, opened 2016-05-16
Fix bug 1285658: adding labels and envs is tricky

benjaminapetersen:
Fix bug 1333118, ensure key,value committed before allowing parent forms to submit.

Simplest fix essentially:
- adds a separate hidden form with its own `<input>` used only as a proxy validation
- if either key or value inputs have any value at all, the hidden form will be set to 'invalid', which will bubble up and invalidate parent forms, ensuring user has to make a decision to submit the new key-value pair before submitting the larger form.
- ensures existing validation on the key-value form continues to work as before

```html
<!-- separate hidden form, allows key-value inputs to use normal validation but bubble up an 'invalid' to parent forms if any values are detected -->
<ng-form hidden name="clean">
  <input name="isClean" ng-model="keyValuesClean"></input>
</ng-form>
```

Screenshots of the various states:
screen shot 2016-05-16 at 12 56 26 pm
screen shot 2016-05-16 at 12 56 31 pm
screen shot 2016-05-16 at 12 56 36 pm
screen shot 2016-05-16 at 12 56 45 pm
screen shot 2016-05-16 at 12 56 40 pm

@jwforres @spadgett for review!

### DIFF
--- a/app/scripts/directives/oscKeyValues.js
+++ b/app/scripts/directives/oscKeyValues.js
@@ -33,6 +33,22 @@ angular.module("openshiftConsole")
       }
       return true;
     };
+
+    // checks if the key,value inputs have any text value.
+    // if so, sets the form name="clean" to an 'invalid' state, which will
+    // invalidate any parent form up the chain.  This should result in an
+    // inability to submit that form until the user commits the new key-value pair.
+    $scope.isClean = _.debounce(function() {
+      $scope.$apply(function() {
+        if(!!$scope.key) {
+          $scope.clean.isClean.$setValidity('isClean', false);
+        } else if(!!$scope.value) {
+          $scope.clean.isClean.$setValidity('isClean', false);
+        } else {
+          $scope.clean.isClean.$setValidity('isClean', true);
+        }
+      });
+    }, 50);
     $scope.addEntry = function() {
       if($scope.key && $scope.value){
         var readonly = $scope.readonlyKeys.split(",");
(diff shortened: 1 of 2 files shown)


jwforres:
not sure how i feel about this :-\  if you've done it in the env vars on from image those aren't as close to the Create button, in fact Create might not visibly be on the same page as those inputs, so you wont see the indication of that changing when you typed into those fields, it wont be obvious to a user in that case why they arent being allowed to submit the form.

jwforres:
Here is an idea, can we wrap those two inputs and the Add button with a `<fieldset>` because fieldset elements support onblur, so we could do something like if you leave the fieldset with data filled out then we can show a warning/error beneath it?

benjaminapetersen:
@jwforres lemme give that a swing and see, I've been thinking about longer pages as well.

benjaminapetersen:
😿  `<fieldset>` and `onblur` / `ng-blur` not gonna work. Booo.  Looking at some other options with validations.

@spadgett made a comment about using that wizard module to break the page down.  Contemplating that approach as well.

jwforres:
Not for the 3.2 fix, don't want to bring in new dependencies

benjaminapetersen:
@jwforres per that last suggestion, its not too hard to do something like this:

screen shot 2016-05-17 at 10 30 14 am

Feels a bit lame, but for a short term fix till we replace the control, perhaps is all we need?

benjaminapetersen:
New idea coming....

PR #18, opened 2016-05-17
Fix bug 1333118, ensure key,value committed before allowing parent forms to submit.

benjaminapetersen:
Nixed the prev solution, presenting new:

- link function passes DOM elements to controller once rendered
- blur events used to check against `document.activeElement` to see if an `osc-key-values` child is active
- show message when user leaves `osc-key-values` with uncommitted changes
- eliminates secondary controller & assimilates it into the main directive controller to simplify access to `$scope` vars
- when uncommitted key-value pairs exist, osc-key-values remains valid, HOWEVER it will invalidate the parent form, ensuring it is not submitted until the user decides to keep or discard all key-value pairs.

@jwforres @spadgett thoughts?

### DIFF
--- a/app/scripts/directives/oscKeyValues.js
+++ b/app/scripts/directives/oscKeyValues.js
@@ -19,51 +19,6 @@ angular.module("openshiftConsole")
       }
     };
   })
-  .controller("KeyValuesController", function($scope){
-    var added = {};
-    $scope.allowDelete = function(value){
-      if ($scope.preventEmpty && (Object.keys($scope.entries).length === 1)) {
-        return false;
-      }
-      if($scope.deletePolicy === "never") {
-        return false;
-      }
-      if($scope.deletePolicy === "added"){
-        return added[value] !== undefined;
-      }
-      return true;
-    };
-    $scope.addEntry = function() {
-      if($scope.key && $scope.value){
-        var readonly = $scope.readonlyKeys.split(",");
-        if(readonly.indexOf($scope.key) !== -1){
-          return;
-        }
-        added[$scope.key] = "";
-        $scope.entries[$scope.key] = $scope.value;
-        $scope.key = null;
-        $scope.value = null;
-        $scope.form.$setPristine();
-        $scope.form.$setUntouched();
-      }
-    };
-    $scope.deleteEntry = function(key) {
-      if ($scope.entries[key]) {
-        delete $scope.entries[key];
-        delete added[key];
-        $scope.form.$setDirty();
-      }
-    };
-    $scope.setErrorText = function(keyTitle) {
-      if (keyTitle === 'path') {
-        return "absolute path";
-      } else if (keyTitle === 'label') {
-        return "label";
-      } else {
-        return "key";
-      }
-    };
-  })
   .directive("oscInputValidator", function(){
 
     var validators = {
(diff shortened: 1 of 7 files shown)


benjaminapetersen:
For references, replaces closed PR #15 (which is the one with the parent form invalidation trigger).

jwforres:
yeah i like where this is going, two things

1) update the error message, "Add or clear this key-value pair"  and make "clear" a link that clears the inputs
2) since we have an error message now, it would be ok to disable the Create button when you have something in this state, the stuff you started do before.  How big of a change was that?

jwforres:
actually we say "Name" in the input, so name-value instead of key-value would be consistent

benjaminapetersen:
Will do!  I can bring that back, it wasn't much of a change.

benjaminapetersen:
@jwforres got the parent-form invalidation back in.
Was also able to re-focus the initial input after each key-value add.

benjaminapetersen:
(now to fix my tests)

benjaminapetersen:
@jwforres fixed my bug.  Now for real fixing my tests since I eliminated the `KeyValuesController `.  Verified it works with multiple instances on the page.

benjaminapetersen:
@jwforres fixed the tests, should be good for another pass.

app/scripts/directives/oscKeyValues.js
```diff
@@ -162,10 +117,112 @@ angular.module("openshiftConsole")
...
+            } else if(!!$scope.value) {
+              $scope.clean.isClean.$setValidity('isClean', false);
+            } else {
+              $scope.clean.isClean.$setValidity('isClean', true);
+            }
+          });
+        }, 100);
+
+        var onBlur = function(nodes) {
+          return _.debounce(function(evt) {
```
jwforres: do we really need debounce for an onBlur ?

app/scripts/directives/oscKeyValues.js
```diff
@@ -162,10 +117,112 @@ angular.module("openshiftConsole")
...
+              $scope.clean.isClean.$setValidity('isClean', false);
+            } else if(!!$scope.value) {
+              $scope.clean.isClean.$setValidity('isClean', false);
+            } else {
+              $scope.clean.isClean.$setValidity('isClean', true);
+            }
+          });
+        }, 100);
+
+        var onBlur = function(nodes) {
```
jwforres: add a comment explaining what this does / why its there

app/views/directives/osc-key-values.html
```diff
@@ -53,6 +59,12 @@
           </a>
         </div>
 
+        <div ng-if="showCommmitWarning" class="has-error">
+          <span class="help-block">
+            Please add or clear this {{(keyTitle || 'key') | lowercase}}-{{(valueTitle || 'value') | lowercase}} pair
```
jwforres: clear should be a link that clears the inputs

app/scripts/directives/oscKeyValues.js
benjaminapetersen: Not really, I will eliminate it and update the `$scope.$apply` to a `$scope.$applyAsync` to avoid potential digest errors.

app/scripts/directives/oscKeyValues.js
benjaminapetersen: will do

app/views/directives/osc-key-values.html
benjaminapetersen: We good with the blue link in the red text @rhamilto @sg00dwin ?

screen shot 2016-05-26 at 12 44 00 pm

benjaminapetersen:
@jwforres so long as we are good with the link color text, this should be ready for final pass.

rhamilto:
The link color text is inherited from Bootstrap/PatternFly.  

Bootstrap:
screen shot 2016-05-26 at 1 07 06 pm

PatternFly:
screen shot 2016-05-26 at 1 07 33 pm

@jwforres and I discussed this, and think the correct course of action is to merge this PR as is and open an issue on PatternFly.  

One possible fix for PatternFly:

screen shot 2016-05-26 at 1 09 12 pm

cc: @andresgalante

andresgalante:
@rhamilto I don't think we should use underline for the link on an form messages, since we don't use underline to represent links anywhere else. I rather go blue links than red underline.
Or maybe use a darker shade or red.
Thanks guys for spotting it. I'll make sure to discuss it in our side with the designers.

benjaminapetersen:
@jwforres @rhamilto works for me, I think it makes sense that this would be something patternfly would handle & we can ignore as a special case.

In that case, still ready for final pass!

jwforres:
LGTM just let me know once you've tested a couple other browsers

benjaminapetersen:
jwforres

benjaminapetersen:
@jwforres IE 10,11,12 all good, firefox good, safari good, I think its good to go.

jwforres:
[merge]  please put together the backport PR for ose/master

benjaminapetersen:
@jwforres will do

PR #20, opened 2016-05-18
Show pipelines on browse build config and build pages

spadgett:
@jwforres @fabianofranz 

pipeline-builds

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -31,7 +31,8 @@ angular.module('openshiftConsole')
       "description":              ["openshift.io/description"],
       "buildNumber":              ["openshift.io/build.number"],
       "buildPod":                 ["openshift.io/build.pod-name"],
-        "jenkinsLogURL":            ["openshift.io/jenkins-log-url"]
+      "jenkinsLogURL":            ["openshift.io/jenkins-log-url"],
+      "jenkinsStatus":            ["openshift.io/jenkins-status-json"]
     };
     return function(annotationKey) {
       return annotationMap[annotationKey] || null;
@@ -39,6 +40,7 @@ angular.module('openshiftConsole')
   })
   .filter('labelName', function() {
     var labelMap = {
+      'buildConfig' : ["openshift.io/build-config.name"],
       'deploymentConfig' : ["openshift.io/deployment-config.name"]
     };
     return function(labelKey) {
@@ -165,6 +167,12 @@ angular.module('openshiftConsole')
         return resource.metadata.labels[key];
       }
       return null;
+    };
+  })
+  .filter('buildConfigForBuild', function(labelNameFilter, labelFilter) {
+    var labelName = labelNameFilter('buildConfig');
+    return function(build) {
+      return labelFilter(build, labelName);
     };
   })
   .filter('icon', function(annotationFilter) {
@@ -833,8 +841,14 @@ angular.module('openshiftConsole')
     };
   })
   .filter('jenkinsLogURL', function(annotationFilter) {
-    return function(build) {
-      return annotationFilter(build, 'jenkinsLogURL');
+    return function(build, asPlainText) {
+      var logURL = annotationFilter(build, 'jenkinsLogURL');
+      if (!logURL || asPlainText) {
+        return logURL;
+      }
+
+      // Link to the Jenkins console that follows the log instead of the raw log text.
+      return logURL.replace(/\/consoleText$/, '/console');
     };
   })
   .filter('buildLogURL', function(isJenkinsPipelineStrategyFilter,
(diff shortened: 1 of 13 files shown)


fabianofranz:
LGTM

The "View Log" link is still broken but that's an issue in the sync plugin (not adding "https" to the URL in the annotation), correct?

spadgett:
Yeah, see https://github.com/fabric8io/openshift-jenkins-sync-plugin/issues/49

@bparees is adding insecure policy redirect to the route in the template.

app/scripts/filters/resources.js
```diff
@@ -833,8 +841,14 @@ angular.module('openshiftConsole')
...
-    return function(build) {
-      return annotationFilter(build, 'jenkinsLogURL');
+    return function(build, asPlainText) {
+      var logURL = annotationFilter(build, 'jenkinsLogURL');
+      if (!logURL || asPlainText) {
+        return logURL;
+      }
+
+      // Link to the Jenkins console that follows the log instead of the raw log text.
+      return logURL.replace(/\/consoleText$/, '/console');
```
jwforres: is this something we can get fixed in the sync plugin? or is it intentional that its sending the raw log text

jwforres: I guess later when we want to have the log appear natively in the console we would need the /consoleText URL anyway

spadgett: Yeah I assume it was intentional. I held off opening a sync plugin issue until I knew what we wanted.

app/styles/_pipeline.less
```diff
@@ -0,0 +1,167 @@
...
+    }
+    min-height: 100px;
+    padding: 10px;
+    .status-icon {
+      margin-bottom: 3px;
+      .inverse {
+        font-size: 8px;
+        margin-bottom: 3px;
+      }
+      &.Complete .inverse {
```
jwforres: upper case class names feel wrong, is this because you are using a class that matches the status we get back?  could we do what we use to do with pods and do .statusComplete .statusFailed,  or just lowercase the status?

app/styles/_pipeline.less
```diff
@@ -0,0 +1,167 @@
...
+    }
+    &:before {
+      right: 0.5em;
+      margin-left: -50%;
+    }
+    &:after {
+      left: 0.5em;
+      margin-right: -50%;
+    }
+    &.ABORTED {
```
jwforres: same with these, can we lower case

app/views/directives/pipeline-status.html
```diff
@@ -0,0 +1,34 @@
...
+      <span class="sr-only">In Progress</span>
+    </span>
+    <span ng-switch-when="NOT_EXECUTED">
+      <!-- No icon -->
+    </span>
+    <span ng-switch-when="PAUSED_PENDING_INPUT">
+      <i class="fa fa-pause" aria-hidden="true"></i>
+      <span class="sr-only">Paused Pending Input</span>
+    </span>
+    <span ng-switch-when="SUCCESS">
```
jwforres: should we have an ng-switch-default of a gray circle with a question mark or something?  i worry about things getting updated and new statuses appearing

app/styles/_pipeline.less
spadgett: I personally like it matching exactly. Otherwise you have to remember the transformation you made  to the status in the view and make sure it matches in the CSS. More chances to introduce a bug.

But I can change it if you disagree.

app/views/directives/pipeline-status.html
spadgett: I show a horizontal gray line with no icon for not executed. I could do the same for unknown here.

jwforres: Thats fine, just want to make sure something is there to catch this case as
opposed to blank space.

app/styles/_pipeline.less
jwforres: On Thu, May 19, 2016 at 1:12 PM, Sam Padgett <notifications@github.com>
wrote:

> In app/styles/_pipeline.less
> <https://github.com/openshift/origin-web-console/pull/20#discussion_r63917813>
> :
>
> > +  .build-summary {
> > +    @media (min-width: 480px) {
> > +      border-right-width: 0;
> > +      flex-grow: 0;
> > +      flex-shrink: 0;
> > +      flex-basis: 200px;
> > +    }
> > +    min-height: 100px;
> > +    padding: 10px;
> > +    .status-icon {
> > +      margin-bottom: 3px;
> > +      .inverse {
> > +        font-size: 8px;
> > +        margin-bottom: 3px;
> > +      }
> > +      &.Complete .inverse {
>
> I personally like it matching exactly. Otherwise you have to remember the
> transformation you made to the status in the view and make sure it matches
> in the CSS. More chances to introduce a bug.
>
> But I can change it if you disagree.
>

I only cared because it looks strange to me, its fine if you prefer it this
way.

spadgett:
@jwforres thanks, updated

jwforres:
hmm travis flaked on the bower install, wonder if we are going to have issues with the archived deps

[merge]

PR #28, opened 2016-05-23
Bug 1333000 - warning for the latest builder image

jhadvig:
Closes https://github.com/openshift/origin/issues/7731

@jwforres PTAL

### DIFF
--- a/app/views/directives/osc-image-summary.html
+++ b/app/views/directives/osc-image-summary.html
@@ -4,6 +4,12 @@
   <div ng-show="resource | annotation:'provider'">Provider: {{ resource | annotation:'provider' }}</div>
   <div ng-show="resource.metadata.namespace">Namespace: {{ resource.metadata.namespace }}</div>
 </div>
-<div class="resource-description gutter-bottom"
-     ng-bind-html="resource | description | linky">
+<div class="resource-description"
+     ng-bind-html="resource | description | linky"
+     ng-class="{'gutter-bottom': tag !== 'latest'}">
+</div>
+
+<div ng-if="tag === 'latest'" class="alert alert-info">
+  <span class="pficon pficon-info" aria-hidden="true"></span>
+  This builder image tracks the latest version of the {{name}} image, which could be updated to a newer major version in the future.  If your application will not run on a different major version go <a href="#" back>back</a> and choose a specific version.
 </div>
--- a/app/scripts/directives/oscImageSummary.js
+++ b/app/scripts/directives/oscImageSummary.js
@@ -6,7 +6,8 @@ angular.module("openshiftConsole")
       restrict: "E",
       scope: {
         resource: "=",
-        name: "="
+        name: "=",
+        tag: "="
       },
       templateUrl: "views/directives/osc-image-summary.html"
     };
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -5572,7 +5572,8 @@ return {
 restrict:"E",
 scope:{
 resource:"=",
-name:"="
+name:"=",
+tag:"="
 },
 templateUrl:"views/directives/osc-image-summary.html"
 };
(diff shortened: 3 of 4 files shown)


spadgett:
/cc @bparees 

I'd adjust the warning text a bit. Maybe just

> This builder image points to the latest version of the {{imageStream.metadata.name}} image, which could be updated to newer major release in the future.

spadgett:
Can you add a screenshot? Also, missing dist changes.

jhadvig:
@spadgett sorry, had done it but forgot to attache :-x
screenshot-32

spadgett:
I feel like that's too strong a warning. It looks like there's something wrong with the template :/

spadgett:
It might be better as an info block on the page after you select the template. @jwforres thoughts?

jwforres:
Hmm, yeah thats going to show up all the time, don't want to desensitize
people to our warning icon

Agree that an alert-info block on the next page is probably visible enough
to do the trick.

bparees:
Agree that the current warning is too strong, i like @spadgett's proposed text.  Maybe add ", impacting your application" to the end to make it clear that if the version changes, their application will be affected (as opposed to someone interpreting it as "ok so it might change in the future, but today it's X, so my application will run on X forever")

jhadvig:
@spadgett here is the update screen from the `createFromImage` page
screenshot-33

jhadvig:
I've udpate the PR based on the chat.
One thing I've podified is adding the `gutter-bottom` class only when the imageTag is not latest. If it is dont add it since it the space between description and info-alert would look kinda odd.
With `gutter-bottom` class:
screenshot-36
Without:
screenshot-35

spadgett:
LGTM, @jwforres OK with the message?

app/views/directives/osc-image-summary.html
```diff
@@ -4,6 +4,12 @@
...
-<div class="resource-description gutter-bottom"
-     ng-bind-html="resource | description | linky">
+<div class="resource-description"
+     ng-bind-html="resource | description | linky"
+     ng-class="{'gutter-bottom': tag !== 'latest'}">
+</div>
+
+<div ng-if="tag === 'latest'" class="alert alert-info">
+  <span class="pficon pficon-info" aria-hidden="true"></span>
+  This builder image points to the latest version of the {{name}} image, which could be updated to newer major release in the future, impacting your application.
```
jwforres: What about:

```
This builder image tracks the latest version of the {{name}} image, which could be updated to a newer major version in the future.  If your application will not run on a different major version go <a>back</a> and choose a specific version.
```

jhadvig: Thought about different msges, but didnt came up with anything better :+1:

bparees: I like @jwforres' suggestion, as long as we're ok being wordy :)

jhadvig:
@jwforres updated :) PTAL

app/views/directives/osc-image-summary.html
```diff
@@ -4,6 +4,12 @@
...
-<div class="resource-description gutter-bottom"
-     ng-bind-html="resource | description | linky">
+<div class="resource-description"
+     ng-bind-html="resource | description | linky"
+     ng-class="{'gutter-bottom': tag !== 'latest'}">
+</div>
+
+<div ng-if="tag === 'latest'" class="alert alert-info">
+  <span class="pficon pficon-info" aria-hidden="true"></span>
+  This builder image tracks the latest version of the {{name}} image, which could be updated to a newer major version in the future.  If your application will not run on a different major version go <a href="project/{{project}}/create" >back</a> and choose a specific version.
```
jwforres: so I believe @spadgett wrote a "back" attribute directive that you can put on a link and it will take you to the previous page in the history, then you wouldnt need to pass project into the osc-image-summary directive

jhadvig:
@jwforres you were right, there was a directive for it. Thanks so much ! :)

jwforres:
[merge]

PR #29, opened 2016-05-23
Fix missing metrics for deployments with one pod

spadgett:
We were incorrectly ignoring data points for deployments where `samples === 1`, which was always true when there is one pod.

@jwforres PTAL

### DIFF
--- a/app/scripts/services/metrics.js
+++ b/app/scripts/services/metrics.js
@@ -43,10 +43,29 @@ angular.module("openshiftConsole")
       return point.start + (point.end - point.start) / 2;
     }
 
+    // Is there engouh data to compare min and max values to calculate a usage
+    // rate for a counter metric like CPU or network?
+    function canCalculateRate(point, config) {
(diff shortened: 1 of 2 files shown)


app/scripts/services/metrics.js
```diff
@@ -43,10 +43,29 @@ angular.module("openshiftConsole")
...
+      // If there isn't a min or max, we can't compare.
+      if (!point.min || !point.max) {
+        return false;
+      }
+
+      // For pod metrics, if samples < 2, min and max will always be the same
+      // because there aren't enough samples in the bucket.
+      // For deployment metrics that are "stacked," samples has a different
+      // meaning. It is set to 1 if there is one pod, even when min and max
+      // have different values, so don't ignore this poing.
```
jwforres: typo, point

jwforres:
just fix the one typo then LGTM

spadgett:
Corrected the typo. [merge]

PR #30, opened 2016-05-24
Bug 1333016 - Add token flag to the `oc login`

jhadvig:
Adding the `--token` flag to the `oc login` command that is recommended after project is created. After revealing the token a warning will show under the commands, same as the one in the Command Line Tools.
For that I've created a separate to make the code DRYer.
Attaching the screens:
With webhook trigger:
screenshot
screenshot-1

Without the trigger:
screenshot-2
screenshot-3

Also refactored a bit the `command-line.html` but without any visual changes.

@spadgett PTAL

### DIFF
--- a/app/scripts/directives/popups.js
+++ b/app/scripts/directives/popups.js
@@ -104,4 +104,13 @@ angular.module('openshiftConsole')
       },
       templateUrl: 'views/directives/_warnings-popover.html'
     };
+  })
+  .directive('tokenWarning', function() {
+    return {
+      restrict: 'E',
+      scope: {
+        showSessionToken: '='
+      },
+      template: '<div ng-show="showSessionToken" class="alert alert-warning"><span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span><strong>A token is a form of a password.</strong>Do not share your API token.</div>'
+    };
   });
--- a/app/views/command-line.html
+++ b/app/views/command-line.html
@@ -33,19 +33,10 @@
                   </p>
                   <p>
                     After downloading and installing it, you can start by logging in using<span ng-if="sessionToken"> this current session token</span>:
-                    <div class="code prettyprint ng-binding" ng-if="sessionToken">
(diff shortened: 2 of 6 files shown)


app/views/command-line.html
```diff
@@ -33,19 +33,10 @@ <h1 id="cli">Command Line Tools</h1>
...
-                    </pre>
+                    <pre class="code prettyprint">oc login {{loginBaseURL}} <span ng-if="sessionToken">--token=<span ng-show="showSessionToken">{{sessionToken}}</span><a href="#" ng-click="toggleShowSessionToken()" ng-show="!showSessionToken">...click to show token...</a></span></pre>
                   </p>
 
-                  <div ng-show="showSessionToken" class="alert alert-warning">
-                    <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
-                    <strong>A token is a form of a password.</strong>
-                    Do not share your API token.
-                  </div>
+                  <token-warning show-session-token='showSessionToken'></token-warning>
```
spadgett: If this is just static content, I don't think we need a directive. I suggest just using ng-include here and in the about page.

```html
<div ng-if="showSessionToken" ng-include="'views/token-warning.html'"></div>
```

where token-warning.html has your markup.

app/scripts/directives/popups.js
```diff
@@ -104,4 +104,13 @@ angular.module('openshiftConsole')
...
       templateUrl: 'views/directives/_warnings-popover.html'
     };
+  })
+  .directive('tokenWarning', function() {
+    return {
+      restrict: 'E',
+      scope: {
+        showSessionToken: '='
+      },
+      template: '<div ng-show="showSessionToken" class="alert alert-warning"><span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span><strong>A token is a form of a password.</strong>Do not share your API token.</div>'
```
spadgett: Needs a space after the period.

liggitt:
In general, I don't want to encourage sharing tokens between the CLI and web console.

spadgett:
@jhadvig I talked to @jwforres after https://github.com/openshift/origin-web-console/pull/30#issuecomment-221259898 I think we want to show `oc login <url>` here and only have the token on the CLI page.

jhadvig:
Closing this one based on conversation with @spadgett and @jwforres. Not a bug

PR #37, opened 2016-05-25
Create deployment-donut directive with scaling controls

spadgett:
Move some logic from the overview-deployment directive to a new deployment-donut directive that can be used on other pages like browse -> deployments. This allows us to add the scaling controls anywhere we show the donut.

Fixes https://github.com/openshift/origin/issues/5295

@jwforres PTAL


jwforres:
[merge]

PR #39, opened 2016-05-25
Express network metrics as KiB/s

spadgett:
@jwforres PTAL. We were previous expressing network metrics as MiB per "bucket"... You'd get very different values depending on time range since bucket duration changed.

screen shot 2016-05-25 at 3 28 44 pm

### DIFF
--- a/app/scripts/services/metrics.js
+++ b/app/scripts/services/metrics.js
@@ -51,6 +51,10 @@ angular.module("openshiftConsole")
         return false;
       }
 
+      if (!point.start || !point.end) {
+        return false;
+      }
+
       // For pod metrics, if samples < 2, min and max will always be the same
       // because there aren't enough samples in the bucket.
       // For deployment metrics that are "stacked," samples has a different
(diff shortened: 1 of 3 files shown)


app/scripts/services/metrics.js
```diff
@@ -78,12 +78,14 @@ angular.module("openshiftConsole")
...
-    function bytesUsed(point, config) {
+    function bytesUsedPerSecond(point, config) {
       if (!canCalculateRate(point, config)) {
         return null;
       }
 
-      return point.max - point.min;
+      var seconds = (point.end - point.start) / 1000;
+      var bytesUsed = point.max - point.min;
+      return bytesUsed / seconds;
```
jwforres: Assume there is no chance that secondd could be zero

jwforres:
Just one question otherwise lgtm

app/scripts/services/metrics.js
spadgett: I don't it's possible, but I can add a guard.

spadgett:
@jwforres I added additional checks to catch 0 values.

spadgett:
[merge]

PR #43, opened 2016-05-26
Improve updates for metrics charts

### LINKED ISSUE #23: Past data appears to change in metrics charts
When we update the metrics chart, we request the full data for the time range (e.g. 1 hour) and redraw the chart. Hawkular will break the data into `n` buckets, and we show the usage for each bucket. This means the exact start and end of each bucket changes on every update, and past data can appear to change slightly since the buckets start and end times have changed.

It would be better to keep the previous data and flow new data to the chart by requesting only metrics from the end timestamp of the last bucket we have. This way previous data wouldn't change. Only new data would be added to the chart. (We can remove buckets from the front of the chart when they fall out of the time range as well.)

@jwforres @fabianofranz

spadgett:
Only request new data since the last data point. Minimize odd animations when the chart updates.

Fixes #23

### DIFF
--- a/app/scripts/directives/metrics.js
+++ b/app/scripts/directives/metrics.js
@@ -241,22 +241,22 @@ angular.module('openshiftConsole')
               total: getLimit(metricID)
             };
 
-            var mostRecentValue = _.last(metricData).value;
-            if (isNaN(mostRecentValue)) {
-              mostRecentValue = 0;
+            var lastValue = _.last(metricData).value;
+            if (isNaN(lastValue)) {
+              lastValue = 0;
             }
             if (metric.convert) {
-              mostRecentValue = metric.convert(mostRecentValue);
+              lastValue = metric.convert(lastValue);
             }
 
             // Round to the closest whole number for the utilization chart.
-            usage.used = d3.round(mostRecentValue);
+            usage.used = d3.round(lastValue);
             if (usage.total) {
               usage.available = Math.max(usage.total - usage.used, 0);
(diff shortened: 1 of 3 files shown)


spadgett:
@jwforres PTAL

jwforres:
[merge]

jwforres:
[merge]

spadgett:
Flake, [merge]

```
Selenium standalone server started at http://172.18.6.94:43349/wd/hub
ERROR - Unable to start a WebDriver session.

/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/atoms/error.js:113
  var template = new Error(this.message);
                 ^
UnknownError: chrome not reachable
  (Driver info: chromedriver=2.14.313457 (3d645c400edf2e2c500566c9aa096063e707c9cf),platform=Linux 3.10.0-229.7.2.el7.x86_64 x86_64) (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 75.34 seconds
Build info: version: '2.44.0', revision: '76d78cf', time: '2014-10-23 20:02:37'
System info: host: 'ip-172-18-6-94.ec2.internal', ip: '172.18.6.94', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-229.7.2.el7.x86_64', java.version: '1.8.0_91'
Driver info: org.openqa.selenium.chrome.ChromeDriver
```

spadgett:
[merge]

PR #44, opened 2016-05-26
Round network metrics to one decimal place

spadgett:
Round network metrics to one decimal place to avoid values like "0.007 KiB/s".

screen shot 2016-05-26 at 7 38 58 pm

@jwforres PTAL


jwforres:
[merge]

PR #47, opened 2016-05-27
[WIP] Show what triggered a build

jhadvig:
The browse page of a build will now show what caused it, in the `Status` section. All new builds objects are going to carry the array of objects that cause the build to trigger when https://github.com/openshift/origin/pull/7518 will get merged. 
Triggers will be:
- Image change
- Build Configuration change
- Generic WebHook
- Github WebHook
- Triggered Manually

All old builds will have `triggeredBy` array empty which will display in web-console as `unknown`

Screens:
screenshot-7
screenshot-6
screenshot-5

### DIFF
--- a/app/views/_webhook-trigger-cause.html
+++ b/app/views/_webhook-trigger-cause.html
@@ -1,0 +1,18 @@
+{{trigger.message === 'GitHub WebHook' ? 'GitHub webhook' : 'Generic webhook'}}:
+<span ng-if="trigger.githubWebHook.revision || trigger.genericWebHook.revision"> {{trigger.githubWebHook.revision.git.message || trigger.genericWebHook.revision.git.message}}</span>
+<osc-git-link
+  ng-if="trigger.githubWebHook.revision || trigger.genericWebHook.revision"
+  class="hash"
+  uri="build.spec.source.git.uri"
+  commit="trigger.githubWebHook.revision.git.commit || trigger.genericWebHook.revision.git.commit">{{trigger.githubWebHook.revision.git.commit || trigger.genericWebHook.revision.git.commit | limitTo:7}}
+</osc-git-link>
+<span ng-if="trigger.githubWebHook.revision || trigger.genericWebHook.revision">
+  authored by {{trigger.githubWebHook.revision.git.author.name || trigger.genericWebHook.revision.git.author.name}},
+</span>
+<span ng-if="trigger.genericWebHook && !trigger.genericWebHook.revision">
+  no revision information,
+</span>
+<a href="" ng-if="!showSecret" ng-click="toggleSecret()"> show obfuscated secret</a>
+<span ng-if="showSecret">
+  {{trigger.githubWebHook.secret || trigger.genericWebHook.secret}}
+</span>
--- a/app/views/browse/_build-details.html
+++ b/app/views/browse/_build-details.html
@@ -34,6 +34,26 @@
             </span>
           </span>
         </dd>
+        <div ng-if='build.spec.triggeredBy.length'>
+          <dt>Triggered by:</dt>
+          <dd>
+            <div ng-repeat="trigger in build.spec.triggeredBy">
+              <div ng-switch="trigger.message">
+                <span ng-switch-when="Manually triggered">Manual build</span>
+                <span ng-switch-when="GitHub WebHook">
+                  <ng-include src=" 'views/_webhook-trigger-cause.html' "></ng-include>
+                </span>
+                <span ng-switch-when="Generic WebHook">
+                  <ng-include src=" 'views/_webhook-trigger-cause.html' "></ng-include>
+                </span>
+                <span ng-switch-when="Image change">
+                  {{trigger.message}} for {{trigger.imageChangeBuild.fromRef.name}}
+                </span>
+                <span ng-switch-default>{{trigger.message}}</span>
+              </div>
+            </div>
+          </dd>
+        </div>
       </dl>
       <h3>Configuration <span class="small" ng-if="buildConfigName">created from <a href="{{buildConfigName | navigateResourceURL : 'BuildConfig' : build.metadata.namespace}}">{{buildConfigName}}</a></span></h3>
       <dl class="dl-horizontal left">
--- a/app/scripts/controllers/build.js
+++ b/app/scripts/controllers/build.js
@@ -14,6 +14,7 @@ angular.module('openshiftConsole')
     $scope.buildConfigName = $routeParams.buildconfig;
     $scope.builds = {};
     $scope.alerts = {};
+    $scope.showSecret = false;
     $scope.renderOptions = {
       hideFilterWidget: true
     };
@@ -132,6 +133,10 @@ angular.module('openshiftConsole')
 
           updateCanBuild();
         }));
+
+        $scope.toggleSecret = function() {
+          $scope.showSecret = true;
+        };
 
         $scope.cancelBuild = function() {
           BuildsService.cancelBuild($scope.build, $scope.buildConfigName, context, $scope);
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -186,10 +186,6 @@
     span[data-icon] {
       color: #888;
     }
-    .hash {
-      font-family: @font-family-monospace;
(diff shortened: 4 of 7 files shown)


jhadvig:
Wasn't sure if we wanna show some additional info for the imagechange and BC change triggers like ref, etc., or some revision info for the webhook.
Tested on top of https://github.com/openshift/origin/pull/7518

jwforres:
Yep definitely, seeing what commit triggered a github webhook would be great

jhadvig:
@jwforres I've updated the PR.
Now in case the github || generic webhooks that are carrying commit info we will show the 'commit ID' and 'commit msg'. I case of the generic webhook that isn't carrying the info we will show just 'No revision or commit information attached' as are the screens showing:
screenshot-12
screenshot-13
Wasnt sure if we wanna also display the author or committer so I leave them out of the popover, since the ID and msg are most important fields. 
For the other types of triggers we wont display any info.

PTAL

spadgett:
"Triggered by: unknown" sounds a bit odd. I might just remove the field if we don't know.

Maybe "Triggered by: Manual Build" instead of "Triggered by: Triggered Manually".

We probably only need the first 7 chars of the commit hash. If we show it inline instead of in a tooltip, we can make it a link. I think that's better for mobile anyway. This is what we use in the pod template.

```html
  <osc-git-link
    class="hash"
    uri="build.spec.source.git.uri"
    commit="build.spec.revision.git.commit">{{build.spec.revision.git.commit | limitTo:7}}</osc-git-link>
```

jhadvig:
@spadgett I've updated the PR, and attaching the screens:

Generic WH without payload
screenshot-20
gitHub WH:
screenshot-19
manual build:
screenshot-18
Image change (config change will look the same, meaning it will just show the 'BuildConfig change')
screenshot-17

jwforres:
We have plenty of space, i would use the space.  Show the commit message and author right there inline.

For image change trigger, say which image changed (i assume we know?)

Are you comparing to what oc describe outputs for all of these?  Because I know we had a discussion on that thread about showing which secret triggered the webhook because you can have more than one generic webhook.  We talked about showing a partially obfuscated secret so you have a chance of knowing which it is without seeing the whole secret.

jhadvig:
@jwforres true that we have plenty of space but wasnt sure how put all the date (for githubwebhook: commit, author, secret;  for imageChange: Image ID, Image Name/Kind; ...), that we show via `oc describe build ...` inline in console.
Instead I've tryed to recycle what we use in the pod-template, when listing the fields, and I kinda like it.
Here are the screens:
screenshot-24
screenshot-25

Also plays nice with mobile:
screenshot-26


cc'ing @spadgett

jhadvig:
@jwforres @spadgett here is the preview for displaying github webhook data inline.
screenshot-28

jwforres:
cc @rhamilto

rhamilto:
I like the inline display.  I think that's a sensible default since that's what we do in most cases throughout the console. 

I recommend changing all instances of "WebHook" to "webhook" since the word isn't a proper noun.  See https://developer.github.com/webhooks/.

It seems like the default behavior with secrets in the console is to use the "Show" link, so that seems like the sensible thing to do over simply displaying the secret without user interaction.

I suggest the following simple tweaks for webhooks with payload so that we're consistent with how we display commit information in pod templates (screenshot below).

screen shot 2016-06-02 at 11 46 26 am


**Triggered by:**         Github webhook:  Added license [9bdc3a2](#) authored by Anonymous User,  [Show secret](#)

Note:  make sure to change the commit hash to use monospace font (like we do in pod templates).

> Here I wasnt sure if we wanna also display the ImageId since it will be the same as the Builder Image value which is displaed below in the Configuration.

I don't think there is a need to duplicate the builder image value since it's already on the page below.  Anyone disagree?

> For the ConfigChange we will only display:
> Triggered by: Config Change

I recommend changing "Config Change" to "Config change" so that the casing is consistent with the other examples (we're using title case it seems for all our labels).

Thoughts?

jhadvig:
@rhamilto thanks so much for sharing ideas :)) really like how the webhook trigger reason looks now :ok_hand:
One thing that we need to figure out is, if we will notify the user that the secret, that will be revealed after clicking on the "Show secret" link, is in obfuscated format, which I think we should, to prevent from any confusions. 
Here are the screens together with and without a tooltip proposal, which will inform user about the fact that the pw is obfuscated:
screenshot-34

screenshot-35
 
screenshot-37
Here Im not sure why the printScreen wont display the cursor attribute in the 'help' state, but it works 
screenshot-38

Or there is also ooption that we change 'Show secret' into 'Show obfuscated secret', so we wont need to show the tooltip, which I kinda incline to. Screen: 
screenshot-36


What do you think ?

I've addressed rest of your comments as well :+1:

rhamilto:
> Or there is also ooption that we change 'Show secret' into 'Show obfuscated secret', so we wont need to show the tooltip, which I kinda incline to.

Agreed.  This seems like the simplest solution.

rhamilto:
All else LGTM.  Nice work, @jhadvig.

jhadvig:
@rhamilto I've updated the PR with 'Show obfuscated secret' link, as discussed. Thanks :)

@jwforres PTAL

app/views/_webhook-trigger-cause.html
```diff
@@ -0,0 +1,9 @@
+{{trigger.message === 'GitHub WebHook' ? 'GitHub webhook' : 'Generic webhook'}} webhook: {{build.spec.revision.git.message}} <osc-git-link
```
jwforres: its printing "webhook" twice,  "GitHub webhook webhook"

app/views/_webhook-trigger-cause.html
```diff
@@ -0,0 +1,9 @@
...
+  class="hash"
+  uri="build.spec.source.git.uri"
+  commit="build.spec.revision.git.commit">{{build.spec.revision.git.commit | limitTo:7}}
+</osc-git-link>
+authored by {{build.spec.revision.git.author.name}}, <a href="" ng-if="!showSecret" ng-click="toggleSecret()">Show obfuscated secret</a>
```
jwforres: i would lowercase the S, it looks weird to me to have a ',' followed by a capital letter

app/views/browse/_build-details.html
```diff
@@ -34,6 +34,32 @@
...
+                <span ng-switch-when="Manually triggered">Manual build</span>
+                <span ng-switch-when="GitHub WebHook">
+                  <ng-include src=" 'views/_webhook-trigger-cause.html' "></ng-include>
+                </span>
+
+                <span ng-switch-when="Generic WebHook">
+                  <span ng-if="trigger.genericWebHook.revision">
+                    <ng-include src=" 'views/_webhook-trigger-cause.html' "></ng-include>
+                  </span>
+                  <span ng-if="!trigger.genericWebHook.revision">Generic webhook without revision</span>
```
jwforres: "Generic webhook, unknown revision"
or
"Generic webhook, no revision information"
or
"Generic webhook, no revision details"
?

jwforres: even without the revision information don't you know what secret triggered it?

app/views/_webhook-trigger-cause.html
```diff
@@ -0,0 +1,9 @@
...
+<span ng-if="showSecret">
+  {{trigger.githubWebHook.secret}}
```
jwforres: what if its a genericWebHook?

app/views/browse/_build-details.html
jhadvig: +1 for the 'Generis webhook, no revision information'

jhadvig:
@jwforres update the PR, PTAL

jwforres:
was just talking with @mfojtik and it sounds like this information is getting moved from spec to status

jhadvig:
@jwforres luckily @mfojtik was just hallucinating :smile: and it should stay as it is :v:

mfojtik:
yes.

jwforres:
ok :-)

jhadvig:
@jwforres if there arent any other comments can we merge this ? :)

jwforres:
[merge]

PR #48, opened 2016-05-27
Bug 1333165 - after creating from a template in console, no idea what values generated parameters got

jhadvig:
Showing the generated parameters on the next-step page
Wasnt fully sure if we want to put the params at the beginning or the end, so attaching both versions + one that failed to create resources:
screenshot-11
screenshot-10
This one is a bit zoomed out...
screenshot-9

@jwforres PTAL


jwforres:
definitely prefer having them on the bottom, not sure how i feel about how much space the table takes up, but I know in some cases the parameter could contain a very long string, or even a multiline string once we support that correctly.

I feel like it needs some sentence there above the table, something like "These parameters were generated automatically and often include things like passwords.  If you will need to reference these values later, copy them to a safe location."

cc @spadgett @rhamilto @ajacobs21e

jhadvig:
@jwforres what about: "These parameters often include things like passwords. If you will need to reference these values later, copy them to a safe location. Some of these parameters were generated automatically."

screenshot-16

Reorganized the sentence a bit since not all the params are generated automatically but dont wanted to end up with something like : "Some of these parameters were generated automatically and often include things like passwords. If you will need to reference these values later, copy them to a safe location."  :anguished:

ajacobs21e:
I like your wording @jhadvig. Is there a more descriptive heading we can use? "Used parameters" makes me think they're not available (but I'm not familiar with the domain). 'app parameters'? 'applied parameters'? 

Is there a way now to find these parameters in the UI? If not, will there be one day?

jhadvig:
@ajacobs21e the only way how to find out these parameters, is to know which resources are referencing in the template, and after the resource is created user can find them in the `Environment` tab of the resource page.
I see your point with the 'Used parameters', but I wouldn't go with 'App paramaters' or 'Application parameters' since template is not necessarily tied to creating application, but rather all kinds of components - HPA, Jobs, ImageStream, ... which are not really an application but a component. For that reason I would prefer 'Used' or 'Applied Parameters', but they are kinda the same :-P

jwforres:
Yep like your wording better @jhadvig

As far as finding them later, parameters don't even necessarily mean
environment variables.  Parameters can end up being used for all kinds of
things like which image to use, which source repository, etc.  So we can't
really direct them where to look for them later unfortunately.   Typically
the ones that are generated are being used in environment variables, or
secrets.

Hmm, i think i prefer Applied Parameters, or Processed Parameters ?
@bparees any opinion here?

spadgett:
> Hmm, i think i prefer Applied Parameters, or Processed Parameters ?

Is just "Parameters" too general?

bparees:
or Parameter Values ?

jhadvig:
Applied Parameter Values ?
@jwforres @spadgett @bparees

ajacobs21e:
I like applied parameter values

jhadvig:
@jwforres PTAL

rhamilto:
The table seems a bit jarring on this page, but it seems like the correct UI presentation since it is how we display name/value pairs elsewhere in the console.  Maybe dropping the white background color on the table cells would help (see attached)?

no-white-cells

Should the passwords be obscured by a click to show link (e.g., Show password)?  It seems like simply outputting passwords to the page is inconsistent with what we do in other password instances (although I see we output MYSQL_PASSWORD values on the environment tab of deployments (see attached)).  Hmm.

screen shot 2016-06-02 at 1 23 04 pm

jwforres:
we dont know whats a password / secret key / etc other than checking the
name for "password" :-\

jhadvig:
@rhamilto thanks, really like how just simple change of bg color does make difference :+1: so I've updated the PR with proposed change.
Unfortunately as @jwforres mentioned we can't tell what is and what isn't the password fur sure, just assume it, but I don't think we want to walk that path.
Though it would be good to be consistent on how we display passwords in console.

rhamilto:
LGTM.  

Understood about the password.  What if we made every param a click-to-show in this context?  That's sort of yucky, but at least it's not revealing passwords?  Could even add a "Show all" link to avoid multiple clicks?  @jwforres, thoughts?

app/styles/_core.less
```diff
@@ -1029,6 +1029,12 @@ h1 small.meta, .build-config-summary .meta {
   color: @gray-light;
 }
 
+.no-background {
```
rhamilto: A "global" fix that would address this any time a .table appears on a gray background would be:

.surface-shaded .table {
  background-color: transparent;
}

Plus, there is then no need to add an additional class to the table.

jhadvig:
@rhamilto I like that :+1:

jwforres:
ok I discussed this with dan and because templates can actually create Secrets, we shouldn't make these visible by default.  Honestly the simplest solution is to have the Header and the message that explains everything and then have a "Show parameters" link that shows the entire table.  WDYT?

rhamilto:
>  Honestly the simplest solution is to have the Header and the message that explains everything and then have a "Show parameters" link that shows the entire table. WDYT?

I like the practicality of that solution, but share @jhadvig's concern that the parameters then will be easy to overlook.  How important is it the user sees the parameters?

jhadvig:
@rhamilto attaching the screens for the change that @jwforres proposed

screenshot-39
screenshot-40

But would be good to somehow highlight the link. 

What about center the link ? Or change the link to a btn ? Mooore screeeens :grin:  : 
Proposal 1
screenshot-41
Proposal 2
screenshot-44
Proposal 3
screenshot-45
Proposal 4
screenshot-46
Proposal 5
screenshot-47

Where I like the 1 and 5 most. Or do you have other proposals ?

rhamilto:
Minor nit:  add a period after "automatically".

Hmm.  I personally don't like centering since English reads from left to right.  

The button does draw the eye, although the blue one is a bit too strong.  So I suggest either sticking with the blue link left aligned or the gray button left aligned.  Opinions @spadgett, @benjaminapetersen, @sg00dwin ?

benjaminapetersen:
@rhamilto @jhadvig @jwforres I like the idea to hide them with a link to show, but I would not hide the table... just use `************` over the values, and click `show` to reveal.  

quick mock:
show-values

sg00dwin:
I prefer @benjaminapetersen idea to show the table with ************** and a Show values link.

jwforres:
ok, sounds like thats the preference then :)  its fine with me

spadgett:
I think we should know what values were generated... How about a message like

> Parameters FOO, BAR, and BAZ have generated values. <a href="">Show values</a>

where clicking would show the table.

jhadvig:
@spadgett you mean to hide the table by default ?

I would go with @benjaminapetersen proposal by hiding the values, but I dont get the `hidden` field in the code sample. Shall the generated parameters be hidden by default and rest will be visible?

spadgett:
It seems to show the same values I just entered starred out. The table doesn't tell me anything new unless I click the show values link, and it might be confusing if none of the values are actually sensitive.

I feel like a message saying these variables were generated with a link to see generated values is more informative.

jwforres:
@liggitt @bparees @ncdc depending on which proposal we go with for generated secrets in templates there may be an effect on this.   If secret creation does the encoding there is no problem, if templates have a base64 generator there will be a small impact

bparees:
@jwforres good point.  not sure what you'd want to do if we create a base64 generator...  would you try to recognize the generator type and decode the value?  or try to recognize the value itself as being base64 and decode it?

jhadvig:
@jwforres @spadgett @rhamilto I've updated the PR.
Based on the discussion we align on the fact we will hide the table by default with a link that will show it.
Also the sentence that describes the `Applied Parameter Values` section will list names of paramarets with generated values.
Screens:
screenshot

screenshot-1

In case of single generated parameter value
screenshot-2

app/scripts/services/templates.js
```diff
@@ -0,0 +1,53 @@
...
+        template = null;
+      }
+    };
+
+  }).service("ProcessedParametersService", function(){
+
+    var params = [];
+    var paramsWithGeneratedValues = [];
+    return {
+      setParams: function(paramaters) {
```
jwforres: misspelled parameters, couple places in this file

app/scripts/services/templates.js
```diff
@@ -0,0 +1,53 @@
...
+        _.each(paramaters, function(param) {
+          params.push({
+            name: param.name,
+            value: param.value
```
jwforres: i think id rather see one pair of set/get methods and just have a 'wasGenerated' or 'generated' added to this map

app/views/create/next-steps.html
```diff
@@ -89,6 +89,18 @@ <h1 ng-if="!pendingTasks(tasks()).length && erroredTasks(tasks()).length">Comple
               </div>
             </div>
 
+            <div ng-if="processedParams.length">
+              <h2>Applied Parameter Values</h2>
+              <p>These parameters often include things like passwords. If you will need to reference these values later, copy them to a safe location.
+                <span ng-if="generatedParams.length > 1">Parameters <span ng-repeat="paramName in generatedParams">{{paramName}}{{$index+1 !== generatedParams.length ? "," : ""}} </span>were generated automatically.</span>
```
jwforres: you can use $last to check whether to add the comma

```
<span ng-repeat="paramName in generatedParams">{{paramName}}<span ng-if="!$last">, </span></span>
```

jwforres:
@bparees is it possible for us to know from the parameter info what type of generator it is? do not want to go down the road of detecting something is a base64 encoded value...

app/scripts/services/templates.js
jhadvig: :+1: was thinking of it, but wanted to be as clear as possible.

app/views/create/next-steps.html
jhadvig: :ok_hand:

bparees:
@jwforres assuming we introduce a new type (like "base64_expression"), you can look at the generator type:
https://github.com/openshift/origin/blob/master/examples/quickstarts/nodejs-mongodb.json#L445

that field should still be available even after the template is processed (ie values generated).

jhadvig:
@jwforres I've addressed the your comments.
Once the introduce the new "base64_expression" like type, will update the logic. For now we should be ok.
PTAL :)

app/scripts/services/templates.js
```diff
@@ -0,0 +1,51 @@
...
+      },
+      clearTemplate: function() {
+        template = null;
+      }
+    };
+
+  }).service("ProcessedParametersService", function(){
+
+    var params = {
+      generated: [],
```
jwforres: hmm these don't seem right, really its
params.all and params.generated

i was hoping it would have been a single array with generated being a flag on the individual params, but i see where that would complicate you changing from the one generated param to many generated params sentences in the view :-\

jwforres:
just fix the one comment and rebase, guessing the merge conflicts are on dist files

jhadvig:
@jwforres I've updated the PR, and put the generated dist files in a separate commit, cause they are huge for some unknown reason to me. Maybe its fine but would rather have your opinion on it.
Will squash afterwards :)

jwforres:
[merge]

PR #49, opened 2016-05-27
Tweak to debounce delay on osc-key-values clean check

benjaminapetersen:
@jwforres increased the debounce just a bit, found at `100` it might not catch the first key entry but works consistently at `200`.


benjaminapetersen:
btw, this is so small ill just do it directly in the backport PR, dont need to worry about merging it before i get that rolling.

jwforres:
[merge]

PR #66, opened 2016-06-15
Container terminal selector Review

MarkDeMaria:
Initial style/functional review of the Container Terminal selector dropdown feature. Functionality is roughly 90% done. The remaining 10% rests on effective testing of crashing containers and disconnecting terminals. Currently, the 'died' value for a terminal is only set when a terminal from the current set is missing from the future (read: updated) set of terminals. Currently the only way for a dead terminal to become alive (not dead) is for it to appear in the next updated list of terminals, not from the client restarting/refreshing the terminal via the reconnect button in the terminal window. The final implementation will take connection and status data directly from the containers to which the terminals are directly connecting.

@benjaminapetersen

### DIFF
--- a/app/scripts/controllers/pod.js
+++ b/app/scripts/controllers/pod.js
@@ -87,6 +87,57 @@ angular.module('openshiftConsole')
       }
     };
 
+    /**
+     *  Will set the newTerm's isVisible/isUsed to true, while hiding the previous
+     */
+    $scope.onTerminalSelectChange = function(newTerm) {
+      // Make all terminals invisible (Becuase we don't have a pointer to the terminal that is currently visible)
+      _.each($scope.containerTerminals, function(term) {
+        term.isVisible = false;
+      });
+
+      newTerm.isVisible = true;
+      newTerm.isUsed = true;
+    };
+
+    var getState = function(containerStatus) {
+      var state = _.get(containerStatus, 'state', {});
+      return _.head(_.keys(state));
+    };
+
+    var makeTerminals = function() {
+      var terminals = [];
+
+      _.each($scope.pod.spec.containers, function(container) {
+        var thisContainerStatus = _.find($scope.pod.status.containerStatuses, { name: container.name });
+        var thisContainerState = getState(thisContainerStatus);
+
+        terminals.push({
+          containerName: container.name,
+          isVisible: false,
+          isUsed: false,
+          containerState: thisContainerState
+        });
+      });
+
+      var currentlyVisible = _.head(terminals);
+      currentlyVisible.isVisible = true;
+      currentlyVisible.isUsed = true;
+
+      $scope.selectedTerminalContainer = currentlyVisible;
+
+      return terminals;
+    };
+
+    var updateTerminals = function(terminals) {
+      _.each(terminals, function(term) {
+        var thisContainerStatus = _.find($scope.pod.status.containerStatuses, { name: term.containerName });
+        var thisContainerState = getState(thisContainerStatus);
+
+        term.containerState = thisContainerState;
+      });
+    };
+
     ProjectsService
       .get($routeParams.project)
       .then(_.spread(function(project, context) {
@@ -106,6 +157,8 @@ angular.module('openshiftConsole')
             pods[pod.metadata.name] = pod;
             ImageStreamResolver.fetchReferencedImageStreamImages(pods, $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, context);
 
+            $scope.containerTerminals = makeTerminals();
+
             // If we found the item successfully, watch for changes on it
             watches.push(DataService.watchObject("pods", $routeParams.pod, context, function(pod, action) {
               if (action === "DELETED") {
@@ -117,6 +170,9 @@ angular.module('openshiftConsole')
               $scope.pod = pod;
               setLogVars(pod);
               setContainerVars();
+
+              updateTerminals($scope.containerTerminals);
+
             }));
           },
           // failure
--- a/app/views/browse/pod.html
+++ b/app/views/browse/pod.html
@@ -99,11 +99,11 @@
                       </span>
 
                       <select
-                          id="selectLogContainer"
-                          ng-if="pod.spec.containers.length > 1"
-                          ng-model="logOptions.container"
-                          ng-options="container.name as container.name for container in pod.spec.containers"
-                          ng-init="logOptions.container = pod.spec.containers[0].name">
+                        id="selectLogContainer"
+                        ng-if="pod.spec.containers.length > 1"
+                        ng-model="logOptions.container"
+                        ng-options="container.name as container.name for container in pod.spec.containers"
+                        ng-init="logOptions.container = pod.spec.containers[0].name">
                       </select>
 
                       <span ng-if="containerStateReason || containerStatusKey">
@@ -119,24 +119,68 @@
                     </log-viewer>
                   </uib-tab>
 
-                  <uib-tab active="selectedTab.terminal"
+                  <uib-tab
+                    active="selectedTab.terminal"
                     select="terminalTabWasSelected = true"
                     ng-init="containers = pod.status.containerStatuses"
-                    ng-if="containersRunning(pod.status.containerStatuses) > 0 && ('pods/exec' | canI : 'get')">
-                    <uib-tab-heading>Terminal</uib-tab-heading>
-                    <div>
-                      <span class="pficon pficon-info" aria-hidden="true"></span>
-                      When you navigate away from this pod, any processes running
-                      in these terminals will quit.
-                    </div>
-                    <div class="pod-container-terminal"
-                      ng-repeat="container in containers | orderBy:'name' track by container.name"
-                      ng-if="container.state.running">
-                      <h3 ng-if="containers.length > 1">{{container.name}}</h3>
-                      <kubernetes-container-terminal pod="pod" container="container.name"
-                        prevent="!terminalTabWasSelected">
-                      </kubernetes-container-terminal>
-                    </div>
+                    ng-if="containersRunning(pod.status.containerStatuses) > 0">
+                      <uib-tab-heading>Terminal</uib-tab-heading>
+
+                      <div class="mar-bottom-md">
+                        <span class="pficon pficon-info" aria-hidden="true"></span>
+                        When you navigate away from this pod, any open terminal connections will be closed.
+                        This will kill any foreground processes you started from the terminal.
+                      </div>
+
+                      <div class="mar-left-xl mar-bottom-lg">
+                        <div class="row">
+                          <div class="pad-left-none pad-bottom-lg col-xs-4 col-lg-3">
+                            <ui-select
+                              ng-model="selectedTerminalContainer"
+                              on-select="onTerminalSelectChange(selectedTerminalContainer)"
+                              class="mar-left-none pad-left-none pad-right-none">
+                             <ui-select-match
+                                class="truncate"
+                                placeholder="Container Name">
+                                  <span class="pad-left-md">
+                                    {{selectedTerminalContainer.containerName}}
+                                  </span>
+                              </ui-select-match>
+                              <ui-select-choices
+                                repeat="term in containerTerminals | filter: $select.search"
+                                ui-disable-choice="(term.containerState !== 'running') && !term.isUsed">
+                                <div row>
+                                  <span ng-bind-html="term.containerName | highlight: $select.search">
+                                  </span>
+                                  <span flex></span>
+                                  <span
+                                    ng-if="term.isUsed && (term.containerState === 'running')"
+                                    ng-class="{'text-muted' : (term.status === 'disconnected')}">
+                                      {{term.status}}
+                                  </span>
+                                  <span
+                                    ng-if="term.containerState !== 'running'"
+                                    ng-class="{'text-muted' : !term.isUsed}">
+                                      {{term.containerState}}
+                                  </span>
+                                </div>
+                              </ui-select-choices>
+                            </ui-select>
+                          </div>
+                        </div>
+
+                        <div class="row" ng-repeat="term in containerTerminals">
+                          <kubernetes-container-terminal
+                            prevent="!terminalTabWasSelected"
+                            ng-if="term.isUsed"
+                            ng-show="term.isVisible"
+                            pod="pod"
+                            container="term.containerName"
+                            status="term.status"
+                            >
+                          </kubernetes-container-terminal>
+                        </div>
+                      </div>
     	            </uib-tab>
 
                   <uib-tab active="selectedTab.events" ng-if="'events' | canI : 'watch'">
--- a/bower.json
+++ b/bower.json
@@ -28,7 +28,7 @@
     "messenger": "1.4.1",
     "kubernetes-label-selector": "1.0.1",
     "kubernetes-topology-graph": "0.0.23",
-    "kubernetes-container-terminal": "0.0.13",
+    "kubernetes-container-terminal": "0.0.14",
     "openshift-object-describer": "1.1.2",
     "layout.attrs": "2.1.0",
     "bootstrap-hover-dropdown": "2.1.3",
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -2909,15 +2909,46 @@ containerEndTime:_.get(c, [ d, "finishedAt" ])
 });
 }
 };
+a.onTerminalSelectChange = function(b) {
+_.each(a.containerTerminals, function(a) {
+a.isVisible = !1;
+}), b.isVisible = !0, b.isUsed = !0;
+};
+var n = function(a) {
+var b = _.get(a, "state", {});
+return _.head(_.keys(b));
+}, o = function() {
+var b = [];
+_.each(a.pod.spec.containers, function(c) {
+var d = _.find(a.pod.status.containerStatuses, {
+name:c.name
+}), e = n(d);
+b.push({
+containerName:c.name,
+isVisible:!1,
+isUsed:!1,
+containerState:e
+});
+});
+var c = _.head(b);
+return c.isVisible = !0, c.isUsed = !0, a.selectedTerminalContainer = c, b;
+}, p = function(b) {
+_.each(b, function(b) {
+var c = _.find(a.pod.status.containerStatuses, {
+name:b.containerName
+}), d = n(c);
+b.containerState = d;
+});
+};
 j.get(c.project).then(_.spread(function(d, h) {
 a.project = d, a.projectContext = h, f.get("pods", c.pod, h).then(function(b) {
 a.loaded = !0, a.pod = b, l(b), m();
 var d = {};
-d[b.metadata.name] = b, g.fetchReferencedImageStreamImages(d, a.imagesByDockerReference, a.imageStreamImageRefByDockerReference, h), k.push(f.watchObject("pods", c.pod, h, function(b, c) {
+d[b.metadata.name] = b, g.fetchReferencedImageStreamImages(d, a.imagesByDockerReference, a.imageStreamImageRefByDockerReference, h), a.containerTerminals = o(), k.push(f.watchObject("pods", c.pod, h, function(b, c) {
 "DELETED" === c && (a.alerts.deleted = {
 type:"warning",
 message:"This pod has been deleted."
-}), a.pod = b, l(b), m();
+}), a.pod = b, l(b), m(), p(a.containerTerminals);
 }));
 }, function(c) {
 a.loaded = !0, a.alerts.load = {
--- a/dist/scripts/templates.js
+++ b/dist/scripts/templates.js
@@ -2488,16 +2488,41 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
     "</span>\n" +
     "</log-viewer>\n" +
     "</uib-tab>\n" +
-    "<uib-tab active=\"selectedTab.terminal\" select=\"terminalTabWasSelected = true\" ng-init=\"containers = pod.status.containerStatuses\" ng-if=\"containersRunning(pod.status.containerStatuses) > 0 && ('pods/exec' | canI : 'get')\">\n" +
+    "<uib-tab active=\"selectedTab.terminal\" select=\"terminalTabWasSelected = true\" ng-init=\"containers = pod.status.containerStatuses\" ng-if=\"containersRunning(pod.status.containerStatuses) > 0\">\n" +
     "<uib-tab-heading>Terminal</uib-tab-heading>\n" +
-    "<div>\n" +
+    "<div class=\"mar-bottom-md\">\n" +
     "<span class=\"pficon pficon-info\" aria-hidden=\"true\"></span>\n" +
-    "When you navigate away from this pod, any processes running in these terminals will quit.\n" +
+    "When you navigate away from this pod, any open terminal connections will be closed. This will kill any foreground processes you started from the terminal.\n" +
     "</div>\n" +
-    "<div class=\"pod-container-terminal\" ng-repeat=\"container in containers | orderBy:'name' track by container.name\" ng-if=\"container.state.running\">\n" +
-    "<h3 ng-if=\"containers.length > 1\">{{container.name}}</h3>\n" +
-    "<kubernetes-container-terminal pod=\"pod\" container=\"container.name\" prevent=\"!terminalTabWasSelected\">\n" +
+    "<div class=\"mar-left-xl mar-bottom-lg\">\n" +
+    "<div class=\"row\">\n" +
+    "<div class=\"pad-left-none pad-bottom-lg col-xs-4 col-lg-3\">\n" +
+    "<ui-select ng-model=\"selectedTerminalContainer\" on-select=\"onTerminalSelectChange(selectedTerminalContainer)\" class=\"mar-left-none pad-left-none pad-right-none\">\n" +
+    "<ui-select-match class=\"truncate\" placeholder=\"Container Name\">\n" +
+    "<span class=\"pad-left-md\">\n" +
+    "{{selectedTerminalContainer.containerName}}\n" +
+    "</span>\n" +
+    "</ui-select-match>\n" +
+    "<ui-select-choices repeat=\"term in containerTerminals | filter: $select.search\" ui-disable-choice=\"(term.containerState !== 'running') && !term.isUsed\">\n" +
+    "<div row>\n" +
+    "<span ng-bind-html=\"term.containerName | highlight: $select.search\">\n" +
+    "</span>\n" +
+    "<span flex></span>\n" +
+    "<span ng-if=\"term.isUsed && (term.containerState === 'running')\" ng-class=\"{'text-muted' : (term.status === 'disconnected')}\">\n" +
+    "{{term.status}}\n" +
+    "</span>\n" +
+    "<span ng-if=\"term.containerState !== 'running'\" ng-class=\"{'text-muted' : !term.isUsed}\">\n" +
+    "{{term.containerState}}\n" +
+    "</span>\n" +
+    "</div>\n" +
+    "</ui-select-choices>\n" +
+    "</ui-select>\n" +
+    "</div>\n" +
+    "</div>\n" +
+    "<div class=\"row\" ng-repeat=\"term in containerTerminals\">\n" +
+    "<kubernetes-container-terminal prevent=\"!terminalTabWasSelected\" ng-if=\"term.isUsed\" ng-show=\"term.isVisible\" pod=\"pod\" container=\"term.containerName\" status=\"term.status\">\n" +
     "</kubernetes-container-terminal>\n" +
+    "</div>\n" +
     "</div>\n" +
     "</uib-tab>\n" +
     "<uib-tab active=\"selectedTab.events\" ng-if=\"'events' | canI : 'watch'\">\n" +
--- a/dist/scripts/vendor.js
+++ b/dist/scripts/vendor.js
@@ -41214,12 +41214,13 @@ prevent:"=",
 rows:"=",
 cols:"=",
 screenKeys:"=",
-autofocus:"=?"
+autofocus:"=?",
+status:"="
 },
 link:function(g, h, i) {
 function j() {
 function a(a) {
-!a && j && (a = "Could not connect to the container. Do you have sufficient privileges?"), a || (a = "disconnected"), j || (a = "\r\n" + a), q.write("[31m" + a + "[m\r\n"), g.$apply(k);
+!a && j && (a = "Could not connect to the container. Do you have sufficient privileges?"), a || (a = "disconnected"), j || (a = "\r\n" + a), q.write("[31m" + a + "[m\r\n"), g.status = "disconnected", g.$apply(k);
 }
 k(), q.reset();
 var b = "", c = g.pod();
@@ -41246,15 +41247,16 @@ q.write(d(b));
 }
 j && (j = !1, m.addClass("hidden"), n.addClass("hidden"), q.cursorHidden = !1, q.showCursor(), q.refresh(q.y, q.y), g.autofocus && q.element && q.element.focus());
 }, p.onclose = function(b) {
-a(b.reason);
+g.status = "disconnected", a(b.reason);
 };
 }, function(b) {
 a(b.message);
-});
+}), g.status = "connected";
 }
 function k() {
-m.addClass("hidden"), n.removeClass("hidden"), q && (q.cursorHidden = !0, q.refresh(q.x, q.y)), p && (p.onopen = p.onmessage = p.onerror = p.onclose = null, p.readyState < 2 && p.close(), p = null), window.clearInterval(o), o = null;
+g.status = "disconnected", m.addClass("hidden"), n.removeClass("hidden"), q && (q.cursorHidden = !0, q.refresh(q.x, q.y)), p && (p.onopen = p.onmessage = p.onerror = p.onclose = null, p.readyState < 2 && p.close(), p = null), window.clearInterval(o), o = null;
 }
+g.status = "disconnected";
 var l = a.element("<div class='terminal-wrapper'>");
 h.append(l);
 var m = a.element("<div class='spinner spinner-white hidden'>"), n = a.element("<button class='btn btn-default fa fa-refresh'>");


benjaminapetersen:
cool, looking it over

app/scripts/controllers/pod.js
```diff
@@ -86,7 +86,115 @@ angular.module('openshiftConsole')
         });
       }
     };
+       
+    $scope.onChange = function(arg) {
```
benjaminapetersen: Perhaps this can be named something like `onTerminalChange` to give it context.

app/scripts/controllers/pod.js
```diff
@@ -86,7 +86,115 @@ angular.module('openshiftConsole')
...
+            // Add this to the list of connected terminals
+            $scope.connectedTerminals.push(ctr);
+          }
+        } else {
+          ctr.isActive = false;
+        }
+      }); 
+    }
+    
+    $scope.tabSelect = function() {
```
benjaminapetersen: I think this function is not being used.

MDeMariaRH: I tested removing it and the terminal window will not change without it. It takes care of the user switching the terminal window to the one that they desire. The updateSelectTerminals handles refreshing the list of terminals available in the select on the event that the list of available terminals changes (Due to a terminal disconnect or container crash).

app/scripts/controllers/pod.js
```diff
@@ -118,6 +226,9 @@ angular.module('openshiftConsole')
               setLogVars(pod);
               setContainerVars();
             }));
+            
+            // TODO Proper location for this call?
+            populateFrontEndTerminalData($scope.pod.status.containerStatuses);
```
benjaminapetersen: You can probably call this function in two places.  Once above and once below the watch:

```javascript
// populate with data from DataService.get (this happens once)
populateFrontEndTerminalData(pod.status.containerStatuses);
watches.push(DataService.watchObject, ...
    // now, update with DataService.watch data (this happens at least once, potentially many times)
    populateFrontEndTerminalData(pod.status.containerStatuses);

```

benjaminapetersen:
Make sure that jshint is running & you are fixing anything it suggests, such as `;` or `!==` vs `!=`.  you can `grunt jshint` if your editor doesn't pick up the `.jshintrc` file.  I believe it automatically runs with `grunt serve` and your terminal will tell you things to fix every time you save.

app/scripts/controllers/pod.js
```diff
@@ -86,7 +86,115 @@ angular.module('openshiftConsole')
...
+      var availableTerminals = _.map(containers, function(container) {
+        return {
+          name: container.name,
+          isActive: false,
+          isConnected: false,
+          died: false
+        };
+      });
+      
+      availableTerminals.push({
```
benjaminapetersen: def made a note to remove this so it isn't accidentally left in.  I typically do todos like this:
```javascript
TODO (bpeterse): thing that I need to do
```

app/scripts/controllers/pod.js
```diff
@@ -86,7 +86,115 @@ angular.module('openshiftConsole')
...
+        name: 'testDead',
+        isActive: false,
+        isConnected: false,
+        died: true
+      });
+            
+      // INTERSECTION
+      _.forEach(availableTerminals, function(newTerm, outerIndex) {
+        
+        var index = _.findIndex($scope.containerTerminals, function(currentTerm) {
```
benjaminapetersen: Can you use `_.find()` here instead?  It might be more clear than working off of an index & testing `-1`.

app/scripts/controllers/pod.js
MDeMariaRH: Changed to 'onTerminalSelectChange.'

app/scripts/controllers/pod.js
benjaminapetersen: If you went with `_.find()`, you can then also update the term you found like this:
```javascript
if(found) {
  angular.extend(availableTerm, found);   // will merge the objects, 2nd objects properties win.
}
```
This eliminates need to manually map the properties.

app/scripts/controllers/pod.js
benjaminapetersen: You need the `terminalTabWasSelected`variable, but I can't see the `tabSelect()` function being used in the view.

benjaminapetersen:
Per our conversation on reducing some loops ~
```javascript 
var makeTerminals = function(containerStatuses, existingTerms) {
  var firstRun = false;
  if(!existingTerms) {
    existingTerms = {};
    firstRun = true;
  }

  
  // keeping existing terminals, append new ones
  // set the first one active if this is the first time called
  _.each(containerStatuses, function(status, i) {
    // using a hash map eliminates some looping & can get passed to
    // view with a tweak to <ui-select choices="">
    if(!terminals[status.containerId]) {
      terminals[container.containerID] = {
        name: container.name,
        // on first run, if this is the first container, set it active
        isActive: (firstRun && (i === 0)) ? true : false,
        isConnected: (firstRun && (i === 0)) ? true : false
        // died: false  <-- not 100% sure we need died yet
      });
    } 
  });
  
  return terminals;
};


DataService.get('pod', function(pod) {
  // first pass on GET makes terminals
  $scope.containerTerminals = makeTerminals(pod.status.containerStatuses);
  DataService.watch('pod', function(pod) {
    // future watch updates maintains current state
    $scope.containerTerminals = makeTerminals(pod.status.containerStatuses, $scope.containerTerminals);  
  })
});



```

MDeMariaRH:
To Kill a Container: use 'docker rm -f'. with 'oc describe pod <podname>' to get container ID.

benjaminapetersen:
Did a quick look, but will wait to do any more till its back in working order.

app/scripts/controllers/pod.js
```diff
@@ -231,11 +133,23 @@ angular.module('openshiftConsole')
         }
       });
```
benjaminapetersen: Hmm. might be a little bit.  Doesn't look like you are using the `firstRun` flag anymore.

MDeMariaRH:
This feature is ready for official review and feedback. @benjaminapetersen 

======================== Major Changes ======================== 
This feature now uses a hashmap instead of a list to maintain its collection of kubernetes container terminals. All infrastructure surrounding the list structure has been retooled to work with a hashmap in both the controller and the view.

======================== Important Notes ========================
This implementation depends on my PR to the github/.com/kubernetes-ui/container-terminal/pull/#14. the KUI did not have a way to send a terminal's status out to the view. This PR addresses that.

======================== UI Screenshots ========================
Overview
image

In Use & Healthy vs. Unused
image

In Use & Healthy vs In Use & Unhealthy
image
image

benjaminapetersen:
- Mentioned trying font colors on the terminal icon rather than background colors to indicate happy/sad state.

@rhamilto @sg00dwin @jwforres @spadgett if anyone else also wants to comment on the screenshots.

spadgett:
Fixes https://github.com/openshift/origin/issues/5429

spadgett:
Depends on https://github.com/kubernetes-ui/container-terminal/pull/14

rhamilto:
Either solely background color or font color are problematic for colorblind users (which is estimated at about 10% of the male population).  So it's probably best to use additional cues to convey this information.  Are there other icons that could be explored?

benjaminapetersen:
Do you think text cues would be helpful:
```
- hello-openshift (connected)
- hello-openshift-fedora (disconnected) ?
```

MDeMariaRH:
Perhaps X for disconnected (http://fontawesome.io/icon/times-circle-o/) and check for healthy (http://fontawesome.io/icon/check-circle-o/)?

jwforres:
@MDeMariaRH was going to suggest that, either that pair or the times-circle and check-circle which are the filled in ones, whichever set looks better

MDeMariaRH:
I think the open icon pair fits better with the donut theme, as well as the question-mark in the top-right of the page. I'll apply coloring in addition to using those icons.

app/views/browse/pod.html
```diff
@@ -124,19 +124,60 @@
                     ng-init="containers = pod.status.containerStatuses"
                     ng-if="containersRunning(pod.status.containerStatuses) > 0">
                     <uib-tab-heading>Terminal</uib-tab-heading>
-                    <div>
+
+                    <div class="mar-bottom-md">
                       <span class="pficon pficon-info" aria-hidden="true"></span>
-                      When you navigate away from this pod, any processes running
-                      in these terminals will quit.
+                      Terminal names with <i class="fa fa-terminal btn-success"></i> are active, healthy terminal sessions to living containers.
```
jwforres: I'd just update the sentence here to something like "When you navigate away from this pod any open terminal connections will be closed.  This will kill any foreground processes you started from the terminal."

If you need to designate what the icons mean, do it in a simple legend underneath,  something like
check-icon - Connected terminal
times-icon - Disconnected terminal

could be inline if we need to save some vertical height.  Alternatively as Ben suggested, adding text after the container name would eliminate the need for explaining the icons.

app/views/browse/pod.html
```diff
@@ -124,19 +124,60 @@
...
-                        prevent="!terminalTabWasSelected">
-                      </kubernetes-container-terminal>
+
+                    <div>
+                      <ui-select
+                        ng-model="selected"
+                        on-select="onTerminalSelectChange(selected)"
+                        class="mar-left-none pad-left-none col-xs-3">
+                        <ui-select-match
+                          placeholder="Select or search">
```
jwforres: I would probably just say "Container name" then it explains what it is you are looking at.   "Select or search" isn't a very useful placeholder, plus its not really search, its typeahead.

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,58 @@ angular.module('openshiftConsole')
...
+      newTerm.value.isActive = true;
+      newTerm.value.isUsed = true;
+    };
+
+    $scope.tabSelect = function() {
+      $scope.terminalTabWasSelected = true;
+    };
+
+    var makeTerminals = function(containerStatuses, existingTerms) {
+      var firstRun = false;
```
jwforres: this whole block could be replaced by

```
var firstRun = !!existingTerms;
```

We always use the `!!` if we want to transform a non-boolean to a boolean.

jwforres: although you'll still need to initialize existingTerms.  thats another one-liner
```
var existingTerms = existingTerms || {};
```
because javascript unfortunately doesnt have an `||=` operator

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,58 @@ angular.module('openshiftConsole')
...
+
+      if(!existingTerms) {
+        existingTerms = {};
+        firstRun = true;
+      }
+
+      // keeping existing terminals, append new ones
+      // set the first one active if this is the first time called
+      _.each(containerStatuses, function(status, i) {
+        // using a hash map eliminates some looping & can get passed to
```
jwforres: un-needed comment

app/views/browse/pod.html
```diff
@@ -124,19 +124,60 @@
...
+                          class="col-xs-3"
+                          repeat="(key, term) in containerTerminals">
+                          <div>
+                            <span
+                              ng-class="{'btn-success' : (term.value.isUsed && (term.value.status === 'connected')),
+                                         'btn-danger' : (term.value.status === 'disconnected'),
+                                         'mar-left-lg' : !term.value.isUsed}">
+                                <i class="fa fa-terminal" ng-show="term.value.isUsed"></i>
+                            </span>
+                            <!--span class="mar-left-lg" ng-show="!term.value.isUsed && !term.value.died"></span>
```
jwforres: remove commented code

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,58 @@ angular.module('openshiftConsole')
...
+          };
+        }
+      });
+
+      var defaultSelected = {
+        key: _.findKey($scope.containerTerminals, { 'isActive': true }),
+        value: _.find($scope.containerTerminals, { 'isActive': true })
+      };
+
+      if(defaultSelected) {
```
benjaminapetersen: can eliminate the if statement here, since you defined the object on 130 and know for certain it exists.

app/views/browse/pod.html
```diff
@@ -124,19 +124,60 @@
...
+                            <span class="pficon-error-circle-o" ng-show="term.value.died"></span-->
+                            <span class="mar-left-sm" ng-bind-html="term.value.name | highlight: $select.search"></span>
+                          </div>
+                        </ui-select-choices>
+                      </ui-select>
+
+                      <br><br> <!-- Padding/margin didn't give me the line separation I wanted. Recommendations? -->
```
benjaminapetersen: Just for some spacing, right?  styles/_spacers.less defines some utillity classes that you can use, such as `.pad-bottom-md`

app/views/browse/pod.html
```diff
@@ -124,19 +124,60 @@
...
+
+                      <div ng-repeat="term in containerTerminals">
+                        <kubernetes-container-terminal
+                          prevent="!terminalTabWasSelected"
+                          ng-if="term.isUsed"
+                          ng-show="term.isActive"
+                          pod="pod"
+                          container="term.name"
+                          status="term.status"
```
benjaminapetersen: Curious, if the `kubernetes-container-terminal` now publishes status as `connected` & `disconnected` (pending your PR), if you can collapse this and the `isUsed` flag together.

benjaminapetersen: Actually nevermind, since `ng-if="term.isUsed` is stopping render on the terminal until the user selects it in the first place. No change needed.

MDeMariaRH:
@benjaminapetersen 

Remaining UI problems:

1. Input box width still extends to width of page. I've tried specifically adding an input tag and adjusting that, but it hasn't worked thus far. Advice welcome!
2. Scrollbar will appear after 8 entries. I think the fix is to change the number of bootstrap rows the select-choices is.

Sidenote: DBCS and bi-directional characters work properly.

benjaminapetersen:
@MDeMariaRH ~
- It extends the width of the page even if you have a bootstrap class on a div around it?  
- just looked at [the demo](http://angular-ui.github.io/ui-select/demo-basic.html), the scroller here seems completely fine.

MDeMariaRH:
@benjaminapetersen 

All styling changes are complete.

-Styling fixed on dropdown size fixed. I had bootstrap classes that were compiling on top of one-other.
-I went and found another scrollable selector and mine matches its scroll behavior. (Scrollbar appears after 8 entries.)

MarkDeMaria:
@benjaminapetersen @jwforres 

Styling Updates
-Icons replaced with their circle-less counterparts. Coloring is unchanged.
-Size of select bar fixed when screen is resized. No longer bugs and goes to 100% width.

Icons
image

Fullscreen Size
image

Half-Screen Size
image

Quarter-Screen Size
image

jwforres:
put connected and disconnected in ( ) otherwise it visually feels like part
of the name.   so   `container-name (connected)`

benjaminapetersen:
@MarkDeMaria quick Q, got an example with 3+ terminals? Would be interesting to see how it looks with a bunch of them, showing 1. available 2. connected 3. active 4. disconnected/failed.  Just to make sure that the drop down is still clear.

I imagine if there are 3-4 green check marks it could potentially get confusing.

liggitt:
to me, the checkmark implies completion... I think I'd prefer just the `(connected)` or `(disconnected)` suffix... not sure the icon buys us a lot

MarkDeMaria:
@liggitt I think the icons are still helpful because it clearly separates those that are actively being used from those that are not. However, I agree that the checkmarks aren't as clear as they could be.

Some alternatives to imply healthy connectivity could be:
http://fontawesome.io/icon/plug/
http://fontawesome.io/icon/wifi/
http://fontawesome.io/icon/signal/

MarkDeMaria:
@benjaminapetersen As request, a larger sample size:

image

benjaminapetersen:
Nice. I think a little extra padding before the parens would be good, and agree the icon might not be exactly right quite yet.

spadgett:
The red x makes it look like the container is crashing. I think I prefer just the connected and disconnected text.

sg00dwin:
How about using light gray (muted) text for disconnected. A couple of options playing with different alignment of (connected) / (disconnected) and also incorporating fa-plug for connected.

**1.**
terminal2
**2.**
terminal4
**3.**
terminal3
**4.**
terminal1

MarkDeMaria:
@sg00dwin I like the third mockup the best, however, the state of 'disconnected' is only given to terminals that were previously in-use, but have suffered a failure of some kind, such as a container crash or a connection interruption. There are three states a terminal can be in:

1. In-use and healthy (connected)
2. In-use and unhealthy (disconnected)
3. Not in-use.

Do you have a suggestion on how to reflect the third state while keeping your greyed-out text styling?

sg00dwin:
Could include a light gray fa-plug with disconnected state.
Updated options including (not in use) state

**1.2**
terminal1 2


**2.2**
terminal2 2


**3.2**
terminal3 2



**4.2**
terminal4 2

MarkDeMaria:
@benjaminapetersen @jwforres @spadgett @liggitt 

Could you offer feedback on @sg00dwin's designs?

Thanks!

spadgett:
The gray text looks like a disabled menu item. I think we should avoid that.

I prefer 2 just showing the text except I wouldn't make the container name gray.

jwforres:
agree with @spadgett's concern (and that option 2.2 is preferred from @sg00dwin 's options).  What if the container-name wasn't gray, but the (disconnected) text was gray?  then there is still a visual differentiator but it doesn't look like the item is disabled.

benjaminapetersen:
Agree with @jwforres last comment, just gray on the (disconnected) seems good!

MarkDeMaria:
Excellent! I will implement 2.2 asap.

MarkDeMaria:
@benjaminapetersen @spadgett @jwforres

Final Review

Selected Terminal
image

Search Bolding & Not In Use
image

Connected & disconnected
image

spadgett:
We can probably get rid of the parens around connected and disconnected since it's right-aligned now. I'd also suggest just showing nothing instead of "(not in use)"

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,49 @@ angular.module('openshiftConsole')
...
 
+    /**
+     *  Will set the newTerm's isActive/isUsed to true, while deactivating the old's
+     */
+    $scope.onTerminalSelectChange = function(newTerm) {
+      // Deactivate all terminals
+      _.each($scope.containerTerminals, function(term) {
+        term.isActive = false;
+      });
+
```
benjaminapetersen: quick reminder, what was it that caused `newTerm` to actually be `newTerm.value` here (and can we squish `.value`)?

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,49 @@ angular.module('openshiftConsole')
...
+    };
+
+    $scope.tabSelect = function() {
+      $scope.terminalTabWasSelected = true;
+    };
+
+    var makeTerminals = function(containerStatuses, existingTerms) {
+      existingTerms = existingTerms || {};
+
+      var firstRun = !!existingTerms;
```
benjaminapetersen: I think line 110 has to be before 108, else you already set `existingTerms` to at least an empty object (so will always be true).

app/views/browse/pod.html
```diff
@@ -124,19 +124,61 @@
                     ng-init="containers = pod.status.containerStatuses"
                     ng-if="containersRunning(pod.status.containerStatuses) > 0">
                     <uib-tab-heading>Terminal</uib-tab-heading>
-                    <div>
+
+                    <div class="mar-bottom-md">
                       <span class="pficon pficon-info" aria-hidden="true"></span>
-                      When you navigate away from this pod, any processes running
-                      in these terminals will quit.
+                      When you navigate away from this pod any open terminal connections will be closed.
```
spadgett: grammar nit: add back the comma after pod here

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,49 @@ angular.module('openshiftConsole')
...
+          existingTerms[status.name] = {
+            name: status.name,
+            // on first run, if this is the first container, set it active
+            isActive: (firstRun && (i === 0)),
+            isUsed: (firstRun && (i === 0))
+          };
+        }
+      });
+
+      var defaultSelected = {
```
benjaminapetersen: nice use of lodash, bet this `key,value` is related to my Q above.

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,49 @@ angular.module('openshiftConsole')
       }
     };
 
+    /**
+     *  Will set the newTerm's isActive/isUsed to true, while deactivating the old's
+     */
+    $scope.onTerminalSelectChange = function(newTerm) {
+      // Deactivate all terminals
```
benjaminapetersen: Prob should update this comment, I don't think this is actually deactivating all the terminals every time you click a new option in the drop down (implies closing the terminal).

app/views/browse/pod.html
```diff
@@ -124,19 +124,61 @@
...
-                        prevent="!terminalTabWasSelected">
-                      </kubernetes-container-terminal>
+
+                    <div class="mar-bottom-lg">
+                      <div class="pad-left-none col-xs-4 col-lg-3">
+                        <ui-select
+                          ng-model="selected"
+                          on-select="onTerminalSelectChange(selected)"
+                          class="mar-left-none pad-left-none pad-right-none"
+                          style="width: 350px;"> <!-- UI select bug https://github.com/angular-ui/ui-select/issues/1678 -->
```
spadgett: I don't understand how the width style is related to the linked bug. Wasn't that fixed in the other PR?

I'd suggest making a terminal-select class and adding all the padding and width styles there instead instead of the mar-* and pad-* styles.

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,49 @@ angular.module('openshiftConsole')
...
+          };
+        }
+      });
+
+      var defaultSelected = {
+        key: _.findKey($scope.containerTerminals, { 'isActive': true }),
+        value: _.find($scope.containerTerminals, { 'isActive': true })
+      };
+
+      $scope.selected = defaultSelected;
```
spadgett: `selected` is too generic. `$scope.selectedTerminalContainer` maybe?

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,49 @@ angular.module('openshiftConsole')
...
+
+    var makeTerminals = function(containerStatuses, existingTerms) {
+      existingTerms = existingTerms || {};
+
+      var firstRun = !!existingTerms;
+
+      _.each(containerStatuses, function(status, i) {
+        if(!existingTerms[status.name]) {
+          existingTerms[status.name] = {
+            name: status.name,
```
spadgett: Maybe

```js
containerName: status.name
```

?

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,49 @@ angular.module('openshiftConsole')
...
+    $scope.tabSelect = function() {
+      $scope.terminalTabWasSelected = true;
+    };
+
+    var makeTerminals = function(containerStatuses, existingTerms) {
+      existingTerms = existingTerms || {};
+
+      var firstRun = !!existingTerms;
+
+      _.each(containerStatuses, function(status, i) {
```
spadgett: Disable the option for containers that aren't running, `!status.state.running`?

https://github.com/angular-ui/ui-select/wiki/ui-select-choices#example-disable-choices

app/scripts/controllers/pod.js
spadgett: I'm not sure what `isActive` means. Isn't this just `isVisible`? We still keep the terminal session alive, right?

spadgett:
@MarkDeMaria will this preserve the order in the dropdown that the containers appear in the pod template array?

app/scripts/controllers/pod.js
jwforres: We do but the connection could die from the server side or somewhere in
between, and that terminal may not be visible when it happens.

spadgett: But the code below sets `isActive` to false for everything except the last selected.

MarkDeMaria:
Depends on #130

app/views/browse/pod.html
MarkDeMaria: @spadgett I put that in prior to rebasing my branch with the PR that fixed this. Fixed!

app/scripts/controllers/pod.js
MarkDeMaria: @benjaminapetersen onTerminalSelectChange receives {key, value} from the view, hence the need to use .value to get at the actual data.

MarkDeMaria:
@spadgett @benjaminapetersen @jwforres

Updated screenshot for connected/disconnected/not in use. As per Sam's request, no parentheses and no 'not in use.' If a terminal isn't used it will not be given a state designation.
image

Also, @spadgett
1) The order is not preserved.
2) For your comment about not selecting a container that isn't running, technically a container that has died is no longer running, and the user would not be able to get back to that terminal window if that were the case. All they'd see is: **'some Container name                  disconnected'** I think being able to select containers that aren't running is needed for this use case, unless I misunderstand your proposal.

spadgett:
For (1), we preserve the order for the container selector in the metrics and logs tab, so I think we should be consistent here.

For (2), I'm OK with that.

spadgett:
Screenshot looks good to me, thanks

MarkDeMaria:
As per my conversation with @spadgett @jwforres @benjaminapetersen 

Current container statuses:
  Waiting
  running
  terminated

Current terminals statuses
  connected
  disconnected
  not in use (don't show anything)

TODO:
!) terminal statuses - keep the way they are

2) Container statuses
  running don't say anything
  Waiting/terminated - declare that in same place as connected/disconnected

  -If you switch to a waiting container, 
    - show the terminal & imply will be ready soon...
  -Disable terminated container options, you can't pick this.
    - don't show terminal window

MarkDeMaria:
@spadgett @jwforres @benjaminapetersen

Terminal Connected vs. Terminal Not In Use (Blank) vs. Container Waiting
image

**Terminal connected and container running, status = connected**
image

**Terminal disconnected and container running, status = disconnected**
image

**Terminal connected or disconnected and container waiting or terminated: Status = waiting or terminated**
[Terminated]
image
[Waiting]
image

**Terminal not in use: Status = Nothing (Blank)**
[Terminal not in use] 
image
[Terminal not in use and container is waiting]
image

benjaminapetersen:
- I like that 'waiting' = you can't select yet.  
- Terminated...?
- The case "disconnected but container still running" = the websocket died for some reason, right? `<kubernetes-container-terminal status="status"` gives a status of `disconnected`. I dont know that there is a way to tell the directive to reconnect, and I'm assuming the websocket crash probably kills running processes.

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,67 @@ angular.module('openshiftConsole')
...
+
+    $scope.tabSelect = function() {
+      $scope.terminalTabWasSelected = true;
+    };
+
+    var makeTerminals = function(existingTerminals) {
+      existingTerminals = existingTerminals || [];
+
+      // Remake list with old list's isVisible/isUsed data
+      _.each($scope.pod.spec.containers, function(spec) {
```
benjaminapetersen: prob should be ` _.each($scope.pod.spec.containers, function(container)`, since there is one spec that contains multiple containers.

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,67 @@ angular.module('openshiftConsole')
...
+        var terminal = _.find(existingTerminals, function(term) {
+          return term.containerName === spec.name;
+        });
+
+        if (!terminal) {
+          existingTerminals.push({
```
benjaminapetersen: Perhaps do the `_.find` once, then reuse the object:
```javascript
var found = _.find($scope.containerTerminals, function(term) {
     return term.containerName === spec.name;
};
 existingTerminals.push({
   isVisible: _.get(found, 'isVisible'
   // etc.
});
```
or, perhhaps a `_.merge({}, {})` kind of thing would work here so you dont have to repeat the key names `isVisible, isUsed, etc.` but lemme look a bit more.

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,67 @@ angular.module('openshiftConsole')
...
+            return terminal.containerName === status.name;
+          }), 'state');
+
+          terminal.containerStatus = Object.getOwnPropertyNames(containerStatus)[0];
+        }
+      });
+
+      var currentlyVisible = _.find($scope.containerTerminals, { 'isVisible': true });
+
+      if (currentlyVisible === undefined) {
```
benjaminapetersen: `if (!currentlyVisible) { }` preferred to `=== undefined`

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,67 @@ angular.module('openshiftConsole')
...
+        existingTerminals[0].isVisible = true;
```
benjaminapetersen: I might suggest: 
```javascript
_.merge(
  _.first(existingTerminals),
  {
    isVisible: true,
    isUsed: true
  })
```

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,67 @@ angular.module('openshiftConsole')
...
+            }), 'isVisible'),
+            isUsed: _.get(_.find($scope.containerTerminals, function(term) {
+              return term.containerName === spec.name;
+            }), 'isUsed'),
+            containerStatus: _.get(_.find($scope.pod.status.containerStatuses, function(term) {
+              return term.containerName === spec.name;
+            }), 'state')
+          });
+        } else {
+          var containerStatus = _.get(_.find($scope.pod.status.containerStatuses, function(status) {
```
benjaminapetersen: Fwiw, shouldn't create a `var` statement inside an `if/else` (in JS).  You might be able to just inline this: `terminal.containerStatus = _.keys(_.get(_.find( /* */ )))`

app/scripts/controllers/pod.js
spadgett: @benjaminapetersen I find deeply nested calls like this hard to follow :/

app/scripts/controllers/pod.js
spadgett: +1, except I'd give it a name other than found.

```js
var term = _.find($scope.containerTerminals, { containerName: spec.name });
```

app/scripts/controllers/pod.js
spadgett: I'd add a helper function somewhere.

```js
var getState = function(containerStatus) {
  var state = _.get(containerStatus, 'state', {});
  return _.head(_.keys(state));
};
```

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,67 @@ angular.module('openshiftConsole')
...
+    };
+
+    var makeTerminals = function(existingTerminals) {
+      existingTerminals = existingTerminals || [];
+
+      // Remake list with old list's isVisible/isUsed data
+      _.each($scope.pod.spec.containers, function(spec) {
+        var terminal = _.find(existingTerminals, function(term) {
+          return term.containerName === spec.name;
+        });
```
spadgett: Suggest the shorthand.

```js
var terminal = _.find(existingTerminals, { containerName: spec.name });
```

app/scripts/controllers/pod.js
spadgett: I'd prefer simply

```js
if (!currentlyVisible) {
  currentlyVisible = _.head(existingTerminals);
  currentlyVisible.isVisible = true;
  currentlyVisible.isUsed = true;
}
```

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,67 @@ angular.module('openshiftConsole')
...
+        } else {
+          var containerStatus = _.get(_.find($scope.pod.status.containerStatuses, function(status) {
+            return terminal.containerName === status.name;
+          }), 'state');
+
+          terminal.containerStatus = Object.getOwnPropertyNames(containerStatus)[0];
+        }
+      });
+
+      var currentlyVisible = _.find($scope.containerTerminals, { 'isVisible': true });
```
spadgett: Nit: you don't need the single quotes around `isVisible` here.

app/scripts/controllers/pod.js
```diff
@@ -117,6 +180,9 @@ angular.module('openshiftConsole')
               $scope.pod = pod;
               setLogVars(pod);
               setContainerVars();
+
+              $scope.containerTerminals = makeTerminals($scope.containerTerminals);
```
spadgett: You should only need to call `makeTerminals` once in the list and just update the container statuses here. The list of containers in the pod won't change. It would simplify `makeTerminals` as well since you don't have to check if the terminal already was created.

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,67 @@ angular.module('openshiftConsole')
...
+        if (!terminal) {
+          existingTerminals.push({
+            containerName: spec.name,
+            isVisible: _.get(_.find($scope.containerTerminals, function(term) {
+              return term.containerName === spec.name;
+            }), 'isVisible'),
+            isUsed: _.get(_.find($scope.containerTerminals, function(term) {
+              return term.containerName === spec.name;
+            }), 'isUsed'),
+            containerStatus: _.get(_.find($scope.pod.status.containerStatuses, function(term) {
```
spadgett: `containerState`

app/views/browse/pod.html
```diff
@@ -119,24 +119,68 @@
...
+                                <span ng-bind-html="term.containerName | highlight: $select.search">
+                                </span>
+                                <span flex></span>
+                                <span
+                                  ng-if="term.isUsed && term.containerStatus === 'running'"
+                                  ng-class="{'text-muted' : (term.status === 'disconnected')}">
+                                    {{term.status}}
+                                </span>
+                                <span
+                                  ng-if="term.isUsed && term.containerStatus !== 'running'"
```
spadgett: Suggest showing waiting or terminated state here even if you haven't picked the terminal (but not running).

app/scripts/controllers/pod.js
benjaminapetersen: works for me!

MarkDeMaria:
The rules are mostly the same:
-Selections are disabled if the terminal has not yet been used and the container state is 'waiting' || 'terminated'
-'connected' || 'disconnected' is reported if the containerState is 'running,' and this text will be muted if the terminal is 'disconnected'
-'waiting' || 'terminated' is reported if the containerState is not 'running'. The text will be muted if the terminal has not yet been used
-No status is reported if the terminal has not yet been selected and the container is running

Major cases below:

**Terminal connected and container running ('connected')** vs
**terminal not connected and container waiting ('waiting', not selectable)** vs
**terminal not connected and container running (blank, no status)**
image

**Terminal previously selected, container terminated & restarted, terminal now disconnected ('disconnected', selectable)**
image

**Terminal not connected, container terminated ('terminated', not selectable). If container was waiting, would report 'waiting'**
image

**Terminal connected & container waiting ('waiting', selectable). If container was terminated, then would report 'terminated'**
image

benjaminapetersen:
Nice. Controller logic seems nice/clean/easy to follow.  +1.
I think your explanation of connected/waiting/disconnected/terminated makes sense.

app/views/browse/pod.html
```diff
@@ -119,24 +119,68 @@
...
+                                </span>
+                                <span flex></span>
+                                <span
+                                  ng-if="term.isUsed && term.containerState === 'running'"
+                                  ng-class="{'text-muted' : (term.status === 'disconnected')}">
+                                    {{term.status}}
+                                </span>
+                                <span
+                                  ng-if="term.containerState !== 'running'"
+                                  ng-class="{'text-muted' : term.isUsed === false}">
```
spadgett: ng-class="{'text-muted' : !term.isUsed}"

app/views/browse/pod.html
```diff
@@ -119,24 +119,68 @@
...
+                           <ui-select-match
+                              class="truncate"
+                              placeholder="Container Name">
+                                <span class="pad-left-md">
+                                  {{selectedTerminalContainer.containerName}}
+                                </span>
+                            </ui-select-match>
+                            <ui-select-choices
+                              repeat="term in containerTerminals | filter: $select.search"
+                              ui-disable-choice="(term.containerState === 'waiting' || term.containerState === 'terminated') && term.isUsed === false">
```
spadgett: ui-disable-choice="term.containerState !== 'running' && !term.isUsed"

app/views/browse/pod.html
```diff
@@ -119,24 +119,68 @@
...
+                                  ng-if="term.containerState !== 'running'"
+                                  ng-class="{'text-muted' : term.isUsed === false}">
+                                    {{term.containerState}}
+                                </span>
+                              </div>
+                            </ui-select-choices>
+                          </ui-select>
+                        </div>
+
+                        <br><br>
```
spadgett: Better to use CSS margins

bower.json
```diff
@@ -54,11 +54,12 @@
     "angular-bootstrap": "0.14.3",
     "moment": "2.10.6",
     "matchHeight": "0.6.0",
-    "font-awesome": "4.6.1"
+    "font-awesome": "4.6.1",
+    "patternfly": "3.5.1"
```
spadgett: You have patternfly in the list twice now.

app/scripts/controllers/pod.js
```diff
@@ -87,6 +87,57 @@ angular.module('openshiftConsole')
...
+      newTerm.isUsed = true;
+    };
+
+    var getState = function(containerStatus) {
+      var state = _.get(containerStatus, 'state', {});
+      return _.head(_.keys(state));
+    };
+
+    var makeTerminals = function() {
+      var existingTerminals = [];
```
spadgett: I'd just name this "terminals"

spadgett:
Just a few small comments, otherwise looks good. Thanks @MarkDeMaria

MarkDeMaria:
@spadgett Changes made.

spadgett:
@MarkDeMaria did you push the updates?

spadgett:
[merge]

PR #67, opened 2016-06-15
Simplify some overview logic

spadgett:
Move handling of the `pipeline.alpha.openshift.io/uses` annotation into BuildsService and clean up the logic for grouping builds in OverviewController. Also remove `$scope.pipelinesByDeployment`, which was no longer used.

@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -13,6 +13,7 @@ angular.module('openshiftConsole')
                         $routeParams,
                         $scope,
                         AlertMessageService,
+                        BuildsService,
                         DataService,
                         DeploymentsService,
                         Logger,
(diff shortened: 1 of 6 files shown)


app/scripts/controllers/overview.js
```diff
@@ -211,81 +212,55 @@ angular.module('openshiftConsole')
       });
     };
 
-    var isRecentBuild = $filter('isRecentBuild');
+    var groupBuildByOutputImage = function(build) {
+      var buildOutputImage = imageObjectRef(build.spec.output.to, build.metadata.namespace);
```
jwforres: output is optional on a build spec, need to handle when its null

app/scripts/services/builds.js
```diff
@@ -115,5 +115,39 @@ angular.module("openshiftConsole")
       return true;
     };
 
+    var annotation = $filter('annotation');
+    BuildsService.prototype.usesDeploymentConfigs = function(buildConfig) {
+      var uses = annotation(buildConfig, 'pipeline.alpha.openshift.io/uses') || '';
```
jwforres: why are you doing the `|| ''`  if you are returning empty array when `!uses`

jwforres:
just the couple minor things, otherwise LGTM, you can merge once updated

spadgett:
Thanks @jwforres, made those updates.

[merge]

PR #71, opened 2016-06-17
Improve overview loading / get started check

spadgett:
Fixes a problem where the get started message can flicker briefly if the first type that loads is empty.

@jwforres PTAL


jwforres:
[merge]

PR #72, opened 2016-06-17
Order services by importance, collapse some when more than two

spadgett:
@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -177,7 +177,7 @@ angular.module('openshiftConsole')
 
     // Set of child services in this project.
     var childServices = {};
-    $scope.isChildService = function(service) {
+    var isChildService = function(service) {
       return !!childServices[service.metadata.name];
     };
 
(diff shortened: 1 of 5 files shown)


app/scripts/controllers/overview.js
```diff
@@ -188,7 +188,48 @@ angular.module('openshiftConsole')
...
+      if (!_.isEmpty(routes)) {
+        score += 5;
+      }
+
+      if (_.has(service, 'metadata.labels.app')) {
+        score += 3;
+      }
+
+      if (!_.isEmpty(childServices)) {
+        score += 1;
```
jwforres: I can't decide on this one... this could cause an entire app block to jump to the top of the page after using the 'Link Service'

app/scripts/controllers/overview.js
```diff
@@ -188,7 +188,48 @@ angular.module('openshiftConsole')
...
+      }
+
+      if (ServicesService.isInfrastructure(service)) {
+        score -= 5;
```
jwforres: should this be -9 ?  would I want infrastructure stuff to appear above a service that didn't have a route?

@smarterclayton opinion?

jwforres: Sorry just remembered that wouldn't happen since those get tiled separately.  So maybe the question here is really should infrastructure services be getting rendered by a separate ng-repeat, just like we do with standalone services and unserviced stuff

spadgett: Yeah, you would need 2 extra ng-repeats, one for infrastracuture services with app label or routes and another for services without, which would make 4 total. I was hesitant to repeat the logic.

app/scripts/controllers/overview.js
spadgett: I'll take it out.

app/scripts/controllers/overview.js
jwforres: i think this is fine for now for this pr but we may want to look at moving
them out later.  we need to decide how important infra stuff is relative to
other edge case stuff.

spadgett:
Updated.

jwforres:
[merge]

PR #82, opened 2016-06-20
What can user do

jhadvig:
Created a new `AuthorizationService` to determine actions that user can/can't do.
The new service consists from two main methods:
- `reviewUserRules` 
 - which serves for reviewing multiple resource/verb pairs (namespaced and also non-namespaced)
 - resource/verb pair combinations are listed in the `$scope.canI` variable which should be set in each controller which will call this method.
 - obtained rules will be cached, until refreshing the page,changing projects, when 3 minute interval passes
- `canI`
 - lightweight version of the first method (response only takes info if user is allowed to perform action on the ressource)
 - servers for reviewing single resource/verb pair (namespaced and also non-namespaced)
 - does not need `$scope.canI` variable, will create one if not present in controller and adds the  result of the resource/verb review to it.

@jwforres PTAL

Implements https://trello.com/c/UQEUgVrY

### DIFF
--- a/app/scripts/controllers/otherResources.js
+++ b/app/scripts/controllers/otherResources.js
@@ -8,7 +8,7 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('OtherResourcesController', function ($routeParams, $scope, AlertMessageService, DataService, ProjectsService, $filter, LabelFilter, Logger, APIService ) {
+  .controller('OtherResourcesController', function ($routeParams, $scope, AlertMessageService, AuthorizationService, DataService, ProjectsService, $filter, LabelFilter, Logger, APIService ) {
     $scope.projectName = $routeParams.project;
     $scope.labelSuggestions = {};
     $scope.alerts = $scope.alerts || {};
@@ -48,6 +48,10 @@ angular.module('openshiftConsole')
     ProjectsService
       .get($routeParams.project)
       .then(_.spread(function(project, context) {
+        $scope.kinds = _.filter($scope.kinds, function(kind){
+          var resource = APIService.kindToResource(kind.kind);
+          return AuthorizationService.canI("list", kind.group ? kind.group + "/" + resource : resource, $scope.projectName);
+        });
         $scope.project = project;
         $scope.context = context;
         $scope.kindSelector.disabled = false;
@@ -70,6 +74,7 @@ angular.module('openshiftConsole')
       if (!selected) {
         return;
       }
+      $scope.selectedResource = (selected.group) ? selected.group + "/" + APIService.kindToResource(selected.kind) : APIService.kindToResource(selected.kind);
       // TODO - We can't watch because some of these resources do not support it (roles and rolebindings)
       DataService.list({
           group: selected.group,
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -126,6 +126,44 @@ angular.module('openshiftConsole')
       return displayName;
     };
   })
+  .filter('canI', function(AuthorizationService) {
+    return function(resource, verb, namespace) {
+      return AuthorizationService.canI(verb, resource, namespace);
+    };
+  })
+  .filter('canIAddToProject', function(AuthorizationService) {
+    return function(namespace) {
+      return AuthorizationService.canIAddToProject(namespace);
+    };
+  })
+  .filter('canIDoAny', function(canIFilter) {
+    var resourceRulesMap = {
+      "buildConfigs": {"buildconfigs": ["delete", "update"], "buildconfigs/instantiate": ["create"]},
+      "builds": {"builds/clone": ["create"], "builds": ["delete", "update"]},
+      "deploymentConfigs": {"extensions/horizontalpodautoscalers": ["create", "update"], "deploymentconfigs": ["create", "update"]},
+      "deployments": {"replicationcontrollers": ["update", "delete"]},
+      "horizontalPodAutoscalers": {"extensions/horizontalpodautoscalers": ["update", "delete"]},
+      "imageStreams": {"imagestreams": ["update", "delete"]},
+      "persistentVolumeClaims": {"persistentvolumeclaims": ["update", "delete"]},
+      "pods": {"pods": ["update", "delete"], "deploymentconfigs": ["update"]},
+      "replicationControllers": {"horizontalpodautoscalers": ["create", "update"], "replicationcontrollers": ["create", "update"]},
+      "routes": {"routes": ["update", "delete"]},
+      "services": {"services": ["update", "create", "delete"]},
+      "projects": {'projects': ['delete', 'update']}
+    };
+    return function(resource) {
+      return _.some(resourceRulesMap[resource], function(verbs, resource) {
+        return _.some(verbs, function(verb) {
+          return canIFilter(resource,verb);
+        });
+      });
+    };
+  })
+  .filter('canIScale', function(canIFilter, isDeploymentFilter) {
+    return function(deployment) {
+      return canIFilter(isDeploymentFilter(deployment) ? "deploymentconfigs/scale" : "replicationcontrollers", "update");
+    };
+  })
   .filter('tags', function(annotationFilter) {
     return function(resource, /* optional */ annotationKey) {
       annotationKey = annotationKey || "tags";
--- a/app/scripts/services/authorization.js
+++ b/app/scripts/services/authorization.js
@@ -1,0 +1,114 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .factory("AuthorizationService", function($q, $cacheFactory, Logger, $interval, DataService){
+    
+    var currentProject = null;
+    var cachedRulesByProject = $cacheFactory('rulesCache', {
+          number: 10
+        });
+
+    var REVIEW_RESOURCES = ["localresourceaccessreviews", "localsubjectaccessreviews", "resourceaccessreviews", "selfsubjectrulesreviews", "subjectaccessreviews"];
+
+    // Transform data from:
+    // rules = {resources: ["jobs"], apiGroups: ["extensions"], verbs:["create","delete","get","list","update"]}
+    // into:
+    // normalizedRules = {"extensions/jobs": ["create","delete","get","list","update"]}
+    var normalizeRules = function(rules) {
+      var normalizedRules = {};
+
+      _.each(rules, function(rule) {
+        _.each(rule.resources, function(resource) {
+          _.each(rule.apiGroups, function(apiGroup) {
+            normalizedRules[(apiGroup === "") ? resource : apiGroup + "/" + resource] = rule.verbs;
+          });
+        });
+      });
+      return normalizedRules;
+    };
+
+    // Check if resource name meets one of following conditions, since those resources can't be create/update via `Add to project` page:
+    //  - 'projectrequests'
+    //  - subresource that contains '/', eg: 'builds/source', 'builds/logs', ...
+    //  - resource is in REVIEW_RESOURCES list 
+    var checkResource = function(resource) {
+      if (resource === "projectrequests" || _.contains(resource, "/") || _.contains(REVIEW_RESOURCES, resource)) {
+        return false;
+      } else {
+        return true;
+      }
+    };
+
+    // Check if user can create/update any resource on the 'Add to project' so the button will be displayed.
+    var canAddToProject = function(rules) {
+      return _.some(rules, function(rule) {
+        return _.some(rule.resources, function(resource) {
+          return checkResource(resource) && !_.isEmpty(_.intersection(rule.verbs ,(["*", "create", "update"])));
+        });
+      });
+    };
+
+    var getProjectRules = function(projectName) {
+      var deferred = $q.defer();
+      currentProject = projectName;
+      var projectRules = cachedRulesByProject.get(projectName);
+      if (!projectRules || projectRules.forceRefresh) {
+        Logger.log("AuthorizationService, loading user rules for " + projectName + " project");
+        var object = {kind: "SelfSubjectRulesReview",
+                      apiVersion: "v1"
+                    };
+        DataService.create('selfsubjectrulesreviews', null, object, {namespace: projectName}).then(
+          function(data) {
+            var normalizedData = normalizeRules(data.status.rules);
+            var canUserAddToProject = canAddToProject(data.status.rules);
+            cachedRulesByProject.put(projectName, {rules: normalizedData,
+                                                        canAddToProject: canUserAddToProject,
+                                                        forceRefresh: false,
+                                                        cacheTimestamp: _.now()
+                                                      });
+            deferred.resolve();
+          }, function() {
+            deferred.reject();
+        });
+      } else {
+        // Using cached data.
+        Logger.log("AuthorizationService, using cached rules for " + projectName + " project");
+        if ((_.now() - projectRules.cacheTimestamp) >= 600000) {
+          projectRules.forceRefresh = true;
+        }
+        deferred.resolve();
+      }
+      return deferred.promise;
+    };
+
+    var getRulesForProject = function(projectName) {
+      return _.get(cachedRulesByProject.get(projectName || currentProject), ['rules']);
+    };
+
+    var canI = function(verb, resource, projectName) {
+      projectName = projectName || currentProject;
+      var rules = getRulesForProject(projectName);
+      if (rules) {
+        if (rules[resource]) {
+          return _.contains(rules[resource], verb) || _.contains(rules[resource], '*');
+        } else if (rules['*']) {
+          return _.contains(rules['*'], verb) || _.contains(rules['*'], '*');
+        } else {
+          return false;
+        }    
+      } else {
+        return false;
+      }
+    };
+
+    var canIAddToProject = function(projectName) {
+      return !!_.get(cachedRulesByProject.get(projectName || currentProject), ['canAddToProject']);
+    };
+
+    return {
+      getProjectRules: getProjectRules,
+      canI: canI,
+      canIAddToProject: canIAddToProject,
+      getRulesForProject: getRulesForProject
+    };
+  });
--- a/app/scripts/services/projects.js
+++ b/app/scripts/services/projects.js
@@ -2,7 +2,7 @@
 
 angular.module('openshiftConsole')
   .factory('ProjectsService',
-    function($location, $q, $routeParams, AuthService, DataService, annotationNameFilter) {
+    function($location, $q, $routeParams, AuthService, DataService, annotationNameFilter, AuthorizationService) {
 
 
       var cleanEditableAnnotations = function(resource) {
@@ -32,10 +32,20 @@ angular.module('openshiftConsole')
                       return DataService
                               .get('projects', projectName, context, {errorNotification: false})
                               .then(function(project) {
-                                context.project = project;
-                                context.projectPromise.resolve(project);
-                                // TODO: fix need to return context & projectPromise
-                                return [project, context];
+                                return AuthorizationService.getProjectRules(projectName)
+                                        .then(function() {
+                                          context.project = project;
+                                          context.projectPromise.resolve(project);
+                                          // TODO: fix need to return context & projectPromise
+                                          return [project, context];
+                                        }, function() {
+                                          $location.url(URI('error')
+                                                    .query({
+                                                      "error" : 'error',
+                                                      "error_description": 'User permissions for project ' + projectName + ' could not be loaded.'
+                                                    })
+                                                    .toString());
+                                        });
                               }, function(e) {
                                 context.projectPromise.reject(e);
                                 var description = 'The project could not be loaded.';
--- a/app/views/browse/_deployment-details.html
+++ b/app/views/browse/_deployment-details.html
@@ -16,7 +16,7 @@
         <status-icon status="deployment | deploymentStatus"></status-icon>
         {{deployment | deploymentStatus}}
         <span style="margin-left: 7px;">
-          <button ng-show="!rollBackCollapsed && (deployment | deploymentStatus) == 'Complete' && !(deployment | deploymentIsLatest : deploymentConfig) && !deployment.metadata.deletionTimestamp" ng-disabled="(deploymentConfigDeploymentsInProgress[deploymentConfigName] | hashSize) > 0" type="button" class="btn btn-default btn-xs" ng-click="rollBackCollapsed = !rollBackCollapsed">Roll Back</button>
+          <button ng-if="'deploymentconfigrollbacks' | canI : 'create'" ng-show="!rollBackCollapsed && (deployment | deploymentStatus) == 'Complete' && !(deployment | deploymentIsLatest : deploymentConfig) && !deployment.metadata.deletionTimestamp" ng-disabled="(deploymentConfigDeploymentsInProgress[deploymentConfigName] | hashSize) > 0" type="button" class="btn btn-default btn-xs" ng-click="rollBackCollapsed = !rollBackCollapsed">Roll Back</button>
           <div ng-show="rollBackCollapsed" class="well well-sm">
             Use the following settings from {{deployment.metadata.name}} when rolling back:
             <div class="checkbox">
@@ -39,7 +39,7 @@
           <!-- disabled until we have an api endpoint for retry
           <button ng-show="(deployment | deploymentStatus) == 'Failed'" type="button" ng-click="retryFailedDeployment(deployment)" ng-disabled="(deploymentConfigDeploymentsInProgress[deploymentConfigName] | hashSize) > 0" class="btn btn-default btn-xs" ng-if="(deployment | deploymentIsLatest : deploymentConfig) && !deployment.metadata.deletionTimestamp">Retry</button>
           -->
-          <button ng-show="(deployment | deploymentIsInProgress) && !deployment.metadata.deletionTimestamp" type="button" ng-click="cancelRunningDeployment(deployment)" class="btn btn-default btn-xs">Cancel</button>
+          <button ng-if="'replicationcontrollers' | canI : 'update'" ng-show="(deployment | deploymentIsInProgress) && !deployment.metadata.deletionTimestamp" type="button" ng-click="cancelRunningDeployment(deployment)" class="btn btn-default btn-xs">Cancel</button>
         </span>
       </dd>
       <dt ng-if-start="deployment | isDeployment">Deployment config:</dt>
@@ -69,7 +69,8 @@
         <replicas status="deployment.status.replicas"
                   spec="deployment.spec.replicas"
                   disable-scaling="!isScalable()"
-                  scale-fn="scale(replicas)">
+                  scale-fn="scale(replicas)"
+                  deployment="deployment">
         </replicas>
         <span ng-if="autoscalers.length">(autoscaled)</span>
       </dd>
@@ -86,11 +87,12 @@
         images-by-docker-reference="imagesByDockerReference"
         builds="builds"
         detailed="true"
-        add-health-check-url="{{(!deploymentConfig || isActive) ? healthCheckURL : ''}}">
+        add-health-check-url="{{((!deploymentConfig || isActive) && ('deploymentconfigs' | canI : 'update')) ? healthCheckURL : ''}}">
       </pod-template>
       <h4>Volumes</h4>
       <p ng-if="!deployment.spec.template.spec.volumes.length">
-        <a ng-href="project/{{project.metadata.name}}/attach-pvc?deployment={{deployment.metadata.name}}">Attach storage</a>
+        <a ng-if="'replicationcontrollers' | canI : 'update'" ng-href="project/{{project.metadata.name}}/attach-pvc?deployment={{deployment.metadata.name}}">Attach storage</a>
+        <span ng-if="!('replicationcontrollers' | canI : 'update')">none</span>
       </p>
       <volumes volumes="deployment.spec.template.spec.volumes" namespace="project.metadata.name"></volumes>
     </div>
@@ -111,12 +113,12 @@
     <span ng-if="warning.reason === 'NoCPURequest'">
       <!-- Prompt to set CPU request on the DC if this is a deployment. -->
       <a ng-href="project/{{projectName}}/set-limits?dcName={{deploymentConfigName}}"
-         ng-if="deploymentConfigName && !deploymentConfigMissing"
+         ng-if="deploymentConfigName && !deploymentConfigMissing && ('deploymentconfigs' | canI : 'update')"
          role="button">Set resource
         <span ng-if="!('cpu' | isRequestCalculated : project)">requests and</span> limits</a>
       <!-- Prompt to set CPU request on the RC if not a deployment. -->
       <a ng-href="project/{{projectName}}/set-limits?rcName={{deployment.metadata.name}}"
-         ng-if="!deploymentConfigName"
+         ng-if="!deploymentConfigName && ('replicationcontrollers' | canI : 'update')"
          role="button">Set resource
         <span ng-if="!('cpu' | isRequestCalculated : project)">requests and</span> limits</a>
     </span>
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -16,10 +16,11 @@
                     data-toggle="tooltip"
                     data-original-title="Building from build configuration {{buildConfig.metadata.name}} has been paused.">
               </span>
-              <div class="pull-right dropdown" ng-if="buildConfig">
+              <div class="pull-right dropdown" ng-if="buildConfig" ng-hide="!('buildConfigs' | canIDoAny)">
                 <!-- Primary Actions -->
                 <button
                     class="btn btn-default hidden-xs"
+                    ng-if="'buildconfigs/instantiate' | canI : 'create'"
                     ng-click="startBuild()">
                   Start Build
                 </button>
@@ -33,22 +34,22 @@
                    class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
                    data-toggle="dropdown"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
                 <ul class="dropdown-menu actions action-button">
-                  <li class="visible-xs-inline">
+                  <li class="visible-xs-inline" ng-if="'buildconfigs/instantiate' | canI : 'create'">
                     <a href=""
                       role="button"
                       ng-click="startBuild()">Start Build</a>
                   </li>
-                  <li>
+                  <li ng-if="'buildconfigs' | canI : 'update'">
                     <a ng-href="{{buildConfig | editResourceURL}}" role="button">Edit</a>
                   </li>
-                  <li>
+                  <li ng-if="'buildconfigs' | canI : 'update'">
                     <edit-link
                       resource="buildConfig"
                       kind="BuildConfig"
                       alerts="alerts">
                     </edit-link>
                   </li>
-                  <li>
+                  <li ng-if="'buildconfigs' | canI : 'delete'">
                     <delete-link
                       kind="BuildConfig"
                       resource-name="{{buildConfig.metadata.name}}"
@@ -78,20 +79,26 @@
                     <div ng-if="unfilteredBuilds && (unfilteredBuilds | hashSize) === 0" class="empty-state-message text-center">
                       <h2>No builds.</h2>
 
-                      <p>
-                        Start a new build to create an image from
-                        <span ng-if="buildConfig.spec.source.type === 'Git'">
-                          source repository
-                          <span class="word-break" ng-bind-html='buildConfig.spec.source.git.uri | githubLink : buildConfig.spec.source.git.ref : buildConfig.spec.source.contextDir | linky'></span>
-                        </span>
-                        <span ng-if="buildConfig.spec.source.type !== 'Git'">
-                          build configuration {{buildConfig.metadata.name}}.
-                        </span>
-                      </p>
+                        <p>
+                          <span ng-if="!('buildconfigs/instantiate' | canI : 'create')">
+                            Builds will create an image from
+                          </span>
+                          <span ng-if="'buildconfigs/instantiate' | canI : 'create'">
+                            Start a new build to create an image from
+                          </span>
+                          <span ng-if="buildConfig.spec.source.type === 'Git'">
+                            source repository
+                            <span class="word-break" ng-bind-html='buildConfig.spec.source.git.uri | githubLink : buildConfig.spec.source.git.ref : buildConfig.spec.source.contextDir | linky'></span>
+                          </span>
+                          <span ng-if="buildConfig.spec.source.type !== 'Git'">
+                            build configuration {{buildConfig.metadata.name}}.
+                          </span>
+                        </p>
 
-                      <button class="btn btn-primary btn-lg" ng-click="startBuild(buildConfig.metadata.name)">
-                        Start Build
-                      </button>
+                        <button class="btn btn-primary btn-lg" ng-click="startBuild(buildConfig.metadata.name)" ng-if="'buildconfigs/instantiate' | canI : 'create'">
+                          Start Build
+                        </button>
+
                     </div>
 
                     <!-- Show the latest build and a chart of the recent build duration and status. -->
@@ -114,7 +121,7 @@
                             <span ng-switch-default>is {{latestBuild.status.phase | lowercase}}.</span>
                           </span>
                         </span>
-                        <a ng-href="{{latestBuild | buildLogURL}}" ng-if="latestBuild | buildLogURL">View Log</a>
+                        <a ng-href="{{latestBuild | buildLogURL}}" ng-if="(latestBuild | buildLogURL) && ('builds/log' | canI : 'get')">View Log</a>
                       </div>
                       <div class="latest-build-timestamp meta text-muted">
                         <span ng-if="!latestBuild.status.startTimestamp">
--- a/app/views/browse/build.html
+++ b/app/views/browse/build.html
@@ -20,14 +20,14 @@
                       data-original-title="Building from build configuration {{buildConfig.metadata.name}} has been paused.">
                 </span>
                 <small class="meta">created <relative-timestamp timestamp="build.metadata.creationTimestamp"></relative-timestamp></small>
-                <div class="pull-right dropdown">
+                <div class="pull-right dropdown" ng-hide="!('builds' | canIDoAny)">
                   <!-- Primary Actions -->
                   <button class="btn btn-default hidden-xs"
                           ng-click="cancelBuild()"
-                          ng-if="!build.metadata.deletionTimestamp && (build | isIncompleteBuild)">Cancel Build</button>
+                          ng-if="!build.metadata.deletionTimestamp && (build | isIncompleteBuild) && ('builds' | canI : 'update')">Cancel Build</button>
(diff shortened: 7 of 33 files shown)


app/scripts/services/authorization.js
```diff
@@ -0,0 +1,169 @@
...
+    // In case no page rules are set, function will determine whether user can create some resources on the page
+    // to know if the 'Add to Project' btn should be shown.
+    AuthorizationService.prototype.setPageRules = function(pageRules) {
+      var self = this;
+      var canCreate = false;
+
+      if (pageRules) {
+        _.each(this.rules, function(rule) {
+          _.each(pageRules, function(verbs, resource) {
+            if (_.indexOf(rule.resources, resource) !== -1) {
```
spadgett: Does this work (here and below)?

`_.includes(rule.resources, resource)`

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,169 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .factory("AuthorizationService", function($q, $log, $interval, DataService, APIService){
```
spadgett: We should use `Logger` instead of `$log`.

jhadvig:
will add `bindata` after review :)

jwforres:
I suspect @liggitt will be interested in this one

app/scripts/services/authorization.js
jhadvig: yeah, forgot about this lodash function. Will have to tattoo it on forearm :smiley:

jwforres:
So first, keep your existing changes in a side branch since you have them working ... 
But I'd like you to try something else since I would prefer to see no changes to the controllers. Here is what I am thinking:

1. Add a filter that passes through to AuthorizationService.canI (similar to what we do with the navigate filters),  AuthorizationService.canI should just return false if a rules review has not yet been performed.  Views should call the canI filter for the individual actions.
2. To determine whether to show the Actions dropdown menu, you'll want to add a filter which is canIDoAny that accepts an array of actions it wants to check and will return true as soon as it determines the user can perform any of the actions.
3. Instead of having all the controllers for namespace scoped pages requesting to have a rules review performed, we should request a rules review as part of ProjectsService.getProject

I've at least run this by @spadgett and @benjaminapetersen and we agree in theory

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,169 @@
...
+      this.wasCanCreateSomeResourceSet = false;
+      this.projectNameRules = null;
+      $interval.cancel(this.intervalTimer);
+    };
+
+    // Determine whether a single action on a specified resource can be performed by user. The review result will be
+    // added into the $scope.canI variable.
+    // In case a resource shall to be added into the $scope.canI variable under a different unique name(eg. not as a 
+    // resource kind) but a different unique name(eg. namespace), use the 'identifier' parameter.  
+    AuthorizationService.prototype.canI = function(ns, verb, kind, $scope, identifier) {
```
liggitt: should this be more specific regarding api group, kind, and resource?

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,169 @@
...
+    function AuthorizationService() {
+      this.rules = null;
+      this.intervalTimer = null;
+      this.forceReload = false;
+      this.canCreateSomeResource = false;
+      this.wasCanCreateSomeResourceSet = false;
+      this.projectNameRules = null;
+    }
+
+    AuthorizationService.prototype.reviewUserRules = function($scope) {
```
spadgett: I know we do this in other places, but we shouldn't pass scope to services. Services shouldn't know anything about a controller's scope.

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,169 @@
...
+      }
+      DataService.create('subjectaccessreviews', null, object, {namespace:ns}).then(
+        function(data) {
+          $scope.canI[identifier][verb] = data.allowed;
+        }, function() {
+          $scope.canI[identifier][verb] = false;
+        });
+    };
+
+    return new AuthorizationService();
```
benjaminapetersen: Since this is returning a single instance of the service, probably don't need to go through all the trouble of creating a constructor function and properties on a prototype... the service can just `return { /* props */ }`.  Would eliminate some `var self = this` kind of things as well.  `imagestreams.js` service is prob a decent example.

spadgett:
Related to https://bugzilla.redhat.com/show_bug.cgi?id=1293791

jhadvig:
@jwforres I've updated the PR(not final) to get additional feedback.
Things I want to ask:
1. I've add the `AuthorizationService.getProjectRules` call to the projects.js service, but not sure if that s the right place, since at the settings page im experiancing that the $digest is not triggered when the $http get call response with the `selfsubjectrulesreviews`.
2. Whats you opinion on showing the 'missing health check' notification, specifically the links to 'Add Health Check'. Also for metrics, as QA asked in the card https://trello.com/c/UQEUgVrY/661-3-only-show-users-actions-they-have-authority-to-perform

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+      }
+    };
+
+    // Method to determine if user can perform specified action on a resource.
+    // If resource is not specified, all resources will be checked if the action
+    // can be performed.
+    AuthorizationService.prototype.canI = function(resource, verb) {
+      var rules = this.getRules();
+      if (rules) {
+        if (resource) {
```
benjaminapetersen: You can probably just `return _.contains(rules[resource], verb)` here since underscore will return true or false for you.

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+          if (_.contains(rules[resource], verb)) {
+            return true;
+          } else {
+            return false;
+          }
+        } else {
+          var canIDoAction = false
```
benjaminapetersen: Might be able to simplify this part as well with something like:
```javascript
return _.some(rules, function(verbs) {
  return _.contains(verbs, verb);
});
```

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+    function AuthorizationService() {
+      this.rules = null;
+      this.intervalTimer = null;
+      this.forceReload = false;
+      this.canICreateSomeResource = false;
+      this.projectNameRules = null;
+    }
+
+    var normalizeRules = function(rules) {
+      var normalizedRules = {};
```
benjaminapetersen: Might be able to eliminate a few lines/a loop here too if you `return _.transform(rules)` or `_.reduce(rules)`

app/scripts/controllers/deployment.js
```diff
@@ -39,6 +39,7 @@ angular.module('openshiftConsole')
       }
     ];
 
+    $scope.isDeploymentConfig = ($routeParams.deploymentconfig) ? true : false;
```
jwforres: isDeploymentConfig isnt really right given how refer to things elsewhere in the code, its isDeployment, or maybe hasDeploymentConfig.

app/scripts/filters/resources.js
```diff
@@ -125,6 +125,50 @@ angular.module('openshiftConsole')
       return displayName;
     };
   })
+  .filter('canI', function(AuthorizationService) {
+    return function(resource, verb) {
+      return AuthorizationService.canI(resource, verb);
+    };
+  })
+  .filter('canIDoAny', function(canIFilter) {
+    var resourceRulesMap = {
```
liggitt: Seems like inconsistent capitalization and pluralization in the keys here would make it difficult to use

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+      _.each(rules, function(rule) {
+        _.each(rule.resources, function(resource) {
+          normalizedRules[resource] = rule.verbs;
```
liggitt: Are API groups considered here?

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+        });
+      });
+      return normalizedRules;
+    };
+
+    AuthorizationService.prototype.getProjectRules = function(projectName) {
+      var self = this;
+      if (!self.rules || projectName !== self.projectNameRules) {
```
liggitt: Pretty sure @jwforres and @spadgett want to check permissions in other projects… also, I was sort of expecting to keep the cache of rules qualified per project and refresh on project navigation, instead of an unqualified cache assumed to apply to the current project

app/scripts/controllers/otherResources.js
```diff
@@ -70,6 +70,7 @@ angular.module('openshiftConsole')
       if (!selected) {
         return;
       }
+      $scope.selectedResource = APIService.kindToResource(selected.kind);
```
jwforres: I think you are going to need the group as well to correctly check things that are not in one of the base APIs.  Have you checked this with HPAs?

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+        // - no rules are cached
+        // - when need to force update the rules
+        // - when switching between projects
+        Logger.log("AuthorizationService, loading user rules");
+        console.log("AuthorizationService, loading user rules");
+        if (self.forceReload) {
+          self.setForceReload(false);
+        }
+        if (!self.intervalTimer) {
+          // Force rules reload every 3 minutes
```
liggitt: Do we expect this to be needed?

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+        // Using cached data.
+        Logger.log("AuthorizationService, using cached rules");
+        self.rules = self.getRules();
+      }
+    };
+
+    // Method to determine if user can perform specified action on a resource.
+    // If resource is not specified, all resources will be checked if the action
+    // can be performed.
+    AuthorizationService.prototype.canI = function(resource, verb) {
```
liggitt: Do you have an example of how we'd use the "canI" function without specifying a resource? Checking you have that verb on all resources you have rules for (which likely is not all resources that exist) is odd.

app/scripts/filters/resources.js
jwforres: We can't hardcode this list, especially for the other resources page because that list is completely dynamic and we will never know all of the 3rd party extension types that may show up.   This filter needs to take in both a resource, and a list of verbs it should check.

app/scripts/filters/resources.js
```diff
@@ -125,6 +125,50 @@ angular.module('openshiftConsole')
...
+    }
+    return function(resource) {
+      var canIDoAny = false;
+      _.each(resourceRulesMap[resource], function(verbs, resource) {
+        _.each(verbs, function(verb) {
+          canIDoAny = canIFilter(resource,verb);
+        });
+      });
+      return canIDoAny;
+    };
```
jwforres: It should just reduce down to the code below.  _.some will short circuit as soon as it gets an element that returns true, which is what we want this to do.
```
    return function(resource, verbs) {
      return _.some(verbs, function(verb) {
        return canIFilter(resource,verb));
      });
    };
```

app/scripts/services/authorization.js
liggitt: I think I'd expect this to return synchronously if the rules for the project were already available, and return a promise chained off of loading the rules if they aren't. I *think* promises that resolve to bools can be handled by angular view ng-if/ng-show attributes

liggitt: Or maybe have a guaranteed synchronous version that returns undefined if the rules haven't been loaded yet

app/scripts/services/projects.js
```diff
@@ -23,6 +23,7 @@ angular.module('openshiftConsole')
           return  AuthService
                     .withUser()
                     .then(function() {
+                      AuthorizationService.getProjectRules(projectName);
```
liggitt: should this return a promise we wait for, along with the project fetch, and only when both are resolved do we consider the project loaded?

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+
+    AuthorizationService.prototype.getProjectRules = function(projectName) {
+      var self = this;
+      if (!self.rules || projectName !== self.projectNameRules) {
+        // Making an API call to get the rules if:
+        // - no rules are cached
+        // - when need to force update the rules
+        // - when switching between projects
+        Logger.log("AuthorizationService, loading user rules");
+        console.log("AuthorizationService, loading user rules");
```
jwforres: take out this console.log

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,94 @@
...
+  .factory("AuthorizationService", function($q, Logger, $interval, DataService, APIService){
+    function AuthorizationService() {
+      this.rules = null;
+      this.intervalTimer = null;
+      this.forceReload = false;
+      this.canICreateSomeResource = false;
+      this.projectNameRules = null;
+    }
+
+    var normalizeRules = function(rules) {
```
jwforres: I know for sure we need to be handling getting back a "*" for a resource or a "*" for the verb.  I think a resource here might also be able to be a specific resource including the name.  @liggitt can you point us to the API doc for what all the possibilities are for these rules

app/scripts/services/projects.js
jwforres: I'm trying to decide, I hate to wait for the rules to load before fetching other stuff but I'm not sure there is a better way to guarantee we get the digest loop triggered off this data loading.

app/scripts/services/authorization.js
liggitt: https://github.com/openshift/origin/blob/master/pkg/authorization/api/v1/types.go#L18 is your best bet. 

verbs can contain specific verbs (like `create`) and/or `*`
resources can contain specific resources (like `pods`) and/or `*`
apiGroups can contain specific groups (like "" for the v1 origin/k8s groups) or `*`
resourceNames can contain specific names (like `mypod`) or `*` or be empty/missing (which means the same as `*`)

For your purposes (99% correct estimates about whether you can do something, erring on the side of letting you try and fail) you can probably ignore attributeRestrictions and nonResourceURLs

app/scripts/services/projects.js
liggitt: If the place we use it in the views is given a promise when the rules haven't resolved yet, does that not trigger a digest loop?

app/scripts/services/authorization.js
jwforres: I guess its just a question of whether we want to force someone to reload the page to force the rules to be refreshed for a project.  Thats probably ok since people's access doesn't change that often.

app/scripts/services/authorization.js
jwforres: Agree with @liggitt here.

app/scripts/services/authorization.js
liggitt: I'd rather remember when we loaded the rules for a project and in the getter, kick a background refresh of the cache if it's older than a certain age, instead of a registered refresh interval. Even then, three minutes seems aggressive

app/scripts/filters/resources.js
jhadvig: @liggitt plural is for pages in which we list the resource, singular for browse page.

@jwforres are you talking about other-resource rules that are listed here, or generally ? Problem is that dropdown often contains action on more then one resource, eg `buildConfig`, where we have to check `buildConfig` for `delete` and `update` action and `buildconfigs/instantiate` for `create` action.  But we can't pass to the filter two arrays {[resources], [verbs]} cause we want to check specific action on specific resource, so it has to be a collection, like the one I've proposed.
Also I've didn't come up with a better place to put the collection since it can't go to a view, or at least it would look ugly:
`ng-hide="!({'buildconfigs': ['delete', 'update'], 'buildconfigs/instantiate': ['create']} | canIDoAny)"`

What we could do is do leave the other-resources out of this, and just use the canI filter to check the specific resource for update and delete action, since we only display those two.

app/scripts/services/authorization.js
jhadvig: Not sure for what reason.

app/scripts/services/authorization.js
jhadvig: @liggitt dont think I follow your concern. The implementation caches the rules for current project, where the ProjectsService will determin which project rules are going to be loaded. When changing the project the rules for the other project will be loaded and cached..
For checking permissions in other projects we should use `subjectaccessreviews` call on specific resource-verb-namespace(dont have the method for that here since I was using it for checking delete permissions of the projects on the projects listing page, but we agreed with @jwforres that we wont do that until we have paging since the page could list lots of projects, which could lead to lots of api calls)

app/scripts/controllers/otherResources.js
jhadvig: I have and it worked. I dont think we need to specify the apiGroup since `selfsubjectrulesreviews` will return all the resources no matter what group they are in. At least thats what I've saw.
@liggitt is that right ?

app/scripts/services/authorization.js
jhadvig: @liggitt agreed that the interval is too short :+1: 
My point was not to hide/show the action btns/links on fly, but rather on next page to avoid any confusion. So it all come to the question if we want to do the changes (if there will be any) on the same page or on next.
Thoughts ?

app/scripts/services/authorization.js
jhadvig: @liggitt so the motivation here was to check if user can `create` or `update` any resource in the project, so we can render the `Add to project` button in the project-header.

app/scripts/services/projects.js
benjaminapetersen: Tend to agree this should be in a promise chain. Might be able to use `$q.all([])` to resolve in parallel, but that could be a pain for updating calling code.  

I'd tend to think line 40 could be:
```javascript
 return [project, context, projectRules];
```
and having a direct handle on a projectRules object for the particular project would be handy.

jhadvig:
@jwforres I've updated the PR. PTAL

app/scripts/controllers/otherResources.js
jwforres: @jhadvig so you are seeing it being returned as "horizontalpodautoscalers" not "extensions/horizontalpodautoscalers" ?   you can have the same "kind" in more than one group, i'd expect to get cases where the kind from one group has a different set of allowed actions than the kind in another group. Other than some special cases, the same kind in different groups actually are different resources.

cc @deads2k

app/scripts/filters/resources.js
jwforres: this request still applies, i want to see this using _.some so that its short circuiting

app/views/browse/deployment.html
```diff
@@ -32,23 +32,24 @@
...
                   <ul class="dropdown-menu actions action-button">
-                    <li>
+                    <li ng-if="'replicationcontrollers' | canI : 'update'">
                       <a ng-href="project/{{project.metadata.name}}/attach-pvc?deployment={{deployment.metadata.name}}"
                          role="button">Attach Storage</a>
                     </li>
                     <!-- Only allow editing health checks if there is no deployment config or this
                          is the active deployment. -->
-                    <li ng-if="!deploymentConfigName || isActive">
+                    <li ng-if="(!deploymentConfigName || isActive) && (hasDeploymentConfig | canIScale)">
```
jwforres: i'd prefer this was `(deployment | canIScale)` where deployment is the actual resource

app/scripts/filters/resources.js
```diff
@@ -125,6 +125,52 @@ angular.module('openshiftConsole')
...
+  })
+  .filter('canIScale', function(canIFilter) {
+    return function(deploymentConfig) {
```
jwforres: this should take the actual deployment resource (the RC), we already have an isDeployment filter.  Reduce this function down to:

```
.filter('canIScale', function(canIFilter, isDeploymentFilter) {
  return function(deployment) {
    return canIFilter(isDeploymentFilter(deployment) ? "deploymentconfigs/scale" : "replicationcontrollers", "update"); 
  }
}
```

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .factory("AuthorizationService", function($q, Logger, $interval, DataService){
+    function AuthorizationService() {
+      this.currentProject = null;
+      this.projects = {};
```
jwforres: this.rulesByProject  to be consistent with our maps in other places

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+        });
+        if (_.contains(rule.verbs, "create") || _.contains(rule.verbs, "update")) {
+          canAddToProject = true;
+        }
+      });
+
+      return [normalizedRules, canAddToProject];
+    };
+
+    var getOldestProject = function(allProjectsRules) {
```
jwforres: you are effectively doing a "least recently created" cache here, i think an LRU cache would work fine and for that angular can do the work for you, @benjaminapetersen can point you to where he used the built-in LRU in data.js

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+
+    AuthorizationService.prototype.getProjectRules = function(projectName) {
+      var self = this;
+      var deferred = $q.defer();
+      this.currentProject = projectName;
+      if (!_.has(self.projects, projectName)) {
+        Logger.log("AuthorizationService, loading user rules for " + projectName + " project");
+
+        // Remove oldest project rules that we cache
+        if (_.size(self.projects) === self.MAX_PROJECTS) {
```
jwforres: angular's LRU will do this work for you

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+                      apiVersion: "v1"
+                    };
+        DataService.create('selfsubjectrulesreviews', null, object, {namespace: projectName}).then(
+          function(data) {
+            var normalizedData = normalizeRules(data.status.rules);
+            self.projects[projectName] = {
+              rules: normalizedData[0],
+              canAddToProject: normalizedData[1],
+              intervalTimer: $interval(function() {
+                                delete self.projects[projectName];
```
jwforres: this could be problematic, this could delete the data while someone is sitting on a particular page, if some other resource causes a digest loop, its not going to re-trigger a rules load, its going to call the canI filters which now have no data to pull from

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+            };
+
+            deferred.resolve(true);
+          }, function() {
+            deferred.reject(false);
+        });
+      } else {
+        // Using cached data.
+        // Just refresh the cacheTimestamp value to a current time. Not the interval.
+        Logger.log("AuthorizationService, using cached rules for " + projectName + " project");
```
jwforres: Instead of https://github.com/openshift/origin-web-console/pull/82/files#diff-493e422fe43d2ad97883bd51eeef2c38R66
Check the cacheTimestamp here and if its older than X amount of time kick off a background refresh of the rules and return what we have already cached.

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+          return false;
+        }    
+      } else {
+        return false;
+      }
+    };
+
+    AuthorizationService.prototype.getCanIAddToProject = function(projectName) {
+      projectName = projectName || this.currentProject;
+      if (_.has(this.projects, projectName)) {
```
jwforres: Make this whole thing a one liner, remember !! forces a boolean return value
```
return !!_.get(this.projects, [projectName || this.currentProject, 'canAddToProject']);
```

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+        return this.projects[projectName].canAddToProject;
+      } else {
+        return false;
+      }
+    };
+
+
+    AuthorizationService.prototype.getRules = function(projectName) {
```
jwforres: getRulesForProject

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+      projectName = projectName || this.currentProject;
```
jwforres: one-liner
```
return _.get(this.projects, [projectName || this.currentProject, 'rules']);
```

_.get handles all the existence checking for you

app/views/other-resources.html
```diff
@@ -47,22 +47,30 @@
                     <td data-title="Labels">
                       <em ng-if="(resource.metadata.labels | hashSize) === 0">none</em>
                       <labels labels="resource.metadata.labels" clickable="true" kind="{{kindSelector.selected.kind | kindToResource : true }}" project-name="{{resource.metadata.namespace}}" limit="3" filter-current-page="true"></labels></td>
+<<<<<<< HEAD
```
jwforres: you've got some bad merge changes here

app/scripts/services/authorization.js
benjaminapetersen: Sure thing! [Angulars LRU](https://docs.angularjs.org/api/ng/service/$cacheFactory) is pretty easy to use, the `DataService()` constructor fn and  `DataService.prototype._data` are where its used in just a few lines of code.

app/views/browse/_deployment-details.html
```diff
@@ -16,7 +16,7 @@
         <status-icon status="deployment | deploymentStatus"></status-icon>
         {{deployment | deploymentStatus}}
         <span style="margin-left: 7px;">
-          <button ng-show="!rollBackCollapsed && (deployment | deploymentStatus) == 'Complete' && !(deployment | deploymentIsLatest : deploymentConfig) && !deployment.metadata.deletionTimestamp" ng-disabled="(deploymentConfigDeploymentsInProgress[deploymentConfigName] | hashSize) > 0" type="button" class="btn btn-default btn-xs" ng-click="rollBackCollapsed = !rollBackCollapsed">Roll Back</button>
+          <button ng-if="'deploymentconfigrollbacks' | canI : 'create'" ng-show="!rollBackCollapsed && (deployment | deploymentStatus) == 'Complete' && !(deployment | deploymentIsLatest : deploymentConfig) && !deployment.metadata.deletionTimestamp" ng-disabled="(deploymentConfigDeploymentsInProgress[deploymentConfigName] | hashSize) > 0" type="button" class="btn btn-default btn-xs" ng-click="rollBackCollapsed = !rollBackCollapsed">Roll Back</button>
```
benjaminapetersen: Wondering if we want an `ng-if` and an `ng-show` on the same node.  I don't mind the `ng-if` as much, but it seems like the `ng-show` is overloaded already. We should probably have a boolean for this and a clearly named function indicating what all this is doing.

I know that the `ng-show` isn't actually part of your PR, adding the `ng-if` just makes the complexity apparent.

benjaminapetersen: I'll open an issue for this since its separate.

jhadvig:
@jwforres I've addressed al the comments && update the PR.
I think its in good shape now. PTAL

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,105 @@
...
+    // Also while looping through the rules, check if user can create/update any resource on the 'Add to project' so the button will be displayed.
+    var normalizeRules = function(rules) {
+      var normalizedRules = {};
+      var canAddToProject = null;
+
+      // Check if resource name meets one of following conditions, since those resources can't be create/update via `Add to project` page:
+      //  - 'projectrequests'
+      //  - subresource taht contains '/', eg: 'builds/source', 'builds/logs', ...
+      //  - resource ending with 'review' string: 'localsubjectaccessreviews', 'selfsubjectrulesreviews', 'subjectaccessreviews' 
+      var checkResource = function(resource) {
```
benjaminapetersen: Could probably create `checkResource()` outside of `normalizeRules()`, since it does the same thing every time called (no gain in recreating it every time `normalizeRules()` is called)

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,105 @@
...
+        if (resource === "projectrequests" || _.contains(resource, "/") || resource.endsWith("reviews")) {
+          return false;
+        } else {
+          return true;
+        }
+      };
+      _.each(rules, function(rule) {
```
benjaminapetersen: triple `_.each()` makes me think there might be a better util for this. I'm also curious about returning the array `[normalizedRules, canAddToProject]`. I need to look at the calling code more, but wondering if this would be more efficient as two separate functions, one to normalize/map, the other to test.

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,105 @@
...
+            canAddToProject = true;
+          }
+        });
+      });
+      return [normalizedRules, canAddToProject];
+    };
+
+    AuthorizationService.prototype.getProjectRules = function(projectName) {
+      var self = this;
+      var deferred = $q.defer();
```
benjaminapetersen: Probably don't need to create a generic deferred here, I would suggest the function return the `DataService.create` (which is a promise), or at least `deferred.resolve(projectRules)` so the resolution is data.  

I believe you can do something like:

```javascript
// return the promise
return DataService
            .create('selfsubjectrulesreviews', / *  args */)
            // tap in here with .then() to do your intermediate work
            .then(function resolve(data) {
              self.cachedRulesByProject.set(projectName, /* stuff */);
              return self.cachedRulesByProject.get(projectName);  // nice to get it from my call to getProjectRules()
            }, function reject(err) {
               // do err handling if you want... or don't and let calling code deal with this
               return err;
            });

```

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,105 @@
...
+    AuthorizationService.prototype.canIAddToProject = function(projectName) {
+      return !!_.get(this.cachedRulesByProject.get(projectName || this.currentProject), ['canAddToProject']);
+    };
+
+
+    AuthorizationService.prototype.getRulesForProject = function(projectName) {
+      return _.get(this.cachedRulesByProject.get(projectName || this.currentProject), ['rules']);
+    };
+
+    return new AuthorizationService();
```
benjaminapetersen: I don't think anyone else cares too much, but I still think we should shy away from the trouble of setting up a `prototype` but then calling the constructor and returning a singleton anyway.  I think our services code can be cleaner if we just `return { /* hash of functions */ }`

app/scripts/services/projects.js
```diff
@@ -32,10 +32,13 @@ angular.module('openshiftConsole')
                       return DataService
                               .get('projects', projectName, context, {errorNotification: false})
                               .then(function(project) {
```
benjaminapetersen: Here is where returning the rules from `AuthorizationService.getProjectRules()` could be useful:
```javascript
AuthorizationService 
  .getProjectRules(projectName)
  then(function resolve(rules) {
    return [project, context, rules]; // might be useful to just get the rules here
  });
```

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,105 @@
...
+          return true;
+        }
+      };
+      _.each(rules, function(rule) {
+        _.each(rule.resources, function(resource) {
+          _.each(rule.apiGroups, function(apiGroup) {
+            normalizedRules[(apiGroup === "") ? resource : apiGroup + "/" + resource] = rule.verbs;
+          });
+
+          if (!canAddToProject && checkResource(resource) && (!_.isEmpty(_.intersection(rule.verbs ,(["*", "create", "update"]))))) {
```
jwforres: nice use of intersection :+1:

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,105 @@
...
+    // into:
+    // normalizedRules = {"extensions/jobs": ["create","delete","get","list","update"]}
+    // Also while looping through the rules, check if user can create/update any resource on the 'Add to project' so the button will be displayed.
+    var normalizeRules = function(rules) {
+      var normalizedRules = {};
+      var canAddToProject = null;
+
+      // Check if resource name meets one of following conditions, since those resources can't be create/update via `Add to project` page:
+      //  - 'projectrequests'
+      //  - subresource taht contains '/', eg: 'builds/source', 'builds/logs', ...
```
jwforres: typo in comment, 'that'

app/views/browse/_deployment-details.html
```diff
@@ -86,11 +87,12 @@
...
         builds="builds"
         detailed="true"
-        add-health-check-url="{{(!deploymentConfig || isActive) ? healthCheckURL : ''}}">
+        add-health-check-url="{{((!deploymentConfig || isActive) && ('deploymentconfigs' | canI : 'update')) ? healthCheckURL : ''}}">
       </pod-template>
       <h4>Volumes</h4>
       <p ng-if="!deployment.spec.template.spec.volumes.length">
-        <a ng-href="project/{{project.metadata.name}}/attach-pvc?deployment={{deployment.metadata.name}}">Attach storage</a>
+        <a ng-if="'replicationcontrollers' | canI : 'update'" ng-href="project/{{project.metadata.name}}/attach-pvc?deployment={{deployment.metadata.name}}">Attach storage</a>
+        <i ng-if="!('replicationcontrollers' | canI : 'update')">none</i>
```
jwforres: why are you using an `<i>` tag here instead of a `<span>`

app/views/browse/build-config.html
```diff
@@ -78,20 +79,22 @@
...
-                          source repository
-                          <span class="word-break" ng-bind-html='buildConfig.spec.source.git.uri | githubLink : buildConfig.spec.source.git.ref : buildConfig.spec.source.contextDir | linky'></span>
-                        </span>
-                        <span ng-if="buildConfig.spec.source.type !== 'Git'">
-                          build configuration {{buildConfig.metadata.name}}.
-                        </span>
-                      </p>
+                      <div ng-if="'buildconfigs/instantiate' | canI : 'create'">
+                        <p>
+                          Start a new build to create an image from
```
jwforres: instead of hiding this entire block, i would just tweak this text a bit, and only hide the start button when the user can't instantiate builds.  maybe make the text 'New builds will create an image from ....'  or even just 'Builds will create an image from...'

app/views/browse/build.html
```diff
@@ -20,14 +20,14 @@
                       data-original-title="Building from build configuration {{buildConfig.metadata.name}} has been paused.">
                 </span>
                 <small class="meta">created <relative-timestamp timestamp="build.metadata.creationTimestamp"></relative-timestamp></small>
-                <div class="pull-right dropdown">
+                <div class="pull-right dropdown" ng-hide="!('builds' | canIDoAny)">
                   <!-- Primary Actions -->
                   <button class="btn btn-default hidden-xs"
                           ng-click="cancelBuild()"
-                          ng-if="!build.metadata.deletionTimestamp && (build | isIncompleteBuild)">Cancel Build</button>
+                          ng-if="!build.metadata.deletionTimestamp && (build | isIncompleteBuild) && ('builds/clone' | canI : 'create')">Cancel Build</button>
```
jwforres: create on builds/clone isnt the right check for cancel build, copy/paste error i'm guessing, same thing on line 42 below.  Should just be checking update on builds i think.

app/views/browse/deployment-config.html
```diff
@@ -129,11 +131,12 @@
...
                               builds="builds"
                               detailed="true"
-                              add-health-check-url="{{healthCheckURL}}">
+                              add-health-check-url="{{('deploymentconfigs' | canI : 'update') ? healthCheckURL : ''}}">
                           </pod-template>
                           <h4>Volumes</h4>
                           <p ng-if="!deploymentConfig.spec.template.spec.volumes.length">
-                            <a ng-href="project/{{project.metadata.name}}/attach-pvc?deploymentconfig={{deploymentConfig.metadata.name}}">Attach storage</a>
+                            <a ng-if="'deploymentconfigs' | canI : 'update'" ng-href="project/{{project.metadata.name}}/attach-pvc?deploymentconfig={{deploymentConfig.metadata.name}}">Attach storage</a>
+                            <i ng-if="!('deploymentconfigs' | canI : 'update')">none</i>
```
jwforres: `<span>` instead of `<i>`

app/views/browse/deployment-config.html
```diff
@@ -150,15 +153,16 @@
...
                                role="button">Set resource
                               <span ng-if="!('cpu' | isRequestCalculated : project)">requests and</span> limits</a>
                           </div>
 
                           <!-- Create autoscaler -->
                           <div ng-if="!autoscalers.length">
-                            <a ng-href="project/{{projectName}}/edit/autoscaler?kind=DeploymentConfig&name={{deploymentConfig.metadata.name}}"
+                            <a ng-if="'extensions/horizontalpodautoscalers' | canI : 'create'" ng-href="project/{{projectName}}/edit/autoscaler?kind=DeploymentConfig&name={{deploymentConfig.metadata.name}}"
                                role="button">Add autoscaler</a>
+                            <i ng-if="!('extensions/horizontalpodautoscalers' | canI : 'create')">none</i>
```
jwforres: `<span>`

app/views/overview/_service-group.html
```diff
@@ -50,14 +50,16 @@ <h4 class="route-title truncate">
...
           <h2>No linked services.</h2>
-          <p>
-            No services are linked to {{service.metadata.name}}.
-            <span ng-if="(services | hashSize) > 1">Add a service to group them together.</span>
-          </p>
-          <div ng-if="(services | hashSize) > 1">
-            <button class="btn btn-primary" ng-show="(services | hashSize) > 1" ng-click="linkService()">
-              Link Service
-            </button>
+          <div ng-if="'services' | canI : 'update'">
```
jwforres: cant decide on this one, what does this look like with just the header vs if you also include the "No services are linked to foo." message?  You might just need hide the button and the "Add a service..." message

jwforres:
Can you go ahead and add this to otherResources.js so that it will only put things in the dropdown that can actually be listed by the user:

```
   // for context, goes underneath the .then
    ProjectsService
      .get($routeParams.project)
      .then(_.spread(function(project, context
...
...
       // add this
        $scope.kinds = _.filter($scope.kinds, function(kind){
          var resource = APIService.kindToResource(kind.kind);
          return AuthorizationService.canI("list", kind.group ? kind.group + "/" + resource : resource, $scope.projectName);
        });
```

You'll also need to include AuthorizationService in the list of deps at the top of the service.  You'll know its working if you can't see PetSets.

Then in constants.js please remove 
```
      "LocalResourceAccessReview",
      "LocalSubjectAccessReview",
      "ResourceAccessReview",
      "SubjectAccessReview",
      "ReplicationControllerDummy",
```
from the blacklist, they'll now be cleanly removed from the list because the List verb isn't allowed on them.  You need to leave DeploymentConfigRollback for now.

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,105 @@
...
+      //  - resource ending with 'review' string: 'localsubjectaccessreviews', 'selfsubjectrulesreviews', 'subjectaccessreviews' 
+      var checkResource = function(resource) {
+        if (resource === "projectrequests" || _.contains(resource, "/") || resource.endsWith("reviews")) {
```
liggitt: checking if it contains `/` is fine, but I'd rather not use endsWith... we should have an explicit list (or hash) of resources we don't care about for this

jwforres:
also remove ReplicaSet, PodTemplate, and ThirdPartyResource from the blacklist

jwforres:
There is another View Log link that needs a canI check here https://github.com/openshift/origin-web-console/pull/82/files#diff-d40b418167447e1277db88f1163bb5d9L117

app/views/browse/pod.html
```diff
@@ -122,7 +122,7 @@
                   <uib-tab active="selectedTab.terminal"
                     select="terminalTabWasSelected = true"
                     ng-init="containers = pod.status.containerStatuses"
-                    ng-if="containersRunning(pod.status.containerStatuses) > 0">
+                    ng-if="containersRunning(pod.status.containerStatuses) > 0 && ('pods/attach' | canI : 'get')">
```
jwforres: this should be pods/exec

app/views/browse/_deployment-details.html
jhadvig: My bad... should check it

app/views/overview/_service-group.html
jhadvig: @jwforres 
screenshot
vs.
screenshot-2

Thinking about it now I would leave the  "No services are linked to foo." there. Thoughts ?

app/scripts/services/authorization.js
jhadvig: The motivation behind having all the logic (data normalization + test if user 'canAddToProject') to go through the received rules only once. On the other hand I can understand that it can get hard to read.

app/scripts/services/projects.js
jhadvig: Thinking about any use of the `rules` in the controller, but cant think of any, since the use of the `canI` filter. Any ideas ?

app/scripts/services/authorization.js
jhadvig: @benjaminapetersen wouldn't that expose all the declared functions, even `normalizeRules()` or `checkResource` ?

app/views/overview/_service-group.html
jwforres: the latter feels more balanced even though a little redundant, lets go with
the second one.  It would be better if "frontend" was a link to the service
to be consistent with the no deployments message, but we can fix that
outside of this PR  @spadgett (later so it doesn't conflict with these
changes)

jhadvig:
@jwforres @benjaminapetersen I've update the PR based on your comments, expect the one that suggests to return the `rules` together with the `project` and `context`, since we have `canI`, `canIDoAny`, `canIScale`, `canIAddToProject` filters that take care of any users actions they have authority to perform

benjaminapetersen:
Cool, I'm good with that.

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+    var canAddToProject = function(rules) {
+      var canAddToProject = null;
+      _.each(rules, function(rule) {
+        _.each(rule.resources, function(resource) {
+          if (!canAddToProject && checkResource(resource) && (!_.isEmpty(_.intersection(rule.verbs ,(["*", "create", "update"]))))) {
+            canAddToProject = true;
+          }
+        });
+        if (canAddToProject) {
+          return false;
```
jwforres: why is this not returning true?

app/scripts/services/authorization.js
```diff
@@ -0,0 +1,121 @@
...
+      }
+    };
+
+    // Check if user can create/update any resource on the 'Add to project' so the button will be displayed.
+    var canAddToProject = function(rules) {
+      var canAddToProject = null;
+      _.each(rules, function(rule) {
+        _.each(rule.resources, function(resource) {
+          if (!canAddToProject && checkResource(resource) && (!_.isEmpty(_.intersection(rule.verbs ,(["*", "create", "update"]))))) {
+            canAddToProject = true;
```
jwforres: i think this would be better if it were nested _.some so that it shortcircuits as fast as possible

```
var canAddToProject = function(rules) {
  return _.some(rules, function(rule) {
    return _.some(rule.resources, function(resource) {
      return checkResource(resource) && !_.isEmpty(_.intersection(rule.verbs ,(["*", "create", "update"])));
    }
  }
}
```

app/scripts/services/authorization.js
jwforres: ignore this comment, i'd rather see https://github.com/openshift/origin-web-console/pull/82/files#r70297639  happen

jwforres:
[merge]

jwforres:
looks like it might have just been a flake [merge]

jwforres:
failed on the same test, @jhadvig can you run `grunt test-integration` locally and see if its passing

jwforres:
[merge] will let it go one more time, but i suspect this may be related to additional delay before the delete action is being shown

jwforres:
[merge]

PR #84, opened 2016-06-20
Make hostname read-only when editing routes

### LINKED ISSUE #80: Hostname should be read-only when editing routes
See https://github.com/openshift/origin/pull/9425

/cc @jwforres

spadgett:
Fixes #80 

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/oscRouting.js
+++ b/app/scripts/directives/oscRouting.js
@@ -36,7 +36,8 @@ angular.module("openshiftConsole")
         route: "=model",
         services: "=",
         showNameInput: "=",
-        routingDisabled: "="
+        routingDisabled: "=",
+        hostReadOnly: "="
       },
       templateUrl: 'views/directives/osc-routing.html',
       controller: function($scope) {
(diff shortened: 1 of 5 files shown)


jwforres:
assuming this just makes the input a read only field?  whenever i see a read only input it makes me wonder why its read only.  any reason not to just switch to just showing the text instead of an input?  possibly with a (?) that explains hostname can't be modified and to create a new route.  or a "Why can't I change this?" link that pops a help section below that explains it and links directly to the "Create Route" page in case they want to create a route with a new hostname.

spadgett:
Static input:

openshift_web_console

Readonly input:

openshift_web_console

spadgett:
Updated to use the static control. @jwforres let me know if you prefer readonly from the screenshots above.

jwforres:
@rhamilto says he likes the readonly input better, i think with the message
visible its ok

spadgett:
Updated to use the read-only input with help text explaining.

jwforres:
[merge]

PR #86, opened 2016-06-20
Remove relative Jenkins build URL workaround

spadgett:
The workaround is no longer needed since the sync plugin now sets an absolute URL. See https://github.com/fabric8io/openshift-jenkins-sync-plugin/pull/86

@jwforres PTAL


jwforres:
[merge]

PR #87, opened 2016-06-21
Deploy images from "Add to Project" page

spadgett:
https://trello.com/c/KMyd8ETi

Rebased from https://github.com/jwforres/origin/pull/44


jorgemoralespou:
@spadgett @jwforres 
>> The other thing that bothers me is we let you set up routes, resource constraints, auto-scaling, etc if you are starting from our builders, why aren't we letting you do that here?

> I'm torn on this. I'm trying to avoid the giant form we have for create from image since the rest can be configured later, but I can see wanting to configure it here.

I agree the whole workflow should be enable from here. This is one of the things that can provide better usability to the UI compared to CLI. Right now you can do all of that from CLI through command line switches, and I see people always executing commands afterwards (being the most obvious the exposure of the service). I think the workflow in UI should allow for configuration of everything at once.

Another important thing is this workflow needs to be tested with a different registry than docker hub and a registry that is secured, so that if a secret is required, can be provided as well. Since it seems that when you search you're querying that remote repository, this should be easily identifiable and provide an easy way for the user to correct.

spadgett:
@jorgemoralespou I agree, but these will have to be follow-ons. We've done some initial work to split the create form into a wizard and plan to reuse it here.

I also agree on secrets, but we don't handle them anywhere in the web console yet. We should open a card to look at secrets in the web console as a whole.

spadgett:
@jwforres PTAL

app/scripts/directives/deployImage.js
```diff
@@ -0,0 +1,184 @@
...
+          $scope.loading = true;
+          ImagesService.findImage($scope.imageName, $scope.context)
+            .then(
+              // success
+              function(response) {
+                $scope.import = response;
+                $scope.loading = false;
+
+                if (_.get(response, 'result.status') !== 'Success') {
+                  $scope.import.error = _.get(response, 'result.message', 'An error occurred.');
```
jwforres: default to 'An error occurred finding the image.'

app/scripts/directives/deployImage.js
```diff
@@ -0,0 +1,184 @@
...
+                  };
+                  $scope.runsAsRoot = ImagesService.runsAsRoot(image);
+                  $scope.ports = ApplicationGenerator.parsePorts(image);
+                  $scope.volumes = ImagesService.getVolumes(image);
+                  $scope.createImageStream = true;
+                }
+              },
+              // failure
+              function(response) {
+                $scope.import.error = $filter('getErrorDetails')(response) || 'An error occurred.';
```
jwforres: same here

app/scripts/directives/deployImage.js
```diff
@@ -0,0 +1,184 @@
...
+            }, function(response) {
+              $scope.import.error = $filter('getErrorDetails')(response) || 'An error occurred.';
+              $scope.loading = false;
+            });
+          }, true);
+
+          $scope.create = function() {
+            var resources = getResources();
+            var titles = {
+              started: "Deploying image " + $scope.import.name + " to project " + $scope.project,
```
jwforres: should these have a period at the end

app/scripts/directives/deployImage.js
```diff
@@ -0,0 +1,184 @@
...
+        };
+
+        var stripTag = $filter('stripTag');
+        var humanizeKind = $filter('humanizeKind');
+
+        // Change image names like "openshift/hello-openshift:latest" to "hello-openshift", which can be used as an app name.
+        var getName = function() {
+          // Remove everything through the last '/'.
+          var nameAndTag = _.last($scope.import.name.split('/'));
+          return stripTag(nameAndTag);
```
jwforres: probably need to be handling pull specs by id, which use `@` not `:`

app/scripts/services/images.js
```diff
@@ -0,0 +1,216 @@
...
+      };
+
+      return DataService.create('imagestreamimports',
+                                null,
+                                importImage,
+                                projectContext).then(function(response) {
+        return {
+          name: _.get(response, 'spec.images[0].from.name'),
+          image: _.get(response, 'status.images[0].image'),
+          tag: _.get(response, 'status.images[0].tag'),
```
jwforres: same question here, should we be handling id based pull specs, or does imagestreamimport not support that?

app/views/directives/deploy-image.html
```diff
@@ -0,0 +1,193 @@
...
+    Deploy an existing image from an image stream tag or Docker pull spec.
+  </p>
+  <form>
+    <fieldset ng-disabled="loading">
+      <div class="radio">
+        <label>
+          <input type="radio" ng-model="mode" value="istag">
+          Image Stream Tag
+        </label>
+        <!-- <span class="text-muted">&mdash; select an image stream tag</span> -->
```
jwforres: just old stuff? remove?

app/views/directives/deploy-image.html
```diff
@@ -0,0 +1,193 @@
...
+            <code>oc policy add-role-to-user system:image-puller system:serviceaccount:{{project}}:default -n {{istag.namespace}}</code>
+          </p>
+        </div>
+      </fieldset>
+      <div class="radio">
+        <label>
+          <input type="radio" ng-model="mode" value="dockerImage">
+          Image Name
+        </label>
+        <!-- <span class="text-muted">&mdash; specify an image name or pull spec</span> -->
```
jwforres: remove

jwforres:
[merge]

PR #90, opened 2016-06-21
Change how service grouping border is created

sg00dwin:
- position it in front of deployment tiles
	- use patternfly blue

@jwforres

### DIFF
--- a/app/styles/_overview.less
+++ b/app/styles/_overview.less
@@ -47,8 +47,18 @@
     margin-top: 5px;
   }
   .service-group {
-    border-left: 3px solid #00b9e4;
     margin-bottom: 20px;
+    position: relative;
+    &:before {
+      content: "";
+      display: block;
+      height: 100%;
+      width: 3px;
+      position: absolute;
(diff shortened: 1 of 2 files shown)


spadgett:
@sg00dwin We picked the blue to match the other items on the page (navbar highlights and donuts). Do you have a screenshot of what it looks like now?

sg00dwin:
It's slightly darker than previous color. I noticed it was what patternfly uses for tiles, but can switch it back.

screen

spadgett:
Can't decide. It doesn't look bad, but is it weird that we have a similar left border in the left nav that is a different color? @jwforres @rhamilto ?

jwforres:
I actually kind of like it, if it weren't for that left bar in the nav :-\

maybe we should get SJ to take a look at it

sg00dwin:
@spadgett, @jwforres I reverted to the original blue, so that I can get the primary fix in. Placing the blue bar above the deployment block left edges.

spadgett:
[merge]

PR #93, opened 2016-06-21
Show alert immediately when deleting project from project list

spadgett:
@jwforres This is the other alert bug we saw. PTAL


jwforres:
[merge]

PR #99, opened 2016-06-22
Upgrading PatternFly and Angular-PatternFly to v3.5.1

rhamilto:
and including PatternFly source less via patternfly.less and patternfly-additions.less so that future updates to PatternFly pick up any new partials added in patternfly.less or patternfly-additions.less.

Note:  I will squash prior to merge, but intentionally left separate commits so the changes are obvious.

@sg00dwin, PTAL.


app/styles/main.less
```diff
@@ -1,85 +1,7 @@
+// PatternFly
 @pf-less-path: "../../bower_components/patternfly/less";
-
```
benjaminapetersen: Nice!  Though I suppose we could ask if we need everything... guessing we probably do so this is cleaner.

rhamilto:
@sg00dwin, PTAnotherL.

sg00dwin:
@rhamilto LGTM

rhamilto:
@jwforres, @spadgett:  PTAL

app/index.html
```diff
@@ -82,6 +82,7 @@
     <script src="bower_components/angular-bootstrap/ui-bootstrap-tpls.js"></script>
     <script src="bower_components/lodash/lodash.js"></script>
     <script src="bower_components/bootstrap/dist/js/bootstrap.js"></script>
+    <script src="bower_components/bootstrap-combobox/js/bootstrap-combobox.js"></script>
```
spadgett: @rhamilto I know we're using bootstrap-select, but are we using bootstrap-combobox?

app/styles/_navbar-alt.less
```diff
@@ -111,19 +113,16 @@
       background: @navbar-pf-alt-color;
     }
   }
-  .nav-item-iconic {
+  .nav .nav-item-iconic {
     // utiltity links
     cursor: pointer;
     line-height: 1;
-    max-height: (@navbar-pf-alt-height - @navbar-pf-alt-border-width); // to keep Firefox from oversizing icons
```
spadgett: Just confirming you've tested on Firefox.

spadgett:
Two questions, otherwise LGTM

jwforres:
also needs a rebase

app/index.html
rhamilto: @spadgett, no, I'm not using bootstrap-combobox, but it's part of PatternFly, so it's getting pulled in accordingly.  We can exclude it if you think it's a problem.

spadgett: To this point, we've only been pulling in the components we're actually using to keep our JS smaller.

rhamilto: @spadgett, I think we probably need to revisit all of PatternFly's dependencies then as there are quite a few I don't think we are using but are also included here.

spadgett: We might step back and look at that outside of this PR. @rhamilto Do you know what is safe to remove? Is Patternfly like Bootstrap where the JS components are optional?

app/styles/_navbar-alt.less
rhamilto: Yep.  And just triple checked for good measure.  :)

screen shot 2016-07-06 at 9 22 27 am

app/index.html
rhamilto: I did a quick inventory, and I consider these suspicious PatternFly JS dependencies:

"bootstrap-combobox": "~1.1.6",
"bootstrap-datepicker": "~1.4.0",
"bootstrap-switch": "~3.3.2",
"bootstrap-touchspin": "~3.0.3",
"bootstrap-treeview": "~1.2.0",
"datatables-colreorder": "~1.1.3",
"datatables-colvis": "~1.1.2",
"moment": "~2.11.0"

That is, these are JS dependencies that are added from PatternFly that I don't think we make use of and have not explicitly been excluded in the wiredep exclude.

screen shot 2016-07-06 at 9 33 13 am

benjaminapetersen: Curious if they have considered the idea of "patternfly core" and "patternfly extensions"... just installing the core but giving a list of optional addons that don't land in your bower_components by default.

rhamilto:
[test]

app/index.html
jwforres: we include moment ourselves, if that is a patternfly dependency now we
could probably fallback to just using what patternfly pulls in, double
check the versions though, as long as patternfly's is newer and backwards
compatible its probably fine.

rhamilto:
@jwforres or @spadgett, it appears Travis is barfing on the PatternFly version change in bower.json.

jwforres:
yep dont worry about travis, will clear its cache after this merges, as
long as Jenkins passes then travis will be able to pass after we fix the
cache

jwforres:
[merge]

PR #101, opened 2016-06-23
Align text to left in the Action td at the other-resources page for mobile

jhadvig:
Fixing small bug where on the other-resources page for the mobile the Action column text was align to right.
Screen of the issue:
screenshot-8

After fix:
screenshot-7

@jwforres @spadgett PTAL

### DIFF
--- a/app/views/other-resources.html
+++ b/app/views/other-resources.html
@@ -47,11 +47,12 @@
                     <td data-title="Labels">
                       <em ng-if="(resource.metadata.labels | hashSize) === 0">none</em>
                       <labels labels="resource.metadata.labels" clickable="true" kind="{{kindSelector.selected.kind | kindToResource : true }}" project-name="{{resource.metadata.namespace}}" limit="3" filter-current-page="true"></labels></td>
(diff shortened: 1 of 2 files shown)


spadgett:
LGTM, but needs dist

jwforres:
its an improvement, but the mobile menu dropdown looks really strange here
 @rhamilto

jhadvig:
@jwforres we could display `Update | Delete` links for the mobile version. But now that I'm thinking about it would be better to solve this globally, since by this we could introduce a precedent how to deal with actions in tables.
Other solution could be a small `Actions` dropdown button which will turn into small btn with `fa-cog` icon for mobile version
Screens:
screenshot-17
screenshot-16
screenshot-15

 @rhamilto any suggestions ?

jhadvig:
@jwforres I've updated the PR
Screens:
screenshot-19
screenshot-18
screenshot-20

jhadvig:
screenshot-21

jwforres:
awesome thanks @jhadvig [merge]

PR #115, opened 2016-06-27
Achieve equal height deployment tiles and enable option to append...

sg00dwin:
multiple tile blocks if needed

@jwforres

### DIFF
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -248,7 +248,7 @@
       border-radius: @border-radius-md;
       padding: 5px;
       position: relative;
-      margin: 10px 0 6px;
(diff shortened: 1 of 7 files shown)


spadgett:
Can you check safari with 2 or more dependent services? Worried about flex basis 0%

sg00dwin:
@spadgett updated so overview-service and no-deployments block use flex-basis: 49% which displays good cross browsers. Safari, FF, Chrome, IE

group

sg00dwin:
@jwforres updated pr correct no deployments block.

updated

jwforres:
issues with standalone services

edge-cases

PR #118, opened 2016-06-28
Stop the production less file from getting generated during watch in grunt serve

jwforres:


jwforres:
@spadgett this fixes that issue we have seen with the dist css getting a diff all the time while switching branches / making changes.  It was running all of the tasks under less and not just the less:development, and since the minification wasn't running it was creating a diff compared to what we check in.

jwforres:
[merge]

spadgett:
Thank you :)

Issue #117, opened 2016-06-28
Must use build config annotation not the label, label can be truncated


jwforres:
We currently use the build config label to associate builds to their build config, but because labels only support 64 chars and build config names can be up to 256 the label may be a truncated value.  The annotation `openshift.io/build-config.name` will have the full value.  Since that annotation was only recently added we need to fallback to looking at the label if the annotation doesn't exist.

jwforres:
@spadgett FYI this affects both the overview and the browse / builds page at least, need to check for any other references

PR #119, opened 2016-06-28
Use build config annotation instead of label when it exists

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/117

### DIFF
--- a/app/scripts/controllers/build.js
+++ b/app/scripts/controllers/build.js
@@ -42,6 +42,8 @@ angular.module('openshiftConsole')
       $scope.selectedTab[$routeParams.tab] = true;
     }
 
+    var buildConfigForBuild = $filter('buildConfigForBuild');
+
     var watches = [];
 
     var setLogVars = function(build) {
@@ -114,24 +116,36 @@ angular.module('openshiftConsole')
         );
 
         watches.push(DataService.watch("builds", context, function(builds, action, build) {
-          $scope.builds = {};
-          // TODO we should send the ?labelSelector=buildconfig=<name> on the API request
-          // to only load the buildconfig's builds, but this requires some DataService changes
-          var allBuilds = builds.by("metadata.name");
-          angular.forEach(allBuilds, function(build, name) {
-            if (build.metadata.labels && build.metadata.labels.buildconfig === $routeParams.buildconfig) {
-              $scope.builds[name] = build;
+          if (!action) {
+            $scope.builds = BuildsService.validatedBuildsForBuildConfig($routeParams.buildconfig, builds.by('metadata.name'));            
+          } else {
+            var buildConfigName = buildConfigForBuild(build);
+            if (buildConfigName === $routeParams.buildconfig) {
+              var buildName = build.metadata.name;
+              switch (action) {
+                case 'ADDED':
+                case 'MODIFIED':
+                  $scope.builds[buildName] = build;
+                  break;
+                case 'DELETED':
+                  delete $scope.builds[buildName];
+                  break;
+              }
             }
-          });
-
-          var buildConfigName;
-          var buildName;
-          if (build) {
-            buildConfigName = build.metadata.labels.buildconfig;
-            buildName = build.metadata.name;
           }
 
           updateCanBuild();
+        },
+        // params object for filtering
+        {
+          // http is passed to underlying $http calls
+          http: {
+            params: {
+              // because build config names can be > 63 chars but label values can't
+              // and we can't do a fieldSelector on annotations.  Plus old builds dont have the annotation.
+              labelSelector: $filter('labelName')('buildConfig') + '=' + _.trunc($routeParams.buildconfig, {length: 63, omission: ''})
+            }
+          }
         }));
 
         $scope.toggleSecret = function() {
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -43,6 +43,7 @@ angular.module('openshiftConsole')
     }
 
     var orderByDate = $filter('orderObjectsByDate');
+    var buildConfigForBuild = $filter('buildConfigForBuild');
     var watches = [];
 
     ProjectsService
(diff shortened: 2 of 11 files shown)


jwforres:
unless @spadgett gets bored on vacation and wants to look at this, @benjaminapetersen can you review this

summary of whats going on:

- anywhere we want to figure out what the build config is from a build resource we should be using the buildConfigForBuild filter
- the filter checks for the existence of the annotation first, if the annotation isn't there it falls back to the correct label (there is a very old pre-release deprecated label called `buildconfig` that we should never use)
- when watching builds for a particular build config we must use the label truncated to the max label length of 63 characters, there is no support for watching based on the annotation
- watched builds must be post filtered and only kept if they either don't have the annotation at all (we have to assume its a build from previous versions before the annotation started getting added), or the annotation is there and matches the build config we expect (because of the truncation on the label we could be getting back builds for another build config if both BCs had names longer than 63 chars)

cc @bparees

bparees:
I didn't check the code, but the bulleted logic sounds right.

benjaminapetersen:
Yup! checking

app/scripts/controllers/build.js
```diff
@@ -114,24 +117,43 @@ angular.module('openshiftConsole')
...
 
         watches.push(DataService.watch("builds", context, function(builds, action, build) {
-          $scope.builds = {};
-          // TODO we should send the ?labelSelector=buildconfig=<name> on the API request
-          // to only load the buildconfig's builds, but this requires some DataService changes
-          var allBuilds = builds.by("metadata.name");
-          angular.forEach(allBuilds, function(build, name) {
-            if (build.metadata.labels && build.metadata.labels.buildconfig === $routeParams.buildconfig) {
-              $scope.builds[name] = build;
+          if (!action) {
```
benjaminapetersen: Curious if this `if/else` block should be wrapped up as a function `validateBuilds()` so we can name what its doing:
```javascript
$scope.builds = validateBuilds(builds, action, build);
```

app/scripts/controllers/build.js
```diff
@@ -114,24 +117,39 @@ angular.module('openshiftConsole')
...
+            $scope.builds = _.pick(builds.by('metadata.name'), function(bld){
```
benjaminapetersen: +1 to _.pick()  😄

app/scripts/controllers/build.js
```diff
@@ -114,24 +117,39 @@ angular.module('openshiftConsole')
...
           updateCanBuild();
+        },
+        // params object for filtering
+        {
+          // http is passed to underlying $http calls
+          http: {
+            params: {
+              // because build config names can be > 63 chars but label values can't
+              // and we can't do a fieldSelector on annotations.  Plus old builds dont have the annotation.
+              labelSelector: 'openshift.io/build-config.name=' + _.trunc($routeParams.buildconfig, {length: 63, omission: ''})
```
benjaminapetersen: I think the `filter('labelName')('buildConfig')` will work here to get `openshift.io/build-config.name`

app/scripts/controllers/buildConfig.js
```diff
@@ -119,7 +127,9 @@ angular.module('openshiftConsole')
         // http is passed to underlying $http calls
         http: {
           params: {
-            labelSelector: 'buildconfig='+$scope.buildConfigName
+            // because build config names can be > 63 chars but label values can't
+            // and we can't do a fieldSelector on annotations.  Plus old builds dont have the annotation.
+            labelSelector: 'openshift.io/build-config.name=' + _.trunc($scope.buildConfigName, {length: 63, omission: ''})
```
benjaminapetersen: same filter `filter('labelName')('buildConfig')`

app/scripts/controllers/build.js
benjaminapetersen: Curious, could ~ 120-139 be put on a buildConfig service and reused in  buildConfig.js?  Thinking `$scope.builds = buildConfigService.validateBuilds(builds, action, build);` might be able to eliminate duplication.

app/scripts/services/navigate.js
```diff
@@ -108,10 +108,12 @@ angular.module("openshiftConsole")
...
 
         var url = "project/" + encodedNamespace + "/" + action + "/";
+        
         switch(kind) {
           case "Build":
-            if (resource.metadata && resource.metadata.labels && resource.metadata.labels.buildconfig) {
-              url += "builds/" + encodeURIComponent(resource.metadata.labels.buildconfig) + "/" + encodedName;
+            var buildConfigName = $filter('buildConfigForBuild')(resource);
+            if (buildConfigName) {
+              url += "builds/" + encodeURIComponent(buildConfigName) + "/" + encodedName;
```
benjaminapetersen: We could optionally use uri.js here to build the path.  
```javascript
url
   .directory('builds')
   .directory(encodeURIComponent(buildConfigName))
   .directory(encodedName)
```

app/scripts/filters/resources.js
```diff
@@ -170,10 +171,10 @@ angular.module('openshiftConsole')
       return null;
     };
   })
-  .filter('buildConfigForBuild', function(labelNameFilter, labelFilter) {
+  .filter('buildConfigForBuild', function(annotationFilter, labelNameFilter, labelFilter) {
```
benjaminapetersen: nice, this is dense!

jwforres:
@benjaminapetersen updated, @spadgett can you double check the navigate.js changes

app/scripts/services/navigate.js
```diff
@@ -98,47 +98,59 @@ angular.module("openshiftConsole")
...
         }
-        var encodedNamespace = encodeURIComponent(namespace);
 
         var name = resource;
         if (resource.metadata) {
           name = resource.metadata.name;
         }
 
-        var encodedName = encodeURIComponent(name);
+        var url = URI("/")
```
spadgett: Does this create a URL with a leading `/` ?

app/scripts/services/navigate.js
```diff
@@ -98,47 +98,59 @@ angular.module("openshiftConsole")
...
             }
             break;
           case "ImageStream":
-            url += "images/" + encodedName;
+            url.segment("images")
+              .segmentCoded(name);
             break;
           default:
-            url += kind.toLowerCase() + "s/" + encodedName;
+            url.segment(kind.toLowerCase() + "s")
```
spadgett: Not new, but `APIService.kindToResource`?

spadgett:
Two comments. navigate.js looks a lot cleaner now.

app/scripts/services/api.js
```diff
@@ -191,8 +199,11 @@ angular.module('openshiftConsole')
         AuthService.withUser();
         return;
       }
-      // Otherwise go to the error page, the server might be down.
-      Navigate.toErrorPage("Unable to load details about the server. If the problem continues, please contact your system administrator.", "API_DISCOVERY", true);
+      // Otherwise go to the error page, the server might be down.  Can't use Navigate.toErrorPage or it will create a circular dependency
+      $window.location.href = URI('error').query({
```
jwforres: in order to use APIService.kindtoResource in navigate.js I had to switch to not using Navigate.toErrorPage, otherwise there was a circular dependency.  This seemed like the better of the two options.

jwforres:
[merge]

PR #121, opened 2016-06-29
Integrating angular-key-value-editor directive for build configs...

rhamilto:
...and deployment configs.  Also updated the Environment tabs of build, deployment, and pod pages to include "Environment Variables" in the heading.

@benjaminapetersen, @jwforres, PTAL.

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -26,7 +26,8 @@ angular
     'ui.ace',
     'extension-registry',
     'as.sortable',
-    'ui.select'
+    'ui.select',
+    'key-value-editor'
   ])
   .constant("mainNavTabs", [])  // even though its not really a "constant", it has to be created as a constant and not a value
                          // or it can't be referenced during module config
@@ -227,7 +228,7 @@ angular
       .when('/project/:project/browse/other', {
         templateUrl: 'views/other-resources.html',
         controller: 'OtherResourcesController'
-      })      
+      })
       .when('/project/:project/browse/persistentvolumeclaims/:pvc', {
         templateUrl: 'views/browse/persistent-volume-claim.html',
         controller: 'PersistentVolumeClaimController'
@@ -389,7 +390,7 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
     next();
     return;
   }
-  
+
   var api = {
     k8s: {},
     openshift: {}
@@ -397,7 +398,7 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
   var apis = {};
   var API_DISCOVERY_ERRORS = [];
   var protocol = window.location.protocol + "//";
-  
+
   // Fetch /api/v1 for legacy k8s resources, we will never bump the version of these legacy apis so fetch version immediately
   var k8sBaseURL = protocol + window.OPENSHIFT_CONFIG.api.k8s.hostPort + window.OPENSHIFT_CONFIG.api.k8s.prefix;
   var k8sDeferred = $.get(k8sBaseURL + "/v1")
@@ -411,7 +412,7 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
       xhr: jqXHR
     });
   });
-  
+
   // Fetch /oapi/v1 for legacy openshift resources, we will never bump the version of these legacy apis so fetch version immediately
   var osBaseURL = protocol + window.OPENSHIFT_CONFIG.api.openshift.hostPort + window.OPENSHIFT_CONFIG.api.openshift.prefix;
   var osDeferred = $.get(osBaseURL + "/v1")
@@ -425,10 +426,10 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
       xhr: jqXHR
     });
   });
-  
+
   // Fetch /apis to get the list of groups and versions, then fetch each group/
   // Because the api discovery doc returns arrays and we want maps, this creates a structure like:
-  // { 
+  // {
   //   extensions: {
   //     name: "extensions",
   //     preferredVersion: "v1beta1",
@@ -440,11 +441,11 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
   //           daemonsets: {
   //             /* resource returned from discovery API */
   //           }
-  //         }  
+  //         }
   //       }
   //     }
   //   }
-  // } 
+  // }
   var apisBaseURL = protocol + window.OPENSHIFT_CONFIG.apis.hostPort + window.OPENSHIFT_CONFIG.apis.prefix;
   var apisDeferred = $.get(apisBaseURL)
   .then(function(data) {
@@ -455,7 +456,7 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
         preferredVersion: apiGroup.preferredVersion.version,
         versions: {}
       };
-      apis[group.name] = group;      
+      apis[group.name] = group;
       _.each(apiGroup.versions, function(apiVersion) {
         var versionStr = apiVersion.version;
         group.versions[versionStr] = {
@@ -475,7 +476,7 @@ hawtioPluginLoader.registerPreBootstrapTask(function(next) {
         }));
       });
     });
-    return $.when.apply(this, apisDeferredVersions);    
+    return $.when.apply(this, apisDeferredVersions);
   }, function(data, textStatus, jqXHR) {
     API_DISCOVERY_ERRORS.push({
       data: data,
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -44,7 +44,14 @@ angular.module('openshiftConsole')
 
     var orderByDate = $filter('orderObjectsByDate');
     var buildConfigForBuild = $filter('buildConfigForBuild');
+    var buildStrategy = $filter('buildStrategy');
     var watches = [];
+
+    // copy buildConfig and ensure it has env so that we can edit env vars using key-value-editor
+    var copyBuildConfigAndEnsureEnv = function(buildConfig) {
+      $scope.updatedBuildConfig = angular.copy(buildConfig);
+      $scope.envVars = buildStrategy($scope.updatedBuildConfig).env || [];
+    };
 
     ProjectsService
       .get($routeParams.project)
@@ -66,6 +73,32 @@ angular.module('openshiftConsole')
               });
             }
 
+            copyBuildConfigAndEnsureEnv(buildConfig);
+
+            $scope.saveEnvVars = function() {
+              $scope.envVars = _.filter($scope.envVars, 'name');
+              buildStrategy($scope.updatedBuildConfig).env = $scope.envVars;
+
+              DataService
+                .update("buildconfigs", $routeParams.buildconfig, $scope.updatedBuildConfig, context)
+                .then(function success(){
+                  // TODO:  de-duplicate success and error messages.
+                  // as it stands, multiple messages appear based on how edit
+                  // is made.
+                  $scope.alerts['saveBCEnvVarsSuccess'] = {
+                    type: "success",
+                    // TODO:  improve success alert
+                    message: $scope.buildConfigName + " was updated."
+                  };
+                }, function error(e){
+                  $scope.alerts['saveBCEnvVarsError'] = {
+                    type: "error",
+                    message: $scope.buildConfigName + " was not updated.",
+                    details: "Reason: " + $filter('getErrorDetails')(e)
+                  };
+                });
+            };
+
             // If we found the item successfully, watch for changes on it
             watches.push(DataService.watchObject("buildconfigs", $routeParams.buildconfig, context, function(buildConfig, action) {
               if (action === "DELETED") {
@@ -75,6 +108,9 @@ angular.module('openshiftConsole')
                 };
               }
               $scope.buildConfig = buildConfig;
+
+              copyBuildConfigAndEnsureEnv(buildConfig);
+
               $scope.paused = BuildsService.isPaused($scope.buildConfig);
             }));
           },
--- a/app/scripts/controllers/deploymentConfig.js
+++ b/app/scripts/controllers/deploymentConfig.js
@@ -63,6 +63,14 @@ angular.module('openshiftConsole')
     });
     AlertMessageService.clearAlerts();
 
+    // copy deploymentConfig and ensure it has env so that we can edit env vars using key-value-editor
+    var copyDeploymentConfigAndEnsureEnv = function(deploymentConfig) {
+      $scope.updatedDeploymentConfig = angular.copy(deploymentConfig);
+      _.each($scope.updatedDeploymentConfig.spec.template.spec.containers, function(container) {
+        container.env = container.env || [];
+      });
+    };
+
     var watches = [];
 
     ProjectsService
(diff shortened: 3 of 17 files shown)


app/scripts/app.js
```diff
@@ -227,7 +228,7 @@ angular
       .when('/project/:project/browse/other', {
         templateUrl: 'views/other-resources.html',
         controller: 'OtherResourcesController'
-      })      
+      })
```
benjaminapetersen: We should possibly double check and see if everyone has the same settings for eliminating extra white space....

app/scripts/controllers/buildConfig.js
```diff
@@ -65,6 +66,46 @@ angular.module('openshiftConsole')
...
+            $scope.updatedBuildConfig = angular.copy($scope.buildConfig);
+            $scope.buildStrategy = $filter('buildStrategy')($scope.updatedBuildConfig);
+            $scope.strategyType = $scope.buildConfig.spec.strategy.type;
+            // note: unlike edit/buildConfig.js, we dont need $filter('envVarsPair') because
+            // key-value-editor uses an array.
+            if(!$scope.buildStrategy.env) {
+              $scope.buildStrategy.env = [];
+            }
+            $scope.envVars = $scope.buildStrategy.env;
+
```
benjaminapetersen: I think some of these comments can be trimmed.

bower.json
```diff
@@ -41,7 +41,8 @@
     "angular-extension-registry": "1.2.6",
     "ng-sortable": "1.3.4",
     "ui-select": "angular-ui-select#0.16.1",
-    "matchHeight": "0.6.0"
+    "matchHeight": "0.6.0",
+    "angular-key-value-editor": "https://github.com/benjaminapetersen/angular-key-value-editor.git#v0.0.9"
```
benjaminapetersen: Just making a note on our conversation about pulling `angular-key-value-editor` under the `openshift` org temporarily, then possibly to `patternfly` as a better long term resting place.

app/scripts/controllers/buildConfig.js
```diff
@@ -65,6 +66,40 @@ angular.module('openshiftConsole')
               });
             }
 
+            $scope.updatedBuildConfig = angular.copy($scope.buildConfig);
+            $scope.buildStrategy = $filter('buildStrategy')($scope.updatedBuildConfig);
```
jwforres: dont have a local variable and a scope variable that use the same name and are not pointing to the same thing

jwforres: actually, does this even need to be on the scope object, the only references to buildStrategy in the views that i am seeing are using the filter, not this object

app/scripts/controllers/buildConfig.js
```diff
@@ -65,6 +66,40 @@ angular.module('openshiftConsole')
...
+            $scope.strategyType = $scope.buildConfig.spec.strategy.type;
```
jwforres: doesn't look like this is being used anywhere

app/scripts/controllers/buildConfig.js
```diff
@@ -65,6 +66,40 @@ angular.module('openshiftConsole')
...
+            if(!$scope.buildStrategy.env) {
+              $scope.buildStrategy.env = [];
+            }
+            $scope.envVars = $scope.buildStrategy.env;
+
+            $scope.saveEnvVars = function() {
+              $scope.envVars = _.reject(  $scope.envVars, function(env) {
+                return env.name === "";
```
jwforres: any reason not to have `return !env.name`  that way if anything ever changed and you got back null / undefined name then it would still reject it

app/scripts/controllers/buildConfig.js
```diff
@@ -65,6 +66,40 @@ angular.module('openshiftConsole')
...
+                  // is made.
+                  $scope.alerts['saveBCEnvVarsSuccess'] = {
+                    type: "success",
+                    // TODO:  improve success alert
+                    message: $scope.buildConfigName + " was updated."
+                  };
+                }, function error(){
+                  $scope.alerts['saveBCEnvVarsError'] = {
+                    type: "error",
+                    // TODO:  improve error alert
```
jwforres: we have a helper filter already for extracting failures from the response, go ahead and implement this TODO please

app/scripts/controllers/buildConfig.js
```diff
@@ -74,6 +109,7 @@ angular.module('openshiftConsole')
                 };
               }
               $scope.buildConfig = buildConfig;
+              $scope.updatedBuildConfig = angular.copy(buildConfig);
```
jwforres: if anything updates the build config it could wipe out any changes you've made in the form, background controllers could potentially trigger an update to the BC's status (example, the latest build number that was run)

app/scripts/controllers/deploymentConfig.js
```diff
@@ -63,6 +63,17 @@ angular.module('openshiftConsole')
     });
     AlertMessageService.clearAlerts();
 
+    // copy deploymentConfig and ensure it has env so that we can edit env vars using key-value-editor
+    var copyDeploymentConfigAndEnsureEnv = function(deploymentConfig) {
+      $scope.updatedDeploymentConfig = angular.copy(deploymentConfig);
+      $scope.updatedDeploymentConfig.spec.template.spec.containers = _.map($scope.updatedDeploymentConfig.spec.template.spec.containers, function(container){
```
jwforres: wouldn't this be easier to read as
```
_.each($scope.updatedDeploymentConfig.spec.template.spec.containers, function(container) {
  container.env = container.env || [];
}
```
updatedDeploymentConfig is already a copy, i don't see the need for creating a new array (which i assume was the reason for using map in the first place)?  @benjaminapetersen  ?

app/scripts/controllers/deploymentConfig.js
```diff
@@ -87,6 +98,34 @@ angular.module('openshiftConsole')
             $scope.deploymentConfig = deploymentConfig;
             updateHPAWarnings();
             ImageStreamResolver.fetchReferencedImageStreamImages([deploymentConfig.spec.template], $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, context);
+            copyDeploymentConfigAndEnsureEnv(deploymentConfig);
+            $scope.saveEnvVars = function() {
+              // copy again to clean off any of angulars $$ vars from the template
+              var updatedDeploymentConfig = angular.copy($scope.updatedDeploymentConfig);
+              updatedDeploymentConfig.spec.template.spec.containers = _.map(updatedDeploymentConfig.spec.template.spec.containers, function(container) {
+                container.env = _.reject(container.env, function(env){
+                  return env.name === "" && env.value ==="";
```
jwforres: same comment as before,  `!env.name && !env.value`

jwforres: actually just `!env.name` this never got updated to only reject when the name is empty, you may want to set an env to an empty value

app/scripts/controllers/deploymentConfig.js
```diff
@@ -87,6 +98,34 @@ angular.module('openshiftConsole')
             $scope.deploymentConfig = deploymentConfig;
             updateHPAWarnings();
             ImageStreamResolver.fetchReferencedImageStreamImages([deploymentConfig.spec.template], $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, context);
+            copyDeploymentConfigAndEnsureEnv(deploymentConfig);
+            $scope.saveEnvVars = function() {
+              // copy again to clean off any of angulars $$ vars from the template
+              var updatedDeploymentConfig = angular.copy($scope.updatedDeploymentConfig);
+              updatedDeploymentConfig.spec.template.spec.containers = _.map(updatedDeploymentConfig.spec.template.spec.containers, function(container) {
```
jwforres: same question on _.map vs just _.each

app/scripts/controllers/deploymentConfig.js
```diff
@@ -87,6 +98,34 @@ angular.module('openshiftConsole')
...
+                  // is made.
+                  $scope.alerts['saveDCEnvVarsSuccess'] = {
+                    type: "success",
+                    // TODO:  improve success alert
+                    message: $scope.deploymentConfigName + " was updated."
+                  };
+                }, function error(){
+                  $scope.alerts['saveDCEnvVarsError'] = {
+                    type: "error",
+                    // TODO:  improve error alert
```
jwforres: implement error message

app/styles/main.less
```diff
@@ -81,6 +81,9 @@
 @import "@{pf-less-path}/type.less";
 @import "@{pf-less-path}/vertical-nav.less";
 
+/* Angular Key Value Editor */
+@import "../../bower_components/angular-key-value-editor/src/less/angular-key-value-editor.less";
```
jwforres: would prefer to keep all the relative path calculations in variables at the very top of the file, create a new var for bower-components-path

app/scripts/controllers/buildConfig.js
benjaminapetersen: Seems like the right thing to do might be to ignore the watch while editing and let the "submit" fail because then the user would know that someone else edited/changed it already?

spadgett: That's what we do elsewhere. It would be nice to give the user a warning though before they go through the trouble editing.

app/scripts/controllers/deploymentConfig.js
benjaminapetersen: The `_.each` version is definitely shorter!  I think this was just a decision of favoring `map` over `each`, but I agree, each is a good call here.

app/scripts/controllers/buildConfig.js
benjaminapetersen: `details: "Reason: " + $filter('getErrorDetails')(e)`

benjaminapetersen:
```javascript
 $scope.alerts['saveBCEnvVarsError'] = {
  type: "error",
  message: $scope.buildConfigName + " was not updated."
  details: "Reason: " + $filter('getErrorDetails')(e)
}

```

app/scripts/controllers/buildConfig.js
benjaminapetersen: Yeah, easy would be a banner "This resource has been updated. refresh?" and let the user decide if it should replace what they have.  Harder is the diff...

@rhamilto just to keep you in the loop on this chatter :)

rhamilto: In addressing the other comments here, I noticed that changes to env vars made via `Edit YAML` are not reflected in the key-value-editor.  I assume this aforementioned bug is related to this same issue.

rhamilto:
@jwforres, I believe I've addressed all comments.  

Given that this only adds the new key-value-editor to the environment tab of build configs and deployment configs, do we want to wait to merge until we've also swapped the instances of osc-key-values for key-value-editor, too?  Or do we want a separate story and PR?  We sort of blew up the scope of the original story...

benjaminapetersen:
I def like separate PRs for chunks of this rather than carrying one big PR, but can also see the concern of having both editors existing during a transition period.  Curious your thoughts as well @jwforres

jwforres:
im ok with separate PRs to keep the review scope down but I would prefer to
not complete the card till we have them all updated.  or add a new card to
update the create form.

benjaminapetersen:
@rhamilto I see the `angular-key-value-editor` entry in `bower.json` is registered now.  Looks like this might be good?

rhamilto:
@benjaminapetersen, I think so, but the build is failing.  And it's not clear to me why.

benjaminapetersen:
aye, taking a look.

benjaminapetersen:
line 915 in the log claims its not available:
```
  Error: [$injector:nomod] Module 'key-value-editor' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
```
hmm.

benjaminapetersen:
I see it in the index.html, app.js and dist files.  Still looking.

benjaminapetersen:
Yup, running `grunt test` locally will result in the same problems.

benjaminapetersen:
@rhamilto, the `test/karma.conf.js` file just needs to be updated and add the main file for `angular-key-value-editor` at ~ line 55:
```javascript
// lots of stuff above
"bower_components/angular-extension-registry/dist/angular-extension-registry.min.js",
"bower_components/angular-key-value-editor/dist/angular-key-value-editor.min.js", // <-- add this line
"bower_components/ng-sortable/dist/ng-sortable.js",
// lots of stuff below
```
tests should run and pass then. you can give it a spin locally with `grunt test`, but it ran for me.

rhamilto:
@benjaminapetersen, thank you!  Success!  @jwforres or @spadgett. ready for merge.

app/scripts/controllers/deploymentConfig.js
```diff
@@ -87,6 +95,33 @@ angular.module('openshiftConsole')
             $scope.deploymentConfig = deploymentConfig;
             updateHPAWarnings();
             ImageStreamResolver.fetchReferencedImageStreamImages([deploymentConfig.spec.template], $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, context);
+            copyDeploymentConfigAndEnsureEnv(deploymentConfig);
+            $scope.saveEnvVars = function() {
+              // copy again to clean off any of angulars $$ vars from the template
```
spadgett: Is this needed? Angular toJSON strips these, which I'd guess is called under the hood when we make the HTTP request.

https://docs.angularjs.org/api/ng/function/angular.toJson

app/scripts/controllers/deploymentConfig.js
```diff
@@ -87,6 +95,33 @@ angular.module('openshiftConsole')
...
+              var updatedDeploymentConfig = angular.copy($scope.updatedDeploymentConfig);
+              _.each(updatedDeploymentConfig.spec.template.spec.containers, function(container) {
+                container.env = _.reject(container.env, function(env){
+                  return !env.name;
+                });
```
spadgett: FYI you should be able to use

```js
container.env = _.filter(container.env, 'name');
```

app/scripts/controllers/buildConfig.js
```diff
@@ -42,8 +42,15 @@ angular.module('openshiftConsole')
       $scope.selectedTab[$routeParams.tab] = true;
     }
 
+    // copy buildConfig and ensure it has env so that we can edit env vars using key-value-editor
+    var copyBuildConfigAndEnsureEnv = function(buildConfig) {
+      $scope.updatedBuildConfig = angular.copy(buildConfig);
+      $scope.envVars = $filter('buildStrategy')($scope.updatedBuildConfig).env || [];
```
spadgett: If you move the func after the `buildStrategy` declaration just below, you won't have the call `$filter`.

```js
$scope.envVars = buildStrategy($scope.updatedBuildConfig).env || [];
```

app/scripts/controllers/deploymentConfig.js
```diff
@@ -63,6 +63,14 @@ angular.module('openshiftConsole')
     });
     AlertMessageService.clearAlerts();
 
+    // copy deploymentConfig and ensure it has env so that we can edit env vars using key-value-editor
+    var copyDeploymentConfigAndEnsureEnv = function(deploymentConfig) {
+      $scope.updatedDeploymentConfig = angular.copy(deploymentConfig);
+      _.each($scope.updatedDeploymentConfig.spec.template.spec.containers, function(container) {
+        container.env = container.env || [];
+      });
```
spadgett: Same comment as below.

app/scripts/controllers/deploymentConfig.js
rhamilto: @benjaminapetersen?

app/scripts/controllers/deploymentConfig.js
rhamilto: Neat!  Fixed (commit forthcoming).

app/scripts/controllers/buildConfig.js
rhamilto: Good catch.  Thanks.

app/scripts/controllers/deploymentConfig.js
rhamilto: As below?  Sorry, I'm not clear on what you're getting at here.

spadgett: Sorry disregard

app/scripts/controllers/deploymentConfig.js
benjaminapetersen: Might have been overly cautious doing the extra copy w/o first trying it.  I'd say remove & test it. If it still works, lets drop the extra angular.copy().  

https://github.com/angular/angular.js/blob/master/src/ng/http.js
Looks like `$http` uses `.toJSON` (via `.serializeValue`)

rhamilto:
@spadgett, thanks for the suggestions.  I've integrated them and think we're now ready for merge.  *crosses fingers*

spadgett:
Can be a follow-on, but I'd like to see a warning if you navigate away without saving. It'd be easy to miss the save button. Maybe also a reset button.

app/scripts/controllers/buildConfig.js
```diff
@@ -66,6 +73,32 @@ angular.module('openshiftConsole')
...
             }
 
+            copyBuildConfigAndEnsureEnv(buildConfig);
+
+            $scope.saveEnvVars = function() {
+              $scope.envVars = _.filter($scope.envVars, 'name');
+              buildStrategy($scope.updatedBuildConfig).env = $scope.envVars;
+
+              DataService
+                .update("buildconfigs", $routeParams.buildconfig, angular.copy($scope.updatedBuildConfig), context)
```
spadgett: Sorry, just noticed: do we need to copy again here? `updatedBuildConfig` is already a copy, right?

spadgett:
I really like the way the new editor works.

spadgett:
One last question on angular.copy, otherwise LGTM.

rhamilto:
> One last question on angular.copy, otherwise LGTM.

Another good catch, @spadgett.  It wasn't needed, so I removed it.

benjaminapetersen:
Nice, looking forward to this landing!

spadgett:
@rhamilto ready to merge?

rhamilto:
👍

spadgett:
[merge]

PR #124, opened 2016-06-30
Refactor BuildsService to remove $scope

benjaminapetersen:
Does a few things:

- stops passing `$scope` to the service
- returns promises from the BuildConfig service methods, giving control of messages to the controllers
- updates service to a singleton

This is part of an old PR from origin refactoring both the BuildService & DeploymentsService.

Further improvements:

- cache our error message objects in a service so we can eliminate the duplication of messages. These could probably be one liners: `$scope.alerts['build-success'] = AlertMessageService.get('build-success')(build);` 

If/when this passes I'll update the DeploymentsService to match.

@jwforres @spadgett


app/scripts/services/builds.js
```diff
@@ -2,152 +2,85 @@
...
+        var req = {
+          kind: "BuildRequest",
+          apiVersion: "v1",
+          metadata: {
+            name: buildName
+          }
+        };
+        return DataService.create("builds/clone", buildName, req, context);
+      },
+      // Q: should this be in a buildConfig service?
```
jwforres: what are you asking here exactly?  The build config is what is being paused, not a particular build.  it means that builds can not be launched from the build config currently

app/scripts/services/builds.js
```diff
@@ -2,152 +2,85 @@
...
+      isPaused: isPaused,
+      // Q: should this be in a buildConfig service?
```
jwforres: same question, whats your concern?

app/scripts/services/builds.js
```diff
@@ -2,152 +2,85 @@
...
-          return;
+        if (isPaused(buildConfig)) {
+          return false;
         }
 
-        if (resource.namespace && resource.namespace !== _.get(buildConfig, 'metadata.namespace')) {
-          return;
+        return true;
+      },
+      // Q: should this be in a buildConfig service?
```
jwforres: ?

app/scripts/services/builds.js
```diff
@@ -2,152 +2,85 @@
...
+        }))) {
+          Logger.warn('Could not parse "pipeline.alpha.openshift.io/uses" annotation', uses);
           return;
         }
 
-        depoymentConfigs.push(resource.name);
-      });
-
-      return depoymentConfigs;
+        return _.compact(
```
jwforres: @spadgett please take a look at this part since you wrote the original

@benjaminapetersen try to keep refactors scoped to one thing at a time, its easier to review whats being changed and limits the regression risk.  On that note, did you modify the contents of any of the other functions in the service? (other than removing anything setting alerts)

app/scripts/services/builds.js
benjaminapetersen: Growth over time, do we want a `buildService` & `buildConfigService`?  I dont think it matters at this point, there isn't a ton here.

app/scripts/services/builds.js
benjaminapetersen: No prob, I can pull out this change.  The rest should be exactly the same other than returning the promises so the controllers can handle the `.then()`

benjaminapetersen:
rebased/addressed comments

app/scripts/services/builds.js
```diff
@@ -2,162 +2,93 @@
...
-      } catch(e) {
-        Logger.warn('Could not parse "pipeline.alpha.openshift.io/uses" annotation', e);
-        return;
-      }
-
-      var depoymentConfigs = [];
-      _.each(uses, function(resource) {
-        if (!resource.name) {
-          return;
+        if (this.isPaused(buildConfig)) {
```
jwforres: I don't think this.isPaused will work now, because you no longer have an instance of an object, need to pull out isPaused into a var function before the map so its available to set onto the map and to be called from this function

benjaminapetersen: I originally did that and had `if (isPaused(buildConfig)) {` but then backed out and tested it.  It still works, I believe because the `return { /* methods */ }` is still an instance of an object, its just not build w/a constructor. 

I'll give it one more test spin though just to make sure I didn't miss something.  I probably prefer to do the standalone functions anyway!

benjaminapetersen:
Rebased/addressed comments again.  @jwforres for consistency I made each fn standalone & used the return like an `exports` at the end of the `BuildsService` file.

jwforres:
sorry i missed this one and now it needs rebasing again :-\

benjaminapetersen:
rebased

jwforres:
[merge]

PR #125, opened 2016-06-30
[WIP][DO NOT MERGE] Fix Issue #109 - Table Empty Messages

### LINKED ISSUE #109: See whether "No <whatever> to show" messages look better outside of the tables
Original issue https://github.com/openshift/origin/issues/5412

Per @spadgett comment, there is an empty state for tables in patternfly:
https://www.patternfly.org/patterns/table-view/

MarkDeMaria:
This is to get the ball rolling on how I should approach these empty state table messages, as some pages have a 'Create Button' option when the table is empty and some pages do not. I know that some pages have tables in their sub-pages (the Events tab of a Deployment, the page for an Image Stream, the Events tab of a Pod, the page for a Service). If I have missed any major pages with tables please let me know!

Fixes #109

Those with Create buttons that I have found:
1. Routes

Those without Create Buttons that I have found:
1. Deployments
2. Events
3. Image Streams
4. Pods
5. Services
6. Storage
7. Other Resources

For pages without 'Create Buttons,' which pages should have create buttons added and which should not? Below are three screenshots - The first is an Empty State Table without a 'create button,' and the other two are tables with 'create buttons,' but one keeps its original 'create button' in the top right while the other removes it until there is something to show in the table.

Builds - No create buttons
image

Routes - Two Buttons
image

Routes - One Button in Table (Other button will appear after there is >=1 route present)
image


MarkDeMaria:
@benjaminapetersen

benjaminapetersen:
Agree, the big blue center button is probably sufficient in these cases.

app/views/browse/routes.html
```diff
@@ -24,7 +24,22 @@
         <div class="container-fluid">
           <div class="row">
             <div class="col-md-12 gutter-top">
-              <table class="table table-bordered table-hover table-mobile">
+              <div ng-if="(routes | hashSize) == 0" class="blank-slate-pf">
```
benjaminapetersen: You may be able to do `ng-if="!(routes | hashSize)"` and `ng-if="routes | hashSize"` and drop the `== 0`.

app/views/builds.html
```diff
@@ -21,7 +21,16 @@
         <div class="container-fluid">
           <div class="row">
             <div class="col-md-12 gutter-top">
-              <table class="table table-bordered table-hover table-mobile">
+              <div ng-if="((buildsByBuildConfig | hashSize) == 0)" class="blank-slate-pf">
```
benjaminapetersen: Same as above, prob don't need the extra `()` and `== 0`.

spadgett:
We should either reuse our existing empty state style (with no background color) or switch the other, existing empty state messages to use `blank-state-pf`.

MarkDeMaria:
@jwforres

jwforres:
you should have both the button in the empty state message and the button in the top right, you shouldn't take away an action from somewhere a user expects it unless they are forbidden from doing that action

jwforres:
You won't be adding create buttons for any pages that do not already have them, if they aren't there already it is because we haven't implemented a GUI creation form for them yet.

benjaminapetersen:
@spadgett @MarkDeMaria I think I prefer the empty state message that doesn't have the grey background.

app/views/browse/routes.html
spadgett: This is in our code so much we might add an `isEmpty` filter.

benjaminapetersen: Agree!

benjaminapetersen:
Fix issue #109 (may need to update the description at the top to get github to automatically close when merged)

spadgett:
Added it to the description.

MarkDeMaria:
@benjaminapetersen @jwforres @spadgett

**Builds**
image

**Builds->Build->Event Tab**
image

**Deployments (Uses PF-Image Icon)**
image

**Deployments->Deployment->Events Tab**
Looks similar to the Build->Event Tab. Current dev env didn't allow me to get a sample of this. Will update when I am able to!

**Events (Uses FA-Calendar Icon)**
image

**Image Streams**
image

**Image Streams->Image Stream**
image

**Pods (Uses PF-Service Icon)**
image

**Pods->Pod->Event Tab**
image

**Routes**
image

**Services**
image

**Services->Service**
Looking for feedback on this page. Is it possible for a service to exist without ports defined? If not, I will not alter this page.

**Storage**
image

**Other Resources**
image

benjaminapetersen:
Do you think any of these would benefit from some help text, like "no builds" has: 'To create your project's first build, visit our Build Configuration Wizard'?  

Text size (example: other resources, "Select a resource from the list above") seems to be a bit too large, fighting with the page heading for attention.  This instance could be updated as "Nothing selected" with smaller help text "Select a resource from the list above".

jwforres:
Agree with Ben on the text seeming very large.

The Events icon isn't right, a calendar icon has too much meaning, it would
make me think we supported calendering or scheduling in the product somehow
(we will have scheduled jobs eventually maybe we would use it there).  Our
events are more like notifications.

benjaminapetersen:
@stevekuznetsov rosie seems a little preoccupied with this PR too.

stevekuznetsov:
The robot has been disciplined. Let me know if new behavior emerges.

benjaminapetersen:
Sad robot...
screen shot 2016-08-01 at 9 56 37 am

MarkDeMaria:
@jwforres @benjaminapetersen 

**Events** - Updated to use monitor icon and h2 instead of h1
image

jwforres:
h2 looks better, can't decide on the monitoring icon, but its better than the calendar icon.

The big box just still doesnt feel right to me on these white pages, lets talk visual design next week when robb and steve are back

spadgett:
> The big box just still doesnt feel right to me on these white pages, lets talk visual design next week when robb and steve are back

Can we reuse the style we already have on other pages? I'd rather not start using class `pf-blank-state` until we're ready to switch to that everywhere.

https://github.com/openshift/origin-web-console/blob/master/app/styles/_core.less#L173-L204

MarkDeMaria:
@jwforres Can we talk visual design with @sg00dwin & @rhamilto on Friday to get this PR rolling again? Thanks!

Issue #129, opened 2016-06-30
Simplify get and watch in controllers


benjaminapetersen:
We have a lot of this:

``` javascript
DataService
  .get('something', $routeParams.something, context)
  .then(function resolve(theThing) {
     // do lots of things...
    watches.push(DataService
      .watchObject('something', $routeParams.something, context, function(theThing) {
         // do lots of the same things again...
      }));      
  }, function reject(err) {

  });
```

I'm guessing this is to get at any cached data if we have it while waiting for the watch.  I'm wondering if we can eliminate the duplication by one of these:

``` javascript
// option 1: 
// ---------------------------------
// get data out of cache (DataService._data(key)) then watch it
$scope.theThing = DataService.getCached(itsName);
DataService
  .watchObject('something', itsName, context, function(theThing) {
    $scope.theThing = theThing;
  }, reject);




// Option 2: reuse resolve/reject functions
// ---------------------------------
var resolve = function(theThing) {
  // do stuff
};
var reject = function(err) {

}

DataService
  .get('something', itsName, context)
  .then(resolve, reject);
DataService
  .watchObject('something', itsName, context, resolve, reject);




// option 3:
// ---------------------------------
// perhaps we have to ensure the GET happens before the WATCH
// so that we dont get out of sync data. 
// has an extra anonymous function but otherwise seems clean
  var resolve = function(theThing) {
    $scope.theThing = theThing;
    // do other happy stuff...
  };
  var reject = function(err) {
    // do sad stuff...
  }

  DataService
    .get('something', itsName, context)
    .then(function(theThing) {
      resolve(theThing);
      DataService
        .watchObject('something', itsName, context, resolve, reject); // reuse resolve, reject
    }, reject);


```

Prob option 3 would win us the most cleanup.

Thoughts @jwforres @spadgett ?

TODOs to resolve this issue
- [x] update `controllers/build.js` (PR #133)
- [x] update `controllers/buildConfig.js`
- [x] update `controllers/deployment.js`
- [ ] update `controllers/deploymentConfig.js`
- [x] update `controllers/image.js`
- [x] update `controllers/persistentVolumeClaims.js`
- [x] update `controllers/pod.js`
- [x] ~~update `controllers/pods.js`~~
- [x] update `controllers/route.js`
- [x] update `controllers/service.js`

benjaminapetersen:
Btw, my initial thought was to do a `.get()` internally to `.watch()` and use the same callback function, but don't think thats a good idea.

[...]

PR #133, opened 2016-07-05
Concept for simplifying get & watch in builds controller

benjaminapetersen:
Related to issue #129.  Attempting to eliminate some copy/paste duplication via reusable resolve & reject functions.

@jwforres @spadgett

### DIFF
--- a/app/scripts/controllers/build.js
+++ b/app/scripts/controllers/build.js
@@ -42,8 +42,6 @@ angular.module('openshiftConsole')
       $scope.selectedTab[$routeParams.tab] = true;
     }
 
-    var buildConfigForBuild = $filter('buildConfigForBuild');
-
     var watches = [];
 
     var setLogVars = function(build) {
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/build.js
```diff
@@ -67,72 +126,17 @@ angular.module('openshiftConsole')
...
-          // success
-          function(build) {
-
-            $scope.loaded = true;
-            $scope.build = build;
-            setLogVars(build);
-            var buildNumber = $filter("annotation")(build, "buildNumber");
-            if (buildNumber) {
-              $scope.breadcrumbs[2].title = "#" + buildNumber;
-            }
```
benjaminapetersen: The request flow may be easier to follow here, but I have no problem with the inline anonymous functions.  Gain is mostly in DRY.

app/scripts/controllers/build.js
```diff
@@ -67,72 +126,17 @@ angular.module('openshiftConsole')
...
-            buildConfigName = build.metadata.labels.buildconfig;
-            buildName = build.metadata.name;
-          }
+            watches.push(DataService.watchObject("builds", $routeParams.build, context, buildResolved));
+            watches.push(DataService.watchObject("buildconfigs", $routeParams.buildconfig, context, buildConfigResolved));
+          }, buildRejected);
 
-          updateCanBuild();
-        }));
+        watches.push(DataService.watch("builds", context, buildsResolved));
```
jwforres: actually... are we even using builds for anything anymore, we use to use it in canBuild to disable the Start Build button, but stopped doing that when build serialization went into the API.  I think we can stop watching builds altogether now.

benjaminapetersen: Oh nice.  Yeah, I didn't even look at that, was just going for a simple ctrl to prototype sharing the get/watch callback but cutting more code is a win! :smile:

app/scripts/controllers/build.js
```diff
@@ -67,72 +126,17 @@ angular.module('openshiftConsole')
...
-            }
-          });
-
-          var buildConfigName;
-          var buildName;
-          if (build) {
-            buildConfigName = build.metadata.labels.buildconfig;
-            buildName = build.metadata.name;
-          }
+            watches.push(DataService.watchObject("builds", $routeParams.build, context, buildResolved));
```
jwforres: now that we've plumbed the labelSelector / fieldSelector stuff, we should probably update the watchObject in DataService to tack on the fieldSelector for the name of the object so we aren't pulling a bunch of data we don't need, doesn't need to be done in this PR, just thinking out loud :)

benjaminapetersen: yeah definitely!

benjaminapetersen:
Cool, will do.  I'll do a file here, a file there & slowly work out some more of these over time.

app/scripts/controllers/build.js
benjaminapetersen: I'll create an issue for tracking that.

benjaminapetersen:
Updated & dropped the watch builds

benjaminapetersen:
and rebased. should be set!

jwforres:
[merge]

PR #134, opened 2016-07-05
Add left margin to alert close button

spadgett:
Prevents alert message text from running up to the edge of the button.

@jwforres PTAL


jwforres:
[merge]

PR #138, opened 2016-07-07
Visualize routes with multiple services on overview

spadgett:
https://trello.com/c/Eddu8kg4

openshift_web_console

/cc @jwforres

### DIFF
--- a/app/views/overview/_dc.html
+++ b/app/views/overview/_dc.html
@@ -4,15 +4,22 @@
       Service:
       <a ng-href="{{service | navigateResourceURL}}">{{service.metadata.name}}</a>
     </div>
-    <div>
+    <div class="service-metadata">
       <span ng-if="!scalableDeploymentByConfig[dcName]">Deploying...</span>
-      <relative-timestamp ng-if="scalableDeploymentByConfig[dcName]" timestamp="scalableDeploymentByConfig[dcName].metadata.creationTimestamp"></relative-timestamp>
+      <div ng-if="scalableDeploymentByConfig[dcName] && weightByService[service.metadata.name]">
+        <ng-include src="'views/overview/_traffic-percent.html'"></ng-include>
+      </div>
     </div>
   </div>
   <div class="deployment-header">
     <div class="rc-header">
-      Deployment:
-      <a ng-href="{{deploymentConfigs[dcName] | navigateResourceURL}}">{{dcName}}</a>
+      <div>
+        Deployment:
+        <a ng-href="{{deploymentConfigs[dcName] | navigateResourceURL}}">{{dcName}}</a>
+      </div>
+      <div>
+        <relative-timestamp ng-if="scalableDeploymentByConfig[dcName]" timestamp="scalableDeploymentByConfig[dcName].metadata.creationTimestamp"></relative-timestamp>
+      </div>
     </div>
     <div column flex class="shield" ng-if="scalableDeploymentByConfig[dcName]"
           ng-class="{ 'shield-lg': (scalableDeploymentByConfig[dcName] | annotation: 'deploymentVersion').length > 3 }">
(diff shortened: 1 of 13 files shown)


spadgett:
@jwforres without the gaps

screen shot 2016-07-07 at 3 20 12 pm

liggitt:
do we want something that will work with more than two services?

spadgett:
@liggitt Yes, we fall back to a different view if there are more than two. @ajacobs21e is working on designs, but currently we just show the progress bar on the right for each.

app/views/overview/_dc.html
```diff
@@ -4,9 +4,13 @@
...
       <a ng-href="{{service | navigateResourceURL}}">{{service.metadata.name}}</a>
     </div>
-    <div>
+    <div class="service-metadata">
       <span ng-if="!scalableDeploymentByConfig[dcName]">Deploying...</span>
-      <relative-timestamp ng-if="scalableDeploymentByConfig[dcName]" timestamp="scalableDeploymentByConfig[dcName].metadata.creationTimestamp"></relative-timestamp>
+      <div ng-if="scalableDeploymentByConfig[dcName] && weightByService[service.metadata.name]">
+        <ng-include src="'views/overview/_traffic-percent.html'"></ng-include>
+      </div>
+      <!-- TODO: Move under the deployment tile -->
```
jwforres: assume we need to do this TODO still

spadgett: Yeah, having a hard time finding a place for it though.

app/views/overview/_traffic-percent.html
```diff
@@ -0,0 +1,17 @@
+<div ng-if="alternateServices.length">
+  <span class="visible-xs visible-sm">
+    Traffic {{(weightByService[service.metadata.name] / totalWeight) | percent}}
```
jwforres: probably a chance of us having a divide by zero here once we get them to allow zero for weight, unless we require at least one non-zero weight

app/views/overview/_traffic-percent.html
```diff
@@ -0,0 +1,17 @@
...
+  </span>
+  <div class="hidden-xs hidden-sm">
+    <span class="traffic-label">Traffic</span>
+    <!-- Calculate the width of the progress bar in px (max 250px) -->
+    <div class="progress progress-sm"
+         ng-style="{ width: ((weightByService[service.metadata.name] / totalWeight * 250) | number) + 'px'}">
```
jwforres: same here, divide by zero chance

spadgett:
Suggestion from @ajacobs21e to use different shades of blue for each service when more than 2. @jwforres @ajacobs21e Thoughts? I think it works when there are 2 services and the bars touch, but not sure about here. We also use different blues to indicate pod status (not ready, running, terminating). I wonder if people will think it's the same for services.

screen shot 2016-07-11 at 1 01 21 pm

spadgett:
Same blue for comparison:

screen shot 2016-07-11 at 1 08 16 pm

jwforres:
hmm the different blues when they are separated look weird to me... looking at it I would expect that the shade means something, possibly that it coincides with the percentage of traffic that service is receiving, like all of those ones that were 17% would be the same shade

spadgett:
@jwforres OK, this is probably ready for another look.

jwforres:
[merge]

ajacobs21e:
> We also use different blues to indicate pod status (not ready, running, terminating)

@spadgett Do you have a master list of all the possible states and how they're visualized in the UI (color, icon, etc)?

spadgett:
@ajacobs21e here are the statuses and colors for the donut

https://github.com/openshift/origin-web-console/blob/1971a40cd061f1095e54043f0d5a859f80448476/app/scripts/directives/podDonut.js#L91-L104

PR #139, opened 2016-07-07
Set a reasonable y-axis max for small metrics values

spadgett:
@jwforres PTAL

### DIFF
--- a/app/scripts/directives/metrics.js
+++ b/app/scripts/directives/metrics.js
@@ -65,6 +65,8 @@ angular.module('openshiftConsole')
             chartPrefix: "memory-",
             convert: bytesToMiB,
             containerMetric: true,
+            // The sparkline y-axis will always extend to at least this value.
+            smallestYAxisMax: 100,
             datasets: [
               {
                 id: "memory/usage",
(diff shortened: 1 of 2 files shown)


app/scripts/directives/metrics.js
```diff
@@ -351,6 +373,9 @@ angular.module('openshiftConsole')
             });
           } else {
             sparklineByMetric[chartId].load(sparklineData);
+            sparklineByMetric[chartId].axis.max({
+              y: yAxisMax
```
jwforres: and i assume when yAxisMax is undefined because the data is bigger that this being undefined will then tell c3 to do its own thing?

jwforres:
just the one question, otherwise LGTM and can mergeg

app/scripts/directives/metrics.js
spadgett: Right, the behavior will be as before for larger values.

spadgett:
[merge]

app/scripts/directives/metrics.js
jwforres: [merge]

PR #140, opened 2016-07-07
Add cluster quota details to the project settings page

jwforres:
clusterquota

### DIFF
--- a/app/scripts/controllers/settings.js
+++ b/app/scripts/controllers/settings.js
@@ -11,13 +11,11 @@
 angular.module('openshiftConsole')
   .controller('SettingsController', function ($routeParams, $scope, DataService, ProjectsService, AlertMessageService, $filter, $location, LabelFilter, $timeout, Logger, annotationFilter, annotationNameFilter) {
     $scope.projectName = $routeParams.project;
-    $scope.quotas = {};
     $scope.limitRanges = {};
     $scope.limitsByType = {};
     $scope.labelSuggestions = {};
     $scope.alerts = $scope.alerts || {};
-    $scope.emptyMessageQuotas = "Loading...";
-    $scope.quotaHelp = "Limits resource usage within the project.";
+    $scope.quotaHelp = "Limits resource usage within this project.";
     $scope.emptyMessageLimitRanges = "Loading...";
     $scope.limitRangeHelp = "Defines minimum and maximum constraints for runtime resources such as memory and CPU.";
     $scope.renderOptions = $scope.renderOptions || {};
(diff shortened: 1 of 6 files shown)


jwforres:
@spadgett PTAL

fyi @deads2k

app/scripts/controllers/settings.js
```diff
@@ -87,6 +87,21 @@ angular.module('openshiftConsole')
           Logger.log("quotas", $scope.quotas);
         });
 
+        DataService.list("appliedclusterresourcequotas", context, function(quotas) {
+          $scope.clusterQuotas = quotas.by("metadata.name");
+          $scope.emptyMessageQuotas = "There are no resource quotas set on this project.";
```
spadgett: The same scope var is set when listing resourcequotas. I think we can set it once outside the list calls.

app/scripts/controllers/settings.js
```diff
@@ -87,6 +87,21 @@ angular.module('openshiftConsole')
...
+          $scope.namespaceUsageByClusterQuota = {};
+          _.each($scope.clusterQuotas, function(quota, quotaName) {
+            if (quota.status) {
+              var namespaceUsage = _.find(quota.status.namespaces, function(ns) {
+                return ns.namespace === $routeParams.project;
+              });
```
spadgett: FYI there is a shorthand syntax

```js
_.find(quota.status.namespaces, { namespace: $routeParams.project });
```

app/scripts/controllers/settings.js
jwforres: hmm, well its set to 'Loading...' until one of the list calls returns

spadgett: Hm, then it will switch to the message for both when one loads even if the other is still pending?

jwforres: yeah im trying to decide what to do with it, because we only show the
message if both lists are empty

app/scripts/directives/quotaUsageChart.js
```diff
@@ -131,30 +127,39 @@ angular.module('openshiftConsole')
 
         var chart;
         var updateChart = function() {
+          var hasCrossProject = typeof $scope.crossProjectUsed !== 'undefined';
```
spadgett: `var hasCrossProject = $scope.crossProjectUsed !== undefined;` ?

app/scripts/controllers/settings.js
spadgett: `DataService.list` doesn't return a promise, does it? Then we could use `$q.all` to set the message.

jwforres: nope it doesnt

jwforres: its ugly but i can handle it with two booleans...

spadgett: Or check if `$scope.clusterQuotas` and `$scope.quotas` are both defined and show a different message in the view?

app/views/settings.html
```diff
@@ -57,17 +57,112 @@
...
+                          <span ng-switch="scope" class="hide-ng-leave">
+                            <span ng-switch-when="Terminating">&mdash; Matches pods that have an active deadline.</span>
+                            <span ng-switch-when="NotTerminating">&mdash; Matches pods that do not have an active deadline.</span>
+                            <span ng-switch-when="BestEffort">&mdash; Matches pods that have best effort quality of service.</span>
+                            <span ng-switch-when="NotBestEffort">&mdash; Matches pods that do not have best effort quality of service.</span>
+                          </span>
+                        </span>
+                      </div>
+                    </dd>
+                  </dl>
```
spadgett: Wondering if this block shouldn't be a directive so it's not repeated.

jwforres: Is your concern mainly around the messages being duplicated? i'd rather add a filter then make this a directive i think

spadgett: The `dl` scopes block I mean.

jwforres:
@spadgett updated

spadgett:
[merge]

PR #141, opened 2016-07-07
Don't store namespace with service link if it's the same

spadgett:
Fixes https://github.com/openshift/origin/issues/9653

@jwforres PTAL


jwforres:
[merge]

PR #143, opened 2016-07-08
Create and edit routes with multiple services

spadgett:
https://trello.com/c/v3jASHVU

screen shot 2016-07-08 at 2 38 40 pm

/cc @jwforres

### DIFF
--- a/app/scripts/controllers/edit/route.js
+++ b/app/scripts/controllers/edit/route.js
@@ -51,7 +51,6 @@ angular.module('openshiftConsole')
         DataService.get("routes", $scope.routeName, context).then(
           function(original) {
             route = angular.copy(original);
-            var serviceName = _.get(route, 'spec.to.name');
             $scope.routing = {
               service: _.get(route, 'spec.to.name'),
               host: _.get(route, 'spec.host'),
@@ -62,9 +61,25 @@ angular.module('openshiftConsole')
 
             DataService.list("services", context, function(services) {
               var servicesByName = services.by("metadata.name");
+              var to = _.get(route, 'spec.to', {});
               $scope.loading = false;
               $scope.services = orderByDisplayName(servicesByName);
-              $scope.routing.service = servicesByName[serviceName];
+              $scope.routing.to = {
+                service: servicesByName[to.name],
+                weight: to.weight
+              };
+              $scope.routing.alternateServices = [];
+              _.each(_.get(route, 'spec.alternateBackends'), function(alternateBackend) {
+                if (alternateBackend.kind !== 'Service') {
+                  Navigate.toErrorPage('Editing routes with non-service targets is unsupported. You can edit the route with the "Edit YAML" action instead.');
+                  return false;
+                }
+
+                $scope.routing.alternateServices.push({
+                  service: servicesByName[alternateBackend.name],
+                  weight: alternateBackend.weight
+                });
+              });
             });
           },
           function() {
@@ -73,16 +88,13 @@ angular.module('openshiftConsole')
 
         // Update the fields in the route from what was entered in the form.
         var updateRouteFields = function() {
-          var serviceName = _.get($scope, 'routing.service.metadata.name');
+          var serviceName = _.get($scope, 'routing.to.service.metadata.name');
           _.set(route, 'spec.to.name', serviceName);
-
-          // Remove the host.generated annotation if the host was edited.
-          if (_.get(route, ['metadata', 'annotations', 'openshift.io/host.generated']) === 'true' &&
-              _.get(route, 'spec.host') !== $scope.routing.host) {
-            delete route.metadata.annotations['openshift.io/host.generated'];
+          var weight = _.get($scope, 'routing.to.weight');
+          if (!isNaN(weight)) {
+            _.set(route, 'spec.to.weight', weight);
           }
 
-          route.spec.host = $scope.routing.host;
           route.spec.path = $scope.routing.path;
 
           var targetPort = $scope.routing.targetPort;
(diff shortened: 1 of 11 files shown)


spadgett:
@jwforres thoughts on the pie chart?

screen shot 2016-07-08 at 4 02 31 pm

spadgett:
@jwforres updated screenshot

screen shot 2016-07-11 at 10 39 28 am

spadgett:
I should remove "Service" from each legend item

spadgett:
@jwforres PTAL

jwforres:
I like this much better than what you had this morning 👍

app/scripts/controllers/edit/route.js
```diff
@@ -73,16 +88,13 @@ angular.module('openshiftConsole')
...
-          var serviceName = _.get($scope, 'routing.service.metadata.name');
+          var serviceName = _.get($scope, 'routing.to.service.metadata.name');
           _.set(route, 'spec.to.name', serviceName);
-
-          // Remove the host.generated annotation if the host was edited.
-          if (_.get(route, ['metadata', 'annotations', 'openshift.io/host.generated']) === 'true' &&
-              _.get(route, 'spec.host') !== $scope.routing.host) {
-            delete route.metadata.annotations['openshift.io/host.generated'];
+          var weight = _.get($scope, 'routing.to.weight');
+          if (!isNaN(weight)) {
```
jwforres: are you not doing number validation on the inputs, or is this just an extra sanity check

app/scripts/directives/oscRouting.js
```diff
@@ -79,13 +80,20 @@ angular.module("openshiftConsole")
...
+          updatePortOptions(newValue);
           // Don't overwrite the target port when editing an existing route unless the user picked a
           // different service.
           if (newValue !== oldValue || !scope.route.targetPort) {
             scope.route.targetPort = _.get(scope, 'route.portOptions[0].port');
           }
+
+          if (scope.services) {
+            // Update the options for alternate services.
+            scope.alternateServiceOptions = _.reject(scope.services, function(service) {
```
jwforres: what happens if they have already added the service as an alternate and they switch it to be the primary?

app/scripts/directives/routeServicePie.js
```diff
@@ -0,0 +1,81 @@
...
+      template: '<div ng-attr-id="{{chartId}}"></div>',
+      link: function($scope) {
+        var chart, config;
+
+        $scope.chartId = _.uniqueId('route-service-chart-');
+
+        config = {
+          bindto: '#' + $scope.chartId,
+          color: {
+            pattern: [$.pfPaletteColors.blue, $.pfPaletteColors.orange, $.pfPaletteColors.green, $.pfPaletteColors.red]
```
jwforres: what does c3 do if it needs more colors, start adjusting the shade of the color?

app/views/directives/osc-routing-service.html
```diff
@@ -0,0 +1,67 @@
...
+             type="number"
+             required
+             min="1"
+             max="256"
+             ng-pattern="/^\d+$/"
+             class="form-control"
+             aria-describedby="{{id}}-weight-help">
+      <div>
+        <span id="{{id}}-weight-help" class="help-block">
+          Weight as an number between 1 and 256 that specifies the relative weight against
```
jwforres: 'Weight is a number...'

app/views/directives/osc-routing-service.html
```diff
@@ -0,0 +1,67 @@
...
+          other route services.
+        </span>
+        <div ng-if="routingServiceForm.weight.$dirty && routingServiceForm.weight.$invalid" class="has-error">
+          <div ng-if="routingServiceForm.weight.$error.number" class="help-block">
+            Must be a number.
+          </div>
+          <div ng-if="routingServiceForm.weight.$error.min" class="help-block">
+            Must greater than or equal to 1.
```
jwforres: 'Must be...'

app/views/directives/osc-routing-service.html
```diff
@@ -0,0 +1,67 @@
...
+      </select>
+      <div>
+        <span ng-attr-id="{{id}}-service-help" class="help-block">
+          <span ng-if="!isAlternate">Service to route to.</span>
+          <span ng-if="isAlternate">Alternate service for route traffic.</span>
+        </span>
+      </div>
+      <div ng-if="(services | hashSize) === 0" class="has-error">
+        <span class="help-block">
+          There are no <span ng-if="is-alternate">additional</span> services in your project to expose
```
jwforres: what if there is only one service in the project?  you can route to that single service in the to, but would have no more to pick from to be "additional",  services|hashSize would be zero and this help block wouldnt show, unless im missing something

app/scripts/controllers/edit/route.js
spadgett: It's not set for routes with only one service.

app/scripts/directives/oscRouting.js
spadgett: The selected alternate option is removed, so dropdown has nothing selected. It's a required field, so they have to pick another before saving.

app/scripts/directives/routeServicePie.js
spadgett: The colors start repeating (5 would be blue, 6 orange, etc.)

app/views/directives/osc-routing-service.html
spadgett: We don't show the option to split across services at all if there is only one, but I think the logic is right? scope.services here is the list of additional services _excluding_ the primary if this is an alternate.

jwforres: ah didnt realize the scope of what this form was, got it

jwforres:
[merge]

PR #144, opened 2016-07-11
Remove pathseg polyfill

spadgett:
No longer needed since c3.js 0.4.11. Verified in Chrome where we saw the original problem. https://github.com/openshift/origin/pull/7373 is no longer an issue

@jwforres PTAL

Fixes https://github.com/openshift/origin/issues/7377


jwforres:
[merge]

benjaminapetersen:
Ah, thx @spadgett, I was supposed to nix that in my bower update PR, my bad.

PR #147, opened 2016-07-11
Eliminate "scope" in ApplicationGenerator

benjaminapetersen:
"scope" was the object returned from the ApplicationGenerator service.  This is a misnomer & is rather confusing. Updated to return an object with "pubic" methods, leaving all underscored methods private.

@jwforres @spadgett for review, noticed while doing key-value-editor replacements.

### DIFF
--- a/app/scripts/services/applicationGenerator.js
+++ b/app/scripts/services/applicationGenerator.js
@@ -4,9 +4,7 @@ angular.module("openshiftConsole")
 
   .service("ApplicationGenerator", function(DataService, Logger, $parse){
 
-    var scope = {};
-
-    scope._generateSecret = function(){
+    var generateSecret = function(){
         //http://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript
         function s4() {
           return Math.floor((1 + Math.random()) * 0x10000)
@@ -16,7 +14,7 @@ angular.module("openshiftConsole")
         return s4()+s4()+s4()+s4();
       };
 
-    scope.parsePorts = function(image) {
+    var parsePorts = function(image) {
       //map ports to k8s structure
       var parsePortsFromSpec = function(portSpec){
         var ports = [];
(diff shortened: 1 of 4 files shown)


spadgett:
you'll need to update applicationGeneratorSpec.js

dist/scripts/templates.js
```diff
@@ -3613,6 +3613,7 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
     "<i class=\"pficon pficon-help\"></i>\n" +
     "</a>\n" +
     "</span></h3>\n" +
+    "{{buildConfig.envVars | json}}\n" +
```
spadgett: Looks like an unrelated change in dist?

benjaminapetersen: ah, bother grunt build :) will fix.

benjaminapetersen:
@spadgett udpated

spadgett:
Still needs applicationGeneratorSpec.js test updates

spadgett:
@benjaminapetersen ping :)

benjaminapetersen:
Thanks @spadgett, will refresh this one quick!

benjaminapetersen:
@spadgett most test easy to fix, but this one fails:
`ApplicationGenerator generating applications from image that includes source should generate a BuildConfig for the source`
with the difference being some generated in triggers:
```javascript
triggers: [{
            generic: {
                secret: 'c77e0419e36c8478'
            },
            type: 'Generic'
        }, {
            github: {
                secret: '26e64b71938ea292'
            },
            type: 'GitHub'
        }, {
            imageChange: {},
            type: 'ImageChange'
        }, {
            type: 'ConfigChange'
        }]
```
vs
```javascript
triggers: [{
            generic: {
                secret: 'secret101'
            },
            type: 'Generic'
        }, {
            github: {
                secret: 'secret101'
            },
            type: 'GitHub'
        }, {
            imageChange: {},
            type: 'ImageChange'
        }, {
            type: 'ConfigChange'
        }]
```
Working through this. I dont know atm why my change would affect this.

benjaminapetersen:
Ah, wait, maybe i do.

benjaminapetersen:
Ok @spadgett this is working again.  Had to think a bit about the monkey patch of `ApplicationGenerator.generateSecret()` after the change. Internally its called as a standalone fn (`generateSecret()`), so replacing the whole method in the test file was no longer working.

app/scripts/services/applicationGenerator.js
```diff
@@ -395,13 +349,71 @@ angular.module("openshiftConsole")
...
+      return resources;
+    };
+
+    return {
+      createRoute: createRoute,
+      generate: generate,
+      parsePorts: parsePorts,
+      generateSecret: generateSecret,
+      getServicePort: getServicePort,
+      // pseudo private
```
spadgett: pseudo private == needed for tests? If so, I'd say that in the comment.

test/spec/services/applicationGeneratorSpec.js
```diff
@@ -3,13 +3,23 @@
 describe("ApplicationGenerator", function(){
   var ApplicationGenerator;
   var inputTemplate;
+  var mathFloor = Math.floor;
 
   beforeEach(function(){
 
     inject(function(_ApplicationGenerator_){
       ApplicationGenerator = _ApplicationGenerator_;
-      ApplicationGenerator._generateSecret = function(){
```
spadgett: Is there a reason we can't simply change this to `ApplicationGenerator.generateSecret` and leave the mock as before?

app/scripts/services/applicationGenerator.js
benjaminapetersen: right... we have tests already written so I didn't want to just remove the tests.  Perhaps they dont even need to be semi-private?

test/spec/services/applicationGeneratorSpec.js
benjaminapetersen: Yup. Since the end of the file does this:

```javascript
return {
  // "exporting" the functions...
  _generateSecret: generateSecret
}
```
Doing `ApplicationGenerator._generateSecret = function(){ ... }` is not effective.  The original function is used internally a number of times: 

Example:
```javascript 
// line 233, this call cannot be mocked
secret: generateSecret()  
``` 
So will never use the mocked function.

app/scripts/services/applicationGenerator.js
benjaminapetersen: updated comment

test/spec/services/applicationGeneratorSpec.js
benjaminapetersen: @spadgett more thoughts on this?

benjaminapetersen:
Hmm... gonna close this. The rebase looks like a train wreck, prob easier to just redo.

PR #148, opened 2016-07-11
Warn users when resource has changed when editing YAML

### LINKED ISSUE #142: Editing deploy config twice via `Edit YAML` option in immediate succession results in a confusing error
screen shot 2016-07-08 at 12 02 01 pm

To reproduce:

* Browse to a deployment config and choose `Edit YAML` from the `Actions` menu in the upper right
* Edit an env var
* Click `Save`
* Click `Actions` and `Edit YAML` again
* Make an additional edit to an env var
* Click `Save`
* Notice the resulting error

I notice the error isn't thrown once the `Deploy` button is re-enabled following the deployment resulting from the first edit. 

@benjaminapetersen

spadgett:
Fixes #142 by warning users when something has changed since they started editing. @jwforres PTAL

openshift_web_console

### DIFF
--- a/app/scripts/controllers/modals/editModal.js
+++ b/app/scripts/controllers/modals/editModal.js
@@ -12,6 +12,15 @@ angular.module('openshiftConsole')
     // Use angular.copy to avoid $$hashKey properties inserted by ng-repeat.
     var resource = angular.copy($scope.resource);
 
+    var getVersion = function(resource) {
+      return _.get(resource, 'metadata.resourceVersion');
+    };
+
+    // Watch for updates to the resource to warn users before the save fails.
+    $scope.$watch('resource', function(newValue) {
+      $scope.resourceChanged = getVersion(newValue) !== getVersion(resource);
+    });
+
     // Hack to make `apiVersion` and `kind` appear at the top.
     //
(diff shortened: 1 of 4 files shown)


spadgett:
Going to change the message to say "copy any changes you've made"

liggitt:
what? no three-way diff showing original, edited, and new?

spadgett:
I was actually thinking about it.....

http://benkeen.github.io/ace-diff/

jwforres:
three-way diff in a modal dialog though?  doesn't seem like we would have the space for that

jwforres:
[merge] this for now, you can open an enhancement issue to look at the diff later

liggitt:
> three-way diff in a modal dialog though? doesn't seem like we would have the space for that

I guess you wouldn't need a three way diff... building a patch between the original content and the edited content, then applying it to the new content, indicating any merge conflicts would be sufficient

spadgett:
If we made it a separate page instead of a dialog, there'd be more room.

https://github.com/openshift/origin/issues/6845#issuecomment-176855327

jwforres:
I'm ok with that

PR #149, opened 2016-07-11
Use same drag handle for edit command as edit env

spadgett:
The styles for the env var editor are better, and this makes the editors more consistent.

openshift_web_console

@jwforres PTAL


jwforres:
[merge]

PR #25, opened 2016-05-20
Fix build config view log link for latest build

spadgett:
The log link was not showing up after some of the pipeline changes.

@jwforres PTAL

PR #151, opened 2016-07-11
Support for Jenkins 'input' stage in pipelines

fabianofranz:
Adding support to manual Jenkins input (with the use of the `input` Jenkinsfile step).

Fixes https://trello.com/c/VgmHJvYQ/676-1-manual-approval-stages-in-pipelines

pipelines_input_1

pipelines_input_2

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -955,6 +955,17 @@ angular.module('openshiftConsole')
       return annotationFilter(build, 'jenkinsBuildURL');
     };
   })
+  .filter('jenkinsInputURL', function(jenkinsBuildURLFilter) {
+    return function(build) {
+      // TODO we should have a jenkinsInputURL annotation, waiting on
+      // https://github.com/fabric8io/openshift-jenkins-sync-plugin/issues/94
+      var buildURL = jenkinsBuildURLFilter(build);
+      if (!buildURL) {
+        return null;
+      }
+      return new URI(buildURL).segment("/input/").toString();
+    };
+  })
   .filter('buildLogURL', function(isJenkinsPipelineStrategyFilter,
                                   jenkinsLogURLFilter,
                                   navigateResourceURLFilter) {
@@ -1007,6 +1018,14 @@ angular.module('openshiftConsole')
       }
 
       return _.indexOf(['ABORTED', 'FAILED', 'SUCCESS'], stage.status) !== -1;
+    };
+  })
+  .filter('pipelineStagePendingInput', function () {
+    return function(stage) {
+      if (!stage) {
+        return false;
+      }
+      return 'PAUSED_PENDING_INPUT' === stage.status;
     };
   })
   .filter('humanizeKind', function (startCaseFilter) {
--- a/app/views/directives/overview-pipeline.html
+++ b/app/views/directives/overview-pipeline.html
@@ -1,41 +1,55 @@
-<div class="overview-pipeline">
-  <div class="timestamp">
-    <relative-timestamp timestamp="build.metadata.creationTimestamp"></relative-timestamp>
-  </div>
-  <div row class="overview-pipeline-details" ng-class="build.status.phase">
-    <div class="build-name">
-      <a ng-href="{{buildConfigName | navigateResourceURL : 'BuildConfig' : build.metadata.namespace}}">{{buildConfigName}}</a>,
-      <a ng-href="{{build | navigateResourceURL}}">#{{build | annotation : 'buildNumber'}}</a>
+<div>
+  <div class="overview-pipeline">
+    <div class="timestamp">
+      <relative-timestamp timestamp="build.metadata.creationTimestamp"></relative-timestamp>
     </div>
-    <div class="build-phase">
-      {{build.status.phase}}
-    </div>
-    <div class="stages">
-      <div ng-if="!jenkinsStatus.stages.length">
-        <div>No stages have started.</div>
+    <div row class="overview-pipeline-details" ng-class="build.status.phase">
+      <div class="build-name">
+        <a ng-href="{{buildConfigName | navigateResourceURL : 'BuildConfig' : build.metadata.namespace}}">{{buildConfigName}}</a>,
+        <a ng-href="{{build | navigateResourceURL}}">#{{build | annotation : 'buildNumber'}}</a>
       </div>
-      <div ng-repeat="stage in jenkinsStatus.stages track by stage.id" class="stage">
-        <span
-            data-toggle="tooltip"
-            data-trigger="hover"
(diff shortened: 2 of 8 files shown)


fabianofranz:
@jwforres @spadgett PTAL

app/scripts/filters/resources.js
```diff
@@ -955,6 +955,17 @@ angular.module('openshiftConsole')
...
   })
+  .filter('jenkinsInputURL', function(jenkinsBuildURLFilter) {
+    return function(build) {
+      // TODO we should have a jenkinsInputURL annotation, waiting on
+      // https://github.com/fabric8io/openshift-jenkins-sync-plugin/issues/94
+      var buildURL = jenkinsBuildURLFilter(build);
+      if (!buildURL) {
+        return null;
+      }
+      return new URI(buildURL).segment("/input/").toString();
```
spadgett: I'm hesitant to assume anything about the Jenkins URL patterns since they could change. I don't see this as a documented API.

https://github.com/jenkinsci/pipeline-stage-view-plugin/tree/master/rest-api

I hope the sync plugin can add this, and this is only a stop-gap.

@bparees FYI

spadgett:
I'm not sure this we should use a warning icon / color for this state, although we want to draw attention to it.

/cc @ajacobs21e for design input

app/scripts/filters/resources.js
jwforres: Agree this is a stopgap, we aren't closing the issue against the sync plugin

ajacobs21e:
@sjcox-rh How did you handle this with Fabric8?

fabianofranz:
The visual changes here are:

1. In overview: the state "paused" icon was changed to orange to direct the user's attention, and the "paused pending input" message now links the Jenkins input page;
2. In overview: close to "View Logs" we now have a "Requires Input" warning message that also links the Jenkins input page;
3. In builds: the state "paused" icon also now in orange, and link to the Jenkins input page in the "build waiting for input" message.

But I'm open for other suggestions. Would a message box like the ones at the bottom of services make sense?

fabianofranz:
We could elect a color from the [Patternfly palette](https://www.patternfly.org/styles/color-palette/#_) specifically to be used in the "paused waiting for action" state. Maybe `pf-gold`, are we using it for anything else already?

fabianofranz:
@sjcox-rh thanks! Looks good to me, my only comment is about the icon change. I think we should keep it clear that the build is really *paused* waiting for user input, and won't continue until you provide what's required. Which is different than regular info/warning, where you build is usually still running but have potential issues the user might be aware of.

Another idea, in case we want to make it stand out more, would be to use an alert box with message and link, right below the build line `sample-pipeline, #25 | Running`.

@jwforres @spadgett @ajacobs21e comments?

jwforres:
@rhamilto is working on an expanded pipeline on the overview when its
running though, so i don't think we would have the collapsed overview state
to worry about

spadgett:
@sjcox-rh Thanks. There's a link to approve or reject the stage. Ideas on where to put the link in the detailed pipeline view?

sjcox-rh:
@spadgett Perhaps under the progress bar and replace the elapsed time text?  Like so:

screen shot 2016-07-13 at 4 15 19 pm

fabianofranz:
@sjcox-rh @spadgett  Note I'm trying to avoid the word "approve" since user inputs can be used for more scenarios - e.g. request some information to be submitted through a text field, require acceptance to terms through a checkbox or even just pausing the stage so that I can manually take a look at it before deploying to production. Jenkins refers to it as "paused for input". 

Any suggestion about a better verb to be used in the link label? I tried "Fulfill Manual Input" which looks meh. :)

spadgett:
Requires input as you have? What does Jenkins show in its UI?

jwforres:
Input needed ?

jwforres:
Or  Input required maybe

fabianofranz:
Jenkins says "paused for input":

fabianofranz:
Latest revision is looking like

pipeline_rev2_1

Notice I didn't change the `Build #30` box on the left yet, I'm not confident about that change because that box is about the *build* state which is still "running", what's in "paused, requires input" state is only the stage. Thoughts?

As per [Jessica's comment](https://github.com/openshift/origin-web-console/pull/151#issuecomment-232443305) the collapsed overview section is likely going away so I'm not doing further changes, but the existing version still works if we eventually decide to use it.

pipeline_rev2_2

app/views/directives/overview-pipeline.html
```diff
@@ -1,41 +1,57 @@
...
+          <a ng-href="{{build | jenkinsInputURL}}" target="_blank">Input Required</a>
+        </div>
+      </div>
+      <div ng-if="build | buildLogURL" class="log-link"><a ng-href="{{build | buildLogURL}}" ng-if="('builds/log' | canI : 'get')" target="_blank">View Log</a></div>
+    </div>
+  </div>
+  <div ng-repeat="stage in jenkinsStatus.stages track by stage.id" class="alert-wrapper" ng-class-odd="'odd-alert'" ng-class-even="'even-alert'" >
+    <div ng-if="stage | pipelineStagePendingInput" class="alert alert-warning toast-pf mar-left-sm">
+      <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
+      <span style="margin-right: 5px;">{{stage.name}} stage requires manual user input to continue.</span>
```
spadgett: class="mar-right-sm"

app/views/directives/overview-pipeline.html
```diff
@@ -1,41 +1,57 @@
...
+      <span>
+        <a href="{{build | jenkinsInputURL}}" target="_blank">Fulfill manual input</a>
```
spadgett: I don't think the span wrapper is needed.

"Provide input"

?

app/views/directives/overview-pipeline.html
spadgett: Maybe

"Stage '{{stage.name}}' requires user input to continue."

app/views/directives/overview-pipeline.html
spadgett: Nit: `ng-href`

spadgett:
A few small comments, otherwise LGTM

fabianofranz:
@spadgett thanks! Comments addressed, squashed into one commit.

spadgett:
[merge]

rhamilto:
Given the changes I'm making to the display of pipelines on the overview, the display of an 'input' stage will look like it does on the BC and build pages, lacking the warning message.

screen shot 2016-07-22 at 12 48 37 pm

@fabianofranz, @jwforres, @spadgett, Do we want to maintain the warning message on the overview?

jwforres:
personally I dont think we need it, there is a much larger visual indicator
now than when it was in the collapsed state

PR #153, opened 2016-07-12
Display template message

jhadvig:
Here is the first draft for the "Post-process template message" card. 
Wasn't sure if we want to also have possibility to edit the message in the `newfromtemplate` page, so I've add the extra input field just so we have the picture. Although not sure if it makes sense to update the message there(maybe if you don't save the template upon the creation process?!?).
screenshot-6

Here is also how we would display the message on the `next-step` page.
screenshot-7
Wasnt sure with the header wording. Wanted to go with 'Template Owner Message' , but went with 'Template Message'

@jwforres thoughts

### DIFF
--- a/app/scripts/controllers/create/nextSteps.js
+++ b/app/scripts/controllers/create/nextSteps.js
@@ -9,7 +9,7 @@
  * Controller of the openshiftConsole
  */
 angular.module("openshiftConsole")
-  .controller("NextStepsController", function($scope, $http, $routeParams, DataService, $q, $location, ProcessedParametersService, TaskList, $parse, Navigate, $filter, imageObjectRefFilter, failureObjectNameFilter, ProjectsService) {
+  .controller("NextStepsController", function($scope, $http, $routeParams, DataService, $q, $location, ProcessedTemplateService, TaskList, $parse, Navigate, $filter, imageObjectRefFilter, failureObjectNameFilter, ProjectsService) {
     var displayNameFilter = $filter('displayName');
     var watches = [];
 
@@ -51,8 +51,10 @@ angular.module("openshiftConsole")
       }
     ];
 
-    $scope.parameters = ProcessedParametersService.getParams();
-    ProcessedParametersService.clearParams();
+    var processedTemplateData = ProcessedTemplateService.getTemplateData();
+    $scope.parameters = processedTemplateData.params;
+    $scope.templateMessage = processedTemplateData.message;
+    ProcessedTemplateService.clearTemplateData();
 
     ProjectsService
       .get($routeParams.project)
--- a/app/scripts/controllers/newfromtemplate.js
+++ b/app/scripts/controllers/newfromtemplate.js
@@ -9,7 +9,7 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('NewFromTemplateController', function ($scope, $http, $routeParams, DataService, ProcessedParametersService, AlertMessageService, ProjectsService, $q, $location, TaskList, $parse, Navigate, $filter, imageObjectRefFilter, failureObjectNameFilter, CachedTemplateService) {
+  .controller('NewFromTemplateController', function ($scope, $http, $routeParams, DataService, ProcessedTemplateService, AlertMessageService, ProjectsService, $q, $location, TaskList, $parse, Navigate, $filter, imageObjectRefFilter, failureObjectNameFilter, CachedTemplateService) {
 
 
     var name = $routeParams.name;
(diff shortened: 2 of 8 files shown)


jhadvig:
will add generated `dist` after the review :)

app/views/create/next-steps.html
```diff
@@ -89,6 +89,11 @@ <h1 ng-if="!pendingTasks(tasks()).length && erroredTasks(tasks()).length">Comple
               </div>
             </div>
 
+            <div ng-if="templateMessage.length">
+              <h2>Template message</h2>
+              <p>{{templateMessage}}</p>
```
liggitt: I would expect whitespace to be preserved here, so they can have a couple paragraphs. I also think we should probably linkify the text (if they want to point to follow up docs)

app/views/newfromtemplate.html
```diff
@@ -39,6 +39,19 @@
                               can-toggle="false"
                               help-text="Each label is applied to each created resource.">
                           </label-editor>
+                          <osc-form-section
+                            header="Message"
+                            about-title="message"
+                            about="Informative message for the user how to utilize the newly created resources. Message can accept parameters in '${PARAMETER}' format. Parameter substitution will be performed on the message before being displayed so that generated credentials and other parameters can be included in the output."
+                            expand="expand"
+                            can-toggle="canToggle">
+                              <input
```
liggitt: should probably be a textarea

jwforres:
We should not be letting them edit that message, thats not the point of it.  It is from a template author as instructions to the user for after the template as been instantiated.

app/views/create/next-steps.html
jwforres: look at osc-image-summary.html  the div with class of resource-description, it preserves newlines and does a  linkify on the text

jwforres:
the message from the template author is arguably the most important piece of information on the next steps page, seems like it should be the first thing they see.  This is what v2 looked like

v2-next-steps


Don't necessarily want the massive green block like we had there, but it was very clear that it was an important message to you that you should look at

cc @ajacobs21e @rhamilto

jhadvig:
@jwforres here are some proposals for the template message on the next-step page
1# Similar to what we used in v2
screenshot-8
2# Same as 1# but with info icon and with the same similar font color
screenshot-12
screenshot-13


3#  Here I've played a bit with colors since we use the green one for success related messages, so picked blue and yellowish that bootstrap is using for warning and info alerts
screenshot-14
screenshot-15

4# And a color pallet that patternfly is using for info alerts
screenshot-18

Personally I like the blue one with black font color which looks really nice:
screenshot-17


cc'ing @spadgett && @sg00dwin for their opinion. Thanks ! :)

jhadvig:
Also not really sure if "Message" is the right word here, since the real meaning is something like "ReadMe" or "Template instructions"

Thoughts ?

jwforres:
I'm not sure we even need a header in the block, its text that the user doesn't really need, they just need the content of the message.

As far as the color palette my preference would be the blue with black text, the default patternfly info colors dont really work here since we have a gray background already.  But I would like an opinion from @ajacobs21e @sjcox-rh and @rhamilto

spadgett:
Some of these options don't have enough color contrast for accessibility. See the "Contrast Ratios" here:

https://www.patternfly.org/styles/color-palette

Glancing at them, I prefer the blue background and black font.

spadgett:
I'm not sure "Template Message" is the right header, but I don't have a good suggestion :(
Maybe just a (i) icon to the left and the message.

jhadvig:
@jwforres @spadgett 
screenshot-16

sjcox-rh:
Black text on blue works for me as well as long as this is the primary content you want the user to focus on while on this page.  I do like the idea of the (i) icon to the left of this message since you are leaning towards no header.  Not sure if the spacing would be thrown off when it's added though.  Worth a try!

spadgett:
I was thinking of a large-ish info icon to the left, then a text block where the text left-aligns with the first line when it wraps.

jhadvig:
@spadgett @sjcox-rh PTAL there three proposal based on you feed bac, thanks so much !
1#
screenshot-19
2#
screenshot-20
3#
screenshot-21

IMHO I really the first one.

spadgett:
@jhadvig I was thinking of the first, except the text is left-aligned as a block. The text doesn't wrap under the icon.

spadgett:
Although I think I'm OK with it as shown as well :)

jhadvig:
@jwforres I've updated the PR based on the feedback from @spadgett and @sjcox-rh

sjcox-rh:
@jhadvig Agree with @spadgett on the alignment of the text next to the icon in the first option!  Looks goods, otherwise!

app/scripts/controllers/create/nextSteps.js
```diff
@@ -51,8 +51,10 @@ angular.module("openshiftConsole")
       }
     ];
 
-    $scope.parameters = ProcessedParametersService.getParams();
-    ProcessedParametersService.clearParams();
+    var processedTemplateData = ProcessedTemplateService.getTemplateData();
```
jwforres: make it return a map not an array, its less fragile

app/scripts/controllers/newfromtemplate.js
```diff
@@ -143,7 +143,7 @@ angular.module('openshiftConsole')
               };
 
               if (!_.isEmpty(config.parameters)) {
```
jwforres: dont rely no the existence of config.parameters in order to set config.message, a template author may have a message but no parameters

jhadvig:
@jwforres comments addressed. PTAL :)

jwforres:
[merge]

jwforres:
@bparees FYI console changes are going in

bparees:
awesome!

PR #155, opened 2016-07-12
Idled deployments 

benjaminapetersen:
**Code is not ready for review.**

This is a PR to get some feedback on visuals for showing an idled deployment.  Here are a few screenshots.  I ended up putting the message below the donut rather than above because it seemed to fit better:

screen shot 2016-07-12 at 11 02 23 am
screen shot 2016-07-12 at 11 02 34 am
Adjacent to autoscaling message for reference:
screen shot 2016-07-12 at 11 01 56 am

- Might make sense to move the `i` with the tooltip to the right of the text rather than next to the button
- Need to play with the wording of the tooltip a bit

@spadgett @jwforres appreciate any feedback!

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -34,7 +34,8 @@ angular.module('openshiftConsole')
       "buildPod":                 ["openshift.io/build.pod-name"],
       "jenkinsBuildURL":          ["openshift.io/jenkins-build-uri"],
       "jenkinsLogURL":            ["openshift.io/jenkins-log-url"],
-      "jenkinsStatus":            ["openshift.io/jenkins-status-json"]
+      "jenkinsStatus":            ["openshift.io/jenkins-status-json"],
+      "idledAt":                  ["idling.alpha.openshift.io/idled-at"]
     };
     return function(annotationKey) {
       return annotationMap[annotationKey] || null;
--- a/app/scripts/directives/deploymentDonut.js
+++ b/app/scripts/directives/deploymentDonut.js
@@ -53,31 +53,32 @@ angular.module('openshiftConsole')
         $scope.$watchGroup(['limitRanges', 'hpa', 'project'], updateHPAWarnings);
         $scope.$watch('rc.spec.template.spec.containers', updateHPAWarnings, true);
 
-        // Debounce scaling so multiple clicks within 500 milliseconds only result in one request.
-        var scale = _.debounce(function () {
-          scaleRequestPending = false;
+        var showScalingError = function(result) {
+          $scope.alerts = $scope.alerts || {};
+          $scope.desiredReplicas = null;
+          $scope.alerts["scale"] =
+            {
+              type: "error",
+              message: "An error occurred scaling the deployment.",
+              details: $filter('getErrorDetails')(result)
+            };
+        };
 
+        // use debouncedScale() unless the returned promise is needed
+        var scale = function () {
+          scaleRequestPending = false;
           if (!angular.isNumber($scope.desiredReplicas)) {
             return;
           }
-
-          var showScalingError = function(result) {
-            $scope.alerts = $scope.alerts || {};
-            $scope.desiredReplicas = null;
-            $scope.alerts["scale"] =
-              {
-                type: "error",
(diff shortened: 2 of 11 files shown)


jwforres:
dont love it as a button

jwforres:
also you shouldnt see scaling arrows of autoscaling is enabled

benjaminapetersen:
Sorry, i should have nixed that. I quick hacked it to get some screenshots rolling :/

benjaminapetersen:
2nd pass:

screen shot 2016-07-12 at 3 43 07 pm
screen shot 2016-07-12 at 3 43 40 pm

But, I dont really care for the link leading the phrase "wake up idled deployment".  I think its clearer if it declares "idled deployment", with an action "wake up" following (but can keep it a link).

liggitt:
my eye is drawn to the huge `0`... which is where people are probably going to be looking for an explanation of why their thing doesn't appear to be running. Would just the text "Idled | <a href="#">Wake up</a>" with an info hover be sufficient?

benjaminapetersen:
@liggitt agree, the `0` grabs attention.  I believe putting things inside the donut is a bit of a pain (canvas rendering?), @spadgett let me pull you in on that.  Meanwhile, lemme look.

jwforres:
had a discussion over here, consensus was something like replace the big 0
with "Idled" at a smaller font size than the number, then for the message
"Idled due to inactivity. <Resume # pods>"  assuming we know what are
resuming back up to.

benjaminapetersen:
Updated, [new annotations for idling here](https://github.com/DirectXMan12/origin/blob/ae153ef58bd4fb2f7c3a9acd6dab82c411360f83/pkg/unidling/api/types.go#L4).

Idled/autoscaled:
idle_autoscaled
Idled/Not autoscaled:
idle_non_autoscaled

jwforres:
we had said 'Resume' was a risky verb to use because they might think their
processes were suspended and not killed

jwforres:
and cant decide whether i would prefer Idle vs idle,  how does it look with
it capitalized

benjaminapetersen:
Idle instead of idle (caps):

screen shot 2016-07-26 at 3 25 06 pm

benjaminapetersen:
Hmm... good thought on "resume".  Trying to think of a better word...

benjaminapetersen:
Depends on https://github.com/openshift/origin/pull/9473

jwforres:
aside from the question about what to replace 'Resume' with i think this PR looks good

jwforres:
Instead of resume:

Launch
Start
Create

leaning towards Start

liggitt:
I like start

app/scripts/filters/resources.js
```diff
@@ -34,7 +34,9 @@ angular.module('openshiftConsole')
       "buildPod":                 ["openshift.io/build.pod-name"],
       "jenkinsBuildURL":          ["openshift.io/jenkins-build-uri"],
       "jenkinsLogURL":            ["openshift.io/jenkins-log-url"],
-      "jenkinsStatus":            ["openshift.io/jenkins-status-json"]
+      "jenkinsStatus":            ["openshift.io/jenkins-status-json"],
+      "idledAt":                  ["idling.alpha.openshift.io/idled-at"],
+      "unidleTarget":             ["idling.alpha.openshift.io/unidle-targets"]
```
jwforres: dont this being used anywhere, what is it supposed to be used for?

benjaminapetersen: Im not sure atm, but I added it for reference since it was part of the idle PR.  I'm guessing its the annotation that would go on the endpoint & tell us what DC/RC is idled.  Since we have the annotation on the DC/RC directly, maybe its nothing the UI would ever use & we can drop it.

jwforres: yeah just drop it for now, we dont even pull endpoints today

benjaminapetersen: Done.

benjaminapetersen:
& updated to "Start"

benjaminapetersen:
@jwforres last pass I think!

jwforres:
[merge]

PR #157, opened 2016-07-12
Add test data for overview edge cases

spadgett:
@jwforres PTAL
@sg00dwin CC


jwforres:
LGTM we can always add to it if we think of more later [merge]

PR #159, opened 2016-07-12
Use build.status.duration for build trends chart

### LINKED ISSUE #27: Pipeline builds don't have a duration when first added to build trends chart
This happens when a new pipeline build completes and is added to the build trends chart while on the build config page. The bar does not show up, and there is no duration value in the tooltip when hovering. Refreshing the page causes the bar to appear.

openshift_web_console

spadgett:
Fixes a problem where pipeline builds that just finished showed duration 0s in the chart.  Also improves the display of the y-axis tick marks by using c3.js defaults.

Fixes #27

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/buildTrendsChart.js
+++ b/app/scripts/directives/buildTrendsChart.js
@@ -84,9 +84,6 @@ angular.module('openshiftConsole')
                 bottom: 0
               },
               tick: {
-                count: 5,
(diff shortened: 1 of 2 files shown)


jwforres:
LGTM

spadgett:
[merge]

PR #161, opened 2016-07-12
Replace osc-key-values with key-value-editor in create from image flow

benjaminapetersen:
Re [this trello card](https://trello.com/c/VYlArUB4) and [this bug ticket](https://bugzilla.redhat.com/show_bug.cgi?id=1354519).
- replaces buildConfig.envVars editor
- replaces deploymentConfig.envVars editor
- Fixes incorrectly referenced DC vars (see bug)

Does not address the key-value-editor within the label-editor (is a separate PR).

@spadgett @jwforres @rhamilto

### DIFF
--- a/app/views/create/fromimage.html
+++ b/app/views/create/fromimage.html
@@ -194,13 +194,15 @@
                                   <i class="pficon pficon-help"></i>
                                 </a>
                               </span></h3>
-                              <osc-key-values
-                                entries="buildConfig.envVars"
-                                delimiter="="
-                                key-validator="env"
-                                delete-policy="added"
-                                key-validation-tooltip="A valid environment variable name is an alphanumeric (a-z and 0-9) string beginning with a letter that may contain underscores."></osc-key-values>
+                              <key-value-editor
+                                entries="buildConfigEnvVars"
+                                key-placeholder="name"
+                                value-placeholder="value"
+                                key-validator="[a-zA-Z][a-zA-Z0-9_]*"
+                                key-validator-error-tooltip="A valid environment variable name is an alphanumeric (a-z and 0-9) string beginning with a letter that may contain underscores."></key-value-editor>
+
                             </osc-form-section>
+
 
                             <!-- /build config -->
 
@@ -210,8 +212,7 @@
                               about="Deployment configurations describe how your application is configured
                                 by the cluster and under what conditions it should be recreated (e.g. when the image changes)."
                               expand="true"
-                              can-toggle="false"
-                              >
+                              can-toggle="false">
                               <div class="animate-drawer" ng-show="$parent.expand">
                                 <h3>Autodeploy when</h3>
                                 <div class="checkbox">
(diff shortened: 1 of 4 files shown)


jwforres:
screenshot of what it looks like with the new collapsed section for the env vars from the image

benjaminapetersen:
DC show/hide existing vars in readonly mode:
screen shot 2016-07-12 at 4 55 44 pm
screen shot 2016-07-12 at 4 55 57 pm

benjaminapetersen:
screen shot 2016-07-12 at 4 55 33 pm

rhamilto:
https://github.com/openshift/angular-key-value-editor/pull/64 resolves the misalignment issue.

jwforres:
@benjaminapetersen the env vars from the image should be editable just not deleteable, so they can choose to set it to "" if they want.  This is because the env var will get set by what the image defines and deleting from our list for the DC will have no effect on that, but letting them set it to "" gives them at least some control.

benjaminapetersen:
Ahh... was thinking they could just add key of same name & value of "" to override.  Didn't do a 'value only editable' feature on it, but I'm sure I can add that.

jwforres:
@benjaminapetersen although it specifically only edit value, which is interesting, they shouldnt be able to modify the key in that case

jwforres:
Probably needs some kind of (i) or help text as to what the env vars from the image actually are

jwforres:
So you'll still need to assume they might duplicate the keys, I think your merge logic for the fully user defined env vars winning still applies

benjaminapetersen:
Updated, screenshots below illustrate change:

1.) Readonly list of deployment config vars, hidden by default. Click link to show.  "Save" action will only store the ones overridden (see above comments & a discussion with @jwforres).
screen shot 2016-07-18 at 10 45 30 am
2.) Screenshot of the DC env tab, showing that only the custom DCs have been saved.
screen shot 2016-07-18 at 11 15 28 am

benjaminapetersen:
@jwforres another pass? I think this is functioning per our latest discussion.

app/views/create/fromimage.html
```diff
@@ -228,17 +229,36 @@
...
-                                    entries="deploymentConfig.envVars"
-                                    delimiter="="
-                                    key-validator="env"
-                                    delete-policy="added"
-                                    key-validation-tooltip="A valid environment variable name is an alphanumeric (a-z and 0-9) string beginning with a letter that may contain underscores."></osc-key-values>
+                                  <p ng-show="DCEnvVarsFromImage.length">
+                                    <a
+                                      href=""
+                                      ng-click="showDCEnvs = (!showDCEnvs)">
+                                      {{showDCEnvs ? 'Hide' : 'Show'}} env vars from image
```
jwforres: I'm not sure I like this text  @spadgett do you have another suggestion?

jwforres:
@benjaminapetersen generally LGTM, just the one question about the link text

app/views/create/fromimage.html
spadgett: I don't have a great suggestion... Maybe "Show image environment"?

benjaminapetersen:
Updated to suggested verbiage ~ 

screen shot 2016-07-18 at 4 59 52 pm

@jwforres @spadgett

jwforres:
[merge]

PR #163, opened 2016-07-13
Bug 1333000: pass image tag to directive

jhadvig:
Pass image tag in order to detect when user picks the 'latest' tag, so the warning saying that the image '... could be updated to a newer major version in the future'.

@spadgett PTAL

https://bugzilla.redhat.com/show_bug.cgi?id=1333000


spadgett:
[merge]

PR #165, opened 2016-07-13
Prevent picking the same route service twice

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1356003

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/oscRouting.js
+++ b/app/scripts/directives/oscRouting.js
@@ -96,6 +96,14 @@ angular.module("openshiftConsole")
           }
         });
 
+        scope.$watch('route.alternateServices', function(alternateServices) {
+          // Find any duplicates.
+          scope.duplicateServices = _(alternateServices).map('service').filter(function(value, index, iteratee) {
+            return _.includes(iteratee, value, index + 1);
+          }).value();
+          formCtl.$setValidity("duplicateServices", !scope.duplicateServices.length);
+        }, true);
+
         var showCertificateWarning = function() {
           if (!scope.route.tls) {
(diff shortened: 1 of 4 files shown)


jwforres:
changes LGTM but we should make sure with the API folks that we want to prevent it, otherwise could get in a case where someone does it through API / CLI and then goes to edit it in the console and cant

spadgett:
@rajatchopra Can you weight in? See

https://bugzilla.redhat.com/show_bug.cgi?id=1356003#c3

> Should the API allow you specify the same service twice in `route.spec.alternateBackends`? I'd expect that to fail validation.

rajatchopra:
This looks good. But the API validation does not deal with it yet. It should. Let me fix up a PR for that soon.

Also, the API calls it alternateBackends, this piece of code calls it alternateServices. Fix one or the other?

rajatchopra:
https://github.com/openshift/origin/pull/9920

spadgett:
> Also, the API calls it alternateBackends, this piece of code calls it alternateServices. Fix one or the other?

Yeah, I purposely chose `alternateServices` to make it clear the web console code only supports services now. We show an error page if you edit a route with anything else.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/edit/route.js#L73-L81

We use the correct property when calling the API.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/edit/route.js#L113-L124

spadgett:
@jwforres okay to merge?

jwforres:
[merge]

PR #166, opened 2016-07-14
Bug 1293791 - Hide 'Add to project' button on overview if user cant create any resource

jhadvig:
screenshot-22

@jwforres PTAL

### DIFF
--- a/app/views/overview.html
+++ b/app/views/overview.html
@@ -14,16 +14,25 @@
               <div ng-if="renderOptions.showGetStarted">
                 <!-- Getting started -->
                 <div class="empty-project text-center">
-                  <h2>Get started with your project.</h2>
-                  <p class="gutter-top">
-                    Use your source or an example repository to build an application
-                    image, or add components like databases and message queues.
-                  </p>
-                  <p class="gutter-top">
-                    <a ng-href="project/{{projectName}}/create" class="btn btn-lg btn-primary">
-                      Add to Project
-                    </a>
-                  </p>
+                  <div ng-if="project.metadata.name | canIAddToProject">
+                    <h2>Get started with your project.</h2>
+                    <p class="gutter-top">
+                      Use your source or an example repository to build an application
+                      image, or add components like databases and message queues.
+                    </p>
+                    <p class="gutter-top">
+                      <a ng-href="project/{{projectName}}/create" class="btn btn-lg btn-primary">
+                        Add to Project
+                      </a>
+                    </p>
+                  </div>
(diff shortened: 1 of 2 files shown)


spadgett:
We purposely avoided saying "the project is empty" because it might not be. There are resources like builds we don't show on the overview.

I think a "Welcome to OpenShift." header and a message telling the user how a project admin can give you edit access might be helpful? Something like this on the projects page

https://github.com/openshift/origin-web-console/blob/master/app/views/projects.html#L51-L52

jhadvig:
@spadgett was thinking about the 'Welcome to OpenShift' as well, but we already display it when you dont have any projects listed in the projects page. 
What about "Welcome to ProjectName project" ?

spadgett:
@jhadvig Unless you just created the project, you've only seen the message once though. (And if you created the project, you probably have edit rights and won't see the message here.)

Open to other suggestions...

jhadvig:
@spadgett I've used the wording that you suggested, although I thing I like the second one more since the code its under the text
screenshot-23
screenshot-24

Thoughts ?

spadgett:
Maybe

"You don't have authority to create resources in project my-project. A project admin can add additional roles to your user ID for this project by running the command..."

@jwforres opinions?

jwforres:
"You don't have authority" makes me think I am trying to do something i shouldn't have, what about:

"If you need to create resources in this project, a project administrator can grant you additional access by running this command:"

jhadvig:
@jwforres I like that. Only other thing is if we want to have the command in the same paragraph as your suggested working or under it?

jwforres:
i like the command on its own line, but i dont really like the single word wrapping into the center in the text in your original example, what does it look like with the updated sentence?

jhadvig:
@jwforres yeah my eyes are bleeding from that one... but the updated version looks pretty solid
screenshot-25

app/views/overview.html
```diff
@@ -14,16 +14,25 @@
...
+                      image, or add components like databases and message queues.
+                    </p>
+                    <p class="gutter-top">
+                      <a ng-href="project/{{projectName}}/create" class="btn btn-lg btn-primary">
+                        Add to Project
+                      </a>
+                    </p>
+                  </div>
+                  <div ng-if="!(project.metadata.name | canIAddToProject)">
+                    <h2>Welcome to {{projectName}} project.</h2>
```
jwforres: i'd probably switch to 'Welcome to project {{projectName}}.'

jhadvig:
@jwforres commet adressed

spadgett:
LGTM [merge]

PR #172, opened 2016-07-14
Create route not using selected service

spadgett:
Introduced by #143 

@jwforres PTAL


spadgett:
Edit route and add to project are OK.

PR #173, opened 2016-07-14
Fix route "Show annotations" link margins

spadgett:
The annotations directive was moved outside of the col div, causing this problem:

screen shot 2016-07-14 at 3 42 00 pm


spadgett:
@jwforres

jwforres:
that doesn't clip the annotations table when its shown does it?

spadgett:
> that doesn't clip the annotations table when its shown does it?

Not in my testing.

jwforres:
[merge]

PR #168, opened 2016-07-14
Only update overview metrics when in view

spadgett:
@jwforres Thoughts on this change?

Requires https://github.com/thenikso/angular-inview

PR #179, opened 2016-07-15
Remember collapsed service group state, open by default

spadgett:
Leave service groups open by default, but remember the collapsed state between sessions. Since #168, metrics are only loaded for deployments in view, so there is less of a performance concern defaulting to open.

Also fixes an issue where clicking the route link collapsed the section.

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/overview/serviceGroup.js
+++ b/app/scripts/directives/overview/serviceGroup.js
@@ -12,10 +12,35 @@ angular.module('openshiftConsole')
       scope: true,
       templateUrl: 'views/overview/_service-group.html',
       link: function($scope) {
-        // Collapse infrastructure services like Jenkins by default on page load.
-        if (ServicesService.isInfrastructure($scope.service)) {
-          $scope.collapse = true;
-        }
+        var localStorageCollapseKey = function() {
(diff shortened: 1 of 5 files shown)


app/scripts/directives/overview/serviceGroup.js
```diff
@@ -12,10 +12,35 @@ angular.module('openshiftConsole')
...
+        var wasCollapsed = function() {
+          var key = localStorageCollapseKey();
+          if (!key) {
+            return false;
+          }
+
+          return localStorage.getItem(key) === 'true';
+        };
+
+        var rememberCollapsedState = function() {
```
jwforres: persistCollapsedState would be more clear to me, remember is ambigous to me whether you are storing or recalling

jwforres:
just the one nit, otherwise LGTM

app/scripts/directives/overview/serviceGroup.js
spadgett: saveCollapsedState ok?

spadgett:
updated

jwforres:
yep works for me [merge]

Issue #146, opened 2016-07-11
Warning icons popping up on DCs now


jwforres:
The message field is now being used to show what triggered the last deployment on a DC.  We need to stop checking for the message field to trigger the warning icon.  Eventually this will be replaced by Conditions but currently no warnings are being set so just comment out the warnings for now until we have ones we can show.

PR #182, opened 2016-07-18
Bug 1356018: dc's warning triangle 'caused by config/image change' is not useful on web

jhadvig:
We dont need the warning anymore since we have a staparate column for the trigger info, which what DeploymentDetails message is for.
Screen:
screenshot-31

@jwforres PTAL

Fixes https://github.com/openshift/origin-web-console/issues/146

### DIFF
--- a/app/views/browse/deployment-config.html
+++ b/app/views/browse/deployment-config.html
@@ -78,7 +78,7 @@
                     </li>
                   </ul>
                 </div>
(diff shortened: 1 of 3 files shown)


jwforres:
We shouldn't be showing the warning icon anywhere now, see my comments in #146 about commenting out for now and not removing, since we will be getting back actual warnings in 1.4 most likely, just on a different field.

Pretty sure the icon shows up on the DC page too, not just on the list

jhadvig:
@jwforres you were right, the icon was on the DC page as well.
I've commented both of them.
PTAL

jwforres:
[merge]

PR #185, opened 2016-07-18
Bug 1357389: canI filter blocks some button from displaying

jhadvig:
The issue was the same as in case of settings page where we had to check if the `project` has been already loaded.
Checked the rest of the buttons where we use the `canI` filter found one additional resource mismatch.
@jwforres PTAL

### DIFF
--- a/app/views/browse/routes.html
+++ b/app/views/browse/routes.html
@@ -7,7 +7,7 @@
       <div class="middle-header header-light">
         <div class="container-fluid">
           <div class="page-header page-header-bleed-right page-header-bleed-left">
-            <div class="pull-right" ng-if="'routes' | canI : 'create'">
+            <div class="pull-right" ng-if="project && 'routes' | canI : 'create'">
               <a ng-href="project/{{project.metadata.name}}/create-route" class="btn btn-default">Create Route</a>
             </div>
             <h1>Routes</h1>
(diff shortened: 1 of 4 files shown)


app/views/browse/routes.html
```diff
@@ -7,7 +7,7 @@
       <div class="middle-header header-light">
         <div class="container-fluid">
           <div class="page-header page-header-bleed-right page-header-bleed-left">
-            <div class="pull-right" ng-if="'routes' | canI : 'create'">
+            <div class="pull-right" ng-if="!project || 'routes' | canI : 'create'">
```
spadgett: @jhadvig can you explain why the check is needed?

jwforres: on the other case it was an ng-hide and it was on the project settings page, so if the project had not been loaded yet you were supposed to hide the actions.

here this is an ng-if so this should be `project && 'routes' | canI : 'create'`

@spadgett the issue is that the canI filter gets evaluated initially to false (the i havent loaded rules yet fallback) before the project has loaded, there is no other data bound at this level in the view so nothing will trigger another digest loop on this part of the view after the rules actually load unless you also make sure the project has loaded first.  Its not ideal.  On all of the individual resource pages the header is underneath an ng-if="whateverResource"  and that resource gets loaded after the project is loaded, so they dont have this problem...

jhadvig: @jwforres thanks for the explanation and also for the fixing the logic.

jhadvig:
@jwforres I've update the PR, based on your comment. PTAL

jwforres:
[merge]

Issue #113, opened 2016-06-24
Improve build pipeline display when pipeline stages wrap


sg00dwin:
@jwforres @spadgett

sg00dwin:
<img width="1058" alt="screen shot 2016-06-24 at 4 06 06 pm" src="https://cloud.githubusercontent.com/assets/1874151/16350685/44593988-3a2e-11e6-94a2-51ff7413ac40.png">

<img width="1038" alt="screen shot 2016-06-24 at 4 44 47 pm" src="https://cloud.githubusercontent.com/assets/1874151/16350689/49299f52-3a2e-11e6-94b7-b2ddf4f29553.png">

<img width="1028" alt="screen shot 2016-06-24 at 4 54 45 pm" src="https://cloud.githubusercontent.com/assets/1874151/16350694/4d22212e-3a2e-11e6-8911-5c178eacef9f.png">

<img width="1027" alt="screen shot 2016-06-24 at 5 02 15 pm" src="https://cloud.githubusercontent.com/assets/1874151/16350698/52c52248-3a2e-11e6-9096-b645881fb9a8.png">

jwforres:
@ajacobs21e @sjcox-rh @rhamilto thoughts on these?  They are options until we can implement scrolling.

rhamilto:
I prefer "Thicker row border", although I think I would reduce the outermost left and right borders back to 1px, as it creates nice, clean separation without unnecessary "noise".

sg00dwin:
Additional options...

<img width="1044" alt="screen shot 2016-06-29 at 3 15 19 pm" src="https://cloud.githubusercontent.com/assets/1874151/16465456/d0c13c72-3e0c-11e6-81e4-ce4278fcad9e.png">
<img width="1051" alt="screen shot 2016-06-29 at 2 49 01 pm" src="https://cloud.githubusercontent.com/assets/1874151/16465457/d0c20e04-3e0c-11e6-8aeb-d8d0d077d361.png">
![new](https://cloud.githubusercontent.com/assets/1874151/16465540/4d1a2f18-3e0d-11e6-9b46-fb5317fcb84f.png)

rhamilto:
Nice work, @sg00dwin!  My opinion remains unchanged, however.  I think "Thicker row border" provides the needed separation without additional noise.

ajacobs21e:
@sg00dwin is horizontal scroll an option?

spadgett:
[...]

PR #186, opened 2016-07-18
Improved build pipeline display when pipeline stages wrap

rhamilto:
320px
400px
600px
800px
1000px
1200px
1600px

### DIFF
--- a/app/views/directives/build-pipeline.html
+++ b/app/views/directives/build-pipeline.html
@@ -1,4 +1,4 @@
-<div row mobile="column" flex>
+<div flex class="build-pipeline">
   <div column class="build-summary">
     <div>
       <span class="status-icon" ng-class="build.status.phase">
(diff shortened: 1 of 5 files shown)


app/views/directives/build-pipeline.html
```diff
@@ -24,27 +24,27 @@
...
-            <div class="pipeline-time" ng-class="stage.status" ng-if="!stage.durationMillis && !(stage | pipelineStagePendingInput)">not started</div>
+    </div>
+    <div class="pipeline">
+      <div class="pipeline-stage" ng-repeat="stage in jenkinsStatus.stages track by stage.id">
+        <div column class="pipeline-stage-column">
+          <div class="pipeline-stage-name" ng-class="build.status.phase">
+            {{stage.name}}
           </div>
+          <pipeline-status ng-if="stage.status" status="stage.status"></pipeline-status>
+          <!-- <div class="pipeline-time" ng-if="stage.durationMillis">{{stage.durationMillis | humanizeDurationValue}}</div> -->
```
spadgett: This isn't new, but do you mind taking out the commented code while here?

spadgett:
Just curious how it looks with an arrow instead of the right triangle.

This is definitely an improvement.

rhamilto:
Arrived at this:

screen shot 2016-07-19 at 3 43 46 pm
screen shot 2016-07-19 at 3 43 25 pm
screen shot 2016-07-19 at 3 43 14 pm
screen shot 2016-07-19 at 3 43 03 pm

PR to follow.

rhamilto:
@jwforres, please take a look.  This PR only addresses the display of stages as I'll need an assist on changing the display of pipelines on the overview page.

rhamilto:
Fixes #113

spadgett:
@rhamilto I have a feeling you won't like this comment... but the gap to the right of the last stage looks a little strange when none of my pipelines wrap :/

openshift_web_console

rhamilto:
@spadgett, I wrestled with that, and I've not been able to come up with a solution that maintains stages that are equal widths without moving the position of the arrow like so:

screen shot 2016-07-21 at 9 40 28 am
screen shot 2016-07-21 at 9 40 43 am

I think the gap on the right is the lesser of two evils.  Thoughts?

spadgett:
[merge]

PR #189, opened 2016-07-18
Create persistent volume claim form on console

zherman0:
This adds a new request button to the storage page that allows users to request a very simple storage claim.  The new form currently only allows for 3 basic inputs, which are sufficient to create a new PVC.
@jwforres @erinboyd

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -219,7 +219,11 @@ angular
       .when('/project/:project/edit', {
         templateUrl: 'views/edit/project.html',
         controller: 'EditProjectController'
-      })      
+      })
+      .when('/project/:project/create-pvc', {
+        templateUrl: 'views/create-persistent-volume-claim.html',
+        controller: 'CreatePersistentVolumeClaimController'
+      })
       .when('/project/:project/attach-pvc', {
         templateUrl: 'views/attach-pvc.html',
         controller: 'AttachPVCController'
--- a/app/scripts/controllers/createPersistentVolumeClaim.js
+++ b/app/scripts/controllers/createPersistentVolumeClaim.js
@@ -1,0 +1,83 @@
+'use strict';
+
+/**
+ * @ngdoc function
+ * @name openshiftConsole.controller:CreatePersistentVolumeClaimController
+ * @description
+ * # CreatePersistentVolumeClaimController
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('CreatePersistentVolumeClaimController', function ($filter, $routeParams, $scope, $window, ApplicationGenerator, DataService, Navigate, ProjectsService) {
+    $scope.alerts = {};
+    $scope.projectName = $routeParams.project;
+    $scope.accessModes="ReadWriteOnce";
+    $scope.claim = {};
+
+    $scope.breadcrumbs = [
+      {
+        title: $scope.projectName,
+        link: "project/" + $scope.projectName
+      },
+      {
+        title: "Storage",
+        link: "project/" + $scope.projectName + "/browse/storage"
+      },
+      {
+        title: "Request Storage"
+      }
+    ];
+
+    ProjectsService
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+        // Update project breadcrumb with display name.
+        $scope.breadcrumbs[0].title = $filter('displayName')(project);
+
+        $scope.createPersistentVolumeClaim = function() {
+          if ($scope.createPersistentVolumeClaimForm.$valid) {
+            $scope.disableInputs = true;
+            var claim = generatePersistentVolumeClaim();
+            DataService.create('persistentvolumeclaims', null, claim, context)
+              .then(function() { // Success
+                // Return to the previous page
+                $window.history.back();
+              },
+              function(result) { // Failure
+                $scope.disableInputs = false;
+                $scope.alerts['create-persistent-volume-claim'] = {
+                    type: "error",
+                    message: "An error occurred requesting storage claim.",
+                    details: $filter('getErrorDetails')(result)
+                };
+              });
+          }
+        };
+
+      function generatePersistentVolumeClaim() {
+        var pvc = {
+          kind: "PersistentVolumeClaim",
+          apiVersion: "v1",
+          metadata: {
+            name: $scope.claim.name,
+            labels: {}
+          },
+          spec: {
+            resources: {
+              requests:{}
+
+            }
+          }
+        };
+
+        pvc.spec.accessModes = [$scope.claim.accessModes || "ReadWriteOnce"] ;
+        var unit =  $scope.claim.unit || "Mi";
+        pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
+
+        return pvc;
+      }
+
+
+    }));
+  });
--- a/app/scripts/directives/oscPersistentVolumeClaim.js
+++ b/app/scripts/directives/oscPersistentVolumeClaim.js
@@ -1,0 +1,28 @@
+"use strict";
+
+angular.module("openshiftConsole")
+  .directive("oscPersistentVolumeClaim", function(){
+    return {
+      restrict: 'E',
+      scope: {
+        claim: "=model"
+      },
+      templateUrl: 'views/directives/osc-persistent-volume-claim.html',
+      link: function(scope) {
+        scope.claim.unit = 'Mi';
+        scope.units = [{
+          value: "Mi",
+          label: "MiB"
+          }, {
+          value: "Gi",
+          label: "GiB"
+          }, {
+          value: "Ti",
+          label: "TiB"
+          }, {
+          value: "Pi",
+          label: "PiB"
+          }];
+        }
+    };
+  });
--- a/app/views/create-persistent-volume-claim.html
+++ b/app/views/create-persistent-volume-claim.html
@@ -1,0 +1,52 @@
+<default-header class="top-header"></default-header>
+<div class="wrap no-sidebar">
+  <div class="sidebar-left collapse navbar-collapse navbar-collapse-2">
+    <navbar-utility-mobile></navbar-utility-mobile>
+  </div>
+  <div class="middle">
+    <!-- Middle section -->
+    <div class="middle-section surface-shaded">
+      <div class="middle-container has-scroll">
+        <div class="middle-content">
+          <div class="container surface-shaded">
+            <div class="col-md-12">
+              <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
+              <alerts alerts="alerts"></alerts>
+
+              <div class="row">
+                <div class="create-route-icon col-md-2 gutter-top hidden-sm hidden-xs">
+                  <span class="fa fa-hdd-o icon-xl"></span>
+                </div>
+                <div class="col-md-8">
+                  <h1>Request Storage</h1>
+                  <div>
+                    <span class="help-block">
+                      Create a request for an administrator defined storage asset by requesting size and access mode attributes for a best fit.
+                    </span>
+                  </div>
+                  <form name="createPersistentVolumeClaimForm">
+                    <div>
+                      <fieldset ng-disabled="disableInputs">
+                        <osc-persistent-volume-claim
+                          model="claim">
+                        </osc-persistent-volume-claim>
+                        <div class="button-group gutter-top gutter-bottom">
+                          <button type="submit"
+                                  class="btn btn-primary btn-lg"
+                                  ng-click="createPersistentVolumeClaim()"
+                                  ng-disabled="createPersistentVolumeClaimForm.$invalid || disableInputs"
+                                  value="">Create</button>
+                          <a class="btn btn-default btn-lg" href="" back>Cancel</a>
+                        </div>
+                      </fieldset>
+                    </div>
+                  </form>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div><!-- /middle-content -->
+      </div><!-- /middle-container -->
+    </div><!-- /middle-section -->
+  </div><!-- /middle -->
+</div><!-- /wrap -->
--- a/app/views/directives/osc-persistent-volume-claim.html
+++ b/app/views/directives/osc-persistent-volume-claim.html
@@ -1,0 +1,95 @@
+<ng-form name="persistentVolumeClaimForm">
+  <fieldset ng-disabled="claimDisabled">
+
+    <!-- Name -->
+    <div class="form-group">
+      <label for="claim-name" class="required">Name</label>
+      <input
+        id="claim-name"
+        class="form-control"
+        type="text"
+        name="name"
+        ng-model="claim.name"
+        ng-required="true"
+        ng-pattern="/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/"
+        ng-maxlength="253"
+        ng-minlength="2"
+        placeholder="my-storage-request"
+        select-on-focus
+        autocorrect="off"
+        autocapitalize="off"
+        spellcheck="false"
+        aria-describedby="claim-name-help">
+      <div>
+        <span id="claim-name-help" class="help-block">A unique name for the storage claim within the project.</span>
+      </div>
+      <div class="has-error" ng-show="persistentVolumeClaimForm.name.$error.pattern && persistentVolumeClaimForm.name.$touched && !claimDisabled">
+        <span class="help-block">
+          Claim names may only contain lower-case letters, numbers, and dashes.
+          They may not start or end with a dash. Max length of 253.
+        </span>
+      </div>
+    </div>
+
+    <div class="form-group" >
+      <label class="required">Access Mode</label><br/>
+      <div class="radio">
+
+        <label class="radio-inline">
+            <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadWriteOnce" aria-describedby="access-modes-help" ng-checked="true" >
+            Single User (RWO)
+        </label>
+
+        <label class="radio-inline">
+          <input type="radio" id="accessModes" name="accessModes" ng-model="claim.accessModes" value="ReadWriteMany" aria-describedby="access-modes-help" >
+            Shared Access (RWX)
+        </label>
+
+        <label class="radio-inline">
+            <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadOnlyMany" aria-describedby="access-modes-help">
+            Read Only (ROX)
+        </label>
+
+      </div>
+      <div>
+        <span id="access-modes-help" class="help-block">Permissions to the mounted volume.</span>
+      </div>
+    </div>
+
+    <!-- capacity -->
+    <div class="form-group">
+      <fieldset class="form-inline compute-resource">
+      <label class="required">Capacity</label>
+      <div ng-class="{ 'has-error': form.$invalid }">
+        <label class="sr-only">Amount</label>
+        <input type="number"
+               name="amount"
+               ng-attr-id="claim-amount"
+               ng-model="claim.amount"
+               ng-required="true"
+               min="0"
+               ng-attr-placeholder="10"
+               class="form-control"
+               ng-attr-aria-describedby="claim-capacity-help">
+        <label class="sr-only" >Unit</label>
+        <select ng-model="claim.unit"
+                name="unit"
+                ng-options="option.value as option.label for option in units"
+                ng-attr-id="claim-capacity-unit"
+                class="form-control inline-select">
+        </select>
+        </div>
+      <div>
+         <span id="claim-capacity-help" class="help-block">
+           Size of the storage request.
+        </span>
+      </div>
+      <div class="has-error" ng-show="persistentVolumeClaimForm.capacity.$error.pattern && persistentVolumeClaimForm.capacity.$touched && !claimDisabled">
+        <span class="help-block">
+          Must be a postive integer.
+        </span>
+      </div>
+      </fieldset>
+    </div>
+  </fieldset>
+</ng-form>
--- a/app/views/storage.html
+++ b/app/views/storage.html
@@ -21,7 +21,12 @@
         <div class="container-fluid">
           <div class="row">
             <div class="col-md-12">
-              <h2>Persistent Volume Claims</h2>
+              <div class="page-header page-header-bleed-right page-header-bleed-left">
+                <div class="pull-right">
+                  <a ng-href="project/{{project.metadata.name}}/create-pvc" class="btn btn-default">Request Storage</a>
+                </div>
+                <h2>Persistent Volume Claims</h2>
+              </div>
               <table class="table table-bordered table-hover table-mobile">
                 <thead>
                   <tr>
--- a/app/index.html
+++ b/app/index.html
@@ -227,6 +227,7 @@
         <script src="scripts/controllers/modals/linkService.js"></script>
         <script src="scripts/controllers/about.js"></script>
         <script src="scripts/controllers/commandLine.js"></script>
+        <script src="scripts/controllers/createPersistentVolumeClaim.js"></script>
         <script src="scripts/directives/date.js"></script>
         <script src="scripts/directives/deleteLink.js"></script>
         <script src="scripts/directives/editLink.js"></script>
@@ -239,6 +240,7 @@
         <script src="scripts/directives/oscImageSummary.js"></script>
         <script src="scripts/directives/oscKeyValues.js"></script>
         <script src="scripts/directives/oscRouting.js"></script>
+        <script src="scripts/directives/oscPersistentVolumeClaim.js"></script>
         <script src="scripts/directives/oscUnique.js"></script>
         <script src="scripts/directives/oscAutoscaling.js"></script>
         <script src="scripts/directives/replicas.js"></script>
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -227,6 +227,9 @@ controller:"CreateRouteController"
 }).when("/project/:project/edit", {
 templateUrl:"views/edit/project.html",
 controller:"EditProjectController"
+}).when("/project/:project/create-pvc", {
+templateUrl:"views/create-persistent-volume-claim.html",
+controller:"CreatePersistentVolumeClaimController"
 }).when("/project/:project/attach-pvc", {
 templateUrl:"views/attach-pvc.html",
 controller:"AttachPVCController"
@@ -5847,6 +5850,50 @@ kubernetes:c.VERSION.kubernetes
 c.withUser(), a.cliDownloadURL = d.CLI, a.cliDownloadURLPresent = a.cliDownloadURL && !_.isEmpty(a.cliDownloadURL), a.loginBaseURL = b.openshiftAPIBaseUrl(), a.sessionToken = c.UserStore().getToken(), a.showSessionToken = !1, a.toggleShowSessionToken = function() {
 a.showSessionToken = !a.showSessionToken;
 };
+} ]), angular.module("openshiftConsole").controller("CreatePersistentVolumeClaimController", [ "$filter", "$routeParams", "$scope", "$window", "ApplicationGenerator", "DataService", "Navigate", "ProjectsService", function(a, b, c, d, e, f, g, h) {
+c.alerts = {}, c.projectName = b.project, c.accessModes = "ReadWriteOnce", c.claim = {}, c.breadcrumbs = [ {
+title:c.projectName,
+link:"project/" + c.projectName
+}, {
+title:"Storage",
+link:"project/" + c.projectName + "/browse/storage"
+}, {
+title:"Request Storage"
+} ], h.get(b.project).then(_.spread(function(b, e) {
+function g() {
+var a = {
+kind:"PersistentVolumeClaim",
+apiVersion:"v1",
+metadata:{
+name:c.claim.name,
+labels:{}
+},
+spec:{
+resources:{
+requests:{}
+}
+}
+};
+a.spec.accessModes = [ c.claim.accessModes || "ReadWriteOnce" ];
+var b = c.claim.unit || "Mi";
+return a.spec.resources.requests.storage = c.claim.amount + b, a;
+}
+c.project = b, c.breadcrumbs[0].title = a("displayName")(b), c.createPersistentVolumeClaim = function() {
+if (c.createPersistentVolumeClaimForm.$valid) {
+c.disableInputs = !0;
+var b = g();
+f.create("persistentvolumeclaims", null, b, e).then(function() {
+d.history.back();
+}, function(b) {
+c.disableInputs = !1, c.alerts["create-persistent-volume-claim"] = {
+type:"error",
+message:"An error occurred requesting storage claim.",
+details:a("getErrorDetails")(b)
+};
+});
+}
+};
+}));
 } ]), angular.module("openshiftConsole").directive("relativeTimestamp", function() {
 return {
 restrict:"E",
@@ -6652,6 +6699,29 @@ link:function(a, b, c, d) {
 a.form = d, a.id = _.uniqueId("osc-routing-service-"), a.$watchGroup([ "model.service", "services" ], function() {
 _.has(a, "model.service") && !_.isEmpty(a.services) || _.set(a, "model.service", _.find(a.services));
 });
+}
+};
+}), angular.module("openshiftConsole").directive("oscPersistentVolumeClaim", function() {
+return {
+restrict:"E",
+scope:{
+claim:"=model"
+},
+templateUrl:"views/directives/osc-persistent-volume-claim.html",
+link:function(a) {
+a.claim.unit = "Mi", a.units = [ {
+value:"Mi",
+label:"MiB"
+}, {
+value:"Gi",
+label:"GiB"
+}, {
+value:"Ti",
+label:"TiB"
+}, {
+value:"Pi",
+label:"PiB"
+} ];
 }
 };
 }), angular.module("openshiftConsole").directive("oscUnique", function() {
--- a/dist/scripts/templates.js
+++ b/dist/scripts/templates.js
@@ -3331,6 +3331,56 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
   );
 
 
+  $templateCache.put('views/create-persistent-volume-claim.html',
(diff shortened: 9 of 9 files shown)


jwforres:
@zherman0 can you add a comment to the PR with screenshots for the current implementation

@spadgett can you review this one

zherman0:
Here is the new storage page with a single new button and the new form for requesting storage/claim.
storage_with_new pvc button
pvc_request storage form

spadgett:
@zherman0 Can you rebase so the PR only contains your commits?

zherman0:
I rebased and just pushed everything so it should be good - I am not sure what failed in the ci test

spadgett:
@zherman0 I still see unrelated commits in the PR. Can you clean up the history and squash to one commit?

zherman0:
I tried to squish it to one commit, but I think I have other stuff in there from the rebase.  Take a look and let me know.

spadgett:
Looks better, can you try the following to fix the `dist/` error?

```
$ hack/clean-deps.sh && hack/install-deps.sh && grunt build
```

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,79 @@
...
+ * @description
+ * # CreatePersistentVolumeClaimController
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('CreatePersistentVolumeClaimController', function ($filter, $routeParams, $scope, $window, ApplicationGenerator, DataService, Navigate, ProjectsService) {
+    $scope.alerts = {};
+    $scope.renderOptions = {
+      hideFilterWidget: true
+    };
```
spadgett: I don't think you need `renderOptions` here.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,79 @@
...
+    ProjectsService
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+        // Update project breadcrumb with display name.
+        $scope.breadcrumbs[0].title = $filter('displayName')(project);
+        var labels = {},
+            orderByDisplayName = $filter('orderByDisplayName');
+
+        DataService.list("services", context, function(services) {
```
spadgett: I'm unclear why the controller is requesting services. You shouldn't need a service to create a PVC? (Here and the route params above.)

app/scripts/services/applicationGenerator.js
```diff
@@ -402,6 +402,55 @@ angular.module("openshiftConsole")
...
     };
 
+  // Public method for creating a persistent volume claim.
+    // Expects pvcOptions from the osc-persistent volume claim directive.
+    scope.createPersistentVolumeClaim = function(pvcOptions, serviceName, labels) {
+      return scope._generatePersistentVolumeClaim({
+        labels: labels || {},
+        params: angular.extend({ include: true }, pvcOptions.persistentVolumeClaimForm)
+      }, pvcOptions.persistentVolumeClaimForm.name, serviceName);
+    };
```
spadgett: If this can only be used with the persistent volume claim form from CreatePersistentVolumeClaimController, I'd probably just put the function in that controller rather than ApplicationGenerator.

app/scripts/services/applicationGenerator.js
```diff
@@ -402,6 +402,55 @@ angular.module("openshiftConsole")
...
+    scope._generatePersistentVolumeClaim = function(input, name, serviceName){
+      if(!input.params.include) {
+        return null;
+      }
+
+      var pvc = {
+        kind: "PersistentVolumeClaim",
+        apiVersion: "v1",
+        metadata: {
+          name: name.$viewValue,
```
spadgett: We shouldn't need to access `$viewValue` directly. Can we just use `$scope.claim.name`? (Assuming this is moved to the PVC controller)

app/views/create-persistent-volume-claim.html
```diff
@@ -0,0 +1,55 @@
...
+      <div class="middle-container has-scroll">
+        <div class="middle-content">
+          <div class="container surface-shaded">
+            <div class="col-md-12">
+              <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
+              <alerts alerts="alerts"></alerts>
+
+              <div class="row">
+                <div class="create-route-icon col-md-2 gutter-top hidden-sm hidden-xs">
+                  <span class="pficon pficon-storage-domain icon-xl"></span>
```
spadgett: Hm, is this the right icon for PVC? @jwforres

app/views/create-persistent-volume-claim.html
```diff
@@ -0,0 +1,55 @@
...
+                </div>
+                <div class="col-md-8">
+                  <h1>Request Storage</h1>
+                  <div>
+                    <span class="help-block">
+                      Create a request for an administrator defined storage asset by requesting size and access mode attributes for a best fit.
```
spadgett: I'd add a "Learn more" link to the docs like we do for other forms.

app/views/create-persistent-volume-claim.html
```diff
@@ -0,0 +1,55 @@
...
+                          services="services"
+                          show-name-input="true">
+                        </osc-persistent-volume-claim>
+                        <div class="button-group gutter-top gutter-bottom">
+                          <button type="submit"
+                                  class="btn btn-primary btn-lg"
+                                  ng-click="createPersistentVolumeClaim()"
+                                  ng-disabled="createPersistentVolumeClaimForm.$invalid || disableInputs || !createPersistentVolumeClaimForm"
+                                  value="">Create</button>
+                          <a class="btn btn-default btn-lg" href="#" back>Cancel</a>
```
spadgett: `href=""` instead of `href="#"`

https://docs.angularjs.org/api/ng/directive/a

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,42 @@
...
+   *     {
+   *       name: "",
+   *       accessMode: "",
+   *       Capacity: ""
+   *     }
+   *
+   */
+  .directive("oscPersistentVolumeClaim", function(){
+    return {
+      require: '^form',
```
spadgett: Since you use `ng-form` in the view, I don't think you need this. You can probably remove `require: '^form'` as well as the `controller` and `link` functions below.

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,42 @@
...
+      restrict: 'E',
+      scope: {
+        route: "=model",
```
spadgett: Why is this called "route"?

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,42 @@
...
+        services: "=",
+        showNameInput: "=",
+        routingDisabled: "="
```
spadgett: You can probably don't need any of these three scope properties (services, showNameInput, or routingDisabled). The fieldset in the create form should take care of disabling the inputs, and we should always show the name input.

app/scripts/services/applicationGenerator.js
```diff
@@ -402,6 +402,55 @@ angular.module("openshiftConsole")
...
+          labels: input.labels,
+          annotations: input.annotations
+        },
+        spec: {
+          to: {
+            kind: "Service",
+            name: serviceName
+          },
```
spadgett: There's no `spec.to` property on PVCs.

app/views/create-persistent-volume-claim.html
```diff
@@ -0,0 +1,55 @@
...
+                <div class="col-md-8">
+                  <h1>Request Storage</h1>
+                  <div>
+                    <span class="help-block">
+                      Create a request for an administrator defined storage asset by requesting size and access mode attributes for a best fit.
+                    </span>
+                  </div>
+                  <form name="createPersistentVolumeClaimForm">
+                    <div ng-if="!services">Loading...</div>
+                    <div ng-if="services">
```
spadgett: You should be able to remove the services check.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+      <!-- regex, maxlength from k8s.io/kubernetes/pkg/util/validation/validation.go -->
+      <input
+        id="claim-name"
+        class="form-control"
+        type="text"
+        name="name"
+        ng-model="claim.name"
+        ng-required="showNameInput"
+        ng-pattern="/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/"
+        ng-maxlength="63"
```
spadgett: Claim names can be 253 characters.

```
ng-maxlength="253"
```

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+        aria-describedby="claim-name-help">
+      <div>
+        <span id="claim-name-help" class="help-block">A unique name for the storage claim within the project.</span>
+      </div>
+      <div class="has-error" ng-show="persistentVolumeClaimForm.name.$error.pattern && persistentVolumeClaimForm.name.$touched && !routingDisabled">
+        <span class="help-block">
+          Claim names may only contain lower-case letters, numbers, and dashes.
+          They may not start or end with a dash.
+        </span>
+      </div>
```
spadgett: Do we need an error message for maxlength as well?

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+    </div>
+
+    <div class="form-group" >
+      <label class="required">Access Mode</label><br/>
+      <!--modes = { 'RWX':'Read/Write Many', 'RWO':'Read/Write Once', 'ROX':'Read-Only Many' };-->
+      <div class="radio" ng-app ng-init="claim.accessModes=ReadWriteOnce">
+        <!--<div ng-init="claim.accessModes=ReadWriteOnce" />-->
```
spadgett: Remove commented out code.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+
+
+        <label class="radio-inline">
+          <fieldset>
```
spadgett: I don't think we should need the fieldset around each radio?

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+          Claim names may only contain lower-case letters, numbers, and dashes.
+          They may not start or end with a dash.
+        </span>
+      </div>
+    </div>
+
+    <div class="form-group" >
+      <label class="required">Access Mode</label><br/>
+      <!--modes = { 'RWX':'Read/Write Many', 'RWO':'Read/Write Once', 'ROX':'Read-Only Many' };-->
+      <div class="radio" ng-app ng-init="claim.accessModes=ReadWriteOnce">
```
spadgett: Why is `ng-app` on this element?

I'd rather initialize the value once in the controller rather than using `ng-init`.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+        <label class="radio-inline">
+          <fieldset>
+            <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadOnlyMany" aria-describedby="service-help">
+            Read Only (ROX)
+          </fieldset>
+        </label>
+
+      </div>
+      <div>
+        <span id="service-help" class="help-block">Accessibility permissions to the mounted volume.</span>
```
spadgett: `id="access-modes-help"`?

Maybe just "Permissions to the mounted volume."

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
+<ng-form name="persistentVolumeClaimForm">
+  <fieldset ng-disabled="routingDisabled">
+
+    <!-- Name -->
+    <div ng-show="showNameInput" class="form-group">
+      <label for="claim-name" class="required">Name</label>
+      <!-- regex, maxlength from k8s.io/kubernetes/pkg/util/validation/validation.go -->
```
spadgett: I don't think the path to validation.go is correct anymore. Maybe just remove the comment here and below.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+      <input
+        id="capacity"
+        class="form-control"
+        type="text"
+        name="capacity"
+        required
+        ng-model="claim.capacity"
+        ng-pattern="/^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$/"
+        ng-maxlength="253"
```
spadgett: I don't think capacity has a maxlength.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+        placeholder="1Gi"
+        autocorrect="off"
+        autocapitalize="off"
+        spellcheck="false"
+        aria-describedby="route-capacity-help">
+      <div>
+        <span id="route-capacity-help" class="help-block">
```
spadgett: `id="claim-capacity-help"`

app/views/create-persistent-volume-claim.html
jwforres: To be consistent with Attach Storage it should be 'fa fa-hdd-o icon-xl'

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,98 @@
...
+      <div>
+        <span id="service-help" class="help-block">Accessibility permissions to the mounted volume.</span>
+      </div>
+    </div>
+
+    <!-- capacity -->
+    <div class="form-group">
+      <label for="capacity" class="required">Capacity</label>
+      <!-- regex, maxlength from k8s.io/kubernetes/pkg/util/validation/validation.go -->
+      <input
```
spadgett: Rather than just a plain input, I'd rather have a text input for the number with a select for the units. We do this for edit resource limits elsewhere. See

https://github.com/openshift/origin-web-console/blob/master/app/views/_compute-resource.html#L3-L19

and

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/editRequestLimit.js#L67-L83

openshift_web_console

app/scripts/app.js
```diff
@@ -250,6 +250,10 @@ angular
         templateUrl: 'views/create-route.html',
         controller: 'CreateRouteController'
       })
+      .when('/project/:project/create-persistent-volume-claim', {
```
spadgett: I'd use `/create-pvc` to be consistent with `/attach-pvc` below.

spadgett:
Can we add a create persistent volume claim action to the empty state message on this page?

screen shot 2016-07-20 at 10 17 07 am

The code is here:

https://github.com/openshift/origin-web-console/blob/master/app/views/attach-pvc.html#L19-L35

Here is an example empty state message that has a button:

https://github.com/openshift/origin-web-console/blob/master/app/views/overview.html#L17-L28

app/scripts/directives/oscPersistentVolumeClaim.js
zherman0: Copied and pasted from the route code and forgot to update - I will change that to claim

app/scripts/directives/oscPersistentVolumeClaim.js
zherman0: I left the controller and link in there for *possible* future enhancements, but I will remove it for now.

spadgett: Thanks, we can always add it back if needed.

app/scripts/services/applicationGenerator.js
zherman0: I moved this out of appgenerator and into the PVC Controller but I still don't have direct access to the name value.  Instead it is an object and the only way I seem to be able to get direct access is through the $viewValue.

spadgett:
@zherman0 When you update to latest master, use `git rebase` or `git pull --rebase` instead of `git pull` or `git merge`. This avoids the extra commits in your pull request.

app/scripts/services/applicationGenerator.js
spadgett: If you have `ng-model` on the input, you should be able to access it as a scope variable like `$scope.claim.name`.

zherman0:
@spadgett - I am rebasing this once more, but do you think we can get this merged in today?

spadgett:
> I am rebasing this once more, but do you think we can get this merged in today?

@zherman0 I've been waiting for you to address https://github.com/openshift/origin-web-console/pull/189#issuecomment-234532832 before reviewing again. The PR has over 30 unrelated commits currently.

app/scripts/controllers/createPersistentVolumeClaim.js
spadgett: @zherman0 Is this another carry-over from the osc routing directive? I don't think you need to do anything with services for PVCs.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,115 @@
...
+        });
+
+        $scope.createPersistentVolumeClaim = function() {
+          if ($scope.createPersistentVolumeClaimForm.$valid) {
+            $scope.disableInputs = true;
+            var claimForm = $scope.createPersistentVolumeClaimForm.persistentVolumeClaimForm;
+            var claim = $scope._generatePersistentVolumeClaim({
+                                        labels: labels || {},
+                                        params: angular.extend({ include: true }, claimForm)
+                                      }, claimForm.name);
```
spadgett: You shouldn't need to pass the form to _generatePersistentVolumeClaim. It should be able to read $scope directly since it's part of the same controller.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,115 @@
...
+                  details: $filter('getErrorDetails')(result)
+                };
+              });
+          }
+        };
+
+        $scope._generatePersistentVolumeClaim = function(input){
+              if(!input.params.include) {
+                return null;
+              }
```
spadgett: No need to input.params.include here. You'll always want to return something when this method is called.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,115 @@
...
+                $scope.alerts['create-persistent-volume-claim'] = {
+                  type: "error",
+                  message: "An error occurred requesting storage claim.",
+                  details: $filter('getErrorDetails')(result)
+                };
+              });
+          }
+        };
+
+        $scope._generatePersistentVolumeClaim = function(input){
```
spadgett: Doesn't need to be a scope function since it's not called from the view. I'd just use

```js
function _generatePersistentVolumeClaim() { 
  // ...
}
```

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,115 @@
...
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+        // Update project breadcrumb with display name.
+        $scope.breadcrumbs[0].title = $filter('displayName')(project);
+        var labels = {},
+            orderByDisplayName = $filter('orderByDisplayName');
+
+        DataService.list("services", context, function(services) {
+//          console.log("Context:(list)", context);
```
spadgett: Remove commented out debugging statement.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,115 @@
...
+              if(!input.params.include) {
+                return null;
+              }
+
+              var pvc = {
+                kind: "PersistentVolumeClaim",
+                apiVersion: "v1",
+                metadata: {
+                  name: $scope.claim.name,
+                  labels: input.labels,
```
spadgett: We should either just use `{}` or add a labels field to the form.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,115 @@
...
+                  annotations: input.annotations
```
spadgett: This will always be unset with the code you have, so you could just remove the annotations.

app/scripts/controllers/createPersistentVolumeClaim.js
spadgett: Can you fix the indentation in this function as well?

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,71 @@
...
+    return {
+      restrict: 'E',
+      scope: {
+        claim: "=model",
+        showNameInput: "=",
+        showAmountInput:"="
+      },
+      templateUrl: 'views/directives/osc-persistent-volume-claim.html',
+      link: function(scope) {
+              var amountAndUnit = $filter('amountAndUnit');
```
spadgett: Can you clean up the indentation so it's consistently 2 spaces?

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,71 @@
...
+              var addUnitOption = function(unit) {
+                if (!_.some(scope.units, { value: unit })) {
+                  scope.units.push({
+                    value: unit,
+                    label: humanizeUnit(unit, scope.type)
+                  });
+                }
+              };
+
+              scope.$watch('defaultValue', function(defaultValue) {
```
spadgett: There's no default value for PVCs unlike resource limits. You should be able to remove this code.

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,71 @@
...
+                    scope.unit = defaultUnit;
+                  }
+                });
+                if (defaultValue) {
+                  setDefault(amountAndUnit(defaultValue, scope.type));
+                }
+              });
+
+              // Set unit options and based on type.
+              switch (scope.type) {
```
spadgett: No need to a switch statement if there is only one default case.

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,71 @@
+"use strict";
+
+angular.module("openshiftConsole")
+  .directive("oscPersistentVolumeClaim", function($filter){
+    return {
+      restrict: 'E',
+      scope: {
+        claim: "=model",
+        showNameInput: "=",
+        showAmountInput:"="
```
spadgett: Would we ever hide this? I think I'd just simplify the directive and just take in "claim" and nothing else until we need it.

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,71 @@
...
+
+              // If unit is not already in options, add it.
+              var addUnitOption = function(unit) {
+                if (!_.some(scope.units, { value: unit })) {
+                  scope.units.push({
+                    value: unit,
+                    label: humanizeUnit(unit, scope.type)
+                  });
+                }
+              };
```
spadgett: This block here was for editing existing resources. I think we can get rid of it here and always default it like you do before.

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,71 @@
...
+                if (defaultValue) {
+                  setDefault(amountAndUnit(defaultValue, scope.type));
+                }
+              });
+
+              // Set unit options and based on type.
+              switch (scope.type) {
+              default:
+                scope.unit = 'Mi';
+                scope.claim.unit = angular.copy(scope.unit);
```
spadgett: Just set scope.claim.unit directly and only have one scope var for it.

app/views/create-persistent-volume-claim.html
```diff
@@ -0,0 +1,55 @@
...
+                    <span class="help-block">
+                      Create a request for an administrator defined storage asset by requesting size and access mode attributes for a best fit.
+                    </span>
+                  </div>
+                  <form name="createPersistentVolumeClaimForm">
+                    <div>
+                      <fieldset ng-disabled="disableInputs">
+                        <osc-persistent-volume-claim
+                          model="claim"
+                          services="services"
```
spadgett: Should not need to pass in services (in fact, the directive does not have it defined for its scope).

app/views/create-persistent-volume-claim.html
```diff
@@ -0,0 +1,55 @@
...
+                          show-name-input="true"
+                          show-amount-input="true">
+                        </osc-persistent-volume-claim>
+                        <div class="button-group gutter-top gutter-bottom">
+                          <button type="submit"
+                                  class="btn btn-primary btn-lg"
+                                  ng-click="createPersistentVolumeClaim()"
+                                  ng-disabled="createPersistentVolumeClaimForm.$invalid || disableInputs || !createPersistentVolumeClaimForm"
```
spadgett:
```
!createPersistentVolumeClaimForm
```

Won't that always eval to false?

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,71 @@
...
+        showNameInput: "=",
+        showAmountInput:"="
+      },
+      templateUrl: 'views/directives/osc-persistent-volume-claim.html',
+      link: function(scope) {
+              var amountAndUnit = $filter('amountAndUnit');
+              var humanizeUnit = $filter('humanizeUnit');
+
+              // Create a unique ID for `label for` and `aria-describedby` attributes.
+              scope.id = _.uniqueId('storage-size-');
```
spadgett: This was part of the compute-resource directive since it was included in the same page multiple times. I'd just remove scope.id here and give each element its own ID. The create PVC will be used once on any page. If you do use scope.id, you should use it for all fields, not just one or two. Easier to remove it though.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,95 @@
...
+
+    <!-- capacity -->
+    <div class="form-group">
+      <fieldset class="form-inline compute-resource">
+      <label ng-if="label">{{label}}</label>
+      <div ng-class="{ 'has-error': form.$invalid }">
+        <label class="sr-only" ng-attr-for="{{id}}">Amount</label>
+        <input type="number"
+               name="amount"
+               ng-attr-id="{{id}}"
```
spadgett: Per my comment above, just give it an ID.

```html
id="storage-amount"
```

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,95 @@
...
+               ng-model="claim.amount"
+               ng-required="showAmountInput"
+               min="0"
+               ng-attr-placeholder="{{placeholder}}"
+               class="form-control"
+               ng-attr-aria-describedby="{{description ? id + '-help' : undefined}}">
```
spadgett:
```html
aria-describedby="claim-capacity-help"
```

app/views/directives/osc-persistent-volume-claim.html
spadgett: or maybe

```
id="claim-amount"
```

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,95 @@
...
+        <label class="sr-only" ng-attr-for="{{id}}-unit">Unit</label>
+        <select ng-model="claim.unit"
+                name="unit"
+                ng-options="option.value as option.label for option in units"
+                ng-attr-id="{{id}}-unit"
```
spadgett: This field shouldn't need an ID.

spadgett:
```html
id="claim-capacity-unit"
```

?

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,95 @@
...
+        </select>
+        </div>
+      <div>
+         <span id="claim-capacity-help" class="help-block">
+           Size of the storage request.
+        </span>
+      </div>
+      <div class="has-error" ng-show="persistentVolumeClaimForm.capacity.$error.pattern && persistentVolumeClaimForm.capacity.$touched && !routingDisabled">
+        <span class="help-block">
+          Must be integer with Gi, Mi, Tb as in 1 Gi for 1 Gigabyte
```
spadgett: This message looks wrong now that we have a separate dropdown for the unit.

app/views/storage.html
```diff
@@ -21,7 +21,12 @@
         <div class="container-fluid">
           <div class="row">
             <div class="col-md-12">
-              <h2>Persistent Volume Claims</h2>
+              <div class="page-header page-header-bleed-right page-header-bleed-left">
+                <div class="pull-right">
+                  <a ng-href="project/{{project.metadata.name}}/create-pvc" class="btn btn-default">Request Storage</a>
+                </div>
+                <h2>Persistent Volume Claims</h2>
+                </div>
```
spadgett: Indentation.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,95 @@
...
+        </span>
+      </div>
+    </div>
+
+    <div class="form-group" >
+      <label class="required">Access Mode</label><br/>
+      <div class="radio">
+
+        <label class="radio-inline">
+            <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadWriteOnce" aria-describedby="service-help" ng-checked="true" >
```
spadgett:
```
aria-describedby="access-modes-help"
```

Here and on the other radio buttons.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,115 @@
...
+ */
+angular.module('openshiftConsole')
+  .controller('CreatePersistentVolumeClaimController', function ($filter, $routeParams, $scope, $window, ApplicationGenerator, DataService, Navigate, ProjectsService) {
+    $scope.alerts = {};
+    $scope.projectName = $routeParams.project;
+    $scope.serviceName = $routeParams.service;
+    $scope.accessModes="ReadWriteOnce";
+    $scope.pvc = {
+      name: $scope.serviceName || ""
+    };
```
spadgett: `$scope.pvc` is not actually read anywhere. You should be able to remove it (when you remove the list services block).

zherman0:
@spadgett - I made changes using your suggestions.  See what you think.  Thanks so much for your input so far; it has been awesome.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,84 @@
...
+ * @name openshiftConsole.controller:CreatePersistentVolumeClaimController
+ * @description
+ * # CreatePersistentVolumeClaimController
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('CreatePersistentVolumeClaimController', function ($filter, $routeParams, $scope, $window, ApplicationGenerator, DataService, Navigate, ProjectsService) {
+    $scope.alerts = {};
+    $scope.projectName = $routeParams.project;
+    $scope.serviceName = $routeParams.service;
```
spadgett: `$scope.serviceName` can be removed.

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -0,0 +1,31 @@
...
+    return {
+      restrict: 'E',
+      scope: {
+        claim: "=model",
+        showNameInput: "=",
+        showAmountInput:"="
+      },
+      templateUrl: 'views/directives/osc-persistent-volume-claim.html',
+      link: function(scope) {
+        scope.id = 'storage-size';
```
spadgett: Should be able to remove `scope.id`.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,95 @@
...
+        </select>
+        </div>
+      <div>
+         <span id="claim-capacity-help" class="help-block">
+           Size of the storage request.
+        </span>
+      </div>
+      <div class="has-error" ng-show="persistentVolumeClaimForm.capacity.$error.pattern && persistentVolumeClaimForm.capacity.$touched && !claimDisabled">
+        <span class="help-block">
+          Must be a postive integer.  Also select units from the drop down.
```
spadgett: Do we need to say to select units? I assume since there is a default value and no empty option there is always a value.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -0,0 +1,84 @@
...
+                $window.history.back();
+                  },
+                  function(result) { // Failure
+                    $scope.disableInputs = false;
+                    $scope.alerts['create-persistent-volume-claim'] = {
+                        type: "error",
+                        message: "An error occurred requesting storage claim.",
+                        details: $filter('getErrorDetails')(result)
+                      };
+                  });
```
spadgett: nit: indentation

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -0,0 +1,95 @@
...
+      <label class="required">Capacity</label>
+      <div ng-class="{ 'has-error': form.$invalid }">
+        <label class="sr-only">Amount</label>
+        <input type="number"
+               name="amount"
+               ng-attr-id="claim-amount"
+               ng-model="claim.amount"
+               ng-required="showAmountInput"
+               min="0"
+               ng-attr-placeholder="1"
```
spadgett: Doesn't need ng-attr- prefix.

```html
placeholder="1"
```

Although I wonder if we shouldn't use a larger value as an example.

spadgett:
Thanks @zherman0, this is looking a lot cleaner. I know I make a lot of comments.

Make sure you squash your commit into one when you're done.

```
$ git checkout claimCreateForm
$ git rebase -i master
```

Then use "fixup" on the extra commits.

zherman0:
@spadgett - I rebased this and squashed it down to one commit with the last tweaks we discussed

spadgett:
@zherman0 Seeing this error in travis:

spadgett:
LGTM, thanks @zherman0

[merge]

Issue #176, opened 2016-07-14
Broken with Origin 1.2 API since introduction of what-can-i-do


rassie:
Even though this is probably not the main use case, I've been successfully using a git checkout of `origin-web-console` with productive OpenShift Origin 1.2 API for several weeks now. However, since the introduction of what-can-i-do in https://github.com/openshift/origin-web-console/pull/82 the console has been broken in this configuration: every attempt to open a project fails with "User permissions for project $PROJECT could not be loaded."

This is apparently due to the type `SelfSubjectRulesReview` missing in 1.2 (introduced in 1.3.0.alpha1). Like I said, I'm fully aware this is not in any way an intended use case, but I would really appreciate some sort of backward-compatibility with the last stable Origin release, since it allows us to preview upcoming user-facing features with very low overhead.

Would it be possible to make what-can-i-do feature optional in the sense of non-blocking for the main functionality?

liggitt:
I could see creating tags in the web console repo to match origin versions, but conditionally supporting multiple server/web console version pairs adds more complexity than I think we want to carry

jwforres:
we can fallback to be permissive if the selfsubjectrulesreview check fails, if the user cant actually do something they will get an error like they did before

PR #190, opened 2016-07-19
Permissive mode for the what-can-i-do authorization

jhadvig:
@jwforres @liggitt I've update the `AuthorizationService` so before loading projects rules it will check with the `ApiService.apiInfo()` if it's aware of the 'selfsubjectrulesreviews'. If not, that means that the console is running against older APIserver, then set the "permissive mode" in which `canI` and `canIAddToProject` will return `true` without checking anything and all the action buttons will be rendered. If the 'selfsubjectrulesreviews' type is present in the APIserver then do the regular checks to determine if the action buttons shall be rendered.
PTAL

Closes https://github.com/openshift/origin-web-console/issues/176

### DIFF
--- a/app/scripts/services/authorization.js
+++ b/app/scripts/services/authorization.js
@@ -1,12 +1,14 @@
 'use strict';
 
 angular.module("openshiftConsole")
-  .factory("AuthorizationService", function($q, $cacheFactory, Logger, $interval, DataService){
+  .factory("AuthorizationService", function($q, $cacheFactory, Logger, $interval, APIService, DataService){
     
     var currentProject = null;
     var cachedRulesByProject = $cacheFactory('rulesCache', {
           number: 10
         });
+    // Permisive mode will cause no checks to be done for the user actions.
+    var permissiveMode = false;
 
     var REVIEW_RESOURCES = ["localresourceaccessreviews", "localsubjectaccessreviews", "resourceaccessreviews", "selfsubjectrulesreviews", "subjectaccessreviews"];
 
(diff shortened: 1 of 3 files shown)


app/scripts/services/authorization.js
```diff
@@ -52,24 +54,32 @@ angular.module("openshiftConsole")
...
+              var normalizedData = normalizeRules(data.status.rules);
+              var canUserAddToProject = canAddToProjectCheck(data.status.rules);
+              cachedRulesByProject.put(projectName, {rules: normalizedData,
+                                                      canAddToProject: canUserAddToProject,
+                                                      forceRefresh: false,
+                                                      cacheTimestamp: _.now()
+                                                    });
+              deferred.resolve();
+            }, function() {
+              deferred.reject();
```
jwforres: This should also turn on permissive mode.  If we tried to get the rules and failed to get the rules then we should fallback to showing the actions.

jhadvig: @jwforres wasn't sure of that... thought that since when the `selfsubjectrulesreviews` is present in the APIserver then we should be determining if to show the action.

liggitt: an error fetching rules doesn't tell us whether or not the user has access. we should err on the side of showing the user actions that might not work

jhadvig:
@jwforres comment addressed. PTAL

app/scripts/services/projects.js
```diff
@@ -38,13 +38,6 @@ angular.module('openshiftConsole')
                                           context.projectPromise.resolve(project);
                                           // TODO: fix need to return context & projectPromise
                                           return [project, context];
-                                        }, function() {
```
jwforres: You might still need an error handler here that returns the [project, context]  since we don't want it to fail resolving the project if the rules review fails.

app/scripts/services/authorization.js
```diff
@@ -52,24 +54,33 @@ angular.module("openshiftConsole")
...
+                                                    });
+              deferred.resolve();
+            }, function() {
+              permissiveMode = true;
+              deferred.resolve();
+          });
+        } else {
+          Logger.log("AuthorizationService, resource 'selfsubjectrulesreviews' is not part of APIserver. Switching into permissive mode.");
+          permissiveMode = true;
+          deferred.resolve();
```
jhadvig: I've replaced `deferred.reject()` with `deferred.resolve()` here

jwforres:
[merge]

PR #191, opened 2016-07-19
Fix clipped descenders in empty-state messages

spadgett:
Fixes this problem:

screen shot 2016-07-19 at 8 21 37 am

@jwforres


jwforres:
[merge]

PR #193, opened 2016-07-20
Achieve equal height deployment, linked/unlinked service, and no deployment tiles.

sg00dwin:


sg00dwin:
@spadgett @jwforres

app/styles/_overview.less
```diff
@@ -97,59 +106,94 @@
...
-      }
-
-      &:nth-child(even) {
-          padding-left: 5px;
-      }
-
+      .flex(@columns: 0 0 49%);
+    } 
+    @media (min-width: @screen-md-min) {
+      .flex(@columns: 0 0 49.375%);
```
spadgett: This is very specific. I'm guessing 50% doesn't work?

app/styles/_overview.less
```diff
@@ -97,59 +106,94 @@
...
+      background: #00b9e4;
+      left: -2px;
+      top: 0;
+      z-index: 1;
     }
+    @media (min-width: @screen-sm-min) {
+      .flex(@columns: 0 0 49%);
+    } 
+    @media (min-width: @screen-md-min) {
+      .flex(@columns: 0 0 49.375%);
```
spadgett: If we have a value like this in more than one place, probably should be a variable.

app/styles/_overview.less
```diff
@@ -97,59 +106,94 @@
...
-      }
+    margin-bottom: 20px;
+    position: relative;
+    &:before {
+      content: "";
+      display: block;
+      height: 100%;
+      width: 3px;
+      position: absolute;
+      background: #00b9e4;
```
spadgett: `@color-pf-light-blue-400` ?

app/views/overview.html
```diff
@@ -42,18 +42,18 @@
                    ng-if="service.metadata.labels.app || routesByService[service.metadata.name].length || childServicesByParent[service.metadata.name].length">
                 <overview-service-group></overview-service-group>
               </div>
-
-              <div row wrap>
+              
+              <div row wrap class="overview-standalone-service">
```
spadgett: `standalone-service-row`? The current name sounds too much like `standalone-service` used below.

app/views/overview.html
```diff
@@ -42,18 +42,18 @@
...
                     ng-if="!service.metadata.labels.app && !routesByService[service.metadata.name].length && !childServicesByParent[service.metadata.name].length"
                     class="standalone-service">
                   <overview-service-group></overview-service-group>
                 </div>
               </div>
-
+            
               <!-- Unserviced DCs, RCs, and pods -->
-              <div row wrap ng-if="(monopodsByService[''] | hashSize) > 0 || (deploymentConfigsByService[''] | hashSize) > 0 || (deploymentsByService[''] | hashSize) > 0">
+              <div row wrap ng-if="(monopodsByService[''] | hashSize) > 0 || (deploymentConfigsByService[''] | hashSize) > 0 || (deploymentsByService[''] | hashSize) > 0" class="overview-monopods">
```
spadgett: "monopod" is not really the right name since it can be other things, too. `class="unserviced-row"` maybe.

app/views/overview/_service.html
```diff
@@ -1,14 +1,16 @@
...
+      <p>
+        There are no deployments or pods for service
+        <a ng-href="{{service | navigateResourceURL}}">{{service.metadata.name}}</a>.
+      </p>
+    </div>
   </div>
 </div>
 
-<div ng-attr-row="{{!service ? '' : undefined}}" ng-attr-wrap="{{!service ? '' : undefined}}" ng-class="{'no-service': !service}">
+<div ng-attr-row="{{!service ? '' : undefined}}" ng-attr-wrap="{{!service ? '' : undefined}}" ng-class="{'no-service': !service, 'service-multiple-targets': (visibleDeploymentsByConfig | hashSize) + (monopodsByService[service.metadata.name] | hashSize) > 1}" class="deployment-block">
```
spadgett: @sg00dwin I need to double check what I have here. I think I missed one case.

app/styles/_overview.less
spadgett: We at least need a comment explaining.

app/views/overview/_service.html
spadgett: Yeah, service -> multiple RCs is not laying out correctly. I'll take a look.

app/styles/_overview.less
sg00dwin: Since it uses flex-basis to set percentage widths for the 2 columns and the remainder is for the gutter in between, padding and negative margins are no longer needed for gutters and positioning. To maintain an approximate gutter width around 15px the flex-basis is adjusted slightly larger when > 992px.

app/styles/_overview.less
sg00dwin: I'll switch to variables

app/styles/_overview.less
spadgett: So something like flex-basis: 50% and margin-right: 15px doesn't work because the second tile wraps?

app/styles/_overview.less
spadgett: Thanks, I'd do it for both 49% and 49.375% values

app/styles/_overview.less
sg00dwin: switched to variable

sg00dwin:
@spadgett updated pr

spadgett:
This changes the rendering of A/B routes which shouldn't have a gap between them.

openshift_web_console

sg00dwin:
@spadgett I did include a .grouped class that when applied to <div class="overview-services grouped"... removes the gap by switching their flex-basis to 50%. 
https://github.com/openshift/origin-web-console/pull/193/files?diff=unified#diff-d3494a108775ccb728f6772fc669a264R335

There may be other parts of the A/B routing that may still need fixing.

spadgett:
@sg00dwin you should be looking for the `single-alternate-service` class since there are other styles applied

https://github.com/spadgett/origin-web-console/blob/master/app/styles/_overview.less#L443-L475

sg00dwin:
@spadgett updated pr using 50% width and padding to control columns. And the gutters now consistent widths of 6px. 
Still need to address the A/B routes issue.

app/views/overview/_service.html
spadgett: @sg00dwin I saw you merged my pull request that fixes this, but I don't see the fix in the PR :(

sg00dwin:
@spadgett updated pr
cc @jwforres

spadgett:
Travis failed with a dist mismatch

sg00dwin:
@spadgett PTAL

app/styles/_overview.less
```diff
@@ -406,54 +479,59 @@
...
+      }
       .alternate-service {
         .deployment-tile {
-          // Remove the gap between the tiles.
-          margin-left: 5px;
           // Prevent the left box-shadow from appearing over the primary service tile.
           box-shadow: 2px 3px 6px -1px rgba(0, 0, 0, 0.1);
-          .service-title {
-            padding-left: 0;
-          }
```
spadgett: Removing this style breaks the layout of routes with multiple services.

spadgett:
dist/ mismatch. Try clean installing assets, then grunt build.

sg00dwin:
@spadgett adding back style. ran clean, install deps, and grunt build

spadgett:
Thanks @sg00dwin [merge]

spadgett:
@sg00dwin It looks like this introduced a small problem with the left border for some services. Can you take a look?

openshift_web_console

sg00dwin:
@spadgett updated pr

screen shot 2016-07-25 at 3 00 51 pm

spadgett:
@sg00dwin The pull request has already merged, so it will have to be a separate pull :)

PR #195, opened 2016-07-20
Add a dedicated pipelines page

spadgett:
screen_shot_2016-07-20_at_2_01_01_pm

@jwforres

### DIFF
--- a/app/scripts/controllers/builds.js
+++ b/app/scripts/controllers/builds.js
@@ -16,7 +16,7 @@ angular.module('openshiftConsole')
     $scope.labelSuggestions = {};
     $scope.alerts = $scope.alerts || {};
     $scope.emptyMessage = "Loading...";
-    $scope.buildsByBuildConfig = {};
+    $scope.latestByConfig = {};
 
     // get and clear any alerts
     AlertMessageService.getAlerts().forEach(function(alert) {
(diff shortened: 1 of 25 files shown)


app/views/browse/deployment.html
```diff
@@ -57,7 +57,8 @@
                         project-name="{{deployment.metadata.namespace}}"
                         alerts="alerts"
                         disable-delete="!!deployment.status.replicas"
-                        hpa-list="hpaForRC">
+                        hpa-list="hpaForRC"
+                        redirect-url="{{deployment | configURLForResource}}">
```
spadgett: @jwforres FYI, I tested this for vanilla RCs. Since `configURLForResource` returns null for replication controllers, it falls back to the default behavior.

spadgett:
@jwforres OK I think this is ready for a look

spadgett:
rebased

app/scripts/controllers/builds.js
```diff
@@ -79,22 +77,15 @@ angular.module('openshiftConsole')
...
+          // Make sure there is a key for every build config we know about
           angular.forEach($scope.buildConfigs, function(buildConfig, buildConfigName){
-            $scope.buildsByBuildConfig[buildConfigName] = $scope.buildsByBuildConfig[buildConfigName] || {};
+            $scope.latestByConfig[buildConfigName] = $scope.latestByConfig[buildConfigName] || null;
           });
         }
 
         function updateFilterWarning() {
-          if (!LabelFilter.getLabelSelector().isEmpty() && $.isEmptyObject($scope.buildsByBuildConfig)) {
+          if (!LabelFilter.getLabelSelector().isEmpty() && $.isEmptyObject($scope.buildConfigs)) {
```
jwforres: buildsByBuildConfig could have had keys for build configs that had been deleted but whose builds still existed, checking if buildConfigs is empty doesn't result in the same answer.  shouldn't this be if latestByConfig is empty?

app/scripts/controllers/pipelines.js
```diff
@@ -0,0 +1,140 @@
...
+        //     params: {
+        //       fieldSelector: 'spec.strategy.type=JenkinsPipeline'
+        //     }
+        //   }
+        // };
+
+        watches.push(DataService.watch("builds", context, function(buildsData) {
+          $scope.buildsLoaded = true;
+          builds = buildsData.by("metadata.name");
+          update();
```
jwforres: pipelines get a lot of updates pushed, we should probably either be debouncing the call to update, or we should only be updating the info that is relevant to the specific build object that is getting the update.

if we are directly using the duration field from the build status then that might cause weird behavior with the time ticking if we debounced it

app/scripts/controllers/pipelines.js
```diff
@@ -0,0 +1,140 @@
...
+        var updateStats = function(buildConfigName, build) {
+          if (!statsByConfig[buildConfigName]) {
+            statsByConfig[buildConfigName] = {
+              count: 0,
+              // total duration of all builds for calculating averages
+              totalDuration: 0
+            };
+          }
+          statsByConfig[buildConfigName].count++;
+          statsByConfig[buildConfigName].totalDuration += BuildsService.getDuration(build);
```
jwforres: why not just also keep an average on here and re-calculate it here when its expected to change?

app/scripts/controllers/pipelines.js
spadgett: I'm worried about reliably being able to calculate duration if the client clock doesn't match the server's. We only have `startTimeMillis` and `durationMillis` for stages.

app/scripts/controllers/builds.js
spadgett: Will fix.

app/scripts/controllers/pipelines.js
spadgett: I can do that. Thinking of filtering out incomplete builds from avg duration as well.

app/scripts/controllers/pipelines.js
spadgett: Overview and build config pages might be worse if we're worried about this, and debouncing in those controllers doesn't fully address the problem. The watch callback still results in a digest loop and any ng-repeats in the view. I think we'd need to throttle or debounce in DataService to fix it properly since DataService triggers the digest loop on watch updates. I'm not sure we want to make that change in 1.3, though.

I suggest we profile and see if it's an issue, then look at where we'd need to fix it.

jwforres: thats fine, done need to do it here

jwforres:
LGTM but I'm pretty sure this is going to conflict with @rhamilto 's PR #219   around the build config URLs.  Need to figure out who should go first vs rebase

spadgett:
> LGTM but I'm pretty sure this is going to conflict with @rhamilto 's PR #219 around the build config URLs

Rebased. Only conflict was in a file @rhamilto deleted.

[merge]

PR #199, opened 2016-07-20
Improve Project List/Selection Page

### LINKED ISSUE #681: Descenders clipped in project list
openshift_web_console

@MarkDeMaria can you check if this is still a problem with #199?

MarkDeMaria:
[Trello Card: Improve Project List Page](https://trello.com/c/L4c4W83T/571-5-improve-project-list-selection-page-evg)

**Reformat tiles**
--Add description (currently truncated at 170 characters)
--Add relative timestamp (# days since creation)
--Placeholder for # of apps
--Placeholder for application status/error overview
**Reformat top bar**
--Add search bar (Searches by unique name, display name, description, and relative timestamp
--Add sort by unique name, display name, and relative timestamp - Ascending and descending, case-insensitive

Fixes #681

**Design by @ajacobs21e** 
Allie's Design

**Screenshots**
Overview (Default sort by unique name)
image

Sort by Display Name
image

Sort by Relative Timestamp
image

Hover - Needs revising
screenshot from 2016-07-20 17-17-14

Search by Description
image

Search by multiple categories (Searching for '4', which appears in multiple names/descriptions)
image

**Points of Discussion**
1. Currently no support for getting the number of apps within a project
2. Currently no support for any status information about a project. Important to consider system load if retrieving health statuses for a large number of projects.
3. Current design does not label display name and unique (machine) name, so telling the two apart is difficult when using the sort dropdown, as one is labeled 'Name' and the other 'Display Name.' The New Project uses this naming convention when soliciting user input.

@benjaminapetersen @spadgett @jwforres


benjaminapetersen:
See PR #200

benjaminapetersen:
@ajacobs21e I imagine the hover effect might need to be tweaked. I don't know that it looks quite right now that there is no space between the tiles.

screen shot 2016-07-21 at 3 39 43 pm

benjaminapetersen:
I believe [data-placement="top"](http://getbootstrap.com/javascript/#four-directions) can fix this.  Might be good to be a single line since its two small words.  

screen shot 2016-07-21 at 3 40 41 pm

I'm actually wondering if we need the tooltip at all.

app/scripts/controllers/projects.js
```diff
@@ -8,6 +8,28 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
+  .filter('findProject', function() {
```
benjaminapetersen: This could probably live under `scripts/filters/findProject.js`.

app/scripts/controllers/projects.js
```diff
@@ -8,6 +8,28 @@
...
+    return function(items, searchString) {
+      var filtered = [];
+
+      if (searchString) {
+        var searchStringLower = searchString.toLowerCase();
+
+        _.forEach(items, function(item) {
+          if (item.metadata.name.toLowerCase().includes(searchStringLower) ||
```
benjaminapetersen: In [filters/resources.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/filters/resources.js#L17) we have a filter for getting at annotations that could be useful here.

```javascript
$filter('annotation')(item, 'displayName');
$filter('annotation')(item, 'description');
```

spadgett:
https://github.com/patternfly/angular-patternfly/issues/212

app/scripts/controllers/projects.js
benjaminapetersen: Maybe:
```javascript
// return early if no search string
if(!searchString) {
  return items;
}
return _.filter(items, function(item) {
  // filter logic
  return ...
});
```

benjaminapetersen:
Yeah, per that issue I'd much rather there be a label to the left of the dropdown & skip the tooltip.

app/scripts/controllers/projects.js
```diff
@@ -16,14 +38,52 @@ angular.module('openshiftConsole')
...
+    var sortProjects = function() {
+      // TODO: currentField is renamed in angular-patternfly 3.0
+      var sortID = _.get($scope, 'sortConfig.currentField.id', 'metadata.name');
+      var adjustedSortOrder = $scope.sortConfig.isAscending;
+
+      // Sort order must be reversed if using timestamp because the info shown to user is the relative timestamp, not the absolute
+      if (sortID === 'metadata.creationTimestamp') {
+        adjustedSortOrder = !adjustedSortOrder;
+      }
+
```
benjaminapetersen: May just be my opinion, but I think you could inline `var sortOrder = adjustedSortOrder ? 'asc' : 'desc';` into the `_.sortByOrder` fn's third arg and eliminate the additional `sortOrder` variable.  It may be easier to follow.  `}, [adjustedSortOrder ? 'asc' : 'desc']);`

benjaminapetersen:
As the page collapses text gets a bit squashed.  Little bit of bootstrap OR a tweak to the flex should fix this (btw, I wouldn't use both at the same time).

screen shot 2016-07-21 at 5 03 50 pm

screen shot 2016-07-21 at 5 03 58 pm

MarkDeMaria:
@benjaminapetersen I talked to @ajacobs21e about including a label next to the sort dropdown a few weeks ago. Allie, did you hear anything back regarding your user test inquiries?

ajacobs21e:
When the toolbar pattern was designed for PF the goal was to keep it simple/ clean. Other design considerations were:
- in smaller viewports, there is no room for those labels, and did we really need to add them at larger sizes?
- sorting is pretty standard in many UIs and frequently doesn't have the "sort" label

When I brought this up to them they agreed to run a usability test on this change. Since this is for 3.4 I didn't plan to run that yet. Is that something I should prioritize, maybe for the first week of August? @jwforres

ajacobs21e:
@MarkDeMaria can you do a hover background color of #ededed instead of the background shadow?

spadgett:
> in smaller viewports, there is no room for those labels, and did we really need to add them at larger sizes?

@ajacobs21e In smaller viewports, the tooltip used in place of a label causes problems. For instance, in iOS, you have to tap the button twice. See https://github.com/patternfly/angular-patternfly/issues/212

I'm fine with or without a label. I just don't like the tooltip because the experience on mobile is poor.

benjaminapetersen:
Agree (though I think I do prefer the label), just dropping the tooltip is prob sufficient!

MarkDeMaria:
Changes:
1. Hover background color #ededed & remove hover shadow
2. Move filter to findProjects.js & improve filter code
3. Various code style improvements

Not-Changed-Yet:
1. Column sizing on extra small screen widths. Meeting with @sg00dwin later.
2. Sort-by label vs. tooltip. Awaiting verdict from @ajacobs21e.

spadgett:
The tooltip fix will need to be an upstream change in angular-patternfly.

benjaminapetersen:
Perhaps use a blue hover similar to what we have in tables?  Not sure about the gray, but not sure the blue will be awesome in this context either.  

screen shot 2016-07-25 at 1 27 24 pm

benjaminapetersen:
I think the background gray can get dropped on this as well, looks like we have white/white in other contexts:

screen shot 2016-07-25 at 1 29 09 pm

ajacobs21e:
@MarkDeMaria yes to removing the tooltip

@benjaminapetersen is the list view default hover color grey or blue?

MarkDeMaria:
@benjaminapetersen @jwforres @spadgett @ajacobs21e 

Design is now mobile friendly. The 'New Project' button turns into a '+' when mobile-sized. I played around with reformatting the filter bar, but many options looked clunky or altered the locations of buttons such that their new locations were jarring or ran counter to the rest of the UI. Thoughts?

image

spadgett:
Thanks @MarkDeMaria 

Looks like you lost the project edit button when you rebased.

I think I'd just use something like this for the creation date at mobile (normal text, left-aligned). 

```html
<span class="text-muted">created [whenever]</span>
```

The big text looks good for wider screens, though.

I'd expect the actions to be in the upper-right of the tile on mobile. We could switch to a kebab menu like we do on the browse pages since we have two actions.

Suggest a darker gray line separating the projects. It's a little too faint.

How does the New Project button look with the full label to the right of the Projects header? Then the filter and sorting controls can take 100% width.

Need some additional margins on the message at the bottom.

benjaminapetersen:
@ajacobs21e for your thoughts on the mobile view!

MarkDeMaria:
-No more + on mobile. Now 'New Project' in line with the 'Projects' header at the top.
-The kebab does not work with our current tile click code. The bottom right is what I have found to be the best location for the edit/trash buttons, as the top line with the display name does not offer enough space for both.

Thoughts?

Updated Mobile Layout
image

(The PR has not been updated to include this code. It's still local because I'm running into a small flex problem with the input box that I'm still working on.)

@benjaminapetersen @ajacobs21e @spadgett

benjaminapetersen:
Button placement decision makes sense to me.  Spacing feels right.

MarkDeMaria:
@spadgett @jwforres

**Full Size**
screenshot from 2016-07-29 12-05-18

**Half Size**
image

**Mobile - IPhone 6**
image

spadgett:
Screenshots look great. Only comment is the highlight color is too close to the background color. (Assuming you're hovering over test in the first screen.)

spadgett:
Might give the creation time more width so it doesn't wrap.

app/scripts/filters/findProject.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .filter('findProject', function($filter) {
```
spadgett: Instead of taking in `$filter`, use `annotationFilter` here. Then you can call it directly below.

benjaminapetersen:
Was just chatting with @jwforres and @spadgett. Lets make sure to test the filtering thoroughly.  Would be good to:

- [ ] ensure the "watch" is working by creating new projects in separate browser tab. original tab should still update
- [ ] ensure filtering still picks up new projects via the watch
- [ ] test some strange conditions such as special characters, ensuring all projects come back into view when all text is eliminated, etc.

benjaminapetersen:
- [ ] "search projects" in the input is missing some left padding. This is prob an easy fix as it should (probably) be automatic... maybe missing or added an extra class?

benjaminapetersen:
- [ ] how does it look with no description text in any blocks?

MarkDeMaria:
Note to self: Marked for deletion out of place
image

benjaminapetersen:
Thinking, we want to make sure that "hover" implies special focus, rather than disabled (gray).  There is a `@table-bg-hover` or `@dropdown-link-active` bootstrap variable that we could swipe (though this is not a table or a dropdown, it seems a consistent hover concept):

screen shot 2016-07-29 at 2 43 26 pm

app/scripts/controllers/projects.js
```diff
@@ -16,14 +16,50 @@ angular.module('openshiftConsole')
     $scope.showGetStarted = false;
     $scope.canCreate = undefined;
 
+    var sortProjects = function() {
```
benjaminapetersen: We are on angular-patternfly 3.5.1, i think this todo has aged & can be removed.

app/scripts/controllers/projects.js
```diff
@@ -16,14 +16,50 @@ angular.module('openshiftConsole')
...
+      var sortID = _.get($scope, 'sortConfig.currentField.id', 'metadata.name');
+      var adjustedSortOrder = $scope.sortConfig.isAscending;
+
+      // Sort order must be reversed if using timestamp because the info shown to user is the relative timestamp, not the absolute
+      if (sortID === 'metadata.creationTimestamp') {
+        adjustedSortOrder = !adjustedSortOrder;
+      }
+
+      $scope.projects = _.sortByOrder($scope.projects, function(project) {
+        return _.get(project, sortID).toLowerCase();
```
benjaminapetersen: to clean this up lets remove the comments

app/scripts/filters/findProject.js
```diff
@@ -0,0 +1,20 @@
...
+    return function(items, searchString) {
+      // return early if no search string
```
benjaminapetersen: i think you can drop the comment here as well, seems straight forward

app/scripts/filters/findProject.js
```diff
@@ -0,0 +1,20 @@
...
+      if(!searchString) {
+        return items;
+      }
+
+      return _.filter(items, function(item) {
+        var searchStringLower = searchString.toLowerCase();
+
```
benjaminapetersen: Unfortunately IE doesn't support [String.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) very well. [_.includes()](https://lodash.com/docs#includes) to the rescue!  There is also a [_.lowerCase()](https://lodash.com/docs#lowerCase) that we might as well use then for a consistent style.

app/scripts/filters/findProject.js
```diff
@@ -0,0 +1,20 @@
...
+angular.module('openshiftConsole')
+  .filter('findProject', function($filter) {
+    return function(items, searchString) {
+      // return early if no search string
+      if(!searchString) {
+        return items;
+      }
+
+      return _.filter(items, function(item) {
+        var searchStringLower = searchString.toLowerCase();
```
benjaminapetersen: Lets move the evaluation of the search string up above the filter as it only needs to be evaluated once.  
```javascript
var searchStringLower = _.lowerCase(searchString);
return _.filter(items, function(item) {
 // else it does the same work over and over...
```

app/styles/_mixins.less
```diff
@@ -94,7 +94,7 @@
   margin-right: 10px;
 }
 
-// additional visiblility utility for xlg 
+// additional visiblility utility for xlg
```
benjaminapetersen: I think the use of `!important` here makes sense, given what its doing.  @sg00dwin @rhamilto feel free to affirm/refute :smile:

app/styles/_projects.less
```diff
@@ -0,0 +1,66 @@
...
+  .projectSearch {
+    flex: 1;
+
+    @media (min-width: @screen-sm-min) {
+      flex: 0;
+      min-width: 200px;
+    }
+  }
+
+  .vertical-divider {
```
benjaminapetersen: You can probably drop this first rule that is only doing the `display:none` and add a [boostrap util class](http://getbootstrap.com/css/#responsive-utilities-classes) `.hidden-sm` to the element.

app/styles/_projects.less
```diff
@@ -0,0 +1,66 @@
...
+    vertical-align: middle;
+    width: 1px;
+  }
+}
+
+.project-names {
+  min-width: 35%;
+}
+
+.tile-project {
```
benjaminapetersen: want to stick with the convention of `project-name`, `project-tile`, `project-*`? rather than reversing to `tile-project` here.

app/styles/_mixins.less
MarkDeMaria: Do you want me to drop this file from the PR? Accidental change.

app/styles/_projects.less
```diff
@@ -0,0 +1,66 @@
...
+  .vertical-divider-lg {
```
benjaminapetersen: Same as above, can prob use the [boostrap util](http://getbootstrap.com/css/#responsive-utilities-classes)

app/scripts/filters/findProject.js
MarkDeMaria: @benjaminapetersen _.lowerCase() is only available in lodash v4.14.1. I cannot find an equivalent lower case function in v3.10.1.

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
     <div class="middle-section surface-shaded">
       <div class="middle-container has-scroll">
         <div class="middle-content">
+          <div ng-if="!showGetStarted" class="container-fluid surface-shaded pad-left-none pad-right-none mar-left-none mar-right-none center-block"> <!-- Safari needs surface-shaded at mobile -->
```
benjaminapetersen: I think with these little util clases, I'd go the route of making a specific class, then using the utils to modify it in particular instances.  so for example:

```less
// a generic class that works in *most* instances
.object-tile {
  margin: 10px 20px;
  // other rules
}
```
but sometimes needs a tweak:
```html
<!-- in this case, i gotta nix the right margin  -->
<div class="object-tile mar-right-none"></div>
```

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
+            <div ng-if="(projects | hashSize) === 0" class="text-muted">Loading...</div>
+            <div ng-if="(projects | hashSize) !== 0" class="gutter-top pad-top-none">
+              <div style="background-color: white;" class="mar-bottom-xl">
```
benjaminapetersen: should prb move the inline style to a css class

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
+                      <span>New project</span>
+                    </a>
+                  </div>
+                  <alerts alerts="alerts"></alerts>
+                </div>
+              </div>
+
+              <div ng-repeat="project in projects | findProject:filterStr">
+                <div row cross-axis="center"
+                     class="tile tile-project tile-click tile-flex tile-project-border mar-left-xl mar-right-xl mar-bottom-none">
```
benjaminapetersen: Prob same as above, I like the generic classes, but prob after ~3 classes need to consider making a specific class to roll some of that up (and hopefully then make it reusable).

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
+                  <!-- TODO (MarkDeMaria): Insert Star funcionality here
```
benjaminapetersen: good note that this is ready for whenever we get the data we need to do it.

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
+                    <div>
+                      <p>{{project.metadata.name}}</p>
+                    </div>
+                    <div class="visible-xs">
+                      <div class="muted" style="margin-top: -5px;" ng-if="project | description">
+                        <truncate-long-text content="project | description" limit="100" use-word-boundary="true"></truncate-long-text>
+                      </div>
+
+                      <div row>
+                        <!-- TODO (MarkDeMaria): insert app retrieval here
```
benjaminapetersen: same, good to have this ready to roll when we have data to support.

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
+                        -->
+
+                        <div class="project-created text-left mar-top-md">
+                          <span class="text-muted h2 mar-bottom-none mar-top-none pad-top-none pad-bottom-none">
+                            Created
+                            <relative-timestamp timestamp="project.metadata.creationTimestamp"></relative-timestamp>
+                          </span>
+                        </div>
+
+                        <span flex></span>
```
benjaminapetersen: nit: can prob make `<span flex></span>` into `<div flex></div>` here so its the same node as its siblings.

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
+                        <div column main-axis="center">
+                          <div class="pad-none">
+                            <span class="vertical-divider-lg"></span>
+                          </div>
+                        </div>
+                      </div>
+                    </div>
+                  </div>
+
+                  <div class="col-xs-1 pad-none hidden-xs hidden-sm">
```
benjaminapetersen: `<div class="project-status">` maybe

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
+                    <!-- TODO (MarkDeMaria): insert app retrieval here
+                    <div class="text-center">
+                      <p class="h1 mar-bottom-none mar-top-none pad-top-none pad-bottom-none">#</p>
+                      <p class="mar-top-none pad-top-sm">Apps</p>
+                    </div>
+                    -->
+                  </div>
+
+                  <div class="col-xs-1 pad-left-none pad-right-none hidden-xs hidden-sm">
```
benjaminapetersen: `<div class="project-status">`

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
-                      resource-name="{{project.metadata.name}}"
-                      project-name="{{project.metadata.name}}"
-                      display-name="{{(project | displayName)}}"
-                      type-name-to-confirm="true"
-                      stay-on-current-page="true"
-                      alerts="alerts"
-                      button-only>
-                    </delete-link>
+
+                  <div class="col-xs-3 col-sm-2 pad-none hidden-xs">
```
benjaminapetersen: `<div class="project-info project-age`>?

app/scripts/controllers/projects.js
```diff
@@ -16,14 +16,50 @@ angular.module('openshiftConsole')
...
+    };
+
+    // Set up the sort configuration for `pf-sort`.
+    $scope.sortConfig = {
+      fields: [{
+        id: 'metadata.name',
+        title: 'Name',
+        sortType: 'alpha'
+      }, {
+        id: 'metadata.annotations["openshift.io/display-name"]',
```
benjaminapetersen: $filter('annotationName')('displayName');   or if you asked for `annotationName` directly instead of `$filter`, you can just `annotationName('displayName')` to make sure this stays the correct key in future (api changes, etc)

app/scripts/controllers/projects.js
```diff
@@ -16,14 +16,50 @@ angular.module('openshiftConsole')
...
+
     AlertMessageService.getAlerts().forEach(function(alert) {
       $scope.alerts[alert.name] = alert.data;
     });
     AlertMessageService.clearAlerts();
 
     AuthService.withUser().then(function() {
       watches.push(DataService.watch("projects", $scope, function(projects) {
-        $scope.projects = projects.by("metadata.name");
+        $scope.projects = _.sortByOrder(projects._data, ['metadata.name'], ['asc']);
```
spadgett: You should access the `_data` property directly. It's internal. Call `projects.by("metadata.name")` before calling sort.

app/scripts/controllers/projects.js
spadgett: I believe `id` here is just a unique id for the sort options.

app/scripts/filters/findProject.js
spadgett: I think it's just a lodash wrapper for `String.toLowerCase()`. We can stick with the native function.

app/scripts/controllers/projects.js
benjaminapetersen: well then, i propose 'waffles' or 'samwise gamgee' or something more interesting.

app/styles/_mixins.less
benjaminapetersen: Ah, yeah if you didn't actually change this, go ahead and remove the file

app/views/projects.html
```diff
@@ -8,52 +8,201 @@
...
+              <div ng-repeat="project in projects | findProject:filterStr">
+                <div row cross-axis="center"
+                     class="tile tile-project tile-click tile-flex tile-project-border mar-left-xl mar-right-xl mar-bottom-none">
+                  <!-- TODO (MarkDeMaria): Insert Star funcionality here
+                  <div>
+                    <i class="fa fa-star-o" aria-hidden="true"></i>
+                  </div>
+                  -->
+
+                  <div class="project-names col-xs-12 col-sm-5 pad-none">
```
benjaminapetersen: Quick comment here to doc our discussion, we should stick with either flex or bootstrap for the layout, but not mix.  So the above `<div row>` can be changed to `<div class="row">` if going with bootstrap children `<div class="col-md-5">`, or these child divs can be changed to something like `<div flex grow="2">` if you wanna keep the above `<div row>` to go all in with flex instead.

jorgemoralespou:
I think that although this screen needed a revamp, the most critical feature for cluster admins is to be able to filter by labels and this PR does leave out both showing  project labels and filtering by labels which is critical and not sure how easy will be to address in the future.

MarkDeMaria:
@jorgemoralespou I believe the web console is not currently given a project's labels when it is given a project. The data we currently have access to is:
image

Is there anything in there that you would find helpful or synonymous with 'label'?

jorgemoralespou:
@MarkDeMaria
Although you can not directly add a label to a project you can add it to a
namespace and will be inherited, and this is what some customers are doing
to organize their projects. There should be the option to add labels to
projects, and I'll open an issue in origin, if there's none already.

As an example, edit the namespace associated to a project and add the
label, you'll see the label in the project itself.

````
apiVersion: v1
kind: Project
metadata:
  annotations:
    openshift.io/description: ""
    openshift.io/display-name: ""
    openshift.io/requester: dev
    openshift.io/sa.scc.mcs: s0:c7,c4
    openshift.io/sa.scc.supplemental-groups: 1000050000/10000
    openshift.io/sa.scc.uid-range: 1000050000/10000
  creationTimestamp: 2016-08-09T18:16:18Z
  labels:
    name: value
  name: example
  resourceVersion: "1617"
  selfLink: /oapi/v1/projects/example
  uid: 5e2764ee-5e5d-11e6-8b26-525400b263eb
spec:
  finalizers:
  - openshift.io/origin
  - kubernetes
status:
  phase: Active
````



2016-08-09 21:46 GMT+02:00 Mark DeMaria <notifications@github.com>:

**Comment by MarkDeMaria** [comment]
@jwforres @benjaminapetersen @spadgett @ajacobs21e 

Full-screen
image

Half-Screen. Note issue with long, non-whitespace names. @rhamilto or @sg00dwin, help would be appreciated when you return!
image

Mobile
image

**Comment by spadgett** [comment]
@MarkDeMaria try `class="truncate"` on the long names

**Comment by spadgett** [comment]
Or `class="word-break"` if you want them to be able to wrap

**Comment by MarkDeMaria** [comment]
@spadgett Unfortunately neither of those do the trick with a string without whitespace.

**Comment by benjaminapetersen** [comment]
@MarkDeMaria I think our `class="word-break"` is supposed to use this rule: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break 

Which should allow it to break the word wherever needed.  We talked about truncating long text in JS before, but I don't think we want to do that because we want to maintain the integrity of the string (perhaps for screen readers, etc).

**Comment by ajacobs21e** [comment]
@MarkDeMaria can you see how it looks at mobile with the date left aligned please?

**Comment by MarkDeMaria** [comment]
@sg00dwin @benjaminapetersen @jwforres 

Word-Break will work at fullscreen and on mobile, but in-between (About medium, as shown) word-break will fight with the pf-list layout. The inner outline on the left is the <h1> tag that word-break is assigned to.
Medium-ish
image
Mobile (Works)
image


Truncate will not work at medium to mobile sizes.
Medium-ish
image
Mobile
image

**Comment by MarkDeMaria** [comment]
@ajacobs21e 

Timestamp div left-justified but text within is center:
image

All left-justified
image

**Comment by benjaminapetersen** [comment]
Def think "Created" label has to be left justified as well (2nd option)

**Comment by benjaminapetersen** [comment]
@MarkDeMaria so `word-break` mostly works, just not @ medium. Curious what css rules are clashing there?

**Comment by rhamilto** [comment]
@jwforres, in attempting to help @MarkDeMaria debug the overflow issue, I started poking at the markup and css for the projects page.   I'd like us to discuss it next week.  I think the use of http://angular-patternfly.rhcloud.com/#/api/patternfly.views.directive:pfListView is probably overkill, and we could much more easily and sanely achieve the desired layout with a lot less markup, css, and headache.

**Comment by benjaminapetersen** [comment]
@rhamilto yeah... not a fan of pf-list much either.  Unfortunately he has redone this one a couple times. :/

**Comment by rhamilto** [comment]
@MarkDeMaria, I'm continuing to poke at this in hopes of finding a simple fix.  I'll keep you posted.

**Comment by MarkDeMaria** [comment]
@ajacobs21e @jwforres @spadgett @benjaminapetersen @rhamilto @sg00dwin 

Tweaks/changes since Rob's responsive bug fixes/cleanup:
1. Add clear button
2. If no display name given, show machine name in display name style/location
3. Empty state table message if filter filters all results
4. Fix bug in toLowerCase crashing if no display name given
5. Steve's mobile suggestions for hiding machine name and changing relative creation timestamp to smaller, muted text below description

**Important discussion points for the future of this page**
@rhamilto Can you lead the discussion on what you believe would be the most valuable changes to this page's design and layout to make it more effective?


Screenshots

Full-screen. Note 'no-display-name' near bottom is a project without a display name given
image

Empty state
image

Mobile Empty State
image


Mobile (Steve's changes)
image

**Comment by jwforres** [comment]
when there is no display name, having both the large text and the small
text the same is redundant, I would remove the small text in that case

**Comment by rhamilto** [comment]
[PatternFly's search widget](https://www.patternfly.org/pattern-library/widgets/#search) uses an "X" icon that gets added to the field once it contains text rather than a separate, omnipresent button.  Any reason you didn't go the PatternFly route?

> @rhamilto Can you lead the discussion on what you believe would be the most valuable changes to this page's design and layout to make it more effective?

Sure.

I think the only remaining issue that's been highlighted is the size of the creation date.  @ajacobs21e, the consensus seems to be that the creation date isn't as important as the very large font size implies.  Any objections to scaling that down to "normal" size?  I ask you as this detail came from your [original design](https://github.com/openshift/origin-web-console/pull/199#issue-166684883).

**Comment by MarkDeMaria** [comment]
Implemented PF's 'X' icon.

**Comment by benjaminapetersen** [comment]
Planning feedback:
- [ ] drop the "created by" date entirely, doesn't seem useful, more space for descriptions.

**Comment by MarkDeMaria** [comment]
Removed timestamp
Fullscreen:
image

Half-ish
image

Mobile
image

**Comment by jwforres** [comment]
just noticed this but for the New Project button at the large resolutions I
would expect it to be top aligned in that corner like it is at mobile.

**Comment by MarkDeMaria** [comment]
@jwforres Fixed.

**Comment by rhamilto** [comment]
Now that we've removed creation date, should we also remove "Age" from the sort dropdown?  It seems a bit odd to be able to sort on an invisible value?

**Comment by MarkDeMaria** [comment]
@jwforres Ready for review.

**Comment by spadgett** [comment]
@MarkDeMaria Noticed some problems with the display name sort when I pulled down your change.

**Comment by MarkDeMaria** [comment]
@spadgett Can you describe what you're seeing?

File: app/scripts/controllers/projects.js [hunk 26]
```diff
@@ -16,13 +16,37 @@ angular.module('openshiftConsole')
     $scope.showGetStarted = false;
     $scope.canCreate = undefined;
 
+    var sortProjects = function() {
+      var sortID = _.get($scope, 'sortConfig.currentField.id', 'metadata.name');
+
+      $scope.projects = _.sortByOrder($scope.projects, function(project) {
+        return _.get(project, sortID).toLowerCase() || _.get(project, 'metadata.name').toLowerCase();
```
spadgett: Shouldn't this be...?

```js
 _.get(project, sortID, '').toLowerCase()
```

Otherwise it will blow up if the project doesn't have a value for `sortID`

spadgett: It looks like `_.sortByOrder` should take an array of functions, not a single function.

app/scripts/controllers/projects.js
```diff
@@ -16,13 +16,37 @@ angular.module('openshiftConsole')
...
+    };
+
     AlertMessageService.getAlerts().forEach(function(alert) {
       $scope.alerts[alert.name] = alert.data;
     });
     AlertMessageService.clearAlerts();
 
     AuthService.withUser().then(function() {
       watches.push(DataService.watch("projects", $scope, function(projects) {
+        $scope.emptyMessage = "No projects to show";
```
spadgett: If this only happens due to a filter, I'd prefer "The current filter is hiding all projects." Ideally with a "Clear filter" link.

spadgett: If the value of `emptyMessage` never changes, just add it to the view markup. No need for a scope var.

app/scripts/filters/findProject.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .filter('findProject', function(annotationFilter) {
```
spadgett: Can you replace this implementation with one that uses `KeywordService.generateKeywords` and `KeywordService.filterForKeywords`?

app/views/projects.html
```diff
@@ -8,83 +8,137 @@
...
-                    <h2 class="project truncate">
-                      <a class="tile-target" href="project/{{project.metadata.name}}">{{(project | uniqueDisplayName : projects)}}</a>
-                      <span ng-if="project.status.phase != 'Active'" data-toggle="tooltip" title="This project has been marked for deletion." class="pficon pficon-warning-triangle-o" style="cursor: help; vertical-align: top; margin-left: 5px;"></span>
-                    </h2>
-                    <div class="muted" style="margin-top: -5px;" ng-if="project | description">
-                      <truncate-long-text content="project | description" limit="512" use-word-boundary="true"></truncate-long-text>
-                    </div>
+                <div ng-if="!showGetStarted">
+                  <div ng-if="(projects | hashSize) === 0" class="empty-state-message">
+                    <h2 class="text-center" id="temporary-loading-message">Loading...</h2>
```
spadgett: Do we need this `id`? I don't see it referenced anywhere.

app/views/projects.html
```diff
@@ -8,83 +8,137 @@
...
-                      kind="Project"
-                      resource-name="{{project.metadata.name}}"
-                      project-name="{{project.metadata.name}}"
-                      display-name="{{(project | displayName)}}"
-                      type-name-to-confirm="true"
-                      stay-on-current-page="true"
-                      alerts="alerts"
-                      button-only>
-                    </delete-link>
+                  <div ng-if="(projects | hashSize) !== 0">
```
spadgett: Or just

```html
<div ng-if="projects | hashSize">
```

app/views/projects.html
```diff
@@ -8,83 +8,137 @@
...
+                    <div class="projects-header">
+                      <div class="projects-bar">
+                        <h1>Projects</h1>
+                        <div class="projects-options">
+                          <div class="projects-add" ng-if="canCreate">
+                            <a href="create-project" class="btn btn-md btn-primary">
+                              <span>New project</span>
```
spadgett: I don't think you need to wrap this in a span?

app/views/projects.html
```diff
@@ -8,83 +8,137 @@
...
+                            <form role="form" class="search-pf has-button">
+                              <div class="form-group has-clear">
+                                <div class="search-pf-input-group">
+                                  <label for="search-projects" class="sr-only">Search</label>
+                                  <input
+                                    type="search"
+                                    class="form-control"
+                                    placeholder="Search"
+                                    id="search-projects"
+                                    ng-init="search.text = ''"
```
spadgett: I'd rather initialize this in the controller.

app/views/projects.html
```diff
@@ -8,83 +8,137 @@
...
+                              alerts="alerts"
+                              button-only>
+                            </delete-link>
+                          </span>
+                        </div>
+                        <div class="list-view-pf-main-info">
+                          <div class="list-view-pf-description project-names">
+                            <div class="list-group-item-heading project-name-item">
+                              <h2 class="h1">
+                                <a class="tile-target" href="project/{{project.metadata.name}}">{{project.metadata.annotations["openshift.io/display-name"] || project.metadata.name}}</a> <!-- {{(project | uniqueDisplayName : projects)}} -->
```
spadgett:
```html
<a class="tile-target" ng-href="project/{{project.metadata.name}}">{{project | displayName}}</a>
```

app/views/projects.html
```diff
@@ -8,83 +8,137 @@
...
+                                <span ng-if="project.status.phase != 'Active'"
+                                  data-toggle="tooltip"
+                                  title="This project has been marked for deletion."
+                                  class="pficon pficon-warning-triangle-o"></span>
```
spadgett: With the list view, is there a chance the icon can wrap on a line by itself below the project name?

I wonder if we still need this since we're watching projects, and it will be removed from the list pretty quickly. @jwforres opinion?

app/views/projects.html
```diff
@@ -8,83 +8,137 @@
...
+                              </h2>
+                              <p class="hidden-xs project-name-item" ng-if="project.metadata.annotations['openshift.io/display-name']">{{project.metadata.name}}</p>
```
spadgett: `ng-if="project | displayName"`

dist.java/java/404.html
```diff
@@ -38,7 +38,7 @@
       <div ng-include src="viewPartial"></div>
     </div>
 
-    <script src="libs-b80880b94c.js"></script>
+    <script src="libs-3d3f8220c5.js"></script>
```
spadgett: This should not be in your changes. Try running `make clean build`

app/views/projects.html
```diff
@@ -8,83 +8,137 @@
...
+                            </div>
+                          </div>
+                        </div>
+                      </div>
+                    </div>
+                    <div ng-if="(projects | findProject:search.text).length === 0" class="text-center">
+                      <h2>{{emptyMessage}}</h2>
+                    </div>
+                    <div class="list-group list-view-pf projects-list">
+                      <div ng-repeat="project in projects | findProject:search.text" class="list-group-item project-info tile-click">
```
spadgett: We call the `findProject` filter 3 times in the view. I'd rather call it once in the view when the keyword values change. We probably want to use `_.debounce` as well so that we don't filter for every keystroke. See

https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/monitoring.js#L328-L331

for an example. If we do this, we can remove the `findProject` filter and just call `KeywordService` directly.

app/views/projects.html
spadgett: Sorry, this should be

```html
ng-if="project | displayName : true"
```

benjaminapetersen:
Catching up on this one, so far so good on these:
- Layout holds up well down to 290px, plenty for any mobile device
- Looks good in IE as well at various sizes
- Clipping issue #681 will be resolved
- Did the sortBy tooltip fix in `sort-pf` ever merge? Hmm...[this did](https://github.com/patternfly/angular-patternfly/issues/212) but [this issue](https://github.com/openshift/origin-web-console/issues/438) says we gotta pick up a new version of angular-patternfly to fix.  Not sure if thats a holdup.

spadgett:
These changes merged in PR #694

PR #205, opened 2016-07-21
Key value editor updates

benjaminapetersen:
This *should* be the last round of key-value-editor updates....

- osc-key-values replaced in create from image flow
- osc-key-values replaced in new from template flow
- osc-key-values replaced in deploy from image
- osc-key-values replaced in edit autoscaler
- osc-key-values replaced in label-editor

I'll comment inline on a couple things.  I did not yet remove `osc-key-values` in this PR.

@jwforres @spadgett @rhamilto


app/scripts/directives/labels.js
```diff
@@ -50,6 +66,32 @@ angular.module('openshiftConsole')
         if (!angular.isDefined(attrs.canToggle)) {
           scope.canToggle = true;
         }
-      }
+      },
+      controller: [
+        '$scope',
+        function($scope) {
+
```
benjaminapetersen: This is a port of the label validation logic from osc-key-values.  The "shared functions" are up above.  This object is a "pseudo regex" with a `.test()` method because its easier to pass to `key-value-editor`, which is using `ng-pattern` and expects a regex or "regex like" object. I went this route instead of passing a function due to wanting the error messages to still key off of `form.input.$error.pattern` w/o having to handle that manually.

jwforres:
[merge]

PR #206, opened 2016-07-21
Fix issue #204, log link throws js error

### LINKED ISSUE #204: Log throws a js error generating the view archive link
When the loggingURL is configured for the web console, going to the Log tab for a pod/build/deployment will throw a JS error when URI.js tries to expand the template.  I suspect this is something to do with our update to a new URI.js, they may have expanded their supported templating language (special characters) and we may have something in that kibana URL that is  incompatible.  Given all of the special characters in the kibana URL we probably need to switch to doing a string replace ourselves using native string replace or lodash stuff instead of using a URI.js template so that we dont hit this again later.

benjaminapetersen:
Fixes #204 

Swapping out `URITemplate()` for `_.template()`, very similar API, I suspect lodash doesn't fuss because it just sees a string and isn't trying to conform to a spec about URIs.

generated:
```
/#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes_container_name,ruby-sample-build-12-build),index:'muh-rubies.*',query:(query_string:(analyze_wildcard:!t,query:'kubernetes_pod_name: ruby-sample-build-12 %26%26 kubernetes_namespace_name: muh-rubies')),sort:!(time,desc))#console_container_name=ruby-sample-build-12-build&console_back_url=https%3A%2F%2Flocalhost%3A9000%2Fproject%2Fmuh-rubies%2Fbrowse%2Fbuilds%2Fruby-sample-build%2Fruby-sample-build-12
```

### DIFF
--- a/app/scripts/services/logLinks.js
+++ b/app/scripts/services/logLinks.js
@@ -60,7 +60,7 @@ angular.module('openshiftConsole')
       };
 
       // broken up for readability:
-      var template = new URITemplate([
+      var template = _.template([
         "/#/discover?",
         "_g=(",
           "time:(",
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
@jwforres @spadgett

jwforres:
works for me, i feel like we have a little more visibility into what lodash is doing, the documentation is pretty clear.

[merge]

Issue #170, opened 2016-07-14
Add to project button white for viewer at some resolutions


spadgett:
When I don't have create / update authority for a project, the add to project button can show up as white (only at some screen widths).

![openshift_web_console_and_issues_ _openshift_origin-web-console](https://cloud.githubusercontent.com/assets/1167259/16849316/e365cb7e-49c7-11e6-869e-a58b1f82e636.png)

/cc @jhadvig @jwforres

spadgett:
@sg00dwin you can add a read-only user to your project with the command

```
oc policy add-role-to-user view test -n myproject
```

Then log in as user test to see.

jhadvig:
@spadgett so one solution would to put the `ng-if` condition from the div on the nested anchor tag. Then we wont hit the issue.
On the other hand there will be visible `<div>` space in case of small screens:
![screenshot-28](https://cloud.githubusercontent.com/assets/1668218/16910607/deedc12c-4cd9-11e6-9b64-57f12c1e68d2.png)

spadgett:
I think we can fix it in CSS to avoid the problem in the screenshot.

PR #196, opened 2016-07-20
Reducing width of project picker at medium screen sizes

rhamilto:
in order to accommodate the wider sidebar

@sg00dwin, sanity check, please?   note:  this does not address the existing issue where the presence of an extension point navbar-utilty causes navbar-utility to wrap at viewport sizes 768 - ~ 810.  I filed https://github.com/openshift/origin-web-console/issues/197 about that.

PR #207, opened 2016-07-21
Make nav-project-menu background transparent to fix …

sg00dwin:
fixes https://github.com/openshift/origin-web-console/issues/170

@spadgett

### DIFF
--- a/app/styles/_project-menu.less
+++ b/app/styles/_project-menu.less
@@ -10,6 +10,7 @@
   .form-group {
     margin-bottom: 0;
     .form-control {
+      background-color: transparent;
       margin-bottom: 0;
       height: auto;
     }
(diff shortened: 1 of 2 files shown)


spadgett:
@sg00dwin Does the dropdown menu size correctly when open? It didn't for me when I tried that change.

sg00dwin:
@spadgett it is resizing correctly for me. the change is just overriding the .form-control background color to transparent, so I don't see how that could be effecting the resize. 

possibly an existing bug non-related?

spadgett:
This is what I mean. The menu when open is wider than it should be.

screen shot 2016-07-22 at 10 30 44 am

sg00dwin:
@spadgett ok see it now, when the 'add to project' button is not shown. So it's related to Robs https://github.com/openshift/origin-web-console/pull/196

I'm looking at what needs to change and include that with this pull.

sg00dwin:
@spadgett updated the pr with a fix for the resized project menu issue. PTAL
screen shot 2016-07-22 at 12 42 38 pm

spadgett:
[merge]

PR #211, opened 2016-07-22
Bump layout.attrs version to latest to pickup mobile size fix.

sg00dwin:
@jwforres PTAL

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -29,7 +29,7 @@
     "kubernetes-topology-graph": "0.0.23",
     "kubernetes-container-terminal": "0.0.14",
     "openshift-object-describer": "1.1.2",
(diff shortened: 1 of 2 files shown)


jwforres:
looks like the update dropped a bunch of webkit specific css, is that expected?

sg00dwin:
Actually those came out in @rhamilto pr https://github.com/benjaminapetersen/layout.attrs/pull/17
so I'm assuming they were no longer needed? that correct @rhamilto ?

rhamilto:
Yep, that's correct.  [Autoprefixer dropped those](https://github.com/postcss/autoprefixer#only-actual-prefixes).

jwforres:
[merge]

PR #218, opened 2016-07-24
Fix problem with parse-error directive close button

spadgett:
New errors were not displayed after the first error is dismissed.

@jwforres PTAL


jwforres:
[merge]

PR #219, opened 2016-07-25
Refactoring pipeline and trigger builds for consistency on overview

rhamilto:
The collapsed pipeline view needed to be refactored given it will now only display when a pipeline !isIncompleteBuild (the full pipeline will display when isIncompleteBuild).  After refactoring the html and less to support this simpler view, it occurred to me what I'd done is create an alternate version of the existing build status display already in use on the overview for triggers, which raised several questions.

screen-shot-2016-07-22-at-4 49 40-pm

Given that we're essentially communicating the same information in a different way, should we instead be using the existing build status display with the appropriate updates?

@jwforres, @spadgett, @ajacobs21e, @sg00dwin?


rhamilto:
pipeline builds and triggered builds are now parallel when isIncompleteBuild.

screen shot 2016-07-26 at 9 21 05 am

earlier phases look like

screen shot 2016-07-26 at 9 20 03 am
screen shot 2016-07-26 at 9 20 18 am

Remaining work:

- [ ] persist latest N isIncompleteBuild pipeline and triggered builds on the overview (currently, they disappear from the overview after 5 minutes)
- [ ] make persisted N isIncompleteBuild pipeline builds dismissible on the overview
- [ ] define N

@jwforres or @spadgett, I could use an assist on the remaining tasks above.

rhamilto:
@jwforres, this is now ready for review.

app/views/_build-timestamp.html
```diff
@@ -0,0 +1 @@
+<relative-timestamp timestamp="build.metadata.creationTimestamp"></relative-timestamp>
```
jwforres: seems excessive to have a partial that just adds a directive, what was the reason for this?

app/views/directives/build-pipeline.html
```diff
@@ -1,49 +1,7 @@
...
-            <a ng-href="{{build | jenkinsInputURL}}" target="_blank">Input Required</a>
-          </div>
-          <div class="pipeline-time" ng-class="stage.status" ng-if="stage.durationMillis && !(stage | pipelineStagePendingInput)">{{stage.durationMillis | timeOnlyDuration}}</div>
-          <div class="pipeline-time" ng-class="stage.status" ng-if="!stage.durationMillis && !(stage | pipelineStagePendingInput)">not started</div>
-        </div>
-      </div>
-    </div>
+<div>
+  <div ng-if="collapseStagesOnCompletion">
+    <div ng-if="build | isIncompleteBuild" ng-include="'views/directives/build-pipeline-expanded.html'"></div>
```
jwforres: if these views are includes and not top level template for a directive then give them the _ prefix on the name, since these are includes within a directive i'd be ok with `views/directives/_build-pipeline-expanded.html`

app/views/_build-timestamp.html
rhamilto: Nope.  Just an oversight.  Good catch. Will fix.

rhamilto:
@jwforres, comments addressed.  Thanks!

jwforres:
[merge]

PR #223, opened 2016-07-25
Fix service-group right border on standalone-services.

sg00dwin:
Set min-width to prevent aggressive truncation of service app name.
Prevent create route link from wrapping.


sg00dwin:
@spadgett PTAL

spadgett:
[merge]

Issue #227, opened 2016-07-26
Other resources list empty for cluster admin


jorgemoralespou:
I have created a project as "dev", and I can see a list of "other resources" I can use 

![image](https://cloud.githubusercontent.com/assets/78350/17135476/3a9cc336-5330-11e6-9fdc-bf1719d8d2c7.png)

If I log in as system:admin I can not see in the list of "other resources" any, even though I should be able to do everything.

![image](https://cloud.githubusercontent.com/assets/78350/17135521/7b99250a-5330-11e6-8c5b-a2b460c0afdf.png)

spadgett:
@jorgemoralespou How are you logging in as cluster admin? I didn't think it was possible in the web console since it requires certificate auth.

/cc @liggitt @jhadvig

jhadvig:
We are not displaying the resource list based on the permissions. Must be some other issue

jorgemoralespou:
You can create a user and add it to system:admin group to be able to login.

````
oc adm policy add-cluster-role-to-user cluster-admin admin
````

jhadvig:
I did just the same thing and I can see the whole list. Are there any other steps ?

jorgemoralespou:
@jhadvig You're right, as this admin user can not see that dropdown in any project

jorgemoralespou:
@jhadvig are you logged as admin user? Are you trying with the vendored ui in origin?

jhadvig:
@jorgemoralespou yes on both questions. Created a cluster user and checked other resource list. It was there.
Weird :-/

spadgett:
@jorgemoralespou Can you check the javascript console and dev tools network tab in your browser to see if there's an error somewhere? View -> Developer -> Developer Tools in Chrome.

spadgett:
Nevermind, I can reproduce.

spadgett:
@jhadvig It looks like we're not handling `rules['*/*']` in `AuthorizationService`?

<img width="215" alt="openshift_web_console" src="https://cloud.githubusercontent.com/assets/1167259/17137312/926c92ee-5308-11e6-991e-dc07f1363566.png">

spadgett:
> We are not displaying the resource list based on the permissions.

We do check.

https://github.com/spadgett/origin-web-console/blob/4ebe815d6c08b3fd86df29d4336e0f8ebf6d5d6a/app/scripts/controllers/otherResources.js#L53

spadgett:
[...]

PR #231, opened 2016-07-26
Modification of authorization rules data structure 

jhadvig:
Added the cluster admin authorization check && DRYed a bit the service
@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/227

### DIFF
--- a/app/scripts/controllers/otherResources.js
+++ b/app/scripts/controllers/otherResources.js
@@ -49,8 +49,16 @@ angular.module('openshiftConsole')
       .get($routeParams.project)
       .then(_.spread(function(project, context) {
         $scope.kinds = _.filter($scope.kinds, function(kind){
-          var resource = APIService.kindToResource(kind.kind);
-          return AuthorizationService.canI("list", kind.group ? kind.group + "/" + resource : resource, $scope.projectName);
+          var resourceAndGroup  = {
+            resource: APIService.kindToResource(kind.kind),
+            group: kind.group || ''
+          };
+          // exclude 'projectrequests', subresources, and REVIEW_RESOURCES from the list
+          if (AuthorizationService.checkResource(resourceAndGroup.resource)) {
+            return AuthorizationService.canI(resourceAndGroup, "list", $scope.projectName);
+          } else {
+            return false;
+          }
         });
         $scope.project = project;
         $scope.context = context;
@@ -74,7 +82,7 @@ angular.module('openshiftConsole')
       if (!selected) {
         return;
       }
-      $scope.selectedResource = (selected.group) ? selected.group + "/" + APIService.kindToResource(selected.kind) : APIService.kindToResource(selected.kind);
+      $scope.selectedResource = {resource: APIService.kindToResource(selected.kind), group: (selected.group || '')};
       // TODO - We can't watch because some of these resources do not support it (roles and rolebindings)
       DataService.list({
           group: selected.group,
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -129,8 +129,8 @@ angular.module('openshiftConsole')
     };
   })
   .filter('canI', function(AuthorizationService) {
-    return function(resource, verb, namespace) {
-      return AuthorizationService.canI(verb, resource, namespace);
+    return function(resource, verb, projectName) {
+      return AuthorizationService.canI(resource, verb, projectName);
     };
   })
   .filter('canIAddToProject', function(AuthorizationService) {
@@ -140,30 +140,59 @@ angular.module('openshiftConsole')
   })
   .filter('canIDoAny', function(canIFilter) {
     var resourceRulesMap = {
-      "buildConfigs": {"buildconfigs": ["delete", "update"], "buildconfigs/instantiate": ["create"]},
-      "builds": {"builds/clone": ["create"], "builds": ["delete", "update"]},
-      "deploymentConfigs": {"extensions/horizontalpodautoscalers": ["create", "update"], "deploymentconfigs": ["create", "update"]},
-      "deployments": {"replicationcontrollers": ["update", "delete"]},
-      "horizontalPodAutoscalers": {"extensions/horizontalpodautoscalers": ["update", "delete"]},
-      "imageStreams": {"imagestreams": ["update", "delete"]},
-      "persistentVolumeClaims": {"persistentvolumeclaims": ["update", "delete"]},
-      "pods": {"pods": ["update", "delete"], "deploymentconfigs": ["update"]},
-      "replicationControllers": {"horizontalpodautoscalers": ["create", "update"], "replicationcontrollers": ["create", "update"]},
-      "routes": {"routes": ["update", "delete"]},
-      "services": {"services": ["update", "create", "delete"]},
-      "projects": {'projects': ['delete', 'update']}
+      'buildConfigs': [
+        {group: '', resource: 'buildconfigs',             verbs: ['delete', 'update']},
+        {group: '', resource: 'buildconfigs/instantiate', verbs: ['create']}
+      ],
+      'builds': [
+        {group: '', resource: 'builds/clone', verbs: ['create']},
+        {group: '', resource: 'builds',       verbs: ['delete', 'update']}
+      ],
+      'deploymentConfigs': [
+        {group: 'extensions', resource: 'horizontalpodautoscalers', verbs: ['create', 'update']},
+        {group: '',            resource: 'deploymentconfigs',        verbs: ['create', 'update']}
+      ],
+      'deployments': [
+        {group: '', resource: 'replicationcontrollers', verbs: ['update', 'delete']}
+      ],
+      'horizontalPodAutoscalers': [
+        {group: 'extensions', resource: 'horizontalpodautoscalers', verbs: ['update', 'delete']}
+      ],
+      'imageStreams': [
+        {group: '', resource: 'imagestreams', verbs: ['update', 'delete']}
+      ],
+      'persistentVolumeClaims': [
+        {group: '', resource: 'persistentvolumeclaims', verbs: ['update', 'delete']}
+      ],
+      'pods': [
+        {group: '', resource: 'pods',              verbs: ['update', 'delete']},
+        {group: '', resource: 'deploymentconfigs', verbs: ['update']}
+      ],
+      'replicationControllers': [
+        {group: 'extensions', resource: 'horizontalpodautoscalers', verbs: ['create', 'update']},
+        {group: '',            resource: 'replicationcontrollers',   verbs: ['create', 'update']}
+      ],
+      'routes': [
+        {group: '', resource: 'routes', verbs: ['update', 'delete']}
+      ],
+      'services': [
+        {group: '', resource: 'services', verbs: ['update', 'create', 'delete']}
+      ],
+      'projects': [
+        {group: '', resource: 'projects', verbs: ['delete', 'update']}
+      ]
     };
     return function(resource) {
-      return _.some(resourceRulesMap[resource], function(verbs, resource) {
-        return _.some(verbs, function(verb) {
-          return canIFilter(resource,verb);
+      return _.some(resourceRulesMap[resource], function(rule) {
+        return _.some(rule.verbs, function(verb) {
+          return canIFilter({resource: rule.resource, group: rule.group}, verb);
         });
       });
     };
   })
   .filter('canIScale', function(canIFilter, isDeploymentFilter) {
     return function(deployment) {
-      return canIFilter(isDeploymentFilter(deployment) ? "deploymentconfigs/scale" : "replicationcontrollers", "update");
+      return canIFilter(isDeploymentFilter(deployment) ? 'deploymentconfigs/scale' : 'replicationcontrollers', 'update');
     };
   })
   .filter('tags', function(annotationFilter) {
--- a/app/scripts/services/authorization.js
+++ b/app/scripts/services/authorization.js
@@ -15,14 +15,16 @@ angular.module("openshiftConsole")
     // Transform data from:
     // rules = {resources: ["jobs"], apiGroups: ["extensions"], verbs:["create","delete","get","list","update"]}
     // into:
-    // normalizedRules = {"extensions/jobs": ["create","delete","get","list","update"]}
+    // normalizedRules = {"extensions": {"jobs": ["create","delete","get","list","update"]}}
     var normalizeRules = function(rules) {
       var normalizedRules = {};
-
       _.each(rules, function(rule) {
-        _.each(rule.resources, function(resource) {
-          _.each(rule.apiGroups, function(apiGroup) {
-            normalizedRules[(apiGroup === "") ? resource : apiGroup + "/" + resource] = rule.verbs;
+        _.each(rule.apiGroups, function(apiGroup) {
+          if (!normalizedRules[apiGroup]) {
+            normalizedRules[apiGroup] = {};
+          }
+          _.each(rule.resources, function(resource) {
+            normalizedRules[apiGroup][resource] = rule.verbs;
           });
         });
       });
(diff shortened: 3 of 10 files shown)


app/scripts/services/authorization.js
```diff
@@ -96,16 +96,22 @@ angular.module("openshiftConsole")
...
         return true;
       } else if (rules) {
         if (rules[resource]) {
-          return _.contains(rules[resource], verb) || _.contains(rules[resource], '*');
+          return checkRulesForResourceVerb(rules, resource, verb);
         } else if (rules['*']) {
-          return _.contains(rules['*'], verb) || _.contains(rules['*'], '*');
+          return checkRulesForResourceVerb(rules, '*', verb);
+        } else if (rules['*/*']) {
+          return true;
```
spadgett: I don't think this is right. We should at least check if it contains `verb` or `'*'`.

@deads2k Can you help me understand the difference between `rules['*']` and `rules['*/*']`?

jhadvig:
@spadgett originally what I wanted to do is to use the ` return checkRulesForResourceVerb(rules, '*/*', verb);` then checking the `*/*` but then I realized that the `*/*` means that the user is able to perform all actions on all resources. But now when Im thinking about it it will be better to to switch the `return true` for `return checkRulesForResourceVerb(...)` just to be safe

spadgett:
>  then I realized that the `*/*` means that the user is able to perform all actions on all resources

I'm not sure that's right. @deads2k or @liggitt can you clarify?

liggitt:
Where are you getting `*/*`? The API doesn't return data like that... are you joining the apiGroup and resource name with a slash?

spadgett:
Ah looks like it's coming from here? Didn't realize we were transforming the data.

https://github.com/spadgett/origin-web-console/blob/589113a123e95e44d4b53c2b3cc5129826685ada/app/scripts/services/authorization.js#L15-L30

Here is what the API returns.

```
      {
        "verbs": [
          "*"
        ],
        "attributeRestrictions": null,
        "apiGroups": [
          "*"
        ],
        "resources": [
          "*"
        ]
      }
```

jhadvig:
We are creating the `*/*` upon the rules normalization process in https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/authorization.js#L25 in which the cluster admin's rules have first object with `apiGroups` equal `*` and so does the `resources` and `verbs` where the normalization process creates `'apiGroups'/'resource' = ['verbs array']` entries

app/scripts/services/authorization.js
```diff
@@ -96,16 +96,22 @@ angular.module("openshiftConsole")
       return _.get(cachedRulesByProject.get(projectName || currentProject), ['rules']);
     };
 
+    var checkRulesForResourceVerb = function(rules, resource, verb) {
+      return _.contains(rules[resource], verb) || _.contains(rules[resource], '*');
+    };
+
     var canI = function(verb, resource, projectName) {
```
liggitt: how is apiGroup being provided to this function? is it included in resource?

app/scripts/services/authorization.js
```diff
@@ -96,16 +96,22 @@ angular.module("openshiftConsole")
...
       if (permissiveMode) {
         return true;
       } else if (rules) {
         if (rules[resource]) {
-          return _.contains(rules[resource], verb) || _.contains(rules[resource], '*');
+          return checkRulesForResourceVerb(rules, resource, verb);
         } else if (rules['*']) {
-          return _.contains(rules['*'], verb) || _.contains(rules['*'], '*');
+          return checkRulesForResourceVerb(rules, '*', verb);
+        } else if (rules['*/*']) {
```
liggitt: it looks like `normalizeRules` *sometimes* includes apiGroup with a slash, which makes it impossible to distinguish between {apiGroup:"builds",resource:"logs"} and {apiGroup:"",resource:"builds/logs"}

app/scripts/services/authorization.js
jhadvig: apiGroup is already part of the rules, to be more specific each key of the `rules` is constructed by `<apiGroup>/<resource>` and its value is an array of verbs

liggitt: it's part of the rules, but I don't see it being provided to `canI` separately, which means it's hard to check if  the literal API group provided to `canI` *or* `*` is allowed without parsing the provided resource. Also, if you're using `/` to delimit the api group, there's no way to distinguish between an api group and a subresource

jhadvig: The `canI` filter directly takes resource in `<apiGroup>/<resource>` format and word 
https://github.com/openshift/origin-web-console/blob/master/app/views/browse/deployment-config.html#L50

The only place we need to distinguish between api group and a subresource is in the `canIAddToProject` but we are paasing there the original rules, not the normalized
https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/authorization.js#L68

app/scripts/services/authorization.js
jhadvig: The only place we need to distinguish between api group and a subresource is in the canIAddToProject but we are paasing there the original rules, not the normalized
https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/authorization.js#L68

app/scripts/services/authorization.js
liggitt: merging the API group and resource into a single arg makes it more difficult to match against rules. To match rules correctly, the passed resource *has* to be parsed into group and resource separately, which means it should not use the same delimiter as subresources.

jhadvig: So then we can change the apigroup, resource delimiter from `/` to `:`, would it be ok ?

liggitt: why not just pass the data in a structured way? you can keep the string as a simple case (api group = ''), and pass structured data in other cases. make the first line of `canI` do this:
```
resource = APIService.toResourceGroupVersion(resource);
```
and wherever you want to pass a non-empty API group, do `canI('create',{group:'extensions',resource:'horizontalpodautoscalers'})`

jhadvig: I wanted to avoid that kind of notation using the `canI` filter, meaning not passing an objects with nested fields, just simple:
- resource without apiGroup: `ng-if="'buildconfigs' |  canI : 'create'"`
- resource with apiGroup `ng-if="'extensions/horizontalpodautoscalers' | canI 'create'"`

 Also searching in the normalized rules is instant since all keys are uniq based on the `apiGroup/resource` format. Quick searching in the rules was the main motivation behind the rules normalization.
I kinda see why the slash delimiter is not the best way, but we are not checking if the given resource is a sub-resource anyway with the `canI` filter.

liggitt: to check resource=foo, apigroup=bar, you have to check four combinations to be accurate:
```
resource=foo, apigroup=bar
resource=foo, apigroup=*
resource=*, apigroup=bar
resource=*, apigroup=*
```
a data structure that facilitates that will also accommodate the first lookup, which is the only one we're currently doing

jhadvig:
@liggitt I've update the PR based on your conversation. Thanks so much for explaining the issue :+1: 
PTAL

cc'ing @spadgett

app/scripts/filters/resources.js
```diff
@@ -127,8 +127,8 @@ angular.module('openshiftConsole')
     };
   })
   .filter('canI', function(AuthorizationService) {
-    return function(resource, verb, namespace) {
-      return AuthorizationService.canI(verb, resource, namespace);
+    return function(resource, verb, groupAndNamespace) {
```
liggitt: group belongs with the resource, not the namespace

app/scripts/filters/resources.js
```diff
@@ -138,23 +138,23 @@ angular.module('openshiftConsole')
...
-      "deployments": {"replicationcontrollers": ["update", "delete"]},
-      "horizontalPodAutoscalers": {"extensions/horizontalpodautoscalers": ["update", "delete"]},
-      "imageStreams": {"imagestreams": ["update", "delete"]},
-      "persistentVolumeClaims": {"persistentvolumeclaims": ["update", "delete"]},
-      "pods": {"pods": ["update", "delete"], "deploymentconfigs": ["update"]},
-      "replicationControllers": {"horizontalpodautoscalers": ["create", "update"], "replicationcontrollers": ["create", "update"]},
-      "routes": {"routes": ["update", "delete"]},
-      "services": {"services": ["update", "create", "delete"]},
-      "projects": {'projects': ['delete', 'update']}
+      "deploymentConfigs": {"horizontalpodautoscalers": {verbs: ["create", "update"], group: "autoscaling"}, "deploymentconfigs": {verbs: ["create", "update"]}},
```
liggitt: no reason for the value here to be a map... putting group and verbs together is weird... this would make more sense:
```
"deploymentConfigs: [
  {group:'autoscaling', resource:'horizontalpodautoscalers', verbs: ["create","update"]},
  {group:'',            resource:'deploymentconfigs',        verbs: ["create", "update"]},
],
...
      return _.some(resourceRulesMap[resource], function(rules) {
        return _.some(rules.verbs, function(verb) {
          return canIFilter(rules.resource,verb, {group: (rules.group || "")});
         });
       });
```

jhadvig:
@liggitt I've update the PR. PTAL

app/scripts/filters/resources.js
```diff
@@ -138,30 +138,30 @@ angular.module('openshiftConsole')
...
+      'deployments': [{group: '', resource: 'replicationcontrollers', verbs: ['update', 'delete']}],
+      'horizontalPodAutoscalers': [{group: 'autoscaling', resource: 'horizontalpodautoscalers', verbs: ['update', 'delete']}],
+      'imageStreams': [{group: '', resource: 'imagestreams', verbs: ['update', 'delete']}],
+      'persistentVolumeClaims': [{group: '', resource: 'persistentvolumeclaims', verbs: ['update', 'delete']}],
+      'pods': [{group: '', resource: 'pods', verbs: ['update', 'delete']}, {group: '', resource: 'deploymentconfigs', verbs: ['update']}],
+      'replicationControllers': [{group: 'autoscaling', resource: 'horizontalpodautoscalers', verbs: ['create', 'update']}, {group: '', resource: 'replicationcontrollers', verbs: ['create', 'update']}],
+      'routes': [{group: '', resource: 'routes', verbs: ['update', 'delete']}],
+      'services': [{group: '', resource: 'services', verbs: ['update', 'create', 'delete']}],
+      'projects': [{group: '', resource: 'projects', verbs: ['delete', 'update']}]
```
liggitt: looks correct, but would be easier to read as:
```
      'buildConfigs': [
      ],
      'builds': [
        {group: '', resource: 'builds/clone', verbs: ['create']}, 
        {group: '', resource: 'builds',       verbs: ['delete', 'update']}
      ],
      'deploymentConfigs': [
        {group: 'autoscaling', resource: 'horizontalpodautoscalers', verbs: ['create', 'update']}, 
        {group: '',            resource: 'deploymentconfigs',        verbs: ['create', 'update']}
      ],
      'deployments': [
        {group: '', resource: 'replicationcontrollers', verbs: ['update', 'delete']}
      ],
      'horizontalPodAutoscalers': [
        {group: 'autoscaling', resource: 'horizontalpodautoscalers', verbs: ['update', 'delete']}
      ],
      'imageStreams': [
        {group: '', resource: 'imagestreams', verbs: ['update', 'delete']}
      ],
      'persistentVolumeClaims': [
        {group: '', resource: 'persistentvolumeclaims', verbs: ['update', 'delete']}
      ],
      'pods': [
        {group: '', resource: 'pods',              verbs: ['update', 'delete']}, 
        {group: '', resource: 'deploymentconfigs', verbs: ['update']}
      ],
      'replicationControllers': [
        {group: 'autoscaling', resource: 'horizontalpodautoscalers', verbs: ['create', 'update']}, 
        {group: '',            resource: 'replicationcontrollers',   verbs: ['create', 'update']}
      ],
      'routes': [
        {group: '', resource: 'routes', verbs: ['update', 'delete']}
      ],
      'services': [
        {group: '', resource: 'services', verbs: ['update', 'create', 'delete']}
      ],
      'projects': [
        {group: '', resource: 'projects', verbs: ['delete', 'update']}
      ]
```

app/scripts/services/authorization.js
```diff
@@ -96,23 +98,37 @@ angular.module("openshiftConsole")
...
+        }
+        var verbs = resources[resource];
+        if (!verbs) { 
+          return false;
+        }
+        return _.contains(verbs, verb) || _.contains(verbs, '*');
+    };
+
+    // canI checks whether any rule allows the specified verb on the specified group-resource (directly or via a wildcard rule).
+    var canI = function(resource, verb, group, projectName) {
```
liggitt: you don't need a separate param for the group... resource should either be a string (which implies group='') or an object with `resource` and `group` fields

app/views/browse/deployment-config.html
```diff
@@ -47,12 +47,12 @@
                       <a ng-href="project/{{projectName}}/set-limits?dcName={{deploymentConfig.metadata.name}}"
                         role="button">Set Resource Limits</a>
                     </li>
-                    <li ng-if="!autoscalers.length && ('extensions/horizontalpodautoscalers' | canI : 'create')">
+                    <li ng-if="!autoscalers.length && ('horizontalpodautoscalers' | canI : 'create' : 'autoscaling')">
```
liggitt: is it possible for this to be `({group:'autoscaling',resource:'horizontalpodautoscalers'} | canI : 'create')`?

app/scripts/filters/resources.js
jhadvig: will prettify !

jhadvig:
@liggitt comments addressed

app/scripts/services/authorization.js
```diff
@@ -96,23 +98,39 @@ angular.module("openshiftConsole")
...
-        } else if (rules['*']) {
-          return _.contains(rules['*'], verb) || _.contains(rules['*'], '*');
-        } else {
-          return false;
-        }    
-      } else {
+      }
+
+      // normalize to structured form
+      var r = APIService.toResourceGroupVersion(_.isObject(resource) ? resource.resource : resource);
```
liggitt: don't need this check... `toResourceGroupVersion` normalizes the input

app/scripts/services/authorization.js
```diff
@@ -96,23 +98,39 @@ angular.module("openshiftConsole")
...
+      if (_.isObject(resource)) {
```
liggitt: don't need this block, `toResourceGroupVersion` handles this for you

app/scripts/filters/resources.js
```diff
@@ -127,8 +127,8 @@ angular.module('openshiftConsole')
     };
   })
   .filter('canI', function(AuthorizationService) {
-    return function(resource, verb, namespace) {
-      return AuthorizationService.canI(verb, resource, namespace);
+    return function(resource, verb, group, projectName) {
```
liggitt: remove the group as a separate parameter

liggitt:
nits, LGTM otherwise. you tested all the pages with view, admin, and cluster-admin roles?

jhadvig:
@liggitt PR updated based on comments.
Tested for all three roles

app/scripts/services/authorization.js
```diff
@@ -96,23 +98,37 @@ angular.module("openshiftConsole")
...
+      }
+
+      // normalize to structured form
+      var r = APIService.toResourceGroupVersion(resource);
+      var rules = getRulesForProject(projectName || currentProject);
+      if (!rules) {
         return false;
       }
-    };
+      console.log(r.resource);
```
liggitt: remove debug logging

liggitt:
nit on debug logging, then LGTM. final approval goes to @spadgett or @jwforres (might want to double check the filter usage in templates)

jhadvig:
Comment addressed.
@spadgett @jwforres PTAL

app/scripts/filters/resources.js
```diff
@@ -138,30 +138,59 @@ angular.module('openshiftConsole')
   })
   .filter('canIDoAny', function(canIFilter) {
     var resourceRulesMap = {
```
benjaminapetersen: fwiw, i like the formatting, improves readability!

app/views/browse/deployment-config.html
```diff
@@ -47,12 +47,12 @@
                       <a ng-href="project/{{projectName}}/set-limits?dcName={{deploymentConfig.metadata.name}}"
                         role="button">Set Resource Limits</a>
                     </li>
-                    <li ng-if="!autoscalers.length && ('extensions/horizontalpodautoscalers' | canI : 'create')">
+                    <li ng-if="!autoscalers.length && ({resource: 'horizontalpodautoscalers', group: 'autoscaling'} | canI : 'create')">
```
spadgett: The JS code makes requests for `horizontalpodautoscalers` in the `extensions` group. @liggitt could the mismatch cause a problem?

app/scripts/controllers/otherResources.js
```diff
@@ -49,8 +49,16 @@ angular.module('openshiftConsole')
       .get($routeParams.project)
       .then(_.spread(function(project, context) {
         $scope.kinds = _.filter($scope.kinds, function(kind){
-          var resource = APIService.kindToResource(kind.kind);
-          return AuthorizationService.canI("list", kind.group ? kind.group + "/" + resource : resource, $scope.projectName);
+          var resource = {
```
spadgett: `resource.resource` is a little awkward, but I don't have a great suggestion. `var resourceAndGroup = ...` maybe?

app/views/browse/deployment-config.html
liggitt: you should check what you use

spadgett: @jhadvig let's update the views to check using the `extensions` group until we change the JS code

soltysh: Generally, as a rule of thumb, we should use the appropriate and stable groups for resources instead of the `extensions` one. The resources in the latter are meant for deprecation sooner or later, whereas the `autoscaling` in case of HPA will keep working. The same applies to other group/resource pairs like `batch/Job`, etc.

jhadvig:
@spadgett I've update the PR based on the comments and you conversation. PTAL

app/views/browse/deployment-config.html
spadgett: Tracked by #255

spadgett:
thanks @jhadvig 

[merge]

Issue #232, opened 2016-07-26
Standalone service showing up as "service group"


spadgett:
This service has no routes or app label, but is treated as a service group on our overview.

<img width="751" alt="openshift_web_console" src="https://cloud.githubusercontent.com/assets/1167259/17139546/f3437bdc-5312-11e6-8566-f11a3c790c97.png">

/cc @jwforres

spadgett:
It has the style. Seems to be a problem in our CSS.

<img width="416" alt="openshift_web_console" src="https://cloud.githubusercontent.com/assets/1167259/17139670/6cfeca30-5313-11e6-87c1-085b99d75f8f.png">

/cc @sg00dwin

jwforres:
@spadgett but a standalone service shouldn't have the grouped services block, was that being hidden with CSS before?

spadgett:
Right it was hidden with CSS. Probably better to switch to an ng if

spadgett:
Hiding the group services box with an ng-if stretches the service tile to 100% width. Will need to work with @sg00dwin on this

spadgett:
@sg00dwin 

<img width="1111" alt="banners_and_alerts_and_openshift_web_console" src="https://cloud.githubusercontent.com/assets/1167259/17154094/8a1c379e-534d-11e6-8770-498cccc12efb.png">

The difference is the first one is a "group" because it has a route, the second one is just a standalone service (and has the `standalone-service` class). The standalone services should tile if more than one without the "no grouped services" message.

Sorry should have caught this on review. See removed styles

https://github.com/openshift/origin-web-console/pull/115/files#diff-d3494a108775ccb728f6772fc669a264L125

and

https://github.com/openshift/origin-web-console/pull/115/files#diff-d3494a108775ccb728f6772fc669a264L439

sg00dwin:
@spadgett @jwforres 
`<overview-service-group>` is used to generate the div structure of 'grouped' services and standalone-services so they reuse a lot of css, but this causes a lot of conflicting display rules and overrides to make them behave/flow differently within the overview.

Can we create a `<overview-standalone-service>` directive and structure accordingly to mitigate those issues?

PR #234, opened 2016-07-26
Fix for standalone-service displayed as service group

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/232

@spadgett PTAL


spadgett:
This fixes the border extending too high, but the whole grouped services block to the right should not be there. I think it's better to fix it with angular in the view rather than using CSS to hide it.

Check in dist and I'll merge this change, though. We can fix the other problem in a follow on.

spadgett:
Removing the fixes declaration from the description so the issue is not closed.

sg00dwin:
@spadgett dist added

spadgett:
[merge]

PR #235, opened 2016-07-27
Render Image Data

stefwalter:
This starts to render image data onto its own page. The URLs contain the tags of an ImageStream such as:

```
/project/my-project/browse/images/busybox:1.23
```

The actual rendering code comes from a shared component, shared with the Atomic Registry.

While this currently works, it's not ready to be merged. Remaining work:

 * [x] Create new repository for bower component (@danmcp) "openshift/registry-image-widgets"
 * [x] Register bower component, remove WIP bower_components commit. (@stefwalter)
 * [x] Use tabs to switch between sections on the browser page.
 * [x] Fix general styling
 * [x] Use non-cockpit code to convert bytes to 'MB'
 * [x] Decide how to render 'To pull this image' with appropriate registry host. (Just placeholder for now).
 * [x] Fixups for neview

CC @jwforres


stefwalter:
Shared bower component is temporarily here: https://github.com/stefwalter/image-widgets/

stefwalter:
@jwforres Are you okay with the bower and github name ```kubernetes-image-widgets``` or something else would be better? I'm thinking forward a bit and hoping that we'll have first class image representations in Kube upstream at some point.

@jwforres How do we get the actual public registry host name that users can use to pull the image onto their machine. When it's not available we substitute a "registry" placeholder.

stefwalter:
[Video](https://www.youtube.com/watch?v=oN5RsX0tFpc)

stefwalter:
Cockpit integration of these same widgets is blocking on https://github.com/cockpit-project/cockpit/pull/4771 ... another pull request to follow on that end.

jwforres:
@stefwalter for the naming of the bower component, i hesitate to call it kubernetes-image-widgets, i dont think there are any plans to upstream imagestreams but @smarterclayton can weigh in here

jwforres:
what was your reason for leaving the : as the delimeter in the URL?  I think i'd rather stick with URL conventions and add the path delimeter so there is a distinction between the image stream page and the image stream tag page, then you wouldnt have to do the special logic in the routes

app/scripts/app.js
```diff
@@ -123,7 +124,13 @@ angular
         controller: 'ImagesController'
       })
       .when('/project/:project/browse/images/:image', {
-        templateUrl: 'views/browse/image.html',
+        templateUrl: function(params) {
+          if (params.image && params.image.indexOf(':') === -1) {
```
jwforres: See https://github.com/openshift/origin-web-console/pull/235#issuecomment-235908594
would rather not have to do this

jwforres:
I noticed the *.map files are being included in the dist of the image widget bower component, I'd prefer those not be included in what we are shipping, it will bloat the final built bindata.go

jwforres:
Ah sorry, realized you temporarily have the bower component embedded here.  As long as the *.map files aren't in the "main" of the bower.json spec of the component then it won't be an issue.

jwforres:
On the config tab where you have the command, I understand that its the command the container will be run with, but I think there needs to be something to tell our users that.

jwforres:
So when we show the container command elsewhere we use the `<code>` style, like what you have for pulling the image, is there a reason for using a different style for the container's command than for the docker pull command?

debugterminal

jwforres:
@smarterclayton @deads2k see @stefwalter 's question about the registry https://github.com/openshift/origin-web-console/pull/235#issuecomment-235542107

how can we know in the console where to point them to as an end user, is it possible today or are we stuck using a placeholder?

deads2k:
> how can we know in the console where to point them to as an end user, is it possible today or are we stuck using a placeholder?

The question is how can i find the hostname is assigned to a route that points to our internal docker registry?  There is no way for a normal user to retrieve that information via the API today.

stefwalter:
@jwforres I'll call the bower component "registry-image-widgets" ... but can you make a new repository called "image-widgets" here: https://github.com/kubernetes-ui

Or is the hosting at kubernetes-ui also a problem? I think certain of the widgets can be used with raw docker manifests, history, layers ... so although this is currently very openshift focused, it can be used without openshift for several cases.

jwforres:
Lets host it under the openshift org, we have a few other bower components
hosted there now too  @danmcp can create the repo   is there a reason to
not call the repo registry-image-widgets since thats what you are calling
the component?   so openshift/registry-image-widgets for the repo

danmcp:
@stefwalter @jwforres https://github.com/openshift/registry-image-widgets

stefwalter:
Thanks @danmcp

app/scripts/app.js
stefwalter: Fixed.

stefwalter:
> So when we show the container command elsewhere we use the <code> style, like what you have for pulling the image, is there a reason for using a different style for the container's command than for the docker pull command?

Fixed.

stefwalter:
Ready for review.

jwforres:
let me know when you are ready for a code review

stefwalter:
@jwforres Should be ready now. I've run 'grunt build' on each commit.

jwforres:
Do you have screenshots of what things look like at mobile resolution?

stefwalter:
> This means any time there is a change to the image stream object it's going to re-fetch the ImageStreamTag. Whether that tag has been updated or not. Can we limit it to only fetching the tag again if we need to? 

Only the one that's currently displayed.  Openshift doesn't allow us to watch ImageStreamTag objects. You can reproduce this with this sorta command, which produces a 404:

```
curl http://127.0.0.1:8001/oapi/v1/watch/namespaces/marmalade/imagestreamtags
```

> I think we know per tag when it was last updated don't we?

How would we know that? We know when the tag data itself changed, but not when the referenced data changes. At least on my test systemst, I see no ```resourceVersion``` or ```uid``` in the ```status.tags``` data.

stefwalter:
> Do you have screenshots of what things look like at mobile resolution?

screen shot 2016-07-30 at 07 19 42
screen shot 2016-07-30 at 07 19 47
screen shot 2016-07-30 at 07 19 49

stefwalter:
> How would we know that? We know when the tag data itself changed, but not when the referenced data changes. At least on my test systems, I see no resourceVersion or uid in the status.tags data.

In addition even the ```ImageStreamTag``` object doesn't have a uid. Nor do the embedded image objects contain their ```resourceVersion``` or ```uid```.

stefwalter:
FWIW, this change only displays rough approximations of image layers until https://github.com/openshift/origin/issues/7880 is fixed or https://github.com/openshift/origin/pull/7956 is merged. But that's not a blocker for this change.

jwforres:
I thought there was something on the ImageStream object that said when the
last sync had happened @smarterclayton

I don't know if there is enough detail there to know that the particular
tag you are interested in has been updated and should be fetched again.

If we don't have enough info there to know when to refetch, then I would at
least want this updated to include a _.debounce so that if we get a bunch
of events across the imagestream websocket within a short time (200ms or
so?) that they don't all trigger separate http requests for the
ImageStreamTag.

stefwalter:
> I thought there was something on the ImageStream object that said when the
last sync had happened @smarterclayton

ImageStreams are updated by lots more than a sync, no?

> If we don't have enough info there to know when to refetch, then I would at
least want this updated to include a _.debounce so that if we get a bunch
of events across the imagestream websocket within a short time (200ms or
so?) that they don't all trigger separate http requests for the
ImageStreamTag.

Added a debounce.

jwforres:
@stefwalter on the mobile screenshots it would be good if the digest and identifier could be forced to wrap, in the console code we have a class "word-break" that does this (sets the best word-break classes for each browser)  since the images are in a separate repo though i not that class doesnt exist outside the console context.  I dont mind fixing this in a follow-up though.

jwforres:
[merge]

jwforres:
@stefwalter Just talked to @smarterclayton and he said there is a 'generation' on status, if that changes then that's when you should re-fetch ImageStreamTag.  Could you do that in a follow-up PR?

stefwalter:
> @stefwalter Just talked to @smarterclayton and he said there is a 'generation' on status, if that changes then that's when you should re-fetch ImageStreamTag. Could you do that in a follow-up PR?

Sure. I'm away for the next few days, but I can follow this up later.

PR #237, opened 2016-07-27
Minor fixes to pod template probe command display

spadgett:
- Prevent trailing whitespace from being highlighted in the `<code>` block
- Use `nowrap` around "1s timeout" to prevent awkward wrapping

Before:

openshift_web_console

After:

openshift_web_console

@jwforres PTAL


jwforres:
[merge]

Issue #224, opened 2016-07-26
Edit build config page looks odd when there's no source repo


spadgett:
It's a header with no content under it. This might not be uncommon for pipeline builds.

<img width="696" alt="openshift_web_console" src="https://cloud.githubusercontent.com/assets/1167259/17122316/6440b59a-52a8-11e6-88e5-67b14838ceec.png">

@jwforres @jhadvig

jhadvig:
@spadgett yeah I guess we can hide the "Source Configuration" header if the build strategy is `JenkinsPipelineBuildStrategy`
I can take a look on it if you dont mind.

spadgett:
@jhadvig `JenkinsPipelineStrategy` builds can have source. I think we just need

```
ng-if="(build | buildStrategy).from"
```

 on the section?

spadgett:
Sorry, I mean

```
ng=if="buildConfig.spec.source.type === 'None'"
```

jhadvig:
@spadgett yep that should do it. Will update and send PR ASAP

PR #239, opened 2016-07-27
Dont show 'Source Configuration' section when source type is 'None'

jhadvig:
@spadgett besides fixing the issues I've cleaned a bit the editor where I've put the 'Jenkins Pipeline Configuration' into standalone section. PTAL
Closes https://github.com/openshift/origin-web-console/issues/224

### DIFF
--- a/app/views/edit/build-config.html
+++ b/app/views/edit/build-config.html
@@ -21,7 +21,7 @@
                 <div class="resource-details">
                   <div class="row">
                     <div class="col-lg-6">
-                      <div class="section">
+                      <div ng-if="buildConfig.spec.source.type !== 'None'" class="section">
                         <h3>Source Configuration</h3>
                         <dl class="dl-horizontal left">
                           <div ng-if="sources.git">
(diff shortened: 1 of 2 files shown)


app/views/edit/build-config.html
```diff
@@ -319,6 +266,59 @@
...
+                            class="form-control"
+                            ng-model="jenkinsfileOptions.type"
+                            ng-options="type.id as type.title for type in jenkinsfileTypes"
+                            aria-describedby="jenkinsfile-type-help">
+                          </select>
+                          <div class="help-block" id="jenkinsfile-type-help">
+                            Use a Jenkinsfile from the source repository or specify the
+                            Jenkinsfile content directly in the build configuration.
+                          </div>
+                        </div>
```
spadgett: I think we should hide this field if source type is 'None'.

spadgett: Or rather, we should hide the field if source type is anything other than Git

jhadvig:
@spadgett updated. PTAL

spadgett:
thanks @jhadvig 

[merge]

PR #240, opened 2016-07-27
Making pipeline builds dismissible on overview

rhamilto:
and reversing sort of triggered builds so that they display in the same order as pipeline builds; additionally, i removed animation on triggered builds so that they don't fade out but rather instantly disappear (which also makes them consistent with pipeline builds).

for dismissible pipeline builds, @benjaminapetersen helped me move the close functionality out of the triggers directive and put it in a separate directive so both the triggers directive and the build pipelines directive can make use of it.

we're still wrestling with persisting the latest complete builds while displaying any in progress builds.  a separate PR with that functionality will follow.

@jwforres, PTAL.


jwforres:
[merge]

jwforres:
[merge]

PR #251, opened 2016-08-02
Bug 1333037 - enable application link after its endpoint has exposed IP address and port

jhadvig:
@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -340,7 +340,6 @@ angular.module('openshiftConsole')
       $scope.renderOptions.showLoading = !loaded && projectEmpty;
     };
 
-
     $scope.viewPodsForDeployment = function(deployment) {
       if (_.isEmpty($scope.podsByDeployment[deployment.metadata.name])) {
         return;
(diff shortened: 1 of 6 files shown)


app/scripts/directives/util.js
```diff
@@ -177,4 +177,18 @@ angular.module('openshiftConsole')
...
       });
     };
+  }).directive('linkDisabled', function(){
+    return {
+      restrict: 'A',
+      scope: {
+        linkDisabled: '='
+      },
+      link: function(scope, element) {
+        element.bind('click', function(event) {
```
jwforres: click handlers fire in the order they are bound, its possible this prevention handler could be bound in a different order than the click handler you are trying to prevent from firing

app/scripts/controllers/overview.js
```diff
@@ -427,6 +426,11 @@ angular.module('openshiftConsole')
           Logger.log("routes (subscribe)", $scope.routesByService);
         }));
 
+        watches.push(DataService.watch("endpoints", context, function(endpointsData) {
```
jwforres: what advantage do we get by watching endpoints, as opposed to looking at the status of the route (which we already have) where we can know if it has been admitted by the router?

jwforres: Do you know if endpoints will show up for pods that are Running but whose containers are not Ready?

app/scripts/filters/resources.js
```diff
@@ -410,6 +410,21 @@ angular.module('openshiftConsole')
       return label;
     };
   })
+  .filter('hasRouteEndpoints', function() {
+    return function(routeName, endpoints) {
+      if (!endpoints || !endpoints[routeName]) {
```
jwforres: pretty sure endpoints are named based on the service, not the route

jhadvig:
Closing this one after conversation with @jwforres

PR #258, opened 2016-08-02
Prevent empty aleternate service selection when creating routes

spadgett:
Select a new item when the currently-selected value is removed from options.

https://bugzilla.redhat.com/show_bug.cgi?id=1362479

@jwforres PTAL


jwforres:
[merge]

PR #271, opened 2016-08-03
Show get started message when only hidden pods exist in project

### LINKED ISSUE #110: Blank overview when I only have build configs and builds
In projects where someone is only running builds and has no DCs, no Services, and no non-infrastructure pods.  You got a blank overview.  I suspect this is because our check on the hashSize of pods doesnt take into account that we dont show you build and deploy pods today.

nothing-here

spadgett:
Fixes #110

@jwforres PTAL


spadgett:
`$scope.monopodsByService` is already filtered using `showMonopod` in the `groupPods` function.

jwforres:
[merge]

PR #274, opened 2016-08-03
Fix log archives link

benjaminapetersen:
Fix bug 1358674 : "archive" link does not automatically select correct  index/pattern

@jwforres @spadgett 
@ewolinetz 

This update will drop the user on a kibana page something like this:
screen shot 2016-08-03 at 11 56 15 am

### DIFF
--- a/app/scripts/services/logLinks.js
+++ b/app/scripts/services/logLinks.js
@@ -49,7 +49,7 @@ angular.module('openshiftConsole')
         var uri = new URI();
         _.each(params, function(param) {
           uri.addSearch(param);
-        });        
+        });
         $window.open(uri.toString(), '_blank');
       };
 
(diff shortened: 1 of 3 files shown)


app/scripts/services/logLinks.js
```diff
@@ -65,12 +65,12 @@ angular.module('openshiftConsole')
         ")",
         "&_a=(",
           //"columns:!(_source),",
-          "columns:!(kubernetes_container_name,<%= containername %>),", 
-          "index:'<%= namespace %>.*',",
+          "columns:!(kubernetes_container_name,message),",
```
jwforres: @ewolinetz 's comment in the bugzilla says this should be log, what is the different between log and message here?

jwforres:
Can we get sign-off from @ewolinetz @sosiouxme @jcantrill that the kibana screenshot ben added looks right now for what you want us to be linking to?

change itself looks good, just the one question about log vs message.  also needs dist built

benjaminapetersen:
@jwforres got the dist
pinged @ewolinetz again
[test]

app/scripts/services/logLinks.js
ewolinetz: I had mistakenly wrote 'log' in the bz. In a later filter in Fluentd we move entries from 'log' to 'message' and then we remove 'log' from the payload that is sent to Elasticsearch.  We can verify that message is correct since there isn't a 'log' column to select from the left side of the screen

ewolinetz:
also to point out, the screen cap is a little misleading since we aren't going to display the `kubernetes_pod_name` and `kubernetes_namespace_name` columns for users.  It is there for verification that we are correctly querying on those fields

jwforres:
[merge]

@benjaminapetersen please work on backporting this for 3.2.1

benjaminapetersen:
@jwforres backport in progress

PR #278, opened 2016-08-03
Build Image Documentation Link in Description

### LINKED ISSUE #243: Linkify links in builder image descriptions
Jessica mentioned that in the future (or are currently) linkifying links in template descriptions. We should also do the same for builder images so that we can link out to relevant documentation.

For instance, with the [Ruby 2.0 builder image](https://github.com/openshift/online/blob/master/templates/image-streams.json#L73) in the OpenShift Online image stream, I may want to point users to the [OpenShift Online Docs on Ruby](https://docs.openshift.com/online/using_images/s2i_images/ruby.html) or directly to the [SCL S2I container readme](https://github.com/sclorg/s2i-ruby-container/tree/master/2.0).

As an example, if I came to the web console as a PHP developer looking to deploy a REST API with Laravel, I know that Laravel has very specific PHP package requirements. If I wanted to find out what is included with our PHP 5.6 S2I builder image my best best would be to look at the actual [source repo for the PHP 5.6 image](https://github.com/sclorg/s2i-php-container/tree/master/5.6). Included packages are not well documented elsewhere, and this would be hard for external users to find.

MarkDeMaria:
Fixes #243 
-Add link to github repo if available on newFromImage page

image

@jwforres

### DIFF
--- a/app/views/directives/osc-image-summary.html
+++ b/app/views/directives/osc-image-summary.html
@@ -4,9 +4,13 @@
   <div ng-show="resource | annotation:'provider'">Provider: {{ resource | annotation:'provider' }}</div>
   <div ng-show="resource.metadata.namespace">Namespace: {{ resource.metadata.namespace }}</div>
 </div>
-<div class="resource-description"
+<div class="resource-description pad-bottom-none"
      ng-bind-html="resource | description | linky"
      ng-class="{'gutter-bottom': tag !== 'latest'}">
+</div>
+<div ng-show="resource | annotation:'sampleRepo'"
+     class="mar-bottom-md"
+     ng-bind-html="resource | annotation:'sampleRepo' | linky">
 </div>
 
 <div ng-if="tag === 'latest'" class="alert alert-info">
--- a/dist/scripts/templates.js
+++ b/dist/scripts/templates.js
@@ -5593,7 +5593,9 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
     "<div ng-show=\"resource | annotation:'provider'\">Provider: {{ resource | annotation:'provider' }}</div>\n" +
     "<div ng-show=\"resource.metadata.namespace\">Namespace: {{ resource.metadata.namespace }}</div>\n" +
     "</div>\n" +
-    "<div class=\"resource-description\" ng-bind-html=\"resource | description | linky\" ng-class=\"{'gutter-bottom': tag !== 'latest'}\">\n" +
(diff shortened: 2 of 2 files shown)


app/views/directives/osc-image-summary.html
```diff
@@ -4,10 +4,14 @@
   <div ng-show="resource | annotation:'provider'">Provider: {{ resource | annotation:'provider' }}</div>
   <div ng-show="resource.metadata.namespace">Namespace: {{ resource.metadata.namespace }}</div>
 </div>
-<div class="resource-description"
+<div class="resource-description pad-bottom-none"
      ng-bind-html="resource | description | linky"
```
jwforres: ah we were already running linky on the description... thats really what i was looking for

@luciddreamz do you also want to have the sample repo show up after the description like @MarkDeMaria has here

spadgett: We also show the sample repository below the git input. I think this might cause confusion? It's not clear in the description that this is just a sample and you can pick any repo.

stevekuznetsov:
[test]

app/views/directives/osc-image-summary.html
luciddreamz: Yeah I don't think we need the sample repo showing up in both places. I really just want to link out to any relevant docs in the description. If anything, underneath the sample repo we could either embed the readme.md or link out to the readme, or mention something to users that they should look at the readme for the sample repo if they use it. For some reason this never seems to be common sense.

luciddreamz: So, for "Build and run PHP 5.6 applications", I wouldn't mind if that link, instead of going to the sample repo, was just a link to either [our docs](https://docs.openshift.com/online/using_images/s2i_images/php.html) or to the [SCLorg repo for PHP 5.6](https://github.com/sclorg/s2i-php-container/tree/master/5.6), possibly with a little book icon next to it indicating docs (see Heroku screenshot example below). Linking to the SCLorg would be easier since we could use the same link for all OpenShift flavors, but this might be some kind of annotation on the image stream anyways. Didn't we have a docs link annotation in v2 cartridges?

Heroku UI, docs link

jwforres:
@luciddreamz to be clear what we are saying is that you can put URLs anywhere in the description and they will automatically be turned into links in the console, this works for both templates and imagestreams.  Its a fix that went in during 3.3.  Given that, is there really anything else you need from us on this?

luciddreamz:
I guess converting a link anywhere in the description to a link is fine. We may want to start having a documentation link as a default annotation for image streams and templates though.

PR #280, opened 2016-08-04
Assign app label for templates using name

spadgett:
Fixes https://github.com/openshift/origin/issues/8325
Uses the same strategy as new-app since https://github.com/openshift/origin/pull/10192

Note that the app label is read-only, but my understanding from @jwforres is we will have a separate follow on to allow overriding default label values more generally.

@jwforres PTAL
@bparees CC


jwforres:
[merge]

Issue #32, opened 2016-05-24
Too much gap between warning icon and reason in events table


spadgett:
![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/15517442/b646f368-21c5-11e6-90d7-f9d781c1626d.png)

@sg00dwin fyi, probably introduced by the fixed table layout

PR #283, opened 2016-08-04
Too much gap between warning icon and reason in events table

jhadvig:
In `_table.less` we are specifying table header width based on the position, but since we are showing different column based on the screen width, so I switched to ID selector.

@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/32

### DIFF
--- a/app/styles/_tables.less
+++ b/app/styles/_tables.less
@@ -145,13 +145,13 @@
 // table for events
 .events-table {
   table-layout: fixed;
-  th:nth-child(1) {width:90px;} // time
-  th:nth-child(2) {width:190px;} // name
-  th:nth-child(3) {width:160px;} // kind
-  th:nth-child(4) {width:10px;padding:0;} // icon
-  th:nth-child(5) {width:150px;} // reason
-  th:last-child   {width:100%;} // message; ensures it gets remaining width
-  td:nth-child(2),td:last-child {
+  th#time {width:90px;}
+  th#kind-name {width:190px;}
+  th#kind {width:160px;}
+  th#severity {width:10px;padding:0;}
+  th#reason {width:150px;}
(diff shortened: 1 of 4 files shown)


app/styles/_tables.less
```diff
@@ -145,13 +145,13 @@
...
-  th:nth-child(5) {width:150px;} // reason
-  th:last-child   {width:100%;} // message; ensures it gets remaining width
-  td:nth-child(2),td:last-child {
+  th#time {width:90px;} // time
+  th#kind-name {width:190px;} // name
+  th#kind {width:160px;} // kind
+  th#severity {width:10px;padding:0;} // icon
+  th#reason {width:150px;} // reason
+  th#message   {width:100%;} // message; ensures it gets remaining width
+  th#kind-name,th#message  {
```
spadgett: This looks much better. I think we can take out some of the comments since it's obvious from the ID.

jhadvig:
@spadgett comment addressed.

app/styles/_tables.less
```diff
@@ -145,13 +145,13 @@
...
-  th:nth-child(4) {width:10px;padding:0;} // icon
-  th:nth-child(5) {width:150px;} // reason
-  th:last-child   {width:100%;} // message; ensures it gets remaining width
-  td:nth-child(2),td:last-child {
+  th#time {width:90px;}
+  th#kind-name {width:190px;}
+  th#kind {width:160px;}
+  th#severity {width:10px;padding:0;}
+  th#reason {width:150px;}
+  th#message   {width:100%;}
```
spadgett: I'd leave the comment "ensures it gets remaining width" here.

jhadvig:
@spadgett sorry... havent noticed the are two comments actually.. my bad

spadgett:
Thanks, LGTM [merge]

PR #289, opened 2016-08-04
Clear insecureEdgeTerminationPolicy editing non-edge routes

### LINKED ISSUE #288: Error changing route from edge to passthrough
This occurs when the route has an insecureEdgeTerminationPolicy set. It's not properly cleared when switching to other termination types.

> An error occurred updating route foo.Route "foo" is invalid: spec.tls.insecureEdgeTerminationPolicy: Invalid value: "Redirect": InsecureEdgeTerminationPolicy is only allowed for edge-terminated routes

spadgett:
Fixes #288

@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/edit/route.js
+++ b/app/scripts/controllers/edit/route.js
@@ -106,6 +106,10 @@ angular.module('openshiftConsole')
 
           if (_.get($scope, 'routing.tls.termination')) {
             route.spec.tls = $scope.routing.tls;
+            if (route.spec.tls.termination !== 'edge') {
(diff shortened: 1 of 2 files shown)


spadgett:
@jwforres hold, looking again I think this is a problem for create as well.

spadgett:
Nevermind, it's not an issue.

https://github.com/spadgett/origin-web-console/blob/12ca5fa174cdaa5011f52df034bbfb5b8acfe569/app/scripts/services/applicationGenerator.js#L152-L154

jwforres:
[merge]

PR #298, opened 2016-08-05
Rewrite BC editor to be able to add GitHub and Generic webhook triggers

jhadvig:
So the reported issue contained two possible solutions, but I didn't liked the idea of diplaying the webhook URL, since user can not modify the URL anyway, just disable/enable the triggers, so there is no point of  showing the URL.
I've added the `Learn more` link and also the link for the configuration tab, although not sure if we want a redirection link besides the breadcrumbs. On the other hand it could come really handy if user would want to open the `Configuration` tab in a new tab.

Here are two drafts:
1#
screenshot-61

2#
screenshot-62

Personally I like the first one since the `Learn more` link is kinda weird next to the header.

@jwforres PTAL

Closes https://github.com/openshift/origin/issues/10098


spadgett:
If I click the configuration tab link, will I lose any edits I've made? Not sure we should encourage the user to navigate away while editing.

I wouldn't put the (i) icon in the help message since we don't in other places. I'd also just use a plain "Lean more" link under the help text to be consistent with other doc links. I'm thinking of something like this:

openshift_web_console

> So the reported issue contained two possible solutions, but I didn't liked the idea of diplaying the webhook URL, since user can not modify the URL anyway, just disable/enable the triggers, so there is no point of showing the URL.

Wondering if we shouldn't have a show webhook URL link somewhere to reveal it if I want to see. As I'm editing the build config, I might want to update the GitHub repo at the same time.

jhadvig:
screenshot-64

jhadvig:
> Wondering if we shouldn't have a show webhook URL link somewhere to reveal it if I want to see. As I'm editing the build config, I might want to update the GitHub repo at the same time.

What about creating having the link to the configuration tab, but force to open it in a new tab, if clicked ?

spadgett:
But if I just enabled the webhook, it won't be there in the new tab.

jhadvig:
True that. OK lets drop that idea. 

> Wondering if we shouldn't have a show webhook URL link somewhere to reveal it if I want to see. As I'm editing the build config, I might want to update the GitHub repo at the same time.

@spadgett you mean to only show the full URL of the webhook ?

spadgett:
I'm thinking of a "Show webhook URLs"  link that will show the copy to clipboard widget with the values directly in the editor.

jhadvig:
@spadgett so here is a preview of the updated BC editor.
screenshot-1

I know we talked to put close icon after the copy-to-clipboard widget but I've realized that if user will by any chance accidentally click on the `X` btn then we would get rid of the webhook and the only way how to undo that would be to Cancel from the editor. So instead I've put there `Remove` btn, which will change to `Undo` and will disable the widget, so if user accidentally remove the webhook, or just changes his mind he caneasily undo the change.
Thoughts ?

spadgett:
"Remove" looks good. It might have been weird with different button styles for copy and remove anyway if we went with the icon.

I wouldn't disable the copy to clipboard input for new values, and I'd always use "Remove" rather than "Undo".

jwforres:
I assumed the disabled with the Undo was after you clicked Remove, is it
not?

spadgett:
> I assumed the disabled with the Undo was after you clicked Remove

Oh must be, I misunderstood.

jhadvig:
@jwforres @spadgett correct after user hits the `Remove` btn the copy-to-clipboard widget will get disabled and `Remove` will change to `Undo`
Only the webhooks that wont be disabled will be save.

jhadvig:
@spadgett @jwforres I've updated the PR with the logic described in previous comments.
It's tested and no data loss shall occur. 
Although there are two things I would like to align on with you:
1. position of the `Learn more` link. Here are screens of link different positions:
**Right to the section header**
screenshot-2
 **Under to the section header**
screenshot-3
I know we discussed this with Sam to place it under the header, but I kinda like the one on the right from the header, since for the headers in different sections we have a tooltip help right to the header.

2. wasn't sure if we want to disable the copy-to-clipboard mechanism the widget is disabled, since user could copy the URL from the input field anyway, even if the widget is disabled. Then the disabling would signal that the webhook will be removed.

PTAL

jwforres:
i prefer to learn more to the right of the header personally.  disabling the copy link would be good if the changes to do it arent big

jwforres:
If the plan is for this PR to fix https://bugzilla.redhat.com/show_bug.cgi?id=1365650  then make sure you update the create from image flow, we still want a checkbox there to turn on the config change trigger, but it should say something like "Launch a build as soon as the build configuration is created."

jwforres:
or "Launch the first build when the build configuration is created."

jhadvig:
> If the plan is for this PR to fix https://bugzilla.redhat.com/show_bug.cgi?id=1365650 then make sure you update the create from image flow, we still want a checkbox there to turn on the config change trigger, but it should say something like "Launch a build as soon as the build configuration is created."

@jwforres not really sure what you mean by adding a checkbox that will add configChange trigger to BC, since in the  'create from image flow' a.k.a 'Deploy Image' tab, no BC will be created, only DC, IS and Service.

jwforres:
Not the Deploy Image tab, the Browse Catalog tab when you select a builder
image.  The "from source" flow

jwforres:
There is already a checkbox in the create from source flow that has the
wrong text on it based on what we know about the config change trigger now

jhadvig:
@jwforres ohh sorry my mistake. Will update as well. Thanks !

jhadvig:
@jwforres I've updated the PR and addressed the comments.
I've put the webhook edit login into a directive to reduce the size of _edit/build-config.html_ and _edit/buildConfig.js_ at least a bit.
Also for the mobile version I've added icons describing the action instead of the *Remove* and *Undo* word action, cause on the small screen the word actions would end up on next line. Screens
screenshot-6
screenshot-7


> i prefer to learn more to the right of the header personally. disabling the copy link would be good if the changes to do it arent big

unfortunately the changes would be need to be bigger, and I would prefer not to do them ATM

 PTAL

jhadvig:
Both... but thats just for screens with width 421px and smaller

spadgett:
@jhadvig You will have to click the buttons twice on iOS if we leave the tooltip. It's not a good experience. Can you use a `title` attribute instead?

jhadvig:
@spadgett yeah I would rather get rid of the tooltip as well, just didn't know how to replace it. Thanks for  hint !

app/scripts/constants.js
```diff
@@ -12,6 +12,7 @@ window.OPENSHIFT_CONSTANTS = {
     "cli":                     "https://docs.openshift.org/latest/cli_reference/overview.html",
     "get_started_cli":         "https://docs.openshift.org/latest/cli_reference/get_started_cli.html",
     "basic_cli_operations":    "https://docs.openshift.org/latest/cli_reference/basic_cli_operations.html",
+    "build-triggers":          "https://docs.openshift.org/latest/dev_guide/builds.html#build-triggers",
```
spadgett: @jwforres @sg00dwin FYI, we'll need to update the enterprise-3.3 branch as well for the new doc link

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -335,42 +333,56 @@ angular.module('openshiftConsole')
...
+          data: trigger
+        };
+      }
+      // Applies to image and config change triggers
+      function constructChangeTrigger(trigger, isPresent) {
+        return {
+          present: isPresent,
+          data: trigger
+        };
+      }
```
spadgett: I think these functions are simple enough I'd just inline them.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -671,49 +683,16 @@ angular.module('openshiftConsole')
...
-      }
-
+      var triggers = [].concat($scope.triggers.githubWebhooks, 
+                              $scope.triggers.genericWebhooks, 
+                              $scope.triggers.imageChangeTriggers,
+                              $scope.triggers.builderImageChangeTrigger,
+                              $scope.triggers.configChangeTrigger);
+      // Filter webhook triggers that are not disabled or imageChange triggers that are present
+      triggers = _.filter(triggers, function(trigger) {
+        return trigger.disabled === false || trigger.present;
```
```js
return !trigger.disabled || trigger.present;
```

app/scripts/directives/editWebhookTriggers.js
```diff
@@ -0,0 +1,33 @@
...
+      templateUrl: 'views/directives/edit-webhook-triggers.html',
+      controller: function($scope) {
+        $scope.addWebhookTrigger = function(type) {
+          var webhook = {
+            disabled: false,
+            data: {
+              type: type
+            }
+          };
+          webhook.data[type.toLowerCase()] = {
```
spadgett: I think I'd rather check for known types instead of assuming that `toLowerCase()` will always give the right property name.

app/scripts/directives/util.js
```diff
@@ -75,19 +75,21 @@ angular.module('openshiftConsole')
...
       restrict: 'E',
       scope: {
-        clipboardText: "="
+        clipboardText: "=",
+        isDisabled: "=?"
       },
       templateUrl: 'views/directives/_copy-to-clipboard.html',
       controller: function($scope) {
         $scope.id = _.uniqueId('clipboardJs');
+        $scope.isDisabled = ($scope.isDisabled) ? true : false;
```
spadgett: Is this needed?

app/styles/_core.less
```diff
@@ -1091,6 +1098,23 @@ a.disabled-link {
...
+    .trigger-url {
+      display: inline-block;
+      vertical-align: middle;
+    }
+    .trigger-actions {
+      display: inline-block;
+      margin-left: 5px;
+      .action-icon,
+      .action-icon:hover {
+        color: #4d5258;
```
spadgett:
```css
color: @color-pf-black-700;
```

app/views/directives/_copy-to-clipboard.html
```diff
@@ -1,9 +1,10 @@
 <div class="input-group copy-to-clipboard">
-  <input id="{{id}}" type="text" class="form-control" value="{{clipboardText}}" readonly select-on-focus>
+  <input id="{{id}}" type="text" class="form-control" value="{{clipboardText}}" ng-disabled="isDisabled" ng-readonly="!isDisabled" select-on-focus>
   <span class="input-group-btn" ng-hide="hidden">
-    <button data-clipboard-target="#{{id}}"
+    <a data-clipboard-target="#{{id}}"
+            ng-disabled="isDisabled"
```
spadgett: Does `ng-disabled` work on links? I didn't think it did.

app/views/directives/_copy-to-clipboard.html
```diff
@@ -1,9 +1,10 @@
...
             data-toggle="tooltip"
             title="Copy to clipboard"
-            class="btn btn-default"><i class="fa fa-clipboard"/></button>
+            class="btn btn-default"><i class="fa fa-clipboard"/></a>
```
spadgett: I'd also add `role="button"` for accessibility

app/views/directives/edit-webhook-triggers.html
```diff
@@ -0,0 +1,30 @@
+<h5>{{type}} webhooks
+  <span class="help action-inline">
+    <a href>
+      <i class="pficon pficon-help" data-toggle="tooltip" aria-hidden="true" data-original-title="{{typeInfo}}">
```
spadgett: I'd also add the text for screen readers in an sr-only span.

```html
<span class="sr-only">{{typeInfo}}</span>
```

spadgett: I'd also put the aria-hidden on the parent `a` element. Otherwise screen readers will read it as an empty link.

app/views/directives/edit-webhook-triggers.html
```diff
@@ -0,0 +1,30 @@
...
+  </span>
+</h5>
+<div ng-repeat="trigger in triggers">
+  <div class="trigger-info">
+    <span class="trigger-url">
+      <copy-to-clipboard is-disabled="trigger.disabled" clipboard-text="bcName | webhookURL : trigger.data.type : trigger.data[type.toLowerCase()].secret : projectName"></copy-to-clipboard>
+    </span>
+    <span class="visible-xs-inline trigger-actions">
+      <a href="" ng-if="!trigger.disabled" class="action-icon" ng-click="trigger.disabled = true; form.$setDirty()">
+        <span class="pficon pficon-close" aria-hidden="true" data-toggle="tooltip" data-original-title="Remove"></span>
```
spadgett: Add an sr-only span with the text for screenreaders

app/views/directives/edit-webhook-triggers.html
```diff
@@ -0,0 +1,30 @@
...
+      </a>
+      <a href="" ng-if="trigger.disabled" class="action-icon" ng-click="trigger.disabled = false">
+        <span class="fa fa-repeat" aria-hidden="true" data-toggle="tooltip" data-original-title="Undo"></span>
```
spadgett: Needs an sr-only span here too.

app/views/directives/edit-webhook-triggers.html
```diff
@@ -0,0 +1,30 @@
...
+    </a>
+  </span>
+</h5>
+<div ng-repeat="trigger in triggers">
+  <div class="trigger-info">
+    <span class="trigger-url">
+      <copy-to-clipboard is-disabled="trigger.disabled" clipboard-text="bcName | webhookURL : trigger.data.type : trigger.data[type.toLowerCase()].secret : projectName"></copy-to-clipboard>
+    </span>
+    <span class="visible-xs-inline trigger-actions">
+      <a href="" ng-if="!trigger.disabled" class="action-icon" ng-click="trigger.disabled = true; form.$setDirty()">
```
spadgett: role="button"

app/views/directives/edit-webhook-triggers.html
```diff
@@ -0,0 +1,30 @@
...
+        <span class="pficon pficon-close" aria-hidden="true" data-toggle="tooltip" data-original-title="Remove"></span>
+      </a>
+      <a href="" ng-if="trigger.disabled" class="action-icon" ng-click="trigger.disabled = false">
```
spadgett: role="button"

app/views/directives/edit-webhook-triggers.html
```diff
@@ -0,0 +1,30 @@
...
+        <span class="fa fa-repeat" aria-hidden="true" data-toggle="tooltip" data-original-title="Undo"></span>
+      </a>
+    </span>
+    <span class="hidden-xs trigger-actions">
+      <a href="" ng-if="!trigger.disabled" ng-click="trigger.disabled = true; form.$setDirty()">Remove</a>
+      <a href="" ng-if="trigger.disabled" ng-click="trigger.disabled = false">Undo</a>
```
spadgett: role="button" on these link as well

app/views/directives/edit-webhook-triggers.html
```diff
@@ -0,0 +1,30 @@
...
+    </span>
+  </div>
+</div>
+<span class="learn-more-inline checkbox">
+  <a href="" ng-click="addWebhookTrigger(type)">Add GitHub webhook</a>
```
spadgett: role="button"

spadgett: Shouldn't this be `Add {{type}} webhook`?

app/views/edit/build-config.html
```diff
@@ -518,56 +518,55 @@
                         </div>
                       </div>
                       <div class="section">
-                        <h3>Triggers</h3>
+                        <h3>Triggers
+                          <a href="{{'build-triggers' | helpLink}}" target="_blank"><span class="learn-more-inline">Learn more<i class="fa fa-external-link"></i></span></a>
```
spadgett: `aria-hidden="true"` on the icon

spadgett:
@jhadvig code looks much cleaner than before moving things into a directive. thanks

app/views/directives/edit-webhook-triggers.html
```diff
@@ -0,0 +1,30 @@
...
+    <a href>
+      <i class="pficon pficon-help" data-toggle="tooltip" aria-hidden="true" data-original-title="{{typeInfo}}">
+      </i>
+    </a>
+  </span>
+</h5>
+<div ng-repeat="trigger in triggers">
+  <div class="trigger-info">
+    <span class="trigger-url">
+      <copy-to-clipboard is-disabled="trigger.disabled" clipboard-text="bcName | webhookURL : trigger.data.type : trigger.data[type.toLowerCase()].secret : projectName"></copy-to-clipboard>
```
spadgett: Same about about `toLowerCase` as I made in the controller.

app/scripts/controllers/edit/buildConfig.js
jhadvig: We need to check if the `trigger.disabled` equals `false`, cause if the trigger wont have `disabled` field set  (eg. imageChange trigger) it will evaluate as `true`, which is not correct.

spadgett: Oh, I see. I think we need to make it more clear in that case. Something like

```js
if (_.has(trigger, 'disabled') && !trigger.disabled) {
  return true;
}
return trigger.present;
```

Then it's more obvious that you're only checking this value if it exists.

It might be better even to have a switch on the different types and only check the property (disabled or present) that we expect.

app/views/directives/_copy-to-clipboard.html
jhadvig: no it wont disable the link. It's just to graphically distinguish the widget in disabled and not-disabled state. we already spoke about it with @jwforres https://github.com/openshift/origin-web-console/pull/298#issuecomment-239510865

jhadvig:

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -335,42 +333,50 @@ angular.module('openshiftConsole')
...
         return imageChangeRef === builderRef;
       }
+      // Applies to webhook triggers
+      function constructWebhookTrigger(trigger, isDisabled) {
+        return {disabled: isDisabled, data: trigger};
+      }
+      // Applies to image and config change triggers
+      function constructChangeTrigger(trigger, isPresent) {
+        return {present: isPresent, data: trigger};
+      }
```
spadgett: Sorry, by "inline" I meant not have the functions, just create the objects below.

app/scripts/controllers/edit/buildConfig.js
spadgett: Can you add a comment explaining the `trigger.disabled === false` check?

spadgett:
thanks @jhadvig 

[merge]

Issue #286, opened 2016-08-04
Task list message has no top margin on some pages


spadgett:
![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/17410748/122626bc-5a43-11e6-8e29-9bf6c4f48445.png)

PR #300, opened 2016-08-05
Remove tasks tag from newfromtemplate page

jhadvig:
@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/286

### DIFF
--- a/app/views/newfromtemplate.html
+++ b/app/views/newfromtemplate.html
@@ -9,7 +9,6 @@
       <div id="scrollable-content" class="middle-container has-scroll">
         <div class="middle-content">
           <div class="container surface-shaded">
-            <tasks></tasks>
             <div class="row">
               <div class="col-md-12">
                 <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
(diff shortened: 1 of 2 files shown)


spadgett:
@jhadvig can you check if it's a problem on other pages?

jhadvig:
@spadgett checked it and we should be safe.
The only views where we display tasks are:
- newfromtemplate
- nextsteps
- overview (and old project.html)

spadgett:
Honestly it's a little strange we show it on newfromtemplate and not other pages. Wonder if we shouldn't just take it out. The right margin also looked wrong if you check the screenshot in #286.

I think we should either show it everywhere OR only on next steps and the overview. Leaning toward just removing it here. @jwforres

jhadvig:
+1 .. have it everywhere could be kinda confusing. Will remove this one

jwforres:
why does it show up here instead of going to the overview or to next steps
like we do for our other pages that are generating multiple things?

jwforres:
what actually causes this to show up here?

jhadvig:
@jwforres when creating app from template, on next you click on the breadcrumb with the template's name

jwforres:
oh then yeah just take it out

spadgett:
+1

jhadvig:
@spadgett update

spadgett:
Do you mind updating the commit message so it matches the actual change?

jhadvig:
@spadgett done :)

spadgett:
Thanks [merge]

PR #301, opened 2016-08-05
Don't warn about HPAs when a CPU limit is set

spadgett:
The request will default to the limit when the pod is created.

https://bugzilla.redhat.com/show_bug.cgi?id=1364309

@jwforres PTAL


jwforres:
[merge]

PR #302, opened 2016-08-05
Honor server sort for builder images

spadgett:
The server returns status tags ordered by semantic version (most recent first).

openshift_web_console

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1325069
Fixes https://github.com/openshift/origin/issues/7847

@jwforres PTAL


jwforres:
that was an easy enough fix :)   [merge]

PR #305, opened 2016-08-05
ImageStreamTag page when tag is from a DockerImage breaks

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1364355


jwforres:
@stefwalter can you take a look at this, was there a reason you were trying to follow the from.name yourself versus letting the API do it for you?  Letting the API handle it seems to work, if i go to the latest tag on one of our example image streams then it correctly pulls back the data for the tag that it actually points to.

stefwalter:
This fix looks correct. Thanks @jwforres

jwforres:
[merge]

PR #306, opened 2016-08-05
Annotate latest builder in catalog

### LINKED ISSUE #242: :Latest tag should state what version it is currently referring to
Jessica mentioned that in the future we will warn developers that :latest is a reference to the newest image we have available, but it is also unclear what version this is currently referring to. 

As an example, if I came to the web console as a PHP developer looking to deploy a REST API with Laravel, I might see php:latest and think it refers to a newer version of PHP than php:5.6. If nothing else this would not be entirely clear. We should tell people what version :latest is currently referring to in addition to warning users that the version will change.

spadgett:
Remove the :latest tile, but annotate the image stream tag that it tracks.

openshift_web_console

Fixes #242

@jwforres PTAL
@luciddreamz CC

### DIFF
--- a/app/scripts/controllers/create.js
+++ b/app/scripts/controllers/create.js
@@ -222,19 +222,41 @@ angular.module('openshiftConsole')
           return;
         }
 
-        // Create a map of spec tags so we can find them efficiently later when
-        // looking at status tags.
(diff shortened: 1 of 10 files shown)


jwforres:
[merge]

PR #310, opened 2016-08-08
Fix the calculation of ImageStreamTag name for lookup

stefwalter:
We were using the ImageStream spec tag 'from' field, when available
but this is not related to the ImageStreamTag name.

https://bugzilla.redhat.com/show_bug.cgi?id=1364355


stefwalter:
Duplicate of #305

spadgett:
Closing as #305 has merged.

PR #320, opened 2016-08-09
Bug 1365391: Add canI filter check to new buttons

jhadvig:
Here is the screen of `pipelines.html` when there are no pipelines and user doesn't have permissions to create anything. Basically it's similar to `overview.html`
screenshot

@jwforres PTAL

### DIFF
--- a/app/views/storage.html
+++ b/app/views/storage.html
@@ -22,7 +22,7 @@
           <div class="row">
             <div class="col-md-12">
               <div class="page-header page-header-bleed-right page-header-bleed-left">
-                <div class="pull-right">
+                <div class="pull-right" ng-if="project && ('persistentvolumeclaims' | canI : 'create')">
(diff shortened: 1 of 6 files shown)


app/views/storage.html
```diff
@@ -22,7 +22,7 @@
           <div class="row">
             <div class="col-md-12">
               <div class="page-header page-header-bleed-right page-header-bleed-left">
-                <div class="pull-right">
+                <div class="pull-right" ng-if="project && 'persistentvolumeclaims' | canI : 'create'">
```
spadgett: Does the filter need parens around it?

```
ng-if="project && ('persistentvolumeclaims' | canI : 'create')"
```

jhadvig: Tested it without them and it worked, but just to match other occurrences of it, I will add them.
Thanks !

jhadvig:
@spadgett updated :)

spadgett:
thanks @jhadvig [merge]

PR #328, opened 2016-08-09
Reduce scale debounce wait time

### LINKED ISSUE #322: Debounced scale calls need the leading option
In many cases you are only bumping the scale up or down by one, and it waits an entire second before making that request.  This is creating an artificial delay in how fast it seems like we can scale in some environments.  I think we should just add the "leading" option so it makes a request initially, and then debounces afterward.

spadgett:
Fixes #322

@jwforres PTAL. Per our conversation, I'm lowering the delay instead of using leading. Faster than 650ms resulted in too many separate requests in my testing.

### DIFF
--- a/app/scripts/directives/deploymentDonut.js
+++ b/app/scripts/directives/deploymentDonut.js
@@ -77,8 +77,8 @@ angular.module('openshiftConsole')
           }
         };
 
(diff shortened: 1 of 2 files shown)


jwforres:
[merge]

PR #330, opened 2016-08-09
Add request storage button to attach PVC empty state message

spadgett:
@jwforres PTAL

Not sure about the button placement relative to the back link. I used "Request Storage" to be consistent with the button on the storage page.

openshift_web_console


jwforres:
dont think i have a better suggestion for the back link [merge]

PR #331, opened 2016-08-09
Don't follow tile-target link when clicking on an icon inside an anchor

### LINKED ISSUE #312: Weird cancel behaviour on edit project
While on project listing page, if you edit a project, and then click cancel, you're sent to the project instead to back to project listing.
This behaviour is weird.
It is clear that is not a "back" button, but still don't look nice to be sent to a different page from the one you got into the edit project form.

spadgett:
Fixes #312

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/util.js
+++ b/app/scripts/directives/util.js
@@ -47,7 +47,7 @@ angular.module('openshiftConsole')
       link: function($scope, element) {
         $(element).click(function (evt) {
           var t = $(evt.target);
-          if (t && t.is('a')){
+          if (t && t.closest("a", element).length) {
             return;
           }
           $('a.tile-target', element).trigger("click");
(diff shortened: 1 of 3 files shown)


app/scripts/directives/util.js
```diff
@@ -50,6 +50,10 @@ angular.module('openshiftConsole')
           if (t && t.is('a')){
             return;
           }
+          // Also check if this is a child of an anchor element (e.g., an icon inside a link).
+          if (t && t.parent() && t.parent().is('a')) {
```
jwforres: this would be better http://api.jquery.com/closest/#closest-selector-context
```
t.closest("a", element)
```

it will look for the closest ancestor that is an `a` tag, but won't look farther up than the element (in this case the tile), then you can just see if the answer is empty

that will handle if you have a parent of a parent of a parent that is an a :)

spadgett:
@jwforres thanks updated

jwforres:
[merge]

spadgett:
[merge]

spadgett:
@MarkDeMaria this should fix the problem you were seeing with the kebab menu on the projects page, although might not matter if we switch to Patternfly list view

PR #337, opened 2016-08-10
Make copy to clipboard input readonly and select text when clicked on

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1365722

### DIFF
--- a/app/views/directives/_copy-to-clipboard.html
+++ b/app/views/directives/_copy-to-clipboard.html
@@ -1,5 +1,5 @@
-<div class="input-group">
-  <input id="{{id}}" type="text" class="form-control" value="{{clipboardText}}">
+<div class="input-group copy-to-clipboard">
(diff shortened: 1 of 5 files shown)


jwforres:
@spadgett PTAL

app/views/directives/_copy-to-clipboard.html
```diff
@@ -1,5 +1,5 @@
-<div class="input-group">
-  <input id="{{id}}" type="text" class="form-control" value="{{clipboardText}}">
+<div class="input-group copy-to-clipboard">
+  <input id="{{id}}" type="text" class="form-control" value="{{clipboardText}}" readonly onclick="this.select()">
```
spadgett: Recommend using directive `select-on-focus`

jwforres:
[merge]

PR #338, opened 2016-08-10
Use relative start times in metrics requests

spadgett:
Requires https://github.com/hawkular/hawkular-metrics/pull/567

/cc @jwforres @mwringe


jwforres:
Change LGTM

spadgett:
Verified that the relative timestamps work with openshift/origin-metrics-hawkular-metrics:latest

[merge]

PR #339, opened 2016-08-10
Warn in delete dialog when we can't cascade

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1365582

@jwforres PTAL


jwforres:
[merge]

Issue #266, opened 2016-08-03
Need tech preview designation for pipelines


jwforres:
Once https://github.com/openshift/origin/pull/10163 merges you will have to hunt to find your password, we should direct them to where they can find it.  This is temporary until we have oauth integration in jenkins.

spadgett:
sorry @jwforres and @bparees, this is my fault

bparees:
have to?  they *get to*.  this is a feature..... :)

jwforres:
I'll start looking at this one

PR #341, opened 2016-08-10
Tech preview designations for pipelines, disable pipeline menu item by default

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/266

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -115,6 +115,7 @@
 .breadcrumb {
   font-size: @font-size-base - 1;
   padding-bottom: 0;
+  margin-bottom: 0;
   strong {
     font-weight: 600;
   }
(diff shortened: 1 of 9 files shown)


jwforres:
@spadgett PTAL

app/styles/_core.less
```diff
@@ -115,6 +115,7 @@
 .breadcrumb {
   font-size: @font-size-base - 1;
   padding-bottom: 0;
+  margin-bottom: 0;
```
jwforres: the bottom margin on breadcrumbs didn't actually have an effect with the h1 as the next element, and looks fine when the alerts show up

spadgett: Won't this change the margin under the breadcrumbs on non-pipeline pages?

spadgett: > the bottom margin on breadcrumbs didn't actually have an effect with the h1 as the next element

Does the "Loading..." message have margin?

spadgett:
[merge]

PR #342, opened 2016-08-10
Issue #167 - Keep Terminal Tab Despite 0 Containers Available

### LINKED ISSUE #167: Show empty state instead of hiding terminal tab
Currently we are hiding the terminal tab rather than showing the empty state message (as we do with logs) when unavailable. 

@MarkDeMaria created the issue :)

MarkDeMaria:
Fixes #167 

Should the number of containers available for a pod drop to 0, the terminal tab will switch to a Patternfly empty state table stating such.

### DIFF
--- a/app/scripts/controllers/pod.js
+++ b/app/scripts/controllers/pod.js
@@ -43,6 +43,8 @@ angular.module('openshiftConsole')
       message: "This terminal has been disconnected. If you reconnect, your terminal history will be lost."
     };
 
+    $scope.noContainersYet = true;
+
     // Must always be initialized so we can check the selectedTab elsewhere
     $scope.selectedTab = {};
     // Check for a ?tab=<name> query param to allow linking directly to a tab.
@@ -140,7 +142,7 @@ angular.module('openshiftConsole')
         }
         else {
           $(window).on('resize.terminalsize', _.debounce(calculateTerminalSize, 100));
-        }        
+        }
         $timeout(calculateTerminalSize, 0);
       }
       else {
@@ -191,6 +193,12 @@ angular.module('openshiftConsole')
       return terminals;
     };
 
+    var updateContainersYet = function(pod) {
+      if ($scope.noContainersYet) {
+        $scope.noContainersYet = $scope.containersRunning(pod.status.containerStatuses) === 0; 
+      }
+    };
+
     var updateTerminals = function(terminals) {
       _.each(terminals, function(term) {
         var thisContainerStatus = _.find($scope.pod.status.containerStatuses, { name: term.containerName });
@@ -220,6 +228,7 @@ angular.module('openshiftConsole')
             ImageStreamResolver.fetchReferencedImageStreamImages(pods, $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, context);
 
             $scope.containerTerminals = makeTerminals();
+            updateContainersYet(pod);
 
             // If we found the item successfully, watch for changes on it
             watches.push(DataService.watchObject("pods", $routeParams.pod, context, function(pod, action) {
@@ -234,6 +243,7 @@ angular.module('openshiftConsole')
               setContainerVars();
 
               updateTerminals($scope.containerTerminals);
+              updateContainersYet(pod);
             }));
           },
           // failure
--- a/app/views/browse/pod.html
+++ b/app/views/browse/pod.html
@@ -128,77 +128,84 @@
                   <uib-tab
                     active="selectedTab.terminal"
                     select="terminalTabWasSelected = true"
-                    ng-init="containers = pod.status.containerStatuses"
-                    ng-if="containersRunning(pod.status.containerStatuses) > 0">
+                    ng-init="containers = pod.status.containerStatuses">
                       <uib-tab-heading>Terminal</uib-tab-heading>
 
-                      <div class="mar-bottom-md">
-                        <span class="pficon pficon-info" aria-hidden="true"></span>
-                        When you navigate away from this pod, any open terminal connections will be closed.
-                        This will kill any foreground processes you started from the terminal.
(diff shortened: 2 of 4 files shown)


app/views/browse/pod.html
```diff
@@ -122,73 +122,84 @@
...
+                                  >
+                                </kubernetes-container-terminal>
+                              </div>
                             </div>
                           </div>
                         </div>
                       </div>
-    	            </uib-tab>
+
+                      <div ng-if="containersRunning(pod.status.containerStatuses) <= 0" class="blank-slate-pf table-log-pods" >
```
spadgett: Can you switch to use the `empty-state-message` class we use elsewhere until we switch everything to `blank-slate-pf`?

Here's an example we use for the debug terminal:

https://github.com/openshift/origin-web-console/blob/e6249b0435b9fea2401cbc7297de16d1cdcbe172/app/views/modals/debug-terminal.html#L11-L28

spadgett: `table-log-pods` ?

spadgett:
Will the current terminal still disappear when the container I'm connected to stops?

MarkDeMaria:
@spadgett 

I revisited my logic and made this work like a one-way switch. The terminal tab will always appear, but if there are no containers running at page-load, then an empty state message will display:
image

If at any point any containers (> 0) come online, the empty state will be replaced by the normal terminal tab contents and will remain there, even if all containers crash again. The logic being that the user may have interacted with a terminal prior to its crash and we want to preserve its contents for inspection.

benjaminapetersen:
still `needs-rebase`, betting something not quite right. seeing some resize stuff in there.

MarkDeMaria:
@benjaminapetersen Fixed my rebase bork.

benjaminapetersen:
@MarkDeMaria hmm, still think that some older code won.  For example, the `<environment>` directive has been replaced by the `<key-value-editor>` directive, but the PR shows the other way around.

app/scripts/controllers/pod.js
```diff
@@ -234,6 +236,10 @@ angular.module('openshiftConsole')
               setContainerVars();
 
               updateTerminals($scope.containerTerminals);
+
+              if ($scope.containersRunning(pod.status.containerStatuses) > 0 ) {
```
jwforres: you need this logic after the makeTerminals call here https://github.com/openshift/origin-web-console/pull/342/files#diff-7f99f2d99896dd6aa2719df2a46a2acfR224

Otherwise you could get a flash of the "No containers running" message between the initial GET succeeding which triggers loaded=true and the list call that watch creates which will then trigger the noContainersYet = false to show the terminal.

app/views/browse/pod.html
```diff
@@ -3,7 +3,7 @@
 
     <!-- Middle section -->
     <div class="middle-section">
-      <div class="middle-container">
+      <div id="scrollable-content" class="middle-container has-scroll">
```
jwforres: remove this diff, bad merge

app/views/browse/pod.html
```diff
@@ -65,13 +65,7 @@
...
                     <div ng-repeat="container in pod.spec.containers">
                       <h3>Container {{container.name}} Environment Variables</h3>
-                      <key-value-editor
-                        entries="container.env"
-                        is-readonly
-                        cannot-add
-                        cannot-sort
-                        cannot-delete
-                        ng-if="container.env.length"></key-value-editor>
+                      <environment env-vars="container.env" ng-if="container.env.length"></environment>
```
jwforres: remove this diff, bad merge

MarkDeMaria:
@jwforres @benjaminapetersen 

Appropriate fixes applied.

Jessica, as far as the location of noContainersYet being updated, did you mean to put it in both places or just one (Underneath the makeTerminals call?)

app/scripts/controllers/pod.js
```diff
@@ -191,6 +193,12 @@ angular.module('openshiftConsole')
       return terminals;
     };
 
+    var updateContainersYet = function(pod) {
+      if ($scope.containersRunning(pod.status.containerStatuses) > 0 ) {
+        $scope.noContainersYet = false;
```
jwforres: Instead of an if check that sets a boolean, just make this a one liner

```
$scope.noContainersYet = $scope.containersRunning(pod.status.containerStatuses) === 0; 
```

app/scripts/controllers/pod.js
```diff
@@ -220,6 +228,7 @@ angular.module('openshiftConsole')
             ImageStreamResolver.fetchReferencedImageStreamImages(pods, $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, context);
 
             $scope.containerTerminals = makeTerminals();
+            updateContainersYet(pod);
```
jwforres: Yep calling this in both places is what I meant.

MarkDeMaria:
@jwforres Changes made.

app/scripts/controllers/pod.js
```diff
@@ -191,6 +193,10 @@ angular.module('openshiftConsole')
       return terminals;
     };
 
+    var updateContainersYet = function(pod) {
+      $scope.noContainersYet = $scope.containersRunning(pod.status.containerStatuses) === 0;
```
jwforres: hmm as it is now i think this is going cause the terminals to disappear if you had already connected terminals that were deleted under the covers.  think in the suggested updates it might have messed up your previous logic.  can you re-test that scenario?

benjaminapetersen: If it would be helpful, ping me for an overview of our e2e test setup to automate this.  Might be helpful to ensure you catch all the cases.  Ive been on a "we should test more" kick and am willing to hash that out a bit. :smile:

MarkDeMaria:
@jwforres

jwforres:
[merge]

Issue #316, opened 2016-08-08
Env var name moves to bottom row when pasted instead of typed


spadgett:
To reproduce:

1. Edit the env vars for a resource like a deployment config
2. Paste in a name value from the clipboard
3. Tab to the value field

Result:

The pasted value moves to the new row below

![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/17490245/e23b3228-5d6f-11e6-8418-4a7af7b9e7dc.png)

@benjaminapetersen Let me know if you can't reproduce.
@jwforres CC

PR #347, opened 2016-08-11
Update version of key-value-editor, fixes paste bug

benjaminapetersen:
Fixes issue #316

Can now paste into the kve & it will land in the correct input.

@jwforres @spadgett


spadgett:
[merge]

PR #348, opened 2016-08-11
Break long identifiers in events table cells

### LINKED ISSUE #346: Regression: long identifiers overflowing events table
screen shot 2016-08-11 at 9 18 22 am

spadgett:
Fixes #346

@jwforres PTAL

### DIFF
--- a/app/styles/_tables.less
+++ b/app/styles/_tables.less
@@ -151,8 +151,8 @@
   th#severity {width:10px;padding:0;}
   th#reason {width:150px;}
   th#message   {width:100%;} // ensures it gets remaining width
-  th#kind-name,th#message  {
-    .word-break;
+  td {
+    .word-break();
   }
   .pficon {
     vertical-align: middle;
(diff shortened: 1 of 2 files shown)


app/styles/_tables.less
```diff
@@ -151,8 +151,8 @@
   th#severity {width:10px;padding:0;}
   th#reason {width:150px;}
   th#message   {width:100%;} // ensures it gets remaining width
-  th#kind-name,th#message  {
-    .word-break;
```
jwforres: do we really want to word-break all of the td's or just the message td?

spadgett: I was thinking we'd want it on all table cells. It only breaks mid-word if there is no natural break.

jwforres: is that true on all the browsers? I'm pretty sure our word-break behaves slightly different in different browsers because it uses the "best" option for each browser

spadgett: No we switched it not to do that a while back. There is a `word-break-all` that behaves differently.

spadgett: See https://github.com/openshift/origin/pull/7883

jwforres: [merge]

spadgett: @stevekuznetsov should the bot have seen this merge comment?

stevekuznetsov: Negative

spadgett:
[merge]

PR #350, opened 2016-08-11
Show empty state message if there is only a DC and nothing else

### LINKED ISSUE #334: Empty overview when you have a deployment config, but no services
I have a DC, BC, and running build. No services or pods, except the builder pod. The overview was blank for me.

screen shot 2016-08-09 at 3 00 53 pm

cc @jwforres

spadgett:
A lonely deployment config with no services or replication controllers will result in an empty overview.

Fixes #334
@jwforres PTAL


jwforres:
[merge]

PR #351, opened 2016-08-11
Ensure DC,BC can display alt text for valueFrom in kve

### LINKED ISSUE #292: Ensure valueFrom in key value editor is visible for DC,BC,RC
Gonna go ahead and make this issue, then run through each scenario & make sure we are able to see `valueFrom`'s. From memory talking with @jwforres, @rhamilto had something working for this but I don't know if it got merged off the top of my head.

benjaminapetersen:
Fix #292 
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1366457

Ensures that there is a readonly alt text available for BCs/DCs that have env vars with a `valueFrom` attribute.  Example:
screen shot 2016-08-11 at 3 28 17 pm

The `$filter` for this is using the same logic that exists in the `<environment>` directive.

@jwforres @spadgett


app/scripts/filters/util.js
```diff
@@ -528,6 +528,31 @@ angular.module('openshiftConsole')
...
+  // valueFrom attribute instead of a value.
+  .filter('altTextForValueFrom', function() {
+    return function(obj) {
+      if(obj.value) {
+        return;
+      }
+      _.each(obj.valueFrom, function(val, key) {
+        console.log(key);
+        if(key === 'configMapKeyRef') {
+          obj.valueAlt = 'Set to the key ' + key + ' in config map ' + val.name;
```
spadgett: Should this be `value.key` here and for secrets?

spadgett:
Unfortunately we lose the help text that we had before

app/scripts/filters/util.js
```diff
@@ -528,6 +528,31 @@ angular.module('openshiftConsole')
...
+  })
+  // currently used to set an alt text value for an env var that has a
+  // valueFrom attribute instead of a value.
+  .filter('altTextForValueFrom', function() {
+    return function(obj) {
+      if(obj.value) {
+        return;
+      }
+      _.each(obj.valueFrom, function(val, key) {
+        console.log(key);
```
jwforres: remove the console.log

jwforres:
and you confirmed if you update other env vars and submit the changes that the things using valueFrom arent getting corrupted?

app/scripts/filters/util.js
```diff
@@ -528,6 +528,31 @@ angular.module('openshiftConsole')
...
+        if(key === 'configMapKeyRef') {
```
jwforres: seems like a switch/case would be better here

app/scripts/filters/util.js
benjaminapetersen: ah will do

app/scripts/filters/util.js
benjaminapetersen: Should it?  The environments directive had this:
```html
<span ng-switch-when="secretKeyRef">
  Set to the key <b>{{value.key}}</b> in secret <b>{{value.name}}</b>.
</span>
```
I can make that change, perhaps to both places then?

spadgett: Isn't that what I just suggested? :)

benjaminapetersen:
@spadgett meaning this text in the `data-content` attrib:
```html
<span class="fa fa-external-link-square" style="cursor: help;" data-toggle="popover" data-trigger="hover" data-content="This is a referenced value that will be generated when a container is created.  On running pods you can check the resolved values by going to the Terminal tab and echoing the environment variable."></span>
```
Hmm... might be good to figure out a way to get that in here somewhere...

app/scripts/filters/util.js
spadgett:
```js
obj.valueAlt = 'Set to the key ' + val.key + ' in config map ' + val.name;
```

spadgett: I think maybe `from` might be a better var name than `val` here.

app/scripts/filters/util.js
benjaminapetersen: can do

app/scripts/filters/util.js
```diff
@@ -528,6 +528,31 @@ angular.module('openshiftConsole')
...
+    };
+  })
+  // currently used to set an alt text value for an env var that has a
+  // valueFrom attribute instead of a value.
+  .filter('altTextForValueFrom', function() {
+    return function(obj) {
+      if(obj.value) {
+        return;
+      }
+      _.each(obj.valueFrom, function(val, key) {
```
spadgett: Expanding on my comment about names, I'd suggest something like

```js
_.each(obj.valueFrom, function(from, refType) { ...
```

then below you have lines like

```js
if (refType === 'configMapKeyRef') { ... }
```

and

```js
obj.valueAlt = 'Set to the key ' + from.key + ' in config map ' + from.name;
```

etc

benjaminapetersen:
@jwforres yup, edited, reordered, deleted, saved, refreshed, yaml editor shows everything still intact:
```yaml
env:
   -
     name: MONGODB_USER
     value: 11111
   -
     name: FAKE_THING
     valueFrom: {configMapKeyRef: {name: abcd, key: 12345 }}
   -
     name: MONGODB_DATABASE
     value: 22222
   -
     name: MONGODB_ADMIN_PASSWORD
     value: 33333
   -
     name: MONGODB_PASSWORD
     value: 44444
   -
     name: fdsafsa
     value: dddd
   -
     name: stuff_and_things
     value: yup
```

I have an example with `configMapKeyRef`, but not the other options.  Have been looking through the [V3 Testfiles](https://github.com/openshift-qe/v3-testfiles) repo for other useful tests.

app/scripts/filters/util.js
```diff
@@ -528,6 +528,31 @@ angular.module('openshiftConsole')
...
       var uri = new URI(url);
       var protocol = uri.protocol();
-      return uri.is('absolute') && (protocol === 'http' || protocol === 'https');      
+      return uri.is('absolute') && (protocol === 'http' || protocol === 'https');
+    };
+  })
+  // currently used to set an alt text value for an env var that has a
+  // valueFrom attribute instead of a value.
+  .filter('altTextForValueFrom', function() {
+    return function(obj) {
```
spadgett: Might also name it `envVar` instead of `obj` unless we expect it to work for other things

benjaminapetersen:
@spadgett cool, I'll update to `from, refType` in the `_.each()`

app/scripts/filters/util.js
benjaminapetersen: Sorry, I meant also in the `<environment>` directive "as well".

spadgett: What I mean is that what you have doesn't match what's in the environment directive. See my example above for what I think it should be.

app/scripts/filters/util.js
benjaminapetersen: Yeah, I wasn't sure if there might be other similar concepts that could use it.  I'll update.

app/scripts/filters/util.js
```diff
@@ -528,6 +528,30 @@ angular.module('openshiftConsole')
...
+  // valueFrom attribute instead of a value.
+  .filter('altTextForValueFrom', function() {
+    return function(envVar) {
+      if(envVar.value) {
+        return;
+      }
+      _.each(envVar.valueFrom, function(from, refType) {
+        switch(refType) {
+          case 'configMapKeyRef':
+            envVar.valueAlt = 'Set to the key ' + refType + ' in config map ' + from.key;
```
spadgett:
```js
envVar.valueAlt = 'Set to the key ' + from.key + ' in config map ' + from.name;
```

app/scripts/filters/util.js
benjaminapetersen: Oop, you are right. Tinkered a bit, looks like this now:

screen shot 2016-08-11 at 5 05 18 pm

I believe thats right/consistent with your suggestion.

spadgett: Yup, thanks :)

benjaminapetersen:
@spadgett still looking at the help text stuff

benjaminapetersen:
New screenshot:
screen shot 2016-08-11 at 7 15 56 pm

Not sure if we want to use the secret agent icon for all of them though.  Makes sense for `secretKeyRef`, but what about for `configMapKeyRef`, `fieldRef` and the default?  Default could be no icon, but then we lose the tooltip help text.

jwforres:
Just use our standard help tooltip icon if there is a tooltip but no
special icon set?

benjaminapetersen:
Using `pficon pficon-help` for everything but `secretKeyRef` (left secretKeyRef the secret agent icon, as above screenshot):
screen shot 2016-08-12 at 10 26 19 am

benjaminapetersen:
[TEST]

benjaminapetersen:
dist, re [test]

app/scripts/filters/util.js
```diff
@@ -528,6 +528,36 @@ angular.module('openshiftConsole')
...
+      }
+      if(!envVar.valueFrom) {
+        return;
+      }
+      envVar.valueIcon = 'pficon pficon-help';
+      envVar.valueIconTooltip = 'This is a referenced value that will be generated when a container is created.  On running pods you can check the resolved values by going to the Terminal tab and echoing the environment variable.';
+      _.each(envVar.valueFrom, function(from, refType) {
+        switch(refType) {
+          case 'configMapKeyRef':
+            envVar.valueAlt = 'Set to the key ' + from.key + ' in config map ' + from.name;
```
jwforres: hmm are you modifying the object we got back from data service or a clone? and we arent actually sending any of these decorated values to the server on a save are we

benjaminapetersen: This filter is modifying whatever it gets.  We've been cloning everything so far & merging in `save()` functions in the controllers.  However, I have not been cleaning off the extra keys.  I've been checking via the YAML editor & the extra keys have been ignored thus far.  

Shall I create a filter to clean the env vars before save of any extra keys on the objects?

spadgett: @benjaminapetersen if you prefix extra key with a `$`, angular won't serialize them back out. then you don't need to worry about it

spadgett: We shouldn't send extra stuff to the server even if it's ignored today. What if a `valueAlt` field is added to the API in the future?

benjaminapetersen: Updated to use [a util function](https://github.com/openshift/angular-key-value-editor/blob/master/src/services/key-value-editor-utils.js#L98) to clean up the entries before saving.  The fn deletes only "known keys" to the kve, and will drop an entry if it has no key or value (since its possible to do that the way it creates new rows).

@jwforres @spadgett

benjaminapetersen:
Also, this PR bumps the version on kve a few points and should fix issue #1366457 (bugzilla) when merged

app/scripts/filters/util.js
spadgett: Curious if there's a reason you prefer this to just prefixing kve properties with a `$`. It seems like it's a bit more error prone trying to clean them up this way.

jwforres: agree with sam here, why not use what angular gives us already

spadgett: Sorry it needs to be two dollar signs, `$$`

https://docs.angularjs.org/api/ng/function/angular.toJson

benjaminapetersen: Oh, because these are indended for a user to manipulate.  I think `$` & `$$` are sort of book keeping prefixes.  IE, I intend a user to be able to set the `valueAlt` or the `valueIcon`, so having them set `$$valueAlt` or `$$valueIcon` seems dirty...

benjaminapetersen: Agree though, is def the right call to fix this now before its an issue!

spadgett: I feel like the problem is we're using the same object to hold both the data and the metadata.

I'd rather copy the keys we know we want into a new object then try to remove keys. Does the editor add additional properties?

benjaminapetersen: I believe I have them all doc'd here: https://github.com/openshift/angular-key-value-editor/blob/master/src/directives/key-value-editor.js#L59

So I think it makes sense for the editor to give a util for cleaning up its metadata.  For us in the web console, also think it makes sense if we prefer an approach where we copy only what we known we want (ie, name, value, valueFrom properties). I can make us a filter for that & update these references.

spadgett: That's a lot of properties. Are we using them all?

I'd rather not have to add extra properties into my data to configure the editor and then call a utility to remove them. Not saying we should change the editor API for this issue, but it's something we might look at going forward.

spadgett:
[merge]

app/scripts/filters/util.js
benjaminapetersen: It is, I believe we have used most.  Conceptually I've been thinking of this control a bit like `data` -> `metadata for UI` -> `data`.  Ie, convert the data to a list of metadata for controlling the UI, interact with it, then convert it back to data when done.  

I'm good with a function that would copy relevant keys rather than delete irrelevant keys.  I def prefer the "purity" of input/output over the direct manipulation.

benjaminapetersen: Shall I make a follow-up issue, or just keep it in mind for now?

Issue #279, opened 2016-08-03
Extra border between header and table on the Storage page


jwforres:
![storageborder](https://cloud.githubusercontent.com/assets/1467629/17377828/8f6df588-5989-11e6-92e5-ed7ae920d488.png)

PR #354, opened 2016-08-12
Differ between page-header and section-header

jhadvig:
@jwforres @sg00dwin PTAL

Closes: https://github.com/openshift/origin-web-console/issues/279

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -53,6 +53,11 @@
       margin-top: 0;
     }
   }
+  .section-header{
+    .page-header();
+    border-bottom: none;
+    padding-bottom: 10px;
+  }
 }
 @media (min-width: @screen-md-min) {
   .console-os {
--- a/app/views/storage.html
+++ b/app/views/storage.html
@@ -21,7 +21,7 @@
         <div class="container-fluid">
           <div class="row">
             <div class="col-md-12">
-              <div class="page-header page-header-bleed-right page-header-bleed-left">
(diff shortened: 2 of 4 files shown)


app/styles/_core.less
```diff
@@ -53,6 +53,15 @@
       margin-top: 0;
     }
   }
+  .section-header{
```
jwforres: would prefer not to re-do all these margins. Its still the page header so i would still expect it to have the page-header class.  Are the only differences dropping the bottom border and adding(changing?) the bottom padding?

app/styles/_core.less
```diff
@@ -53,6 +53,15 @@
...
+    margin-bottom: 0;
+    margin-left: -20px;
+    margin-right: -20px;
+    margin-top: 20px;
+    padding-left: 20px;
+    padding-right: 20px;
+    padding-bottom: 9.5px;
```
jwforres: try not to use half pixels, they behave differently across devices / browsers

app/styles/_core.less
jwforres: sorry nevermind about the page header, forgot storage had multiple headers, i still hate to re-do all the margins though, could we do something like

```
.section-header {
  .page-header();
  border-bottom: none;
  padding-bottom: 10px; // or 9px whatever it needs to be
}
```

jwforres: you'll also need to check all the mobile margins

jwforres:
[merge]

PR #358, opened 2016-08-12
Don't add `openshift` to the istag selector twice

### LINKED ISSUE #357: Runtime error on add to project page if user is part of openshift namespace
This happens when we build the namespace menu on the "Deploy Image" tab since we insert "openshift" always. The same item appears twice.

```
angular.js:11706 Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: namespace in $select.items, Duplicate key: string:openshift, Duplicate value: openshift
http://errors.angularjs.org/1.3.20/ngRepeat/dupes?p0=namespace%20in%20%24select.items&p1=string%3Aopenshift&p2=openshift
    at angular.js:63
    at ngRepeatAction (angular.js:24772)
    at Object.$watchCollectionAction [as fn] (angular.js:14279)
    at Scope.$digest (angular.js:14412)
    at Scope.$apply (angular.js:14675)
    at done (angular.js:9725)
    at completeRequest (angular.js:9915)
    at XMLHttpRequest.requestLoaded (angular.js:9856)
```

spadgett:
Fixes #357

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/istagSelect.js
+++ b/app/scripts/directives/istagSelect.js
@@ -29,7 +29,11 @@ angular.module("openshiftConsole")
         $scope.isNamesByNamespace = {};
 
         DataService.list("projects", {}, function(projectData) {
-          $scope.namespaces = ['openshift'].concat(_.keys(projectData.by('metadata.name')).sort());
+          var sortedNamespaces = _.keys(projectData.by('metadata.name')).sort();
+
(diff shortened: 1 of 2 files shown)


jwforres:
[merge]

PR #361, opened 2016-08-12
Shouldnt show Idle text if RC is scaled up, even when annotation exists

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1366115


jwforres:
@spadgett PTAL

spadgett:
[merge]

PR #362, opened 2016-08-12
Add "Cancel deployment" action to overview

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1365666

openshift_web_console

openshift_web_console

### DIFF
--- a/app/scripts/controllers/modals/confirmModal.js
+++ b/app/scripts/controllers/modals/confirmModal.js
@@ -13,12 +13,14 @@ angular.module('openshiftConsole')
                                                  $uibModalInstance,
                                                  message,
                                                  details,
-                                                 buttonText,
-                                                 buttonClass) {
+                                                 okButtonText,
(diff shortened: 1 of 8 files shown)


spadgett:
@jwforres PTAL

jwforres:
im just trying to decide if it needs the full "Cancel deployment" text for the link, what if it just said "cancel"  lower case so it doesnt seem so visually important and flows better with the previous text ?

jwforres:
otherwise this LGTM

spadgett:
Here it is with lowercase cancel for comparison

openshift_web_console

jwforres:
i definitely prefer the cancel instead of cancel deployment, could go
either way on Cancel vs cancel i think, will leave that up to you

spadgett:
Slight preference for Cancel so it's consistent with other action links in the web console

spadgett:
[merge]

PR #366, opened 2016-08-15
Always show services with dependents as top-level services on overview

spadgett:
Fixes a bug where if

  Service A -> Service B -> Service C

Service C would not appear on the overview.

@jwforres PTAL


jwforres:
[merge]

Issue #365, opened 2016-08-15
Environment variable editor in Environment tab for BC and DC should check permissions


jhadvig:
In the environment variable editor in Environment tab for BC and DC user is able to add envVars but after hitting the `Save` btn, permission error will pop up:
![screenshot-5](https://cloud.githubusercontent.com/assets/1668218/17662870/fb2d9d06-62e8-11e6-8de0-81bfb3aa6f12.png)

Short term solution would be to have two `key-value-editor` tags and for both check the permissions.

Long term solution would be for isReadlony attr in `key-value-editor` to take the boolean: https://github.com/openshift/angular-key-value-editor/issues/78

PR #367, opened 2016-08-15
Check is user can update envVars + replace environment tags for key-value-editor

jhadvig:
Update the old `environment` tags for the `key-value-editor` + checking if user has permissions to update the BC or DC to determin if the `key-value-editor` should be rendered as readonly
@jwforres @benjaminapetersen PTAL

Closes https://github.com/openshift/origin-web-console/issues/365

### DIFF
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -390,17 +390,26 @@
                     <uib-tab-heading>Environment</uib-tab-heading>
                     <h3>Environment Variables</h3>
                     <ng-form name="bcEnvVars">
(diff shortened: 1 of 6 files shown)


jwforres:
[merge]

benjaminapetersen:
+1 😄 

Yeah, that ng-if stuff, sorry I haven't solved that one yet.  Hopefully can fix in near future, but gonna link the issue you opened:
https://github.com/openshift/angular-key-value-editor/issues/78

jhadvig:
@benjaminapetersen thanks ! :)

Issue #364, opened 2016-08-12
Tiles can overlap in overview


spadgett:
![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/17636178/a87d7bda-60a9-11e6-9e4d-22fe3fc65ca9.png)

/cc @jwforres

spadgett:
This was at about ~1000px wide

PR #369, opened 2016-08-15
Move port nowrap from around entire block to individual info so that it doesn't cause issue

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/364


sg00dwin:
@jwforres @spadgett PTAL

spadgett:
[merge]

Issue #344, opened 2016-08-11
Impossible to add environment variable starting with "app" like application


jorgemoralespou:
I'm deploying an image from an imagestream and trying to create an environment variable called "application". When I start typing and reach "app" the field goes gray and not editable.

I'm using 1.3.0-alpha.3

![image](https://cloud.githubusercontent.com/assets/78350/17582806/ccb802a6-5fad-11e6-9ad4-9b7f7c3104c9.png)

benjaminapetersen:
Ah, yeah, for labels we made "app" and "template" readonly :frown:

This is somewhat of a duplicate of this bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1362047, which we will address by eliminating the readonly state on these two "special" labels

PR #370, opened 2016-08-15
Have a separate read-only table for system labels on create flows and allow overriding them

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/344
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1362047

### DIFF
--- a/app/scripts/controllers/create/createFromImage.js
+++ b/app/scripts/controllers/create/createFromImage.js
@@ -41,6 +41,8 @@ angular.module("openshiftConsole")
       }
     ];
 
+    var appLabel = {name: 'app', value: ''};
+
     ProjectsService
       .get($routeParams.project)
       .then(_.spread(function(project, context) {
@@ -77,6 +79,7 @@ angular.module("openshiftConsole")
             portOptions: []
           };
           scope.labels = [];
+          scope.systemLabels = [appLabel];
           scope.annotations = {};
           scope.scaling = {
             replicas: 1,
@@ -183,6 +186,9 @@ angular.module("openshiftConsole")
 
         $scope.$watch('scaling.autoscale', checkCPURequest);
         $scope.$watch('container', checkCPURequest, true);
+        $scope.$watch('name', function(newValue) {
+          appLabel.value = newValue;
+        });
 
         initAndValidate($scope);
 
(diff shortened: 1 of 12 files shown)


jwforres:
@spadgett PTAL

jwforres:
note the update to the app generator test spec since we no longer expect app generator to define the app label

app/scripts/controllers/create/createFromImage.js
```diff
@@ -77,6 +77,8 @@ angular.module("openshiftConsole")
             portOptions: []
           };
           scope.labels = [];
+          scope.appLabel = {name: 'app', value: ''};
```
spadgett: nit: could be a local var rather than a scope var since it's not used in the view

app/scripts/directives/deployImage.js
```diff
@@ -95,7 +98,7 @@ angular.module("openshiftConsole")
 
           $scope.$watch('app.name', function() {
             _.set(
-              _.find($scope.labels, function(label) {
+              _.find($scope.systemLabels, function(label) {
```
spadgett: Not new, but this could be

```js
_.find($scope.systemLabels, { name: 'app' })
```
```

File: app/scripts/controllers/newfromtemplate.js [hunk 3]
```diff
@@ -176,7 +177,10 @@ angular.module('openshiftConsole')
 
         $scope.createFromTemplate = function() {
           $scope.disableInputs = true;
-          $scope.template.labels = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries($scope.labels));
+          var userLabels = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries($scope.labels));
+          var systemLabels = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries($scope.systemLabels));
+          $scope.template.labels = _.extend(systemLabels, userLabels);
```
spadgett: This mutates `systemLabels`, but I don't think it's a problem since `mapEntries` and `compactEntries` probably create a copy.

spadgett:
LGTM

spadgett:
Travis is complaining about a dist mismatch

app/scripts/controllers/newfromtemplate.js
jwforres: yeah they are converting from an array to a map

jwforres:
yeah forgot ben updated the kve

jwforres:
[merge]

PR #371, opened 2016-08-15
Fix bug flowing data to metrics chart

spadgett:
The logic around trimming data points from the front of the chart was incorrect since we switched to relative timestamps.

Fixes https://github.com/openshift/origin/issues/10419
@jwforres PTAL


spadgett:
This also makes the pod and deployment charts consistent in number of data points and how often they update.

spadgett:
[merge]

PR #374, opened 2016-08-16
Improve service linking dialog

### LINKED ISSUE #285: Streamline simple service grouping scenarios
Now, if I have 2 services and I want to group them it makes me select the 2nd one. It would be good/nice if user didn't have to expand drop down and select the only option.

Some possible ways to improve:
 * if only 2, don't prompt and just group
 * if only 2, have dialog show and prefill with only option...user just can hit ok

Ideally some of the services would be hidden in larger deployments, like pipelines, which give me all the jenkins gunk

spadgett:
* Autofocus the ui-select dropdown so you can just start typing
* Pre-select the service if there's only one to link

Fixes #285

@jwforres PTAL
@sspeiche CC


jwforres:
[merge]

PR #375, opened 2016-08-16
Don't show all events for builds when build pod annotation not set

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1367306

@jwforres PTAL


jwforres:
[merge]

PR #379, opened 2016-08-16
Don't show metrics for builds on monitoring page

spadgett:
The charts usually are empty or contain only one or two data points. We
could add more helpful metrics at some point (like maximum memory and
CPU used during the build), but currently the charts aren't useful.

@jwforres PTAL


jwforres:
[merge]

PR #380, opened 2016-08-16
Change terminal command to set the TERM env

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1367337

### DIFF
--- a/app/scripts/controllers/pod.js
+++ b/app/scripts/controllers/pod.js
@@ -128,16 +128,18 @@ angular.module('openshiftConsole')
       });
     };
 
-    if (!characterBoundingBox.height || !characterBoundingBox.width) {
-      Logger.warn("Unable to calculate the bounding box for a character.  Terminal will not be able to resize.");
-    }
-    else {
-      $(window).on('resize.terminalsize', _.debounce(calculateTerminalSize, 100));
-    }
-
     $scope.$watch('selectedTab.terminal', function(terminalTabSelected) {
       if (!!terminalTabSelected) {
+        if (!characterBoundingBox.height || !characterBoundingBox.width) {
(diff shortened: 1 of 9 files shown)


jwforres:
@spadgett PTAL

kubectl / oc rsh command defaults to xterm today if it doesnt find a different TERM env in your cli

jwforres:
this makes things like `clear` actually behave the way you expect

spadgett:
```
bower not-cached    https://github.com/kubernetes-ui/container-terminal.git#1.0.0
bower resolve       https://github.com/kubernetes-ui/container-terminal.git#1.0.0
bower ENORESTARGET  No tag found that was able to satisfy 1.0.0
```

spadgett:
term.js is an xterm clone, so TERM=xterm looks right

spadgett:
Should we update debug-terminal.html as well?

jwforres:
ok updated. I also made the change to not listen for resize till the terminal tab is selected

app/scripts/controllers/pod.js
```diff
@@ -128,17 +128,19 @@ angular.module('openshiftConsole')
...
+        if (!characterBoundingBox.height || !characterBoundingBox.width) {
+          Logger.warn("Unable to calculate the bounding box for a character.  Terminal will not be able to resize.");
+        }
+        else {
+          $(window).on('resize.terminalsize', _.debounce(calculateTerminalSize, 100));
+        }        
         $timeout(calculateTerminalSize, 0);
       }
+      else {
+        $(window).off('resize.terminalsize');
```
spadgett: Do we still need the `off()` call in destroy? What if I navigate away with the terminal tab selected?

spadgett:
[merge]

Issue #273, opened 2016-08-03
Blue border can extend past bottom of tile in no deployments message


spadgett:
![screen shot 2016-08-03 at 11 24 19 am](https://cloud.githubusercontent.com/assets/1167259/17371186/fa4d1f4a-596c-11e6-9fbe-ecd8d5570d1e.png)

/cc @sg00dwin @jwforres

spadgett:
Hm, it's not just the no deployments message.

![screen shot 2016-08-04 at 2 25 00 pm](https://cloud.githubusercontent.com/assets/1167259/17413822/c5568efe-5a50-11e6-8319-46b3a85872fe.png)

jhadvig:
Looked at this one and fixing the issue on the first screen https://github.com/openshift/origin-web-console/issues/273#issue-169160315 is quite simple, just removing `min-height`s in
https://github.com/openshift/origin-web-console/blob/master/app/styles/_overview.less#L157-L161

The second one would be a bit complicated.. one solution would be to compare the heights of services and swap them. But not sure if thats the ideal solution though :confused:

PR #382, opened 2016-08-16
Fix for equal height tiles and no deployment blocks

sg00dwin:
fixes https://github.com/openshift/origin-web-console/issues/273

This fixes 90% of the scenarios by setting min-heights for .overview-services and .no-deployment-block

To fix this so that each grows/shrinks for all scenarios using flex, then the underlying markup structure for no deployment blocks will need to be different from the general service-group blocks since fundamentally they behave differently.


sg00dwin:
@spadgett PTAL 
@jwforres

spadgett:
[merge]

PR #391, opened 2016-08-16
Let user clear changes to Environment tab forms

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1364015

Also fixes the pristine state on these forms after a Save, and now properly does a clone for RCs so thats its not dirtying the object we get back from data service.


jwforres:
@spadgett

spadgett:
[merge]

PR #394, opened 2016-08-16
Changes to show deployment connector arrow at mobile

sg00dwin:
Moved right arrow from markup to css, so that it can be switched from down to right arrow and still be semantically correct.

Refactored the css so that it's mobile first.

screen shot 2016-08-16 at 5 18 17 pm

### DIFF
--- a/app/styles/_overview.less
+++ b/app/styles/_overview.less
@@ -368,10 +368,33 @@
         .text-center();
       }
     }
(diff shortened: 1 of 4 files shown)


sg00dwin:
@spadgett @jwforres PTAL

jwforres:
[merge]

PR #409, opened 2016-08-17
[WIP] Add ng-aria

benjaminapetersen:
`<key-value-editor>` accessibility research spawned some curiosity.  WIP as I have only read up and installed it, still tinkering.

@spadgett feel like I remember you mentioning this in the past.  I'll look in the origin repo and see if there was an issue or something.

@jwforres 

(not proposing to merge this week)

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -94,6 +94,7 @@
     <script src="bower_components/angular-animate/angular-animate.js"></script>
     <script src="bower_components/angular-touch/angular-touch.js"></script>
     <script src="bower_components/angular-route/angular-route.js"></script>
+    <script src="bower_components/angular-aria/angular-aria.js"></script>
     <script src="bower_components/angular-bootstrap/ui-bootstrap-tpls.js"></script>
     <script src="bower_components/angular-drag-and-drop-lists/angular-drag-and-drop-lists.js"></script>
     <script src="bower_components/lodash/lodash.js"></script>
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -11,6 +11,7 @@
  */
 angular
   .module('openshiftConsole', [
+    'ngAria',
     'ngAnimate',
     'ngCookies',
     'ngResource',
@@ -416,6 +417,13 @@ angular
       .otherwise({
         redirectTo: '/'
       });
+  })
+  .config(function($ariaProvider) {
+    $ariaProvider.config({
+      tabindex: false,
+      bindRoleForClick: false,
+      bindKeydown: false
+    });
   })
   .constant("API_CFG", _.get(window.OPENSHIFT_CONFIG, "api", {}))
   .constant("APIS_CFG", _.get(window.OPENSHIFT_CONFIG, "apis", {}))
--- a/bower.json
+++ b/bower.json
@@ -12,6 +12,7 @@
     "angular-animate": "1.5.11",
     "angular-touch": "1.5.11",
     "angular-route": "1.5.11",
+    "angular-aria": "1.5.11",
     "angular-bootstrap": "0.14.3",
     "angular-patternfly": "3.21.0",
     "uri.js": "1.18.0",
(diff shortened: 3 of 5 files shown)


spadgett:
+1 looks like a good change (for post 1.3)

jwforres:
yep seems reasonable to me

benjaminapetersen:
The checklist for accessibility originally had this, per @spadgett on [this comment](https://github.com/openshift/origin/issues/5988#issuecomment-160721193).

benjaminapetersen:
[test]
Rebased this one.  Working through the UI everything seems in order, no funny business yet.
@jwforres @spadgett

jwforres:
looking back at the list of what it adds, my biggest concerns are with ngClick and ngDblClick because its automatically adding tabindex to things.  I don't think we want that.  Example, it would make our clickable tiles be in the keyboard tabindex, but we have links within the tiles which link to the same thing.  We don't want both the tile and the link to be in the tabindex.

Unfortunately it doesn't look like you can tell it to ignore individual instances of ng-click and instead have to globally turn off of those particular features, see https://docs.angularjs.org/guide/accessibility

liggitt:
you *really* don't want to get in the business of managing tabindex if you can at all help it

benjaminapetersen:
Agreed on managing tabindex, lets not do that.

So I believe setting `tabindex="-1"` will basically act as a "ignore" for ng-Aria (and [eliminates its tab focus ability](https://www.w3.org/TR/html5/editing.html#sequential-focus-navigation-and-the-tabindex-attribute)).  If we hate that (I kinda do), we can make an attribute directive `<aria-ignore>` that could set that attribute for us behind the scenes (and we won't forget why we set it) on whatever element we use it with. 

I can add that to this PR and do a round of lots of tabbing through the UI.  Feel free to chime in with any other known tab disasters :)

benjaminapetersen:
[Some chatter @ the angular repo about this issue](https://github.com/angular/angular.js/issues/14602).

jwforres:
@liggitt are you suggesting we should let ng-aria manage it for us, or that
we shouldn't let ng-aria muck with the tabindex?

liggitt:
The latter

jwforres:
its not just the tabindex, its also adding a keypress event and
role=button.  I think I would be more comfortable just disabling the
ngClick support altogether, there are three flags you disable, the details
were in that developer guide i linked before

benjaminapetersen:
Works for me, I'll update this.

benjaminapetersen:
Added `$ariaProvider.config()`, turned off a couple things.  Inspecting source I can def see the additional `aria-*` attributes added to elements.  Tabbing around the interface (before enabling, after enabling) it seems seamless.  So far I haven't found any quirky interactions.

benjaminapetersen:
looks like a flake, [TEST]

spadgett:
> ECONFLICT Unable to find suitable version for angular

benjaminapetersen:
Hmm. neat. will fix.

benjaminapetersen:
matched `1.3.20` w/the other angular core modules, should be good to go this time. 
[TEST]

benjaminapetersen:
Fail on:
```
PhantomJS 2.1.1 (Linux) ERROR
  Error: [$injector:modulerr] Failed to instantiate module openshiftConsole due to:
  [$injector:modulerr] Failed to instantiate module ngAria due to:
  [$injector:nomod] Module 'ngAria' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
  http://errors.angularjs.org/1.3.20/$injector/nomod?p0=ngAria
```
Hmm. angular-aria installed, in `bower.json`, in `index.html`, dependency in `app.js`...
[TEST] again Rosie while I go fish around...

benjaminapetersen:
rebased

benjaminapetersen:
rebased, includes the config block:
```javascript
  .config(function($ariaProvider) {
    $ariaProvider.config({
      tabindex: false,
      bindRoleForClick: false,
      bindKeydown: false
    });
  })
```

benjaminapetersen:
tinkering with tests...

benjaminapetersen:
rebased

PR #413, opened 2016-08-17
Bug fix where IE11 collapsed the page contents when the scrollbar appears

rhamilto:
There is no need to turn off scrolling and then re-enable it as all
pages should scroll by default.  Per @sg00dwin, this is a remnant of a
previous design where the page header was fixed.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1363592

@jwforres, PTAL


jwforres:
This breaks Follow and Go to Top on logs both in the Logs tab and in the chromeless view

rhamilto:
D'oh.  Sorry.  I completely forgot to test that after getting caught up helping @sg00dwin with the overview animation.  Lemme take another swing at this.

spadgett:
@jwforres `log-viewer` needed to scroll `#container-main` instead of `#scrollable-content` after this change. I swept the code, and there are no other references to `scrollable-content`, so I removed the ID.

rhamilto:
@jwforres, PTAL. 

Note:  this highlights [an existing bug where empty containers are taking up vertical space at the bottom of the overview page when no deployments yet exist](https://github.com/openshift/origin-web-console/issues/421).  I think it went unnoticed before because the background was gray.  Now it's very obvious.  I discussed it with @spadgett this morning, and we think #421 is the lesser of the two evils.

spadgett:
@rhamilto I'll take a look at #421

jwforres:
[merge]

PR #417, opened 2016-08-18
Last 2 kve link changes

benjaminapetersen:
@jwforres 

(which ever of these auto updates bugzilla:)
Fix #1366457 
Fixes #1366457


spadgett:
[merge]

spadgett:
Wondering if the editor should fall back to a generic "Add value" if `add-row-link` is missing.

benjaminapetersen:
@spadgett +1, if `add-row-link` doesn't have a value like `add-row-link="Add something"` it will fall back to "add row".

Issue #418, opened 2016-08-18
At narrower screen widths, there's no enough room for deployment arrow


spadgett:
![screen_shot_2016-08-18_at_9_53_43_am](https://cloud.githubusercontent.com/assets/1167259/17776317/f8014590-6529-11e6-91b0-19d901f5ef65.png)

/cc @jwforres

PR #419, opened 2016-08-18
Adjust position of deployment arrow at narrow width to be top aligned with the donuts

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/418

@spadgett PTAL


spadgett:
[merge]

PR #420, opened 2016-08-18
Updating akve to v2.4.1 to resolve Bootstrap variable override issue

rhamilto:
@jwforres or @spadgett, PTAL.

fyi, @benjaminapetersen


spadgett:
@rhamilto thank you for fixing this

[merge]

spadgett:
[merge]

benjaminapetersen:
Ah neat.  Prob we could/should just pull in the built css from kve in `index.html` rather than the less in our `main.less` file, but this works!

PR #422, opened 2016-08-18
Limit number of open websockets in IE

jwforres:


jwforres:
@spadgett PTAL

I went ahead and polled HPAs on the deployment page just in case since we are right at the limit.  There is still an API issue with closing websockets for log endpoints specifically https://github.com/openshift/origin/issues/10449#issuecomment-240865153

app/scripts/filters/util.js
```diff
@@ -560,4 +560,19 @@ angular.module('openshiftConsole')
...
+    return function() {
+      return isIE;
+    };
+  })
+  .filter('isEdge', function() {
+    var ua = navigator.userAgent;
+    var isEdge = /chrome.+? edge/i.test(ua);
+    return function() {
+      return isEdge;
+    };
```
spadgett: I'm fine with this, although we used a constant for iOS:

https://github.com/openshift/origin-web-console/blob/master/app/scripts/app.js#L319

dist/scripts/vendor.js
```diff
@@ -40680,7 +40680,7 @@ s.write(d(b));
 }
 j && (j = !1, m.addClass("hidden"), n.addClass("hidden"), s.cursorHidden = !1, s.showCursor(), s.refresh(s.y, s.y), g.autofocus && s.element && s.element.focus());
 }, p.onclose = function(b) {
-g.status = "disconnected", a(b.reason);
+console.log("websocket closed", b), g.status = "disconnected", a(b.reason);
```
spadgett: Looks like you have a stray console log in the dist

spadgett:
Just one minor comment. LGTM

jwforres:
hmm yeah i dont want to have to add the constant to the injection list for each browser i want to check, maybe in 3.4 we should add something to the Constants service instead for all of these

[merge]

Issue #423, opened 2016-08-19
Blue border can extend past bottom of tile


spadgett:
![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/17813024/97cb7cd0-65f7-11e6-8abe-9a43f9142030.png)

Related to #273

rhamilto:
a quick, dirty hack to fix the white edges bug is to add

```
  .middle { // hack to quickly, safely fix a safari bug
    background-color: red;
  }
```

to [line 18 of _overview.less](https://github.com/openshift/origin-web-console/blob/master/app/styles/_overview.less#L18).

spadgett:
How does that affect pages with a white background?

rhamilto:
@spadgett, it doesn't as it is specific to the overview.

spadgett:
Ah OK, didn't see it was under `.overview`. I think we should go ahead with that change. Can you open a PR?

Issue #421, opened 2016-08-18
Empty divs taking up vertical space on overview when no deployments yet exist


rhamilto:
the appearance of the overview with #413 
![screen shot 2016-08-18 at 3 58 22 pm](https://cloud.githubusercontent.com/assets/895728/17789878/14461c84-6562-11e6-9f73-4c9c3d0b3241.PNG)

the offending markup
![screen shot 2016-08-18 at 9 11 05 am](https://cloud.githubusercontent.com/assets/895728/17789883/1896d29c-6562-11e6-86ce-a6a8e99e62c4.PNG)

rhamilto:
This bug only occurs when the empty containers are scrolled off the page (that is, the viewport is shorter than the page contents).

spadgett:
I'm able to reproduce by unchecking all triggers creating from source.

![screen shot 2016-08-19 at 8 55 28 am](https://cloud.githubusercontent.com/assets/1167259/17810377/caf82b10-65ea-11e6-8784-94af6eb9af20.png)

spadgett:
OK, the bug I'm seeing is a little different and impacts a lot of pages. The `surface-shaded` style doesn't go to the bottom of the page in many cases. After a certain point, there's a white background:

![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/17810793/445dfaaa-65ed-11e6-938d-2c6bf4ccc680.png)

rhamilto:
@spadgett, possible browser caching issue?  I can only reproduce on the overview when there aren't any deployments...

spadgett:
Did you pull latest master? Only seeing it since the IE has scroll fix

rhamilto:
@spadgett, yup.  Latest commit in my local master is https://github.com/openshift/origin-web-console/commit/4d411df7541262b055757656f1bc843576c9f75f

spadgett:
Ah, my fault. I had a local change...

rhamilto:
@spadgett, phew!

I did discover we've got a couple of different visual defects on the overview page in Safari.  paging @sg00dwin?

<img width="1233" alt="screen shot 2016-08-19 at 10 32 06 am" src="https://cloud.githubusercontent.com/assets/895728/17813229/7afc223e-65f8-11e6-9dc7-d134b953ae6c.PNG">

spadgett:
I had opened #423, but it doesn't cover everything in the screenshot :/

rhamilto:
[...]

PR #425, opened 2016-08-19
Band aid fix to address gray background issues on overview

rhamilto:
Partially addresses #421 and #423 

@spadgett, PTAL


spadgett:
[merge]

PR #430, opened 2016-08-19
Swap environment table for kve readonly

benjaminapetersen:
Just about missed this `<environment>` directive hidden in the collapsible toggle on next steps.  One thing that I dislike here though is seeing "value" placeholder where they values are actually empty.  

screen shot 2016-08-19 at 1 46 04 pm

screen shot 2016-08-19 at 1 46 33 pm

@jwforres


spadgett:
> One thing that I dislike here though is seeing "value" placeholder where they values are actually empty.

I think we already have this problem already. If I set a env var on a DC with no value, the placeholder is shown.

benjaminapetersen:
Yup. :/

benjaminapetersen:
Updated with new kve that hides placeholder in readonly mode:

screen shot 2016-08-19 at 3 27 06 pm

[test]

jwforres:
[merge]

PR #445, opened 2016-08-22
kve bump, validation fix on edit build config

### LINKED ISSUE #431: Env var editor headers still visible when last var removed
When I remove the last variable, the headers are still present, which looks a little strange. Should `x` on the only row should just clear the inputs? Or maybe we hide the headers when there are no values.

screen shot 2016-08-19 at 3 23 54 pm

/cc @jwforres

benjaminapetersen:
Fix #431

bugzilla:
Fix #1368917 
Fix issue #1368917

- Fixes a 'clean' on the env list in edit buildConfig
- Bumps KVE up a couple notches:
   - Fixes missed ID change resulting in some failed validation feedback
   - adds a guard against a runtime error in a `$watch`
   - ensures the last inputs do not disappear on delete if using `add-row-link` attrib

@jwforres @spadgett


jwforres:
[merge]

PR #454, opened 2016-08-23
Fix unequal service header heights for multi-service routes in Firefox

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1369404

Here is what was causing the unequal heights. Adding additional right margin to the icon fixes it.

screen shot 2016-08-23 at 9 22 47 am

@jwforres PTAL


jwforres:
[merge]

PR #456, opened 2016-08-23
Fix swallowed keys in BC editor

benjaminapetersen:
Eliminates the `$filter('envVarsPair')` which was swallowing duplicate keys when converting array to hash, and is probably no longer needed since we are not using `<osc-key-values>`.

@jwforres @spadgett @jhadvig


benjaminapetersen:
[test]

jwforres:
[merge]

PR #457, opened 2016-08-23
Don't show debug action for completed pods or build pods

spadgett:
@jwforres PTAL

We might move this logic into the controller, but I didn't want to do that right before 1.3.


jwforres:
[merge]

PR #462, opened 2016-08-23
Stop passing containerName on deployment logs

jwforres:

PR #464, opened 2016-08-23
Don't send container parameter when requesting build logs

spadgett:
See also #462

@jwforres PTAL
@liggitt CC


jwforres:
[merge]

PR #465, opened 2016-08-23
Update kve, fix bug #1369315

benjaminapetersen:
Bugzilla:
Fix #1369315
Fix issue #1369315

Tested in Chrome, FF, Safari, IE via browserstack, should be good on the inputs.
@jwforres @spadgett


jwforres:
[merge]

PR #466, opened 2016-08-23
Restore service-group-header and deployment-tile left border style

spadgett:
Fixes the order the styles are declared so as not to accidentally override the left border width.

@jwforres


jwforres:
[merge]

PR #469, opened 2016-08-24
Fix bug #1367713 - JMX/Threads console is blank

astefanutti:


spadgett:
@astefanutti The index.html segment was removed to fix another bug. See

https://github.com/openshift/origin/commit/8c3be50fbd7df6c5410b51b19772826ffeb8b822

Have you been able to build and verify this fix?

astefanutti:
@spadgett I've tested it on Chrome, didn't know about the previous Safari issue. Let me test it on Safari as well...

astefanutti:
@spadgett the thing is that with openshift/origin@8c3be50 the trailing slash is removed which causes the error.

spadgett:
I think a URI.js update might be normalizing the URL to remove the trailing slash.

spadgett:
@astefanutti This might work?

```js
            var targetURI = new URI().path(BaseHref)
                                     .segment('java')
                                     // Add an empty segment to force a trailing slash, which is required.
                                     .segment('')
                                     .hash(token)
                                     .query({
                                       jolokiaUrl: jolokiaUrl,
                                       title: title,
                                       returnTo: returnTo
                                     });
```

astefanutti:
@spadgett thanks. Let me try that out...

spadgett:
@astefanutti Can you verify that the font icons load properly as well? They aren't loading for me.

astefanutti:
@spadgett adding an empty segment seems to work. I've just repushed with it.

On my local environment, font icons load correctly on Chrome but not on Safari.

jwforres:
@astefanutti even in chrome we see it trying to load the font icon from ../components/ path which doesn't exist

app/scripts/extensions/javalink/javaLink.js
```diff
@@ -66,7 +66,9 @@
             var title = container.name || 'Untitled Container';
             var token = AuthService.UserStore().getToken() || '';
             var targetURI = new URI().path(BaseHref)
-                                     .segment('java/') // Must have a trailing slash to avoid runtime errors in Safari
+                                     .segment('java')
+                                     // Add an empty segment to force a trailing slash, which is required.
+                                     .segment('/')
```
spadgett: I believe this should be

```js
.segment('')
```

See https://github.com/medialize/URI.js/commit/2b604927f997299789c50badadbbb295c3bab754

jwforres:
@astefanutti are you running it embedded in the openshift binary? because we get a release of openshift-jvm with everything packaged into a zip and thats what gets embedded into openshift, and in that setup the fonts are in a /fonts directory

astefanutti:
@spadgett my mistake (still works though). I've just repushed.

The original regression may have been introduced with  medialize/URI.js@0f8e416ee649cda2c76d34fd26e7702beb82258a.

astefanutti:
@jwforres no, I'm running it locally.

jwforres:
This is what gets released to us from you guys:

```
curl -s https://codeload.github.com/hawtio/openshift-jvm/tar.gz/v1.0.45-build | tar -xz -C openshift-jvm --strip-components=1
```

And in that zip if you look at what the style.css is referencing for the fontawesome font, its pointing to a path that doesn't exist.  Looks like it has the same problem with the glyphicons for bootstrap.

jwforres:
@astefanutti so looking back in the history we can see there use to be two copies of the fontawesome CSS, one had the correct paths and one had the incorrect paths, between v1.0.44 and v1.0.45 the duplicate copy was taken out, but the one that was taken out was the one with the correct paths.  I'm going to open a separate PR that rolls openshift back to 1.0.42 which is what we were at before @gashcrumb bumped the version last week.  Unless there is a strong reason for us to update to a later version?

jwforres:
[merge] for this fix

astefanutti:
@jwforres I've just created hawtio/openshift-jvm#4 that should correct the path and add a couple of other missing resources. Once @gashcrumb feedback is OK, I presume he'll do a release, as v1.0.45 contains a number of fixes and improvements compared to v1.0.42.

PR #471, opened 2016-08-24
Trim prefilled name for deploy image to 24 characters

### LINKED ISSUE #470: Deploy image can prefill a name that is too long
If you search for an image name that is longer than 24 characters, the web console will prefill that value as the name. It's longer than the maxlength on the input, however, and the create button will be disabled.

/cc @jwforres

spadgett:
Fixes #470

@jwforres PTAL


jwforres:
[merge]

PR #473, opened 2016-08-24
Do not add log lines for empty messages

liggitt:
Found when debugging https://github.com/openshift/origin/issues/10601

An initial ping frame with 0-length data is being sent when streaming log messages. That should not result in an empty line. Actual empty log lines will at least contain a "\n"


spadgett:
LGTM

liggitt:
[merge]

PR #479, opened 2016-08-25
Dont check bcEnvForms is pristine on pipeline page

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1370079

### DIFF
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -134,7 +134,7 @@ angular.module('openshiftConsole')
               }
               $scope.buildConfig = buildConfig;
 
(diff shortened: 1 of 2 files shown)


jwforres:
@spadgett PTAL  when the bcEnvVars form doesn't exist it just falls back to the behavior that was happening before the update detection was added

spadgett:
LGTM. We might follow up later to avoid doing the copy / ensure env altogether for pipeline builds, but this seems like the safer fix for 1.3.

jwforres:
agreed, i didnt want to get into that change right now in case it had an effect on editing the pipeline

[merge]

PR #482, opened 2016-08-25
Override the java console origin logo for OCP

jwforres:


spadgett:
Is this a manual update we'll need to make whenever the Java console changes?

spadgett:
I guess it'll be a merge conflict we have to resolve when rebasing the enterprise branch?

dist.java/java/logo-banner-override.css
```diff
@@ -0,0 +1,8 @@
+.navbar-brand img {
```
spadgett: Wonder if we shouldn't inline these styles in index.html to prevent ever accidentally deleting this file from a hack script or grunt task.

jwforres:
@spadgett for reference https://github.com/openshift/ose/blob/enterprise-3.2/hack/install-assets.sh#L53

tweaked the command to add the extra '' because the existing sed commands blow up on OSx, thats what we kept hitting using the install-assets from the enterprise repo

spadgett:
Thanks, I think this is better. LGTM

Issue #13, opened 2016-05-16
openshift shared resource namespace is actually configurable


jwforres:
The 'openshift' namespace is actually configurable in the master config OpenShiftSharedResourceNamespace

We need to be passing that configuration through the dynamic config.js and using that as the namespace we pull from, and look for anywhere else that we make assumptions about what namespace that is.

jwforres:
cc @spadgett

PR #483, opened 2016-08-25
Refactoring of BC editor IST pickers

jhadvig:
@spadgett here is the PR.
Will probably do some additional method wording changes, testing etc... so no hurry with the review :)

@jwforres fyi

### DIFF
--- a/app/scripts/controllers/edit/buildConfig.js
+++ b/app/scripts/controllers/edit/buildConfig.js
@@ -15,9 +15,6 @@ angular.module('openshiftConsole')
     $scope.emptyMessage = "Loading...";
     $scope.sourceURLPattern = SOURCE_URL_PATTERN;
     $scope.options = {};
-    $scope.builderOptions = {};
-    $scope.outputOptions = {};
-    $scope.imageSourceOptions = {};
     $scope.jenkinsfileOptions = {
       type: 'path'
     };
@@ -26,34 +23,18 @@ angular.module('openshiftConsole')
       ImageStreamImage: "Image Stream Image",
       DockerImage: "Docker Image Repository"
     };
+
     $scope.buildFromTypes = [
-      {
-        "id": "ImageStreamTag",
-        "title": "Image Stream Tag"
-      },
-      {
-        "id": "ImageStreamImage",
-        "title": "Image Stream Image"
-      },
-      {
-        "id": "DockerImage",
-        "title": "Docker Image Repository"
-      }
+      "ImageStreamTag",
+      "ImageStreamImage",
+      "DockerImage"
     ];
     $scope.pushToTypes = [
-      {
-        "id": "ImageStreamTag",
-        "title": "Image Stream Tag"
-      },
-      {
-        "id": "DockerImage",
-        "title": "Docker Image Repository"
-      },
-      {
-        "id": "None",
-        "title": "--- None ---"
-      }
+      "ImageStreamTag",
+      "DockerImage",
+      "None"
     ];
+
     $scope.jenkinsfileTypes = [{
       "id": "path",
       "title": "From Source Repository"
@@ -93,16 +74,12 @@ angular.module('openshiftConsole')
       title: "Edit"
     });
 
-    $scope.buildFrom = {
-      projects: [],
-      imageStreams: [],
-      tags: {},
-    };
-    $scope.pushTo = {
-      projects: [],
-      imageStreams: [],
-      tags: {},
-    };
+    $scope.imageOptions = {
+      from: {},
+      to: {},
+      fromSource: {}
+    }
+
     $scope.sources = {
       "binary": false,
       "dockerfile": false,
@@ -125,7 +102,6 @@ angular.module('openshiftConsole')
       "Parallel",
       "SerialLatestOnly"
     ];
-    $scope.availableProjects = [];
 
     AlertMessageService.getAlerts().forEach(function(alert) {
       $scope.alerts[alert.name] = alert.data;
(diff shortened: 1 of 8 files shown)


jhadvig:
@spadgett I think the PR is ready for review. 
As part of the work I've updated the `istagSelect.js` directive so if the `istag` scope variable contains info about imageStreamTag it will pre-populate the pickers with the info so we dont lose the data. Also I've added the _bool_ variable to that directive's scope called `customTagValue` that will enable to type a custom tag value into the field. Also the typed tag will be separated by a *New Tag* group to graphically distinguish the new tag from the ones that already exist. Here is the screen:
screenshot-10

I've reduced the `edit/buildConfig.js` controller heavily and I think its pretty readable now. 

Did quite a lot of testing as well.

PTAL

spadgett:
Nice to see us remove so much code without losing any capability.

jhadvig:
@spadgett indeed :) ... will squash and add the bindata after review if you dont mind

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -154,89 +130,53 @@ angular.module('openshiftConsole')
...
-                (buildFrom.kind === "ImageStreamImage") ? buildFrom.name : "",
-                (buildFrom.kind === "ImageStreamTag") ? buildConfig.metadata.namespace + "/" + buildFrom.name : buildFrom.name);
-            } else {
-              $scope.builderOptions = $scope.setPickedVariables($scope.builderOptions, "None", buildConfig.metadata.namespace, "", "", "", "");
-            }
+            var setImageOptions = function(imageOptions, imageData) {
+              imageOptions.type = (imageData && imageData.kind) ? imageData.kind : "None";
+              var ist = {},
+                  isi = "",
+                  di = "";
```
spadgett: These names are a bit too short for me. Maybe `istag`, `isimage`, and `dockerImage`? `istag` and `isimage` then match the CLI abbreviations.

spadgett: It's also easy to misread `isi` as `ist` since they look similar.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -154,89 +130,53 @@ angular.module('openshiftConsole')
...
+                ist = {namespace: "", imageStream: "", tagObject: {tag: ""}};
+              }
 
-            $scope.builderImageStream = {
-              namespace: $scope.builderOptions.pickedNamespace,
-              imageStream: $scope.builderOptions.pickedImageStream,
-              tag: $scope.builderOptions.pickedTag,
-            };
+              if (imageOptions.type === "ImageStreamImage") {
+                isi = (imageData.namespace) ? imageData.namespace + "/" + imageData.name : buildConfig.metadata.namespace + "/" + imageData.name;
```
```js
isi = (imageData.namespace || buildConfig.metadata.namespace) + "/" + imageData.name;
```

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -154,89 +130,53 @@ angular.module('openshiftConsole')
...
+              imageOptions.imageStreamTag = ist;
+              imageOptions.imageStreamImage = isi;
+              imageOptions.dockerImage = di;
             };
 
-            $scope.options.forcePull = !!$scope.buildStrategy.forcePull;
+            setImageOptions($scope.imageOptions.from, $scope.buildStrategy.from);
+            setImageOptions($scope.imageOptions.to, $scope.updatedBuildConfig.spec.output.to);
 
             if ($scope.sources.images) {
```
spadgett: Not new, but I'd feel better if we check `_.isEmpty()`. The API omits the field if empty, but I'd prefer to be safe to avoid a runtime error below.

spadgett: Nevermind, I see we only ever index into the array directly after checking length.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -648,30 +303,33 @@ angular.module('openshiftConsole')
       }
     };
 
-    $scope.constructImageObject = function(optionsModel) {
+    var constructImageObject = function(optionsModel) {
       var imageObject = {};
-      if (optionsModel.pickedType === "ImageStreamTag") {
+      if (optionsModel.type === "ImageStreamTag") {
```
spadgett: Not new, but this can be a switch instead of if, else if, else if.

app/scripts/directives/istagSelect.js
```diff
@@ -21,31 +25,102 @@ angular.module("openshiftConsole")
       restrict: 'E',
       scope: {
         istag: '=model',
-        selectDisabled: '='
+        selectDisabled: '=',
+        includeOpenshift: '=',
+        customTagValue: '='
```
spadgett: I'd name this to `allowCustomTag` to make it clear you're not setting a value.

app/scripts/directives/istagSelect.js
```diff
@@ -21,31 +25,102 @@ angular.module("openshiftConsole")
       restrict: 'E',
       scope: {
         istag: '=model',
-        selectDisabled: '='
+        selectDisabled: '=',
+        includeOpenshift: '=',
```
spadgett: It's a mouthful, but maybe `includeSharedResourceNamespce`? It's actually customizable (see #13).

app/views/edit/build-config.html
```diff
@@ -143,74 +143,30 @@
...
                               <div class="form-group">
-                                <label for="imageSourceFrom">Image Source From</label>
-                                <select
-                                  class="form-control simpleselector"
-                                  ng-model="imageSourceOptions.pickedType"
-                                  ng-options="type.id as type.title for type in imageSourceTypes"
-                                  ng-change="assambleInputType('imageSource', imageSourceOptions.pickedType)">
-                                </select>
+                                <label>Image Source From</label>
+                                <ui-select required ng-model="imageOptions.fromSource.type" search-enabled="false">
```
spadgett: You have `search-enabled=false` but are filtering based on `$select.search` below.

app/views/edit/build-config.html
```diff
@@ -327,87 +283,47 @@
                           <div>
                             <div class="form-group">
                               <label for="buildFrom">Build From</label>
-                              <select
-                                class="form-control simpleselector"
-                                ng-model="builderOptions.pickedType"
-                                ng-options="type.id as type.title for type in buildFromTypes"
-                                ng-change="assambleInputType('builder', builderOptions.pickedType)">
-                              </select>
+                              <ui-select required ng-model="imageOptions.from.type" search-enabled="false">
```
spadgett: Same comment as above on `search-enabled="false"`

app/views/edit/build-config.html
```diff
@@ -422,75 +338,27 @@
                           <div>
                             <div class="form-group">
                               <label for="buildFrom">Push To</label>
-                              <select
-                                class="form-control simpleselector"
-                                ng-model="outputOptions.pickedType"
-                                ng-options="type.id as type.title for type in pushToTypes"
-                                ng-change="assambleInputType('output', outputOptions.pickedType)">
-                              </select>
+                              <ui-select required ng-model="imageOptions.to.type" search-enabled="false">
```
spadgett: Same comment on `search-enabled="false"`

app/views/edit/build-config.html
```diff
@@ -582,11 +450,14 @@
                           </span>
                         </h3>
                         <div class="form-group">
-                          <select
-                            class="form-control"
-                            ng-model="updatedBuildConfig.spec.runPolicy"
-                            ng-options="type as (type | sentenceCase) for type in runPolicyTypes">
-                          </select>
+                          <label class="sr-only">Run policy type</label>
+                          <ui-select required ng-model="updatedBuildConfig.spec.runPolicy" search-enabled="false">
```
spadgett: Here too :)

app/scripts/directives/istagSelect.js
```diff
@@ -21,31 +25,102 @@ angular.module("openshiftConsole")
...
+        $scope.prepopulate = ($scope.istag.namespace && $scope.istag.imageStream && $scope.istag.tagObject.tag) ? true : false ;
+
+        var prepopulate = function(ns) {
+          $scope.isByNamespace[ns] = {};
+          $scope.isNamesByNamespace[ns] = [];
+
+          if (!_.contains($scope.namespaces, ns)) {
+            $scope.namespaces.push(ns);
+            $scope.isNamesByNamespace[ns] = $scope.isNamesByNamespace[ns].concat($scope.istag.imageStream);
+            $scope.isByNamespace[ns][$scope.istag.imageStream] = {status: {tags: [{tag: $scope.istag.tagObject.tag}]}};
```
spadgett: I'd return early here instead of having a big else block.

app/scripts/directives/istagSelect.js
```diff
@@ -21,31 +25,102 @@ angular.module("openshiftConsole")
...
+        customTagValue: '='
       },
       templateUrl: 'views/directives/istag-select.html',
-      link: function($scope){
+      controller: function($scope){
         $scope.isByNamespace = {};
         $scope.isNamesByNamespace = {};
 
+        // Check if the istag object contains data about namespace/imageStream/tag so the ui-select will be pre-populated with them
+        $scope.prepopulate = ($scope.istag.namespace && $scope.istag.imageStream && $scope.istag.tagObject.tag) ? true : false ;
```
spadgett: Shouldn't need to be a scope var since it's not used in the view. Although you'd need to give it a different name since you have a function with that name.

```js
var prepopulate = $scope.istag.namespace && $scope.istag.imageStream && scope.istag.tagObject.tag;
```

app/scripts/directives/istagSelect.js
```diff
@@ -21,31 +25,102 @@ angular.module("openshiftConsole")
...
               return;
             }
 
-            DataService.list('imagestreams', { namespace: namespace }, function(isData) {
-              $scope.isByNamespace[namespace] = isData.by('metadata.name');
-              $scope.isNamesByNamespace[namespace] = _.keys($scope.isByNamespace[namespace]).sort();
-            });
+            if ($scope.prepopulate) {
+              prepopulate(namespace);
+              $scope.prepopulate = false;
```
spadgett: I'd return early here instead of having the big else block.

app/scripts/directives/istagSelect.js
```diff
@@ -21,31 +25,102 @@ angular.module("openshiftConsole")
...
+                  return !tag.items;
+                });
+                $scope.isByNamespace[$scope.istag.namespace][$scope.istag.imageStream].status.tags.unshift({tag: search});
+              }
+            };
+
+            $scope.groupTags = function(tagObject) {
+              if ($scope.customTagValue) {
+                if (tagObject.items) {
+                return 'AVAILABLE TAGS';
```
spadgett: I feel like we should use `text-transform: capitalize;` style on the group header instead of returning all caps here if it's what we want.

Indentation is off here.

app/scripts/directives/istagSelect.js
```diff
@@ -21,31 +25,102 @@ angular.module("openshiftConsole")
...
+                } else {
+                  return 'NEW TAG';
+                }
+              } else {
```
spadgett: I'd remove the else here since you always return in the if above

spadgett: Maybe just

```js
if ($scope.customTagValue) {
  return tagObject.items ? 'Available Tags' : 'New Tag';
}

return '';
```

spadgett:
Thanks @jhadvig this looks much cleaner.

app/scripts/directives/istagSelect.js
jhadvig: no need to capitalize the text, found you that the ui-select groups have capitalized header anyway

spadgett:
Maybe "Current Tags" instead of "Available Tags" in the group header?

jhadvig:
@spadgett I've added an additional commit with suggested changes. PTAL

app/scripts/directives/istagSelect.js
```diff
@@ -69,23 +70,23 @@ angular.module("openshiftConsole")
...
               return;
             }
 
-            if ($scope.prepopulate) {
+            if (shouldPrepopulate) {
               prepopulate(namespace);
-              $scope.prepopulate = false;
+              shouldPrepopulate = false;
+              return;
             } else {
```
spadgett: You can remove the `else` if you're returning above.

app/views/edit/build-config.html
```diff
@@ -149,14 +149,14 @@
                                 <label>Image Source From</label>
                                 <ui-select required ng-model="imageOptions.fromSource.type" search-enabled="false">
                                   <ui-select-match>{{$select.selected | startCase}}</ui-select-match>
-                                  <ui-select-choices repeat="type in (imageSourceTypes | filter : $select.search)">
+                                  <ui-select-choices repeat="type in imageSourceTypes">
                                     <div ng-bind-html="type | startCase"></div>
```
spadgett: If you're not highlighting a search match, you shouldn't need to use `ng-bind-html`. Does this work?

```html
<ui-select-choices repeat="type in imageSourceTypes">
  {{type | startCase}}
</ui-select-choices>
```

Here and the selects below.

jhadvig: yep it does :+1:

spadgett:
LGTM, please squash and add dist. We'll need to wait until after 1.3 to merge.

jhadvig:
@spadgett sounds like a plan :)

spadgett:
@jwforres Do we have a card for QE to regression test?

jhadvig:
@spadgett done

spadgett:
Squashed commit LGTM. @jhadvig Ready for merge?

jhadvig:
@spadgett yes... let's do this :rocket:

spadgett:
[merge]

PR #494, opened 2016-08-29
canI check for adding HPAs to an RC

### LINKED ISSUE #492: "Add autoscaler" link has no `canI` check
We show the "Add autoscaler" link for an RC without any `canI` checks on if the user can create HPAs.

See https://github.com/spadgett/origin-web-console/blob/b0cc4952881c3aeedba4b786cb42937b8115480c/app/views/browse/_deployment-details.html#L127-L130

/cc @jwforres

spadgett:
Fixes #492

@jwforres 

I used "none" to be consistent with volumes, but maybe it should be "no autoscalers"

openshift_web_console

### DIFF
--- a/app/views/browse/_deployment-details.html
+++ b/app/views/browse/_deployment-details.html
@@ -126,7 +126,13 @@
   <!-- Create autoscaler -->
   <div ng-if="!autoscalers.length">
     <a ng-href="project/{{projectName}}/edit/autoscaler?kind=ReplicationController&name={{deployment.metadata.name}}"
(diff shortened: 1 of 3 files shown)


jwforres:
hmm yeah that reads strangely.  I feel like we could be more descriptive here, something like

'Autoscaling is currently disabled.  There are no autoscalers for this deployment configuration (or RC).'

jwforres:
LGTM, for post 1.3

jhadvig:
Good catch ! 
Would it make sense to also change the _Volumes_ to something more descriptive ?

jwforres:
[merge]

PR #506, opened 2016-09-01
Don't set container name for build log options on monitoring page

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1372195

@jwforres We were incorrectly setting the build pod name as the container name for build log options on the monitoring page. This caused the view archive link to show up with bad parameters.


jwforres:
[merge]

PR #508, opened 2016-09-01
Browse pages for k8s deployments and replica sets

### LINKED ISSUE #525: Can attach storage directly to an RC owned by a DC
We should change the action to attach storage to the DC instead. See https://github.com/openshift/origin-web-console/pull/508/files/1860434f9a525a8ed1f430fe58d2720f9452f79b#r77872462

spadgett:
Add browse pages for deployments and replica sets. This does not update the overview or monitoring page.

Part of https://trello.com/c/HGvCTI96
Fixes #525


app/views/browse/deployment.html
```diff
@@ -1,142 +1,266 @@
...
+                            <div ng-if="deployment.status.updatedReplicas">
+                              {{deployment.status.updatedReplicas}} up to date
+                            </div>
+                            <div ng-if="deployment.status.availableReplicas || deployment.status.unavailableReplicas">
+                              <span ng-if="deployment.status.availableReplicas">{{deployment.status.availableReplicas}} available<span ng-if="deployment.status.unavailableReplicas">,</span></span>
+                              <span ng-if="deployment.status.unavailableReplicas">{{deployment.status.unavailableReplicas}} unavailable</span>
+                            </div>
+                          </dd>
+                          <dt>Paused:</dt>
+                          <dd>{{!!deployment.spec.paused}}</dd>
```
jwforres: don't love seeing true/false here. for now can you switch it to Yes/No.  I think long term I'd rather not show it at all when paused is false, and when paused is true highlight it some other way.

app/views/browse/deployment.html
```diff
@@ -1,142 +1,266 @@
...
+                          <dt>Selectors:</dt>
+                          <dd>
+                            <selector selector="deployment.spec.selector"></selector>
+                          </dd>
+                          <dt>Strategy:</dt>
+                          <dd>{{deployment.spec.strategy.type | sentenceCase}}</dd>
+                          <dt ng-if-start="deployment.spec.strategy.rollingUpdate">Max Unavailable:</dt>
+                          <dd>
+                            {{deployment.spec.strategy.rollingUpdate.maxUnavailable}}
+                            <span data-toggle="tooltip"
```
jwforres: i'd rather see this help icons with the label not the value

spadgett: open to changing this, but it felt pretty weird with the colon :/ not sure what's best

spadgett:
Rebased.

app/scripts/controllers/replicaSet.js
```diff
@@ -0,0 +1,448 @@
...
+                        Navigate,
+                        keyValueEditorUtils,
+                        kind) {
+    var hasDC = false;
+    var displayKind = $filter('humanizeKind')(kind);
+    switch (kind) {
+    case 'ReplicaSet':
+      $scope.breadcrumbs = [
+        {
+          title: "Replica Sets",
```
spadgett: @jwforres forgot to make this change to all of the edit pages. might want a utility that builds breadcrumbs from the resource?

app/scripts/controllers/attachPVC.js
```diff
@@ -252,26 +250,14 @@ angular.module('openshiftConsole')
             $scope.alerts = {};
 
             // save deployment or deployment config
```
jwforres: outdated comment,  // save resource

app/scripts/controllers/attachPVC.js
```diff
@@ -252,26 +250,14 @@ angular.module('openshiftConsole')
...
-                  displayError("An error occurred attaching the persistent volume claim to deployment.", getErrorDetails(result));
-                }
-              );
-            }
+            DataService.update(resourceGroupVersion, resource.metadata.name, $scope.attach.resource, context).then(
+              function() {
+                $window.history.back();
+              },
+              function(result) {
+                displayError("An error occurred attaching the persistent volume claim to deployment config.", getErrorDetails(result));
```
jwforres: message says deployment config in it, should either be generic and say resource, or embed the lower-cased humanized kind here

app/scripts/controllers/deployment.js
```diff
@@ -13,204 +13,136 @@ angular.module('openshiftConsole')
...
-    // Check if the metrics service is available so we know when to show the tab.
-    MetricsService.isAvailable().then(function(available) {
-      $scope.metricsAvailable = available;
-    });
-
-    var setLogVars = function(deployment) {
-      $scope.logCanRun = !(_.includes(['New', 'Pending'], $filter('deploymentStatus')(deployment)));
-    };
-
+    // copy deploymentConfig and ensure it has env so that we can edit env vars using key-value-editor
```
jwforres: deployment not deploymentConfig

app/scripts/controllers/deployment.js
```diff
@@ -223,190 +155,80 @@ angular.module('openshiftConsole')
...
-            $scope.$watch('deployment.spec.selector', function() {
-              selector = new LabelSelector($scope.deployment.spec.selector);
-              updatePodsForDeployment();
-            }, true);
-
-            watches.push(DataService.watch("pods", context, function(podData) {
-              pods = podData.by('metadata.name');
-              updatePodsForDeployment();
+            // Watch replica sets for this deployment
+            // TODO: Use the label selector
```
jwforres: should this TODO be to use the controller ref?

app/scripts/controllers/deployment.js
```diff
@@ -223,190 +155,80 @@ angular.module('openshiftConsole')
...
+          limitRanges = response.by("metadata.name");
+          updateHPAWarnings();
+        });
 
-        // Sets up subscription for imageStreams
         watches.push(DataService.watch("imagestreams", context, function(imageStreams) {
           $scope.imageStreams = imageStreams.by("metadata.name");
           ImageStreamResolver.buildDockerRefMapForImageStreams($scope.imageStreams, $scope.imageStreamImageRefByDockerReference);
-          ImageStreamResolver.fetchReferencedImageStreamImages($scope.podTemplates, $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, context);
+          // If the dep config has been loaded already
```
jwforres: deployment not dep config

app/scripts/controllers/deployment.js
```diff
@@ -223,190 +155,80 @@ angular.module('openshiftConsole')
...
-          updateHPAWarnings();
-        }, {poll: limitWatches, pollInterval: 60 * 1000}));
-
-        // List limit ranges in this project to determine if there is a default
-        // CPU request for autoscaling.
-        DataService.list("limitranges", context, function(response) {
-          limitRanges = response.by("metadata.name");
+        }, context, function(hpa) {
+          $scope.autoscalers =
+            HPAService.filterHPA(hpa.by("metadata.name"), 'Deployment', $routeParams.deployment);
```
jwforres: does filterHPA handle the deployment being null in case this fires before deployment is loaded

app/scripts/controllers/deployment.js
```diff
@@ -223,190 +155,80 @@ angular.module('openshiftConsole')
...
+          Logger.log("builds (subscribe)", $scope.builds);
+        }));
 
         $scope.scale = function(replicas) {
           var showScalingError = function(result) {
             $scope.alerts = $scope.alerts || {};
             $scope.alerts["scale"] = {
               type: "error",
-              message: "An error occurred scaling the deployment.",
+              message: "An error occurred scaling the deployment config.",
```
jwforres: deployment

app/scripts/controllers/deployment.js
jwforres: ignore, missed that it was sending the routeParam not the object

app/scripts/controllers/deployments.js
```diff
@@ -28,119 +39,151 @@ angular.module('openshiftConsole')
...
-        $scope.$apply(function() {
-          $scope.deploymentConfigs = labelSelector.select($scope.unfilteredDeploymentConfigs);
+          Logger.log("deployments (subscribe)", $scope.deployments);
+        }));
+
+        watches.push(DataService.watch({
+          group: "extensions",
+          resource: "replicasets"
+        }, context, function(replicaSets) {
+          $scope.unfilteredReplicaSets = replicaSets.by("metadata.name");
```
jwforres: add a TODO comment that this will need to be updated to only the replicasets that aren't part of a deployment

app/scripts/controllers/deployments.js
```diff
@@ -28,119 +39,151 @@ angular.module('openshiftConsole')
...
           }
-        }
 
-        // Extract the causes from the encoded deployment config
-        if (deployment) {
-          if (action !== "DELETED") {
-            deployment.causes = $filter('deploymentCauses')(deployment);
+          $scope.deploymentsByDeploymentConfig = DeploymentsService.associateDeploymentsToDeploymentConfig($scope.deployments, $scope.deploymentConfigs, true);
+          if ($scope.deploymentsByDeploymentConfig['']) {
+            $scope.unfilteredReplicationControllers = $scope.deploymentsByDeploymentConfig[''];
```
jwforres: add label suggestions for these

spadgett:
@jwforres pushed a commit with the changes for breadcrumbs

spadgett:
Updated for comments

app/scripts/controllers/deployments.js
```diff
@@ -28,119 +39,151 @@ angular.module('openshiftConsole')
...
+          return false;
+        };
+
+        LabelFilter.onActiveFiltersChanged(function(labelSelector) {
+          // trigger a digest loop
+          $scope.$apply(function() {
+            $scope.deploymentConfigs = labelSelector.select($scope.unfilteredDeploymentConfigs);
+            $scope.deploymentsByDeploymentConfig = DeploymentsService.associateDeploymentsToDeploymentConfig($scope.deployments, $scope.deploymentConfigs, true);
+            if ($scope.deploymentsByDeploymentConfig['']) {
+              $scope.unfilteredReplicationControllers = $scope.deploymentsByDeploymentConfig[''];
```
jwforres: i dont think its new, but i think there is a bug here.  If I have a DC with one set of labels trigger a few RCs, then I change the labels on the DC, trigger a few RCs.  Then I filter on labels that will filter out the DC and the new RCs, but not filter out the old RCs, they will end up visible in the Other Replication Controllers list.

app/scripts/controllers/replicaSet.js
```diff
@@ -0,0 +1,448 @@
...
+          link: "project/" + $routeParams.project + "/browse/deployments"
+        }
+      ];
+
+      $scope.resource = {
+        group: "extensions",
+        resource: "replicasets"
+      };
+      $scope.healthCheckURL = Navigate.healthCheckURL($routeParams.project,
+                                                      "ReplicaSet",
```
jwforres: if you passed the kind var in here instead wouldn't these two be identical and you could move this out of the switch statement?

app/scripts/controllers/replicaSet.js
```diff
@@ -0,0 +1,448 @@
...
+        };
+
+        var updateHPAWarnings = function() {
+            HPAService.getHPAWarnings($scope.deployment, $scope.autoscalers, limitRanges, project)
+                      .then(function(warnings) {
+              $scope.hpaWarnings = warnings;
+            });
+        };
+
+        var updateDC = function(rc) {
```
jwforres: looked like you updated other references to rc vars to be rs, update this one too?

jwforres: TODO comment that this needs to be updated to work for RS -> Deployment

spadgett: > looked like you updated other references to rc vars to be rs, update this one too?

I left this one as `rc` because it's always a replication controller if it has a deployment config (at least for now).

app/scripts/controllers/replicaSet.js
spadgett: > if you passed the kind var in here instead wouldn't these two be identical and you could move this out of the switch statement?

No, one has to pass the "extensions" group.

app/scripts/controllers/replicaSet.js
```diff
@@ -0,0 +1,448 @@
...
+            $scope.alerts["load"] = {
+              type: "error",
+              message: "The " + kind + " details could not be loaded.",
+              details: "Reason: " + $filter('getErrorDetails')(e)
+            };
+          }
+        );
+
+        function extractPodTemplates() {
+          angular.forEach($scope.deployments, function(deployment, deploymentId){
```
jwforres: don't think this is new, but we really only want the pod template for the main deployment, this is pulling podTemplates for all deployments which means we are pulling image stream images we dont actually need

app/scripts/controllers/replicaSet.js
```diff
@@ -0,0 +1,448 @@
...
+            $scope.podTemplates[deploymentId] = deployment.spec.template;
+          });
+        }
+
+        watches.push(DataService.watch($scope.resource, context, function(deployments, action, deployment) {
+          $scope.deployments = deployments.by("metadata.name");
+          extractPodTemplates();
+          ImageStreamResolver.fetchReferencedImageStreamImages($scope.podTemplates, $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, context);
```
jwforres: related to the previous comment, this really should be in the get and watchObject and not here since we only need the pod template image data for the main RC/RS

app/scripts/controllers/replicaSet.js
```diff
@@ -0,0 +1,448 @@
...
+          $scope.emptyMessage = "No deployments to show";
+          $scope.deploymentsByDeploymentConfig = DeploymentsService.associateDeploymentsToDeploymentConfig($scope.deployments);
+
+          var deploymentConfigName;
+          var deploymentName;
+          if (deployment) {
+            deploymentConfigName = $filter('annotation')(deployment, 'deploymentConfig');
```
jwforres: TODO comment to make this work for RS -> Deployment

app/scripts/controllers/replicaSet.js
```diff
@@ -0,0 +1,448 @@
...
+        }, {poll: limitWatches, pollInterval: 60 * 1000}));
+
+        // List limit ranges in this project to determine if there is a default
+        // CPU request for autoscaling.
+        DataService.list("limitranges", context, function(response) {
+          limitRanges = response.by("metadata.name");
+          updateHPAWarnings();
+        });
+
+        $scope.startLatestDeployment = function(deploymentConfig) {
```
jwforres: is this even used from this page?

spadgett: Not new, but doesn't look like it.

app/scripts/controllers/deployments.js
spadgett: Looks like we have two deployments service methods that do the association differently.

app/scripts/filters/resources.js
```diff
@@ -149,12 +149,13 @@ angular.module('openshiftConsole')
         {group: '', resource: 'builds/clone', verbs: ['create']},
         {group: '', resource: 'builds',       verbs: ['delete', 'update']}
       ],
-      'deploymentConfigs': [
+      'deployments': [
         {group: 'extensions', resource: 'horizontalpodautoscalers', verbs: ['create', 'update']},
-        {group: '',            resource: 'deploymentconfigs',        verbs: ['create', 'update']}
+        {group: '',            resource: 'deployments',             verbs: ['create', 'update']}
```
jwforres: shouldnt deployments be in the extension group

app/scripts/controllers/deployments.js
spadgett: Opened issue #522.

app/scripts/controllers/replicaSet.js
spadgett: Added TODO

app/scripts/controllers/replicaSet.js
spadgett: Opened issue #523

app/scripts/filters/resources.js
```diff
@@ -1084,12 +1089,16 @@ angular.module('openshiftConsole')
...
       ReplicationController: 'rc',
       Route: 'route',
       Service: 'svc',
     };
     return function(kind) {
-      return abbreviated[kind] || kind;
+      if (!kind) {
+        return kind;
+      }
+      return abbreviated[kind] || kind.toLowerCase();
```
jwforres: kind.toLowerCase is inconsistent with what the CLI displays in similar scenarios, it should really be printing the resource, e.g. if it cant abbreviate something it should say deployments not deployment.  

where are we using abbreviateKind today?

spadgett: It's used for the events sidebar for a specific involved object like "rc/foo" or "deployment/bar"

spadgett: I can change it to "deployments/bar"

app/views/attach-pvc.html
```diff
@@ -35,11 +33,10 @@ <h2 class="text-center">No persistent volume claims.</h2>
                     To claim storage from a persistent volume, refer to the documentation on <a target="_blank" ng-href="{{'persistent_volumes' | helpLink}}">using persistent volumes</a>.
                   </p>
 
-                  <p ng-if="attach.deploymentConfig"><a href="{{attach.deploymentConfig | navigateResourceURL}}">Back to deployment config</a></p>
-                  <p ng-if="attach.deployment"><a href="{{attach.deployment | navigateResourceURL}}">Back to deployment</a></p>
+                  <p ng-if="attach.resource"><a ng-href="{{attach.resource | navigateResourceURL}}">Back to {{name}}</a></p>
```
jwforres: cant decide but maybe `Back to {{kind | humanizeKind}} {{name}}`

app/views/browse/_pod-details.html
```diff
@@ -79,7 +79,7 @@
       <div ng-if="pod.spec.volumes.length">
```
jwforres: not new, but this makes the below link to attach storage unreachable, also missing the canI check on that link

app/views/browse/_replica-set-details.html
```diff
@@ -89,10 +88,18 @@
...
-        <a ng-if="'replicationcontrollers' | canI : 'update'" ng-href="project/{{project.metadata.name}}/attach-pvc?deployment={{deployment.metadata.name}}">Attach storage</a>
-        <span ng-if="!('replicationcontrollers' | canI : 'update')">none</span>
-      </p>
+      <div ng-if="!deployment.spec.template.spec.volumes.length">
+        <div ng-if="kind === 'ReplicaSet'">
+          <a ng-if="{ group: 'extensions', resource: 'replicasets' } | canI : 'update'"
+             ng-href="project/{{project.metadata.name}}/attach-pvc?kind=ReplicaSet&name={{deployment.metadata.name}}&group=extensions">Attach storage</a>
+          <span ng-if="!({ group: 'extensions', resource: 'replicasets' } | canI : 'update')">none</span>
+        </div>
+        <div ng-if="kind === 'ReplicationController'">
```
jwforres: i dont think this is new, but this is letting you AttachStorage to an RC that's owned by a DC, looks like we have the same problem with the Attach Storage action showing up in the dropdown menu

app/views/browse/_replica-set-details.html
```diff
@@ -89,10 +88,18 @@
...
       <h4>Volumes</h4>
-      <p ng-if="!deployment.spec.template.spec.volumes.length">
-        <a ng-if="'replicationcontrollers' | canI : 'update'" ng-href="project/{{project.metadata.name}}/attach-pvc?deployment={{deployment.metadata.name}}">Attach storage</a>
-        <span ng-if="!('replicationcontrollers' | canI : 'update')">none</span>
-      </p>
+      <div ng-if="!deployment.spec.template.spec.volumes.length">
+        <div ng-if="kind === 'ReplicaSet'">
+          <a ng-if="{ group: 'extensions', resource: 'replicasets' } | canI : 'update'"
+             ng-href="project/{{project.metadata.name}}/attach-pvc?kind=ReplicaSet&name={{deployment.metadata.name}}&group=extensions">Attach storage</a>
+          <span ng-if="!({ group: 'extensions', resource: 'replicasets' } | canI : 'update')">none</span>
```
jwforres: resource | canI : 'update'

jwforres: there are several places here both for RS and RC where you could use resource instead

app/views/browse/_replica-set-details.html
spadgett: Opened issue #525

app/views/browse/_replica-set-details.html
spadgett: > here are several places here both for RS and RC where you could use resource instead

Yeah, I added `$scope.resource` later. I'll update the view.

app/views/browse/_replication-controller-actions.html
```diff
@@ -0,0 +1,47 @@
...
+  </button>
+  <a href=""
+     class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
+     data-toggle="dropdown"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
+  <ul class="dropdown-menu actions action-button">
+    <li ng-if="'replicationcontrollers' | canI : 'update'">
+      <a ng-href="project/{{project.metadata.name}}/attach-pvc?kind=ReplicationController&name={{deployment.metadata.name}}"
+         role="button">Attach Storage</a>
+    </li>
+    <li ng-if="!deploymentConfigName && ('replicationcontrollers' | canI : 'update')">
```
jwforres: so we are inconsistent in how we are handling these, for Set Resource Limits we completely hide the action on any RC that is from a DC, but for Edit Health Checks we enable the action if you are the active RC for a DC but send you over to editing the DC's health checks.

spadgett: Also not a new issue :(  But shouldn't be hard to fix.

app/views/browse/deployment.html
```diff
@@ -1,141 +1,266 @@
...
+                      project-name="{{deployment.metadata.namespace}}"
+                      alerts="alerts"
+                      hpa-list="autoscalers">
+                    </delete-link>
+                  </li>
+                </ul>
+              </div>
+              <small class="meta" ng-if="deployment">created <relative-timestamp timestamp="deployment.metadata.creationTimestamp"></relative-timestamp></small>
+            </h1>
+            <labels labels="deployment.metadata.labels" clickable="true" kind="deployments" title-kind="deployment configs" project-name="{{deployment.metadata.namespace}}" limit="3"></labels>
```
jwforres: shouldnt title-kind be "deployments" too? in which case i think you can drop title-kind if i'm remembering how the directive works

app/views/browse/_replica-set-details.html
spadgett: I went ahead and added a fix in this PR.

app/views/browse/_replication-controller-actions.html
spadgett: I added a Set Resource Limits action to update the DC (when there is a DC).

app/views/browse/deployment.html
```diff
@@ -1,141 +1,266 @@
...
+                            {{deployment.spec.strategy.rollingUpdate.maxSurge}}
+                            <span data-toggle="tooltip"
+                              title="The maximum number of pods that can be created above the desired number of pods."
+                              class="pficon pficon-help text-muted small"></span>
+                          </dd>
+                          <dt>Min Ready:</dt>
+                          <dd>
+                            {{deployment.spec.minReadySeconds || 0}} sec
+                            <span data-toggle="tooltip"
+                              title="The minimum number of seconds for which a newly created pod should be ready without any of its containers crashing for it to be considered available."
```
jwforres: hmm this sentence is hard to parse, isnt the definition of ready that the containers are started, so if its ready for X seconds then it was ready for X seconds with no containers crashing?
Should this just be:
`The minimum number of seconds a new pod must be ready before it will be considered available.`

spadgett: Yeah, I used the exact wording from the Kubernetes doc for these tooltips. I agree your suggestion is better.

app/views/browse/deployment.html
```diff
@@ -1,141 +1,266 @@
...
+                    </div>
+                    <div class="row">
+                      <div class="col-lg-6">
+                        <h3>Template</h3>
+                        <pod-template
+                            pod-template="deployment.spec.template"
+                            images-by-docker-reference="imagesByDockerReference"
+                            builds="builds"
+                            detailed="true"
+                            add-health-check-url="{{ ({ group: 'extensions', resource: 'deployments' } | canI : 'update') ? healthCheckURL : '' }}">
```
jwforres: I dont think we have a canI check for the action in this warning that pops up in the pod template

app/views/browse/deployment.html
spadgett: Although I'm not sure of the behavior if you don't have readiness probes.

app/views/browse/deployment.html
spadgett: That's what this is doing. The pod template directive doesn't know what resource to check, so it has to be outside.

spadgett: If there's no health check URL, the pod template won't link to the action.

app/views/browse/replica-set.html
```diff
@@ -0,0 +1,151 @@
...
+            <div ng-if="deployment">
+              <h1>
+                {{deployment.metadata.name}}
+                <span
+                  ng-if="deploymentConfigMissing"
+                  class="pficon pficon-warning-triangle-o"
+                  style="cursor: help; vertical-align: middle;"
+                  data-toggle="tooltip"
+                  data-trigger="hover"
+                  title="The deployment's deployment config is missing."
```
jwforres: 'The replica set's deployment is missing.'

app/views/browse/replica-set.html
```diff
@@ -0,0 +1,151 @@
...
+                  aria-hidden="true">
+                </span>
+                <span ng-if="deploymentConfigMissing" class="sr-only">Warning: The deployment's deployment config is missing.</span>
```
jwforres: replica set / deployment

app/views/browse/replica-set.html
jwforres: ah right this needs to work for the old stuff too, nvm leave it till we fix the rest of the Dep->RS stuff

spadgett: Isn't this always a replication controller and deployment config?

app/views/browse/deployment.html
jwforres: ignore me, i've clearly been looking at this too long already

app/views/deployments.html
```diff
@@ -96,8 +98,66 @@
...
+                      <th>Strategy</th>
+                    </tr>
+                  </thead>
+                  <tbody ng-repeat="k8sDeployment in k8sDeployments | orderObjectsByDate : true">
+                    <tr>
+                      <td data-title="Name">
+                        <a ng-href="{{k8sDeployment | navigateResourceURL}}">{{k8sDeployment.metadata.name}}</a>
+                      </td>
+                      <td data-title="Last Version">
+                        #{{k8sDeployment | annotation : 'deployment.kubernetes.io/revision'}}
```
jwforres: we usually try to fallback cleanly in some way of there if an annotation is missing, for now could just say Unknown.  once we can map the RSes in this should be the link to the latest RS and can be the name of that RS if the annotation is missing like we do for DC->RC

app/views/deployments.html
```diff
@@ -96,8 +98,66 @@
...
+                      </td>
+                      <td data-title="Strategy">
+                        {{k8sDeployment.spec.strategy.type | sentenceCase}}
+                      </td>
+                    </tr>
+                  </tbody>
+                </table>
+              </div>
+              <div ng-if="replicaSets | hashSize" id="replica-sets">
+                <h3>Replica Sets</h3>
```
jwforres: once we do the roll-up under Deployments this should probably say Other Replica Sets, like we do for RCs

app/views/directives/breadcrumbs.html
```diff
@@ -1,7 +1,6 @@
 <ol class="breadcrumb" ng-if="breadcrumbs.length">
   <li ng-repeat="breadcrumb in breadcrumbs" ng-class="{'active': !$last}">
-    <a ng-if="!$last && breadcrumb.link" href="{{breadcrumb.link}}">{{breadcrumb.title}}</a>
-    <a ng-if="!$last && !breadcrumb.link" href="" back>{{breadcrumb.title}}</a>
```
jwforres: you lost this behavior for back, did you update things that were relying on that?

spadgett: I didn't mean to commit this change. Will undo.

app/views/modals/delete-resource.html
```diff
@@ -34,6 +34,14 @@
       </div>
 
       <div ng-switch="kind">
+        <div ng-switch-when="Deployment">
+          <strong>Note:</strong> None of the replica sets created by this deployment will be deleted.
+
+          To delete the deployment and all of its replica sets, you can run the command
+          <pre class="code prettyprint mar-top-md">oc delete deployment {{resourceName}} -n {{projectName}}</pre>
```
jwforres: did you check if that actually works, i don't know if kubectl/oc actually behaves like DCs/RCs in this regard

spadgett: > did you check if that actually works

Yes, it tried it. It cascades.

app/scripts/controllers/attachPVC.js
```diff
@@ -98,7 +96,12 @@ angular.module('openshiftConsole')
                 $scope.attach.containers.individual[container.name] = true;
               });
               $scope.attach.resource = resource;
-              rebuildBreadcrumb();
+              $scope.breadcrumbs = BreadcrumbsService.getBreadcrumbs({
+                object: resource,
+                project: project,
+                subpage: 'Edit Health Checks',
```
jwforres: Attach Storage

app/views/browse/deployment.html
```diff
@@ -109,26 +109,32 @@
...
-                          <dd ng-if-end>
-                            {{deployment.spec.strategy.rollingUpdate.maxSurge}}
+                          <dt>
+                            Max Surge:
                             <span data-toggle="tooltip"
                               title="The maximum number of pods that can be created above the desired number of pods."
                               class="pficon pficon-help text-muted small"></span>
+                          </dt>
+                          <dd ng-if-end>
+                            {{deployment.spec.strategy.rollingUpdate.maxSurge}}
```
jwforres: is maxSurge required, can it be unbounded?

app/views/browse/deployment.html
```diff
@@ -109,26 +109,32 @@
...
-                          <dd>
-                            {{deployment.spec.strategy.rollingUpdate.maxUnavailable}}
+                          <dt ng-if-start="deployment.spec.strategy.rollingUpdate">
+                            Max Unavailable:
                             <span data-toggle="tooltip"
                               title="The maximum number of pods that can be unavailable during the update process."
                               class="pficon pficon-help text-muted small"></span>
+                          </dt>
+                          <dd>
+                            {{deployment.spec.strategy.rollingUpdate.maxUnavailable}}
```
jwforres: same question, is this required?

app/views/browse/deployment.html
```diff
@@ -109,26 +109,32 @@
...
-                          <dd>
-                            {{deployment.spec.strategy.rollingUpdate.maxUnavailable}}
+                          <dt ng-if-start="deployment.spec.strategy.rollingUpdate">
+                            Max Unavailable:
                             <span data-toggle="tooltip"
                               title="The maximum number of pods that can be unavailable during the update process."
                               class="pficon pficon-help text-muted small"></span>
+                          </dt>
+                          <dd>
+                            {{deployment.spec.strategy.rollingUpdate.maxUnavailable || 1}}
```
jwforres: handle maxUnavailable and maxSurge being set to 0 vs unset

jwforres:
add a spec test to test the breadcrumb service

jwforres:
the rest LGTM

spadgett:
Added BreadcrumbsService [test]s

https://github.com/openshift/origin-web-console/pull/508/files#diff-592a16f6565fa6356b312fa17470d7f3R1

jwforres:
LGTM, can see if travis passes on the new tests and then should be good to merge

spadgett:
[merge]

PR #509, opened 2016-09-01
Manipulating the sort direction of events

### LINKED ISSUE #439: Sort by on events page should switch to the default sort order that makes sense for different sort attributes
Timestamp is our default which is Z->A (newest to oldest, which is a confusing indicator anyway)  If I switch to sorting by Reason i would expect that to default to A->Z .

rhamilto:
Fixes #439 

@jwforres, @spadgett, @benjaminapetersen this is a bit hacky, but I've been unable to find another path.  Thoughts?

### DIFF
--- a/app/scripts/directives/events.js
+++ b/app/scripts/directives/events.js
@@ -31,10 +31,20 @@ angular.module('openshiftConsole')
         };
 
         var sortedEvents = [];
+        var currentID = _.get($scope, 'sortConfig.currentField.id');
+        var defaultIsReversed = {
+          lastTimestamp: true
+        };
         var sortEvents = function() {
-          // TODO: currentField is renamed in angular-patternfly 3.0
-          var sortID = _.get($scope, 'sortConfig.currentField.id', 'lastTimestamp'),
-              order = $scope.sortConfig.isAscending ? 'asc' : 'desc';
+          var sortID = _.get($scope, 'sortConfig.currentField.id', 'lastTimestamp');
+          // only change if sort dropdown field is changed
+          if (currentID !== sortID) {
+            // set currentID to sortID
+            currentID = sortID;
+            // reverse the sort
+            $scope.sortConfig.isAscending = !(defaultIsReversed[currentID]);
+          }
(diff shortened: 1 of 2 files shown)


app/scripts/directives/events.js
```diff
@@ -31,10 +31,24 @@ angular.module('openshiftConsole')
...
-          var sortID = _.get($scope, 'sortConfig.currentField.id', 'lastTimestamp'),
-              order = $scope.sortConfig.isAscending ? 'asc' : 'desc';
+          var sortID = _.get($scope, 'sortConfig.currentField.id', 'lastTimestamp');
+          // only change if sort field is changed
+          if (currentID !== sortID) {
+            // set currentID to sortID
+            currentID = sortID;
+            // if currentID is lastTimestamp and isAscending is true, set
+            // isAscending to false so records display reverse sorted by time
+            if (currentID === 'lastTimestamp' && $scope.sortConfig.isAscending === true) {
```
jwforres: i would take this approach

```
var defaultIsReversed = {
  lastTimestamp: true
};
if (current ID !== sortID) {
  $scope.sortConfig.isAscending = !(defaultIsReversed[currentID]);
}
```

That lets you add more reversed sorts easily in the future without having to check them all, and lets you do the assignment in one single line without the nested if/else block

rhamilto:
Neat!  Thanks, @jwforres.  This one is ready for review and merge after freeze.

jwforres:
LGTM

benjaminapetersen:
I'd prob suggest making use of returns from the functions themselves so we can do something like:
```javascript
// can reduce a var & may be easier to read/maintain
var update = function() {
  $scope.sortedEvents = sortEvents(); 
  $scope.filteredEvents = filterForKeyword();
};
```
`sortEvents()` would just have to do this then:
```javascript
var sortEvents = function() {
  /* do stuff, then... */
  // return the sort, rather than set a variable outside the functions scope
  return _.sortByOrder($scope.events, [sortID], [order]);
};
```
I like to ask a function to do a job and give me back what it created (avoid side effects if possible).

jwforres:
[merge]

jwforres:
[merge]

PR #510, opened 2016-09-02
Adding code formatting instructions

rhamilto:
Per a conversation @benjaminapetersen, @sg00dwin, and I had yesterday,
we realized we should highlight this information in the README as we’re
not always consistent, despite the fact that .editorconfig exists in
the repo.

@jwforres, PTAL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,22 @@ a different API server, copy `app/config.js` to
 not tracked and will be used instead if it exists.
 
 #### Before opening a pull request
+
+Please configure your editor to use the
+following settings to avoid common code inconsistencies and dirty
+diffs:
+
+* Use soft-tabs set to two spaces.
+* Trim trailing white space on save.
+* Set encoding to UTF-8.
+* Add new line at end of files.
+
+Or [configure your editor](http://editorconfig.org/#download) to
+utilize [`.editorconfig`](https://github.com/openshift/origin-web-console/blob/master/.editorconfig),
+which will apply these settings automatically.
+
+Then:
+
 1. If needed, run `grunt build` to update the files under the dist directory
 2. Run the spec tests with `grunt test`
 3. Run the integrations tests (your api server must be running) `grunt test-integration`


README.md
```diff
@@ -54,6 +54,22 @@ a different API server, copy `app/config.js` to
 not tracked and will be used instead if it exists.
 
 #### Before opening a pull request
+
+Please configure your editor to using the
```
jwforres: to use

jwforres:
Minor grammar nit, otherwise LGTM

README.md
```diff
@@ -54,6 +54,22 @@ a different API server, copy `app/config.js` to
...
+following settings to avoid common code inconsistencies and dirty
+diffs:
+
+* Use soft-tabs set to two spaces.
+* Trim trailing white space on save.
```
spadgett: Not saying we shouldn't, but it will create a lot of noisy diffs until the trailing space is gone from all of our files.

rhamilto: Agreed, @spadgett.  But it will also prevent new noisy diffs going forward.  :)

benjaminapetersen: I think it will only create the noise in files you open & save, right?  So hopefully won't be insanely noisy...

rhamilto:
Good catch, @jwforres.  Clearly it's a rainy Friday afternoon.  ;-)

README.md
spadgett: > But it will also prevent new noisy diffs going forward. :)

Devil's advocate, but I've seen very few times where anyone has added trailing whitespace to a line they haven't changed. So it hasn't created noise in the diffs.

Trimming the whitespace always on save can cause whitespace merge conflicts, however, when there are no otherwise conflicting changes.

spadgett: > I think it will only create the noise in files you open & save, right?

Right, so potentially every file in your PR :)

benjaminapetersen: Small PRs for the win!! 😸 
Big PRs for the cries a lot.  😿

jwforres:
[merge]

jwforres:
[merge]

PR #514, opened 2016-09-06
Don't break back button navigating to error page

spadgett:
Without this fix, the back button almost always redirects you immediately back to the error page.

@jwforres


jwforres:
[merge]

jwforres:
[merge]

PR #516, opened 2016-09-06
Make sure latest data is always sent to callback when using cached data

jwforres:


jwforres:
@spadgett PTAL

cc @liggitt

app/scripts/services/data.js
```diff
@@ -334,27 +334,41 @@ angular.module('openshiftConsole')
...
-        }, 0);
-      }
-      else {
-        $timeout(function() {
+      $timeout(function() {
+        // re-pull the data from the cache in case it has changed on us from the watch
+        var data = this._data(key);
+        if (!data) {
+          // Extremely unlikely, would require that the watch was closed and other watches pushed
+          // this data out of the cache
```
jwforres: the alternative in this scenario is to go do the GET if we've lost the entire cached data map in the interim, but it means a significantly bigger refactor, if are going to cherry-pick this into 1.3.1 i'd feel better with a less risky refactor

app/scripts/services/data.js
```diff
@@ -578,9 +592,13 @@ DataService.prototype.createStream = function(resource, name, context, opts, isR
     else {
       if (callback) {
         var existingData = this._data(key);
+        var resourceVersion = this._resourceVersion(key);
         if (existingData) {
```
spadgett: I might even remove the existingData var 

```js
if (this._data(key)) ...
```

so no one changes it back?

jwforres:
updated

spadgett:
LGTM thanks

spadgett:
[merge]

Issue #515, opened 2016-09-06
Remove trigger button wraps to next line on mobile


spadgett:
![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/18279764/a8f87b12-7424-11e6-8f50-26ace8ee019b.png)

/cc @jhadvig

spadgett:
I suggest we remove the extra left margin in the subsections. I prefer it left aligned.

jhadvig:
@spadgett yeah that would make sense

PR #518, opened 2016-09-07
Remove trigger button wraps to next line on mobile

jhadvig:
Remove the extra left margin + set `max-width` for mobile. Screen:
screenshot-22

Tested it with different screen resolutions and should be ok for them.

@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/515


spadgett:
[merge]

Thanks @jhadvig

PR #527, opened 2016-09-08
Use Heapster rate metrics for CPU and network

### LINKED ISSUE #349: Metrics charts should not require samples every 30s
To calculate CPU and network usage rates, we require that heapster sample at least every 30s for a one hour chart. We should look at using Hawkular's rate APIs or getting the raw metrics data from Hawkular to remove this limitation.

We might want to make the time ranges customizable as well. For instance, if Hawkular is configured to keep metrics for fewer than 7 days, we shouldn't show the 7 days option in the time range dropdown.

/cc @mwringe @smarterclayton @DirectXMan12

spadgett:
Update to cpu/usage_rate, network/tx_rate, and network/rx_rate metrics

Fixes #349
@jwforres PTAL

### DIFF
--- a/app/scripts/directives/deploymentMetrics.js
+++ b/app/scripts/directives/deploymentMetrics.js
@@ -47,21 +47,21 @@ angular.module('openshiftConsole')
           label: "CPU",
           units: "millicores",
           chartPrefix: "cpu-",
-          descriptor: 'cpu/usage',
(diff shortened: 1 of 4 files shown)


spadgett:
@DirectXMan12 @mwringe FYI

jwforres:
[merge]

PR #530, opened 2016-09-12
WIP - enable PKCE

liggitt:
Switches the web console from using an implicit token flow to using an authorization code flow in combination with PKCE (a hash of a privately held "verifier" is bound to the authorization code, and the original verifier must be presented when exchanging the authorization code for a token).

TODO:
- [ ] tests for crypto pieces
- [ ] find sha256 hash algorithm implemented in javascript
- [ ] switch from plain PKCE to S256 PKCE
- [ ] dist build

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -21,6 +21,7 @@ window.OPENSHIFT_CONFIG = {
   },
   auth: {
     oauth_authorize_uri: "https://localhost:8443/oauth/authorize",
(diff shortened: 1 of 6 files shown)


liggitt:
superceded by https://github.com/openshift/origin-web-console/pull/1342 and https://github.com/openshift/origin-web-common/pull/7

PR #498, opened 2016-08-30
Remove nodeName from debug pod

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1371489

@jwforres PTAL

PR #532, opened 2016-09-13
Don't set spec.host when creating debug pod

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1371489

Don't set spec.host when creating debug pod. Unfortunately I missed this in my previous fix, #498. Also don't set label `debug.openshift.io/name`, which isn't used by `oc debug`.

@jwforres I did a diff on the JSON between the web console and CLI. I tested in staging by POSTing directly to the API server with `spec.host` removed, and it worked. PTAL


jwforres:
[merge]

PR #539, opened 2016-09-14
[WIP]: Initial cut of restoring topology view

rhamilto:
This needs more work, but is a starting point.  @jwforres and I believe the only way to resolve the orphaned tooltip issue below is to update [kubernetes-ui/topology-graph](https://github.com/kubernetes-ui/topology-graph).

Known issues:
-  Tooltips do not open in IE/Edge
-  Open tooltips are orphaned when the visibility of the parent is
  toggled off using the legend

To do:
- [ ] Add build assets

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -28,6 +28,39 @@ angular.module('openshiftConsole')
     $scope.renderOptions = $scope.renderOptions || {};
     $scope.renderOptions.showLoading = true;
     $scope.renderOptions.showGetStarted = false;
+    $scope.renderOptions.showToolbar = false;
+    $scope.renderOptions.overviewMode = 'tiles';
+
+    /*
+     * HACK: The use of <base href="/"> that is encouraged by angular is
+     * a cop-out. It breaks a number of real world use cases, including
+     * local xlink:href. Use location.href to get around it, even though
+     * these SVG <defs> are local in the template.
+     */
+
+     // must be named kinds to work with topology-icon directive
+    $scope.kinds = {
+      DeploymentConfig: location.href + "#vertex-DeploymentConfig",
+      Pod: location.href + "#vertex-Pod",
+      ReplicationController: location.href + "#vertex-ReplicationController",
+      Route: location.href + "#vertex-Route",
+      Service: location.href + "#vertex-Service"
+    };
+    // a separate map is required since the toplogy-icon directive deletes keys
+    // from the kinds map
+    $scope.legendKinds = {
+      DeploymentConfig: location.href + "#vertex-DeploymentConfig",
+      Pod: location.href + "#vertex-Pod",
+      ReplicationController: location.href + "#vertex-ReplicationController",
+      Route: location.href + "#vertex-Route",
+      Service: location.href + "#vertex-Service"
+    };
+
+    $scope.topologySelection = null;
+
+    /* Filled in by updateTopology */
+    $scope.topologyItems = { };
+    $scope.topologyRelations = [ ];
 
     $scope.alerts = $scope.alerts || {};
     AlertMessageService.getAlerts().forEach(function(alert) {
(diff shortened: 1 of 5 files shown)


benjaminapetersen:
Possible  fix the orphaned popovers is the focus trigger:
```javascript
placement: 'top',
trigger: 'manual focus' // can have multiple triggers
```
docs indicate that a click anywhere else (focus change) should [dismiss](http://getbootstrap.com/javascript/#dismiss-on-next-click).

app/scripts/controllers/overview.js
```diff
@@ -498,6 +538,173 @@ angular.module('openshiftConsole')
...
+              if (g.length) {
+                if (!(g.data('bs.popover'))) {
+                  g.popover({
+                    container: 'body',
+                    content: function() {
+                      return $filter('sentenceCase')($scope.topologySelection.kind) + ' <a href="' + $filter('navigateResourceURL')($scope.topologySelection) + '">' + $scope.topologySelection.metadata.name + '</a>';
+                    },
+                    html: true,
+                    placement: 'top',
+                    trigger: 'manual'
```
benjaminapetersen: to put my comment in context, maybe try updating this? `trigger: 'manual focus'`

app/scripts/controllers/overview.js
```diff
@@ -498,6 +538,173 @@ angular.module('openshiftConsole')
...
+
+        $scope.$on("select", function(ev, resource) {
+          $scope.$apply(function() {
+            $scope.topologySelection = resource;
+          });
+        }, true);
+
+        $scope.$watch('renderOptions.overviewMode', function(mode) {
+          if (mode === 'topology') {
+            $(window).on('click.topologyItem', function(evt) {
```
benjaminapetersen: You  might be able to move this to a directive & get the DOM stuff out of the controller, even if you add it as an attribute to the `div` wrapping the topology stuff:
```html
<!-- pass it the selected topology item -->
<div
    topology-popover="topologySelection" 
    ng-if="renderOptions.overviewMode === 'topology'
```

app/scripts/controllers/overview.js
```diff
@@ -498,6 +538,173 @@ angular.module('openshiftConsole')
...
+        var updateTimeout = null;
+
+        function updateTopology() {
+          updateTimeout = null;
+
+          var topologyRelations = [];
+          var topologyItems = { };
+
+          // Because metadata.uid is not unique among resources
+          function makeId(resource) {
```
benjaminapetersen: can `makeId` be moved up out of `updateTopology`? It prob only needs to be declared once & would shrink `updateTopology` a bit.

app/scripts/controllers/overview.js
```diff
@@ -498,6 +538,173 @@ angular.module('openshiftConsole')
...
+            return resource.kind + resource.metadata.uid;
+          }
+
+          // Add the services
+          angular.forEach($scope.services, function(service) {
+            topologyItems[makeId(service)] = service;
+          });
+
+          var isRecentDeployment = $filter('isRecentDeployment');
+          var isVisibleDeployment = function(deployment) {
```
benjaminapetersen: Same as above, `isVisibleDeployment` could probably be declared outside of `updateTopology`

app/scripts/controllers/overview.js
```diff
@@ -498,6 +538,173 @@ angular.module('openshiftConsole')
...
+            var dc = $scope.deploymentConfigs[dcName];
+            if (!dc) {
+              return false;
+            }
+
+            // Show the deployment if it's recent (latest or in progress) or if it's scalable.
+            return isRecentDeployment(deployment, dc) || $scope.isScalableDeployment(deployment);
+          };
+
+          // Add everything related to services, each of these tables are in
```
benjaminapetersen: It may also be nice/easier to maintain if we broke out each of these next sets of `[].forEach` loops into separate named functions.

app/scripts/controllers/overview.js
```diff
@@ -498,6 +538,173 @@ angular.module('openshiftConsole')
...
+                    return;
+                  }
+
+                  topologyItems[makeId(resource)] = resource;
+                });
+              }
+            });
+          });
+
+          // Things to link to services. Note that we can push as relations
```
benjaminapetersen: same, separate fn?

app/scripts/controllers/overview.js
rhamilto: We tried that, @benjaminapetersen.  Unfortunately, it didn't work.  :(  I can't remember why exactly...

PR #541, opened 2016-09-15
Overview metrics updates

spadgett:
- Show average per pod instead of total
- Don't rely on label selectors to query Hawkular metrics as these won't work for label expressions (used by replica sets)

@jwforres PTAL


app/views/directives/deployment-metrics.html
```diff
@@ -45,7 +45,7 @@
     </p>
   </div>
 
-  <div ng-repeat="metric in metrics" ng-if="!noData && !metricsError" class="metrics-full">
+  <div ng-repeat="metric in metrics" ng-show="!noData && !metricsError" class="metrics-full">
```
jwforres: whats the reason behind switching to show

spadgett: The div with the chart ID has to exist before calling c3.generate. It worked with ng-if before because we called $timeout in the controller, but this PR removes the $timeout.

jwforres:
[merge]

PR #556, opened 2016-09-20
A visualized graph of PV usage over time in the Metrics table

zhaosijun:
I am following the story [https://trello.com/c/6gpwW6uJ/253-pv-monitoring].

### DIFF
--- a/app/scripts/directives/podMetrics.js
+++ b/app/scripts/directives/podMetrics.js
@@ -20,7 +20,7 @@ angular.module('openshiftConsole')
       },
       templateUrl: 'views/directives/pod-metrics.html',
       link: function(scope) {
-        scope.includedMetrics = scope.includedMetrics || ["cpu", "memory", "network"];
+        scope.includedMetrics = scope.includedMetrics || ["cpu", "memory", "network", "volumes"];
         var donutByMetric = {}, sparklineByMetric = {};
         var intervalPromise;
         var getMemoryLimit = $parse('resources.limits.memory');
@@ -32,6 +32,8 @@ angular.module('openshiftConsole')
 
         // Set to true when the route changes so we don't update charts that no longer exist.
         var destroyed = false;
+        
+        var volumeDataSets = [];
 
         scope.uniqueID = _.uniqueId('metrics-chart-');
 
@@ -98,6 +100,20 @@ angular.module('openshiftConsole')
             ]
           });
         }
+        generateVolumeDataSets();
+        if (_.includes(scope.includedMetrics, "volumes")) {
+          scope.metrics.push({
+            label: "Volumes",
+            units: "MiB",
+            chartPrefix: "volumes-",
+            chartType: "spline",
+            convert: ConversionService.bytesToMiB,
+            // The sparkline y-axis will always extend to at least this value.
+            // Avoid spikey charts when rounding very small network usage values.
+            smallestYAxisMax: 10000,
+            datasets: volumeDataSets
+          });
+        }
 
         // Set to true when any data has been loaded (or failed to load).
         scope.loaded = false;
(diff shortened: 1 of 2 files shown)


jwforres:
cc @erinboyd

app/scripts/directives/podMetrics.js
```diff
@@ -130,6 +146,18 @@ angular.module('openshiftConsole')
         // Show last hour by default.
         scope.options.timeRange = _.head(scope.options.rangeOptions);
 
+        function generateVolumeDataSets() {
+          angular.forEach(scope.pod.spec.volumes, function(volumeData) {
+            if (volumeData.persistentVolumeClaim !== undefined) {
```
spadgett: Maybe just

```js
if (!volumeData.persistentVolumeClaim) {
  return;
}
```

and return early?

app/scripts/directives/podMetrics.js
```diff
@@ -130,6 +146,18 @@ angular.module('openshiftConsole')
         // Show last hour by default.
         scope.options.timeRange = _.head(scope.options.rangeOptions);
 
+        function generateVolumeDataSets() {
+          angular.forEach(scope.pod.spec.volumes, function(volumeData) {
```
jwforres: we've switched to using _.each instead of angular.forEach for any new code

app/scripts/directives/podMetrics.js
jwforres: for style consistency
```
if (volumeData.persistentVolumeClaim)
```
there's no reason to distinguish between null and undefined which are both falsey, if the persistentVolumeClaim sub-object exists then that if check will pass.  The only time we specifically check for null/undefined is when we need to distinguish between non-existence and zero integer values.  In those cases we have an isNil filter that handles both null and undefined.

app/scripts/directives/podMetrics.js
```diff
@@ -130,6 +146,18 @@ angular.module('openshiftConsole')
...
+            if (volumeData.persistentVolumeClaim !== undefined) {
+              volumeDataSets.push({
+                id: "filesystem/usage",
```
spadgett: I believe `filesystem/usage` is a pod-level metric. If so, we should update `getMetricType()` in `MetricsService`.

spadgett: Won't this return the same values for each volume if there is more than one? I don't see anything identifying the specific volume being passed to `MetricsService`.

app/scripts/directives/podMetrics.js
```diff
@@ -98,6 +100,20 @@ angular.module('openshiftConsole')
             ]
           });
         }
+        generateVolumeDataSets();
+        if (_.includes(scope.includedMetrics, "volumes")) {
```
spadgett: Only include the metric if `volumeDataSets.length`?

app/scripts/directives/podMetrics.js
zhaosijun: Thanks.

app/scripts/directives/podMetrics.js
zhaosijun: Thanks.

zhaosijun: In my opinion, the volumeDataSets will only include persistent Volumes which "persistentVolumeClaim" can identify.

app/scripts/directives/podMetrics.js
zhaosijun: I fetch all metric definitions and found  "filesystem/usage" is also a container-level metric just like "cpu/usage" and "memery/usage". 
In my opinion, the volumeDataSets will include all persistent volumes. The volumes graph will display all pv metrics.

zhaosijun: According to the "filesystem/usage" metric definitions, we can only get the pod filesystem usage. Maybe need to add the "pvc/usage" metric definitions.  I am following the cadvisor for this. Thanks.

app/scripts/directives/podMetrics.js
zhaosijun: Just like this. Thanks

app/scripts/directives/podMetrics.js
zhaosijun: In my opinion, pod share the persistent volumes with its replications. I found  it take "Downsampling" to get all deployment metrics to makes the user experience better. Pod mabye have many pvcs.

zhaosijun: yes. Only include the metric if volumeDataSets.length. Thanks.

sspeiche:
Having a hard time visualizing this, any screenshots?

PR #557, opened 2016-09-20
Don't request isimage more than once on page load

### LINKED ISSUE #548: ImageStreamResolver can make multiple requests for the same image stream image
See https://github.com/openshift/origin-web-console/issues/523#issuecomment-245369576

spadgett:
@jwforres The watch callback happens almost immediately after the get callback in the browse pages, which means we were calling `fetchReferencedImageStreamImages` a second time before the first call completed. I changed the code to just call it from the watch, and removed some unecessary `$scope` vars. PTAL

Partially addresses #548

### DIFF
--- a/app/scripts/controllers/deployment.js
+++ b/app/scripts/controllers/deployment.js
@@ -19,13 +19,13 @@ angular.module('openshiftConsole')
                         Navigate,
                         ProjectsService,
                         keyValueEditorUtils) {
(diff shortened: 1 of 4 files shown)


jwforres:
We can still get into race situations if we are say coming back from editing a DC, because the watch updates will fire more requests to get the ImageStreamImage

jwforres:
[merge]

PR #565, opened 2016-09-20
Group pods by replica sets on the overview

spadgett:
This adds replica sets to the overview. It does not group replica sets by deployment.

https://trello.com/c/HGvCTI96

openshift_web_console

@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -36,7 +36,16 @@ angular.module('openshiftConsole')
     AlertMessageService.clearAlerts();
 
     var watches = [];
-    var routes, services, deploymentConfigs, deployments, pods, buildConfigs, builds, horizontalPodAutoscalers, hpaByDC, hpaByRC;
+    var routes,
+        services,
+        deploymentConfigs,
+        deployments,
+        replicaSets,
+        pods,
+        buildConfigs,
+        builds,
+        horizontalPodAutoscalers,
+        hpaByResource;
 
     var isJenkinsPipelineStrategy = $filter('isJenkinsPipelineStrategy');
     var annotation = $filter('annotation');
@@ -130,30 +139,35 @@ angular.module('openshiftConsole')
       });
     };
 
+    var groupReplicaSets = function() {
+      if (!services || !replicaSets) {
+        return;
+      }
+
+      $scope.replicaSetsByService = DeploymentsService.groupByService(replicaSets, services);
+    };
+
     var groupHPAs = function() {
-      hpaByDC = {};
-      hpaByRC = {};
-      angular.forEach(horizontalPodAutoscalers, function(hpa) {
+      hpaByResource = {};
+      _.each(horizontalPodAutoscalers, function(hpa) {
         var name = hpa.spec.scaleRef.name, kind = hpa.spec.scaleRef.kind;
         if (!name || !kind) {
           return;
         }
 
-        switch (kind) {
-        case "DeploymentConfig":
-          hpaByDC[name] = hpaByDC[name] || [];
-          hpaByDC[name].push(hpa);
-          break;
-        case "ReplicationController":
-          hpaByRC[name] = hpaByRC[name] || [];
-          hpaByRC[name].push(hpa);
-          break;
-        default:
-          Logger.warn("Unexpected HPA scaleRef kind", kind);
+        // TODO: Handle groups and subresources in hpa.spec.scaleRef
+        // var groupVersion = APIService.parseGroupVersion(hpa.spec.scaleRef.apiVersion) || {};
+        // var group = groupVersion.group || '';
+        // if (!_.has(hpaByResource, [group, kind, name])) {
+        //   _.set(hpaByResource, [group, kind, name], []);
+        // }
+        // hpaByResource[group][kind][name].push(hpa);
+
+        if (!_.has(hpaByResource, [kind, name])) {
+          _.set(hpaByResource, [kind, name], []);
         }
+        hpaByResource[kind][name].push(hpa);
       });
-      $scope.hpaByDC = hpaByDC;
-      $scope.hpaByRC = hpaByRC;
     };
 
     // Filter out monopods we know we don't want to see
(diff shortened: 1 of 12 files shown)


app/scripts/services/deployments.js
```diff
@@ -367,7 +367,7 @@ angular.module("openshiftConsole")
       return scalableName === deployment.metadata.name;
     };
 
-    DeploymentsService.prototype.groupByService = function(/* deployments or deployment configs */ resources, services) {
+    DeploymentsService.prototype.groupByService = function(resources, services) {
```
jwforres: add a comment that its resources that replicate pods and have a selector defined

app/scripts/services/pods.js
```diff
@@ -40,6 +40,34 @@ angular.module("openshiftConsole")
...
+      //
+      // Returns: hash where the UID of the owner is the key and a hash of pods
+      //          is the value
+      groupByOwnerUID: function(pods, owners) {
+        var podsByOwnerUID = {};
+        // Look at each pod.
+        _.each(pods, function(pod) {
+          var found = false;
+          _.each(owners, function(owner) {
+            var selector = new LabelSelector(owner.spec.selector);
```
jwforres: instead of generating this selector for each owner for every single pod, can you do an initial pass on the owners and create a map of UID to LabelSelector object?

app/scripts/services/pods.js
```diff
@@ -40,6 +40,34 @@ angular.module("openshiftConsole")
...
+            if (selector.matches(pod)) {
+              _.set(podsByOwnerUID, [owner.metadata.uid, pod.metadata.name], pod);
+              found = true;
+              return false;
```
jwforres: i'm wondering now whether we should be breaking out early like this, I know this isn't new logic, but in theory you can have more than one owner for a pod if you end up with two controllers fighting over the same pod.  Fighting controllers will be fixed eventually but not in 1.4.  Frankly multiple pod owners should probably be showing up as a warning.

app/views/overview/_rc.html
```diff
@@ -1,9 +1,9 @@
-<div class="deployment-tile" ng-if="deployment.kind === 'ReplicationController'">
+<div class="deployment-tile">
```
jwforres: just curious, will this rc partial work for petset too now that its agnostic to kind?

app/scripts/controllers/overview.js
```diff
@@ -410,6 +397,19 @@ angular.module('openshiftConsole')
...
+      if (!horizontalPodAutoscalers) {
+        return null;
+      }
+
+      var kind = _.get(object, 'kind'),
+          name = _.get(object, 'metadata.name');
+      if (!_.has($scope, ['hpaByKind', kind, name])) {
+        _.set($scope, ['hpaByKind', kind, name], []);
+      }
+      return _.get($scope, ['hpaByKind', kind, name]);
```
spadgett: It avoids this error from the deploymentDonut $watch.


> angular.js:11706 Error: [$rootScope:infdig] 10 $digest() iterations reached. Aborting!
Watchers fired in the last 5 iterations: [[{"msg":"fn: regularInterceptedExpression","newVal":[],"oldVal":[]},{"msg":"hpa","newVal":"<<already seen>>","oldVal":"<<already seen>>"}],[{"msg":"fn: regularInterceptedExpression","newVal":[],"oldVal":"<<already seen>>"},{"msg":"hpa","newVal":"<<already seen>>","oldVal":"<<already seen>>"}],[{"msg":"fn: regularInterceptedExpression","newVal":[],"oldVal":"<<already seen>>"},{"msg":"hpa","newVal":"<<already seen>>","oldVal":"<<already seen>>"}],[{"msg":"fn: regularInterceptedExpression","newVal":[],"oldVal":"<<already seen>>"},{"msg":"hpa","newVal":"<<already seen>>","oldVal":"<<already seen>>"}],[{"msg":"fn: regularInterceptedExpression","newVal":[],"oldVal":"<<already seen>>"},{"msg":"hpa","newVal":"<<already seen>>","oldVal":"<<already seen>>"}]]

app/scripts/controllers/overview.js
```diff
@@ -188,12 +193,13 @@ angular.module('openshiftConsole')
...
 
     var groupPods = function() {
-      if (!pods || !deployments) {
+      if (!pods || !deployments || !replicaSets) {
         return;
       }
 
-      $scope.podsByDeployment = PodsService.groupByReplicationController(pods, deployments);
-      $scope.monopodsByService = PodsService.groupByService($scope.podsByDeployment[''], services, showMonopod);
+      var allOwners = _.toArray(deployments).concat(_.toArray(replicaSets));
```
spadgett: I did that originally. This seemed cleaner though and it has the same time complexity since we iterate over the list in the service anyway.

app/views/overview/_rc.html
spadgett: I think it will, although I haven't tested.

app/scripts/controllers/overview.js
jwforres: agree the code is cleaner as is, and not worried about time complexity as much as the extra memory allocation of concating the arrays vs an array of arrays, even then its probably not enough to even fret over, especially if we fix the watch debounce

app/scripts/controllers/overview.js
jwforres: fun, ok

spadgett: Let me see if there's a better fix.

spadgett:
> In general I wish we could have made even more of this generic, when we go to add petsets I'd love it if we didn't have to have another map of petSetsByService. That said, with the complexity of how RCs roll-up to DCs and RSs roll-up to Deps I'm not sure how realistic that goal is.

I wish we could generalize some of this logic. I was thinking of taking a swing at it when I group replica sets by deployment on the overview.

app/scripts/controllers/overview.js
```diff
@@ -347,50 +353,31 @@ angular.module('openshiftConsole')
...
-    };
-
     $scope.isScalableDeployment = function(deployment) {
       return DeploymentsService.isScalable(deployment,
-                                            deploymentConfigs,
-                                            $scope.hpaByDC,
-                                            $scope.hpaByRC,
-                                            $scope.scalableDeploymentByConfig);
+                                           deploymentConfigs,
+                                           _.get($scope, 'hpaByKind.DeploymentConfig'),
```
spadgett: Probably. Or make group be a separate key, group -> kind -> name -> HPAs

spadgett:
@jwforres updated

jwforres:
[merge]

Issue #528, opened 2016-09-08
Different icon representing storage on create-pvc page


rhamilto:
The icon used on the create-pvc page (fa-hdd-o) is a different icon than the one we use to denote storage in the primary nav (pficon-container-node).  Presumably, these should be the same icon since they're meaning the same thing.

![screen shot 2016-09-08 at 3 02 38 pm](https://cloud.githubusercontent.com/assets/895728/18362961/0bff2b0e-75d6-11e6-8501-6c5365f32fae.PNG)

![screen shot 2016-09-08 at 3 04 39 pm](https://cloud.githubusercontent.com/assets/895728/18362963/101a1104-75d6-11e6-9034-42c220e78a8c.PNG)

sg00dwin:
@rhamilto @jwforres  Which should be designated for storage? The nav uses pficon-container-node and there's also a pficon-storage-domain

http://www.patternfly.org/styles/icons/

jwforres:
@ajacobs21e

ajacobs21e:
@jwforres @sg00dwin @rhamilto  I'm not familiar enough with the terminology to know which icon makes more sense. Is one more intuitive?

ajacobs21e:
do any of these make more sense?
http://www.patternfly.org/styles/icons/cheatsheet/#_

pficon-volume
fa-database
pficon-cluster

PR #572, opened 2016-09-21
Removing large icon from create pages based on discussion

sg00dwin:
fixes https://github.com/openshift/origin-web-console/issues/528

I kept the form content centered, but switched width to col-md-10 from col-md-8 to account for removing the icon column.

### DIFF
--- a/app/views/attach-pvc.html
+++ b/app/views/attach-pvc.html
@@ -37,11 +37,7 @@
                 </div>
 
                 <div class="row" ng-show="pvcs && pvcs.length && attach.resource">
-                  <div class="create-storage-icon col-md-2 gutter-top hidden-sm hidden-xs">
-                    <span class="fa fa-hdd-o icon-xl"></span>
-                  </div>
-
-                  <div class="col-md-8">
+                  <div class="col-md-10 col-md-offset-1 gutter-top">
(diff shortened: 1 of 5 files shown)


sg00dwin:
@rhamilto @jwforres

app/views/create/fromimage.html
```diff
@@ -14,10 +14,7 @@
               <div class="col-md-12">
                 <div class="osc-form">
                   <div class="row">
-                    <div class="col-md-2 template-name gutter-top hidden-sm hidden-xs">
-                      <span class="fa fa-cubes"></span>
-                    </div>
-                    <div class="col-md-8">
```
spadgett: This might be the only one I keep. If we remove it from this page, we should remove it from the template page as well. But I'd rather use the technology icon for the builder / template.

@jwforres thoughts?

jwforres:
yeah i agree, lets keep these and finally actually fix them to use the
customized icon that you see in the catalog

jwforres:
or at least dont remove them completely, id be ok with coming up with an
alternative design for placement / size of the technology icon on these two
pages

sg00dwin:
@jwforres updated pr, reverted app/views/create/fromimage.html

spadgett:
[merge]

PR #573, opened 2016-09-22
Bug 1374165: Unable to upload the same file the second time after clear in ace editor

jhadvig:
We have to clear the `input[type=file]` value after clearing the filed with the `Clear value` button, since the value will stay the same the `change` event wont be triggered.

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/oscFileInput.js
+++ b/app/scripts/directives/oscFileInput.js
@@ -24,7 +24,8 @@ angular.module('openshiftConsole')
         dropZoneName = scope.dropZoneId + "-drag-and-drop-zone",
         dropZoneSelectorName = "#" + dropZoneName,
         highlightDropZone = false,
-        showDropZone = false;
+        showDropZone = false,
+        inputFileField = element.find('input[type=file]')[0];
 
         // Add/Remove dropZone based on if the directive element is disabled
         scope.$watch('disabled', function() {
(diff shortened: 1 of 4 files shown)


app/scripts/directives/oscFileInput.js
```diff
@@ -68,8 +69,12 @@ angular.module('openshiftConsole')
           });
         }
 
+        $('.clear-btn').on('click', function() {
```
jwforres: 1. you should scope the `$('.clear-btn')` to only look within the range of this directive's DOM elements, if there were more than one of these on the page you wouldn't want them to affect each other.  To scope it you just pass the element as the second parameter `$('.clear-btn', element)`
2. we are trying to get into the habit of scoping calls to jquery on/off.  This lets you have multiple click handlers on the same node that can be independently turned off.  To scope a click handler it looks something like this `$('.clear-btn', element).on('click.fileInputClear')`   notice the `click.fileInputClear`  the first part before the `.` is the event, the part after is whatever you want to call your scope
3. If you add a jquery event handler in a directive you should clean it up within the `scope.$on('$destroy'` section, so add this `$('.clear-btn', element).off('click.fileInputClear')`

spadgett: @jhadvig We should call a scope function from the existing ng-click handlerrather than using a jQuery event handler. Not good to mix these.

jhadvig:
@jwforres @spadgett thanks for comments.. will remember those.
PTAL

spadgett:
[merge]

PR #575, opened 2016-09-22
Clean up watch on BC editor scope destroy

### LINKED ISSUE #574: Build config editor doesn't unwatch build config on scope destroy
The build config editor does not clean up a watch it opens. This is problem in particular for IE, which limits how many websockets can be open.

/cc @jwforres

spadgett:
Fixes #574
@jwforres PTAL


spadgett:
Swept the code, and this is the only place I saw where we missed the unwatch.

jwforres:
[merge]

Issue #529, opened 2016-09-09
Runtime error importing YAML for resource you don't have authority to GET


spadgett:
I hadn't run reconcile cluster roles, so I didn't have authority to do anything with pet sets. It looks like it tries to update the resource on any non-404 GET errors, which fails.

<img width="711" alt="screen shot 2016-09-09 at 7 52 08 am" src="https://cloud.githubusercontent.com/assets/1167259/18386092/87e91932-7662-11e6-9223-285f88a68f31.png">

/cc @jwforres

jhadvig:
@spadgett could you please share the steps how to reproduce ? Thanks :)

spadgett:
One way to see it is to try to create a PV as a normal user.

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: pv0003
spec:
  capacity:
    storage: 5Gi
  accessModes:
    - ReadWriteOnce
  persistentVolumeReclaimPolicy: Recycle
  nfs:
    path: /tmp
    server: 172.17.0.2
```

PR #576, opened 2016-09-22
Runtime error importing YAML for resource you don't have authority to GET

jhadvig:
since the check is async. we have to check for error occurrence after all check are done.
@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/529


spadgett:
[merge]

PR #582, opened 2016-09-22
Persist tab state without adding to browser history

jwforres:


jwforres:
@spadgett WDYT of this change?  I can't think of a reason that reloadOnSearch matters on the pages where we have tabs or filters.

spadgett:
+1 `reloadOnSearch: false` seems like exactly what we want

benjaminapetersen:
@jwforres, i think `uib-tab` supports `select=` as a callback, so something like:
```html
select="selectTab('environment')"
```
and the JS could just be:
```javascript
$scope.selectTab = function(what) {
      $location.replace().search('tab', what);
};
```
might be nice, less code, no additional $watch?

jwforres:
@benjaminapetersen in the version we have its on each tab, it adds it to every tab in the view

jwforres:
I agree its less code for the controller, but it doesn't feel right to me to make each tab in the view responsible for making sure the callback gets fired.  In my ideal world the uib-tabset element should have had a callback option for when any of its tabs are selected

benjaminapetersen:
@jwforres, yup, each would get `<uib-tab select="onSelect('tab-name')"`

benjaminapetersen:
Hmm. I think if our views are declarative & we get all the work into a service its a win, but I'm good if you prefer the current approach!

jwforres:
@spadgett can be the deciding vote

benjaminapetersen:
Cool cool!  I'll readily admit I'm an awful controller troll/nag :smile:

jwforres:
@benjaminapetersen how do you feel about the latest version

app/scripts/directives/util.js
```diff
@@ -178,4 +178,24 @@ angular.module('openshiftConsole')
         }
       });
     };
+  })
+  .directive('persistTabState', function($routeParams, $location) {
```
spadgett: Probably worth a comment saying `reloadOnSearch` should be false for the route to use this directive.

benjaminapetersen:
Oooo nice, even less in the controllers!  +1
Only thought, perhaps put it on the tabset itself for context:
```html
<uib-tabset class="mar-top-none" persist-tab-state>
```
Either way though, def looks good to me.

jwforres:
[merge]

PR #585, opened 2016-09-23
Bug 1378851 - Unable to update PushTo section for ImageStreamTag in BC editor

jhadvig:
The bug was causing by runtime error, since we need to check if the namespace & is & tag are not empty, why are the default values.

@spadgett PTAL

### DIFF
--- a/app/scripts/directives/istagSelect.js
+++ b/app/scripts/directives/istagSelect.js
@@ -35,9 +35,7 @@ angular.module("openshiftConsole")
         $scope.isNamesByNamespace = {};
 
         // Check if the istag object contains data about namespace/imageStream/tag so the ui-select will be pre-populated with them
-        var shouldPrepopulate = _.has($scope, 'istag.namespace') &&
-                                _.has($scope, 'istag.imageStream') &&
-                                _.has($scope, 'istag.tagObject.tag');
+        var shouldPrepopulate = _.get($scope, 'istag.namespace') && _.get($scope, 'istag.imageStream') && _.get($scope, 'istag.tagObject.tag');
 
         var prepopulate = function(ns) {
           $scope.isByNamespace[ns] = {};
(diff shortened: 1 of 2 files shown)


app/scripts/directives/istagSelect.js
```diff
@@ -35,9 +35,12 @@ angular.module("openshiftConsole")
         $scope.isNamesByNamespace = {};
 
         // Check if the istag object contains data about namespace/imageStream/tag so the ui-select will be pre-populated with them
-        var shouldPrepopulate = _.has($scope, 'istag.namespace') &&
-                                _.has($scope, 'istag.imageStream') &&
-                                _.has($scope, 'istag.tagObject.tag');
```
spadgett: Should be able to simply change `_.has` to `_.get`

jhadvig:
I thought so as well, but we need to check whether the ns & is & tag are set and not just empty values `""` since those are the default values when IST is not set.

spadgett:
> _.get will get the value but not check for its actual value

`_.get` will return `""`, which evaluates to false

jhadvig:
@spadgett updated.. thanks ! PTAL

app/scripts/directives/istagSelect.js
```diff
@@ -35,9 +35,7 @@ angular.module("openshiftConsole")
...
+        var shouldPrepopulate = (_.get($scope, 'istag.namespace') && _.get($scope, 'istag.imageStream') && _.get($scope, 'istag.tagObject.tag')) ? true : false;
```
spadgett: @jhadvig I find this harder to understand than before. Generally you shouldn't have

```js
var foo = <expression> ? true : false;
```

It should usually just be

```js
var foo = <expression>;
```

I'd keep it like it was before, but replacing `_.has` with `_.get`.

jhadvig: but the problem is this expression with `_.get` will return string and not bool.

spadgett: But that's OK. A non-empty string is truthy in JavaScript.

https://developer.mozilla.org/en-US/docs/Glossary/Truthy

jhadvig: I understand that, but thought we want to have the `shouldPrepopulate` value set explicitely to bool. 
Will update due to your suggestion. Thanks !

jhadvig:
@spadgett updated. PTAL

spadgett:
[merge]

PR #586, opened 2016-09-23
Bug 1378703: No need to show error about editing again when bc is deleted

jhadvig:
Wasn't sure if we shall do it this way(deleting the opposing alert) or just nuke all alerts... but would rather go with this one since its safer.

@spadgett PTAL

### DIFF
--- a/app/scripts/controllers/edit/buildConfig.js
+++ b/app/scripts/controllers/edit/buildConfig.js
@@ -198,13 +198,13 @@ angular.module('openshiftConsole')
             // If we found the item successfully, watch for changes on it
             watches.push(DataService.watchObject("buildconfigs", $routeParams.buildconfig, context, function(buildConfig, action) {
               if (action === 'MODIFIED') {
-                $scope.alerts["background_update"] = {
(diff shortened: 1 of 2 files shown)


spadgett:
I mean the same alert ID for updated and deleted.

jhadvig:
@spadgett updated

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -198,13 +198,13 @@ angular.module('openshiftConsole')
             // If we found the item successfully, watch for changes on it
             watches.push(DataService.watchObject("buildconfigs", $routeParams.buildconfig, context, function(buildConfig, action) {
               if (action === 'MODIFIED') {
-                $scope.alerts["background_update"] = {
+                $scope.alerts["updated/created"] = {
```
spadgett: `"updated/deleted"`?

jhadvig:
@spadgett sorry for that... had a long day
PTAL

spadgett:
[merge]

PR #587, opened 2016-09-23
Improve loading message position on browse pages

spadgett:
Avoid having the header jump after the page loads and keep a consistent top margin below the breadcrumbs when there's no header.

@jwforres PTAL


jwforres:
so some of them you moved below middle-content and others you left under the alerts, whats the difference between those

spadgett:
I moved the ones below only when the header is displayed before the content loads. So it looks like

my-build-config
Loading...

instead of

Loading...
my-build-config

It keeps the header from jumping when the loading message disappears.

jwforres:
[merge]

PR #588, opened 2016-09-23
Fix runtime error from labels on create from image page

### LINKED ISSUE #512: Runtime error on nextSteps page only when a java app is selected
To reproduce: 
- follow the new project flow
- choose one of the java apps (`wildfly 10`, `wildfly 9`, `wildfly 8`)
- enter a name & use the sample git repository

The following error will appear in the console:
 
screen shot 2016-09-02 at 4 19 33 pm

This only happens with a java app.  I suspect it is passing an entries list as an `{ }` instead of an `[ ]` at some point, but I have not yet tracked it down.  It appears to happen right at the button click (API call), so I suspect it may have something to do with mapping an `[ ]` back to an `{ }` (labels?).

@jwforres @spadgett

benjaminapetersen:
Fix #512 

@jwforres @spadgett

### DIFF
--- a/app/scripts/controllers/create/createFromImage.js
+++ b/app/scripts/controllers/create/createFromImage.js
@@ -78,7 +78,7 @@ angular.module("openshiftConsole")
             include: true,
             portOptions: []
           };
-          scope.labels = [];
+          scope.userDefinedLabels = [];
           scope.systemLabels = [appLabel];
           scope.annotations = {};
           scope.scaling = {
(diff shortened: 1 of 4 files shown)


spadgett:
[merge]

benjaminapetersen:
ah, @spadgett I commented on the original issue, I think the other ones are fine, the output of the transformation is always assigned to a new object, not back to the original `$scope.labels`, so no type issue.

spadgett:
ok thanks [merge]

PR #592, opened 2016-09-26
For #556 A visualized graph of PV usage over time in the Metrics table

zhaosijun:
@spadgett In order to get pod pv metrics I set heapster source argument  "kubernetes.summary_api" and add metrics provider to the nfs volume plugin.  Refer to these links. [https://trello.com/c/6gpwW6uJ/253-pv-monitoring]  [https://github.com/kubernetes/kubernetes/pull/33468]  [https://github.com/openshift/origin-metrics/pull/228]


app/scripts/directives/podMetrics.js
```diff
@@ -130,6 +146,19 @@ angular.module('openshiftConsole')
...
+        function generateVolumeDataSets() {
+          _.each(scope.pod.spec.volumes, function(volumeData) {
+            if (!volumeData.persistentVolumeClaim) {
+              return;
+            }
+            volumeDataSets.push({
+              id: "filesystem/usage",
+              label: volumeData.name,
+              data: []
+            });
```
spadgett: It still looks to me like the metrics request for each volume will return the same data, so you'll have a multiline chart with identical lines if you have more than one volume. Can you check this case?

If you can show usage per volume, great. If we can't, we should make this a simple area chart with total usage.

zhaosijun:
@spadgett  PTAL Thanks
image

zhaosijun:
I excute these commands
oc rsh hawkular-cassandra-1-dy07t
du -mh /cassandra_data 
du -mh /cassandra_data_test
image

spadgett:
@zhaosijun Is this pull request replacing #556?

We won't be able to merge anything that relies on upstream changes until they're are available in the origin repository.

zhaosijun:
@spadgett yes.

mwringe:
Is there anything else that needs to be done in Origin Metrics for this to be merged? The requested change there has already been merged.

zhaosijun:
@mwringe  The request  relies on upstream kubernetes. I think this request can be closed. Afrer the upstream kubernetes  are available I can follow this requested change.

mwringe:
@zhaosijun I am confused, Heapster is now reporting the volume usages. What else needs to be done with upstream Kubernetes? Can you point me to the issue issue?

zhaosijun:
@mwringe Last months I have followed the source of  volunme usages. The volume usages comes fromsummary metrics that provided by kubelet server. Refer to [https://github.com/openshift/origin-metrics/pull/228](url)
The volume metrcis is provided by the volume plugins in the kubernetes. And most volume plugins such as nfs compute the metrics by the command 'du'. Refer to [https://github.com/kubernetes/kubernetes/pull/33468](url)    
I think the volume metrics can be provided by other ways. This request need not to relies on upstream kubernetes.

theute:
@zhaosijun We would still need to collect PV metrics, I see that the kube PR has been closed for inactivity. Could you please explain where we stand now ?

zhaosijun:
@theute I think collecting PV metrics in kubernetes is heavyweigh. I think the volume metrics can be provided by other ways. This request need not to relies on upstream kubernetes.

mwringe:
The metrics are already collected by Heapster at the moment and stored into Hawkular. So the metrics are already available to the console to use.

theute:
Ok, cool !

PR #593, opened 2016-09-26
Bug 1378716 - Invalid url can be saved by editing bc page

jhadvig:
@jwforres PTAL

### DIFF
--- a/app/views/edit/build-config.html
+++ b/app/views/edit/build-config.html
@@ -40,7 +40,7 @@
                                   required>
                               </div>
                               <div>
-                                <span class="text-warning" ng-if="form.sourceUrl.$dirty && !sourceURLPattern.test(updatedBuildConfig.spec.source.git.uri)">Git repository should be a URL.</span>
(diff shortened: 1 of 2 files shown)


app/views/edit/build-config.html
```diff
@@ -472,7 +472,7 @@
                     <button
                       type="submit"
                       class="btn btn-primary btn-lg"
-                        ng-disabled="form.$invalid || form.$pristine || disableInputs">
+                        ng-disabled="form.$invalid || form.$pristine || disableInputs || isSourceURLInvalid()">
```
spadgett: We made this a warning rather than an error on the "add to project" page. Suggest the same here. It's a complicated regex, and in at least one case, we had it wrong. I don't think we should block users from submitting the form.

spadgett: In other words, I don't agree with the bug (unless the warning is not showing up). Added some comments in Bugzilla.

jhadvig: Yeah and I think that there was a similar issue in the past, but couldnt find it. I can close this one the and put the BZ on NOTABUG if you agree @spadgett ?

jhadvig: I think the current warning is sufficient. Closing this one.

PR #577, opened 2016-09-22
Separate edit YAML page

spadgett:
This PR changes the YAML editor to be a full page rather than a dialog. This gives the editor more space as well as making it bookmarkable.

Fixes https://github.com/openshift/origin/issues/6845

screen shot 2016-09-22 at 11 28 21 am

@jwforres PTAL

PR #594, opened 2016-09-26
Fix Edit YAML action for replica sets

spadgett:
Regression from #577

@jwforres PTAL. I checked all of the other pages, and this is the only one broken.


jwforres:
[merge]

PR #602, opened 2016-09-27
Check basic quota problems before new-app from image or deploy image

jwforres:
Partially implements https://trello.com/c/mfeqgKvL

### DIFF
--- a/app/scripts/controllers/create/createFromImage.js
+++ b/app/scripts/controllers/create/createFromImage.js
@@ -13,10 +13,12 @@ angular.module("openshiftConsole")
       LimitRangesService,
       MetricsService,
       HPAService,
+      QuotaService,
       TaskList,
       failureObjectNameFilter,
       $filter,
       $parse,
+      $uibModal,
       SOURCE_URL_PATTERN,
       keyValueEditorUtils
     ){
@@ -184,6 +186,19 @@ angular.module("openshiftConsole")
           $scope.showCPURequestWarning = !HPAService.hasCPURequest([$scope.container], $scope.limitRanges, project);
         };
 
+        var quotas, clusterQuotas;
+
+        DataService.list("resourcequotas", context, function(quotaData) {
+          quotas = quotaData.by("metadata.name");
+          Logger.log("quotas", quotas);
+        });
+
+        // TODO clean up anything not needed here
+        DataService.list("appliedclusterresourcequotas", context, function(clusterQuotaData) {
+          clusterQuotas = clusterQuotaData.by("metadata.name");
+          Logger.log("cluster quotas", clusterQuotas);
+        });
+
         $scope.$watch('scaling.autoscale', checkCPURequest);
         $scope.$watch('container', checkCPURequest, true);
         $scope.$watch('name', function(newValue) {
@@ -192,56 +207,8 @@ angular.module("openshiftConsole")
 
         initAndValidate($scope);
 
-        var ifResourcesDontExist = function(apiObjects, namespace){
-          var result = $q.defer();
-          var successResults = [];
-          var failureResults = [];
-          var remaining = apiObjects.length;
-
-          function _checkDone() {
-            if (remaining === 0) {
-              if(successResults.length > 0){
-                //means some resources exist with the given nanme
-                result.reject(successResults);
-              }
-              else {
-                //means no resources exist with the given nanme
-                result.resolve(apiObjects);
-              }
-            }
-          }
-
-          apiObjects.forEach(function(apiObject) {
-            var resource = APIService.objectToResourceGroupVersion(apiObject);
-            if (!resource) {
-              failureResults.push({data: {message: APIService.invalidObjectKindOrVersion(apiObject)}});
-              remaining--;
-              _checkDone();
-              return;
-            }
-            if (!APIService.apiInfo(resource)) {
-              failureResults.push({data: {message: APIService.unsupportedObjectKindOrVersion(apiObject)}});
-              remaining--;
-              _checkDone();
-              return;
-            }
-            DataService.get(resource, apiObject.metadata.name, {namespace: (namespace || $routeParams.project)}, {errorNotification: false}).then(
-              function (data) {
-                successResults.push(data);
-                remaining--;
-                _checkDone();
-              },
-              function (data) {
-                failureResults.push(data);
-                remaining--;
-                _checkDone();
-              }
-            );
-          });
-          return result.promise;
-        };
-
-        var createResources = function(resources){
+        var generatedResources;
+        var createResources = function(){
           var titles = {
             started: "Creating application " + $scope.name + " in project " + $scope.projectDisplayName(),
             success: "Created application " + $scope.name + " in project " + $scope.projectDisplayName(),
@@ -252,7 +219,7 @@ angular.module("openshiftConsole")
           TaskList.clear();
           TaskList.add(titles, helpLinks, $routeParams.project, function(){
             var d = $q.defer();
-            DataService.batch(resources, context)
+            DataService.batch(generatedResources, context)
               //refactor these helpers to be common for 'newfromtemplate'
               .then(function(result) {
                     var alerts = [];
@@ -297,9 +264,42 @@ angular.module("openshiftConsole")
           Navigate.toNextSteps($scope.name, $scope.projectName, $scope.usingSampleRepo() ? {"fromSample": true} : null);
         };
 
-        var elseShowWarning = function(){
-          $scope.nameTaken = true;
-          $scope.disableInputs = false;
+        var launchConfirmationDialog = function(alerts) {
+          var modalInstance = $uibModal.open({
+            animation: true,
+            templateUrl: 'views/modals/confirm.html',
+            controller: 'ConfirmModalController',
+            resolve: {
+              modalConfig: function() {
+                return {
+                  alerts: alerts,
+                  message: "Problems were detected while checking your application configuration.",
+                  okButtonText: "Create Anyway",
+                  okButtonClass: "btn-danger",
+                  cancelButtonText: "Cancel"
+                };
+              }
+            }
+          });
+
+          modalInstance.result.then(createResources);
+        };
+
+        var showWarningsOrCreate = function(result){
+          // Now that all checks are completed, show any Alerts if we need to
+          var quotaAlerts = result.quotaAlerts || [];
+          var errorAlerts = _.filter(quotaAlerts, {type: 'error'});
+          if ($scope.nameTaken || errorAlerts.length) {
+            $scope.disableInputs = false;
+            $scope.alerts = quotaAlerts;
+          }
+          else if (quotaAlerts.length) {
+             launchConfirmationDialog(quotaAlerts);
+             $scope.disableInputs = false;
+          }
+          else {
+            createResources();
+          }
         };
 
         $scope.projectDisplayName = function() {
@@ -308,6 +308,7 @@ angular.module("openshiftConsole")
 
         $scope.createApp = function(){
           $scope.disableInputs = true;
+          $scope.alerts = {};
           $scope.buildConfig.envVars = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries($scope.buildConfigEnvVars));
           $scope.deploymentConfig.envVars = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries($scope.DCEnvVarsFromUser));
           var userLabels = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries($scope.userDefinedLabels));
@@ -316,15 +317,23 @@ angular.module("openshiftConsole")
 
           var resourceMap = ApplicationGenerator.generate($scope);
           //init tasks
-          var resources = [];
+          generatedResources = [];
           angular.forEach(resourceMap, function(value){
             if(value !== null){
               Logger.debug("Generated resource definition:", value);
-              resources.push(value);
+              generatedResources.push(value);
             }
           });
-          ifResourcesDontExist(resources, $scope.projectName, $scope)
-            .then(createResources, elseShowWarning);
+
+          var nameTakenPromise = ApplicationGenerator.ifResourcesDontExist(generatedResources, $scope.projectName);
+          var checkQuotaPromise = QuotaService.getLatestQuotaAlerts(generatedResources, context);
+          // Don't want to wait for the name checks to finish before making the calls to quota
+          // so kick off the requests above and then chain the promises here
+          var setNameTaken = function(result) {
+            $scope.nameTaken = result.nameTaken;
+            return checkQuotaPromise;
+          };
+          nameTakenPromise.then(setNameTaken, setNameTaken).then(showWarningsOrCreate, showWarningsOrCreate);
         };
       }));
   });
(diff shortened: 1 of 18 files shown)


jwforres:
@spadgett ready for a first pass.  I need to double check the isTerminating and isBestEffort are working correctly.

app/scripts/filters/resources.js
```diff
@@ -1274,15 +1274,36 @@ angular.module('openshiftConsole')
...
+          return !isNilFilter(request) && usageValueFilter(request) !== 0;
+        });
+        var hasNonZeroLimit = _.some(_.get(container, 'resources.limits'), function(limit){
+          return !isNilFilter(limit) && usageValueFilter(limit) !== 0;
+        });
+        return hasNonZeroRequest || hasNonZeroLimit;
+      });
+    };
+  })
+  .filter('isTerminatingPod', function(isNilFilter) {
```
spadgett: We have an existing `isTerminating` filter, but with different logic.

https://github.com/spadgett/origin-web-console/blob/e8e3ecdc1dbe23297674577da7180d48b0c6a3c2/app/scripts/filters/resources.js#L850-L854

Should we update the existing filter instead?

app/scripts/filters/resources.js
```diff
@@ -1274,15 +1274,36 @@ angular.module('openshiftConsole')
...
+    return function(pod) {
+      // a pod is terminating if activeDeadlineSeconds is set, ADS can be zero
+      return !isNilFilter(pod.spec.activeDeadlineSeconds);
```
spadgett: Maybe

```js
return _.has(pod, 'spec.activeDeadlineSeconds');
```

?

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+              matches = !terminatingPod;
+              break;
+            case "BestEffort":
+              matches = bestEffortPod;
+              break;
+            case "NotBestEffort":
+              matches = !bestEffortPod;
+              break;
+          }
+          return !matches;
```
spadgett: I think this is easier to understand with `_.every`. It should break out early as well. Maybe

```js
var matchesPod = function(scope) {
  switch(scope) {
    case "Terminating":
      return terminatingPod;
    case "NotTerminating":
      return !terminatingPod;
    case "BestEffort":
      return bestEffortPod;
    case "NotBestEffort":
      return !bestEffortPod;
  }
  return true;
};
var scopes = quota.spec.quota ? quota.spec.quota.scopes : quota.spec.scopes;
return _.every(scopes, matchesPod);
```

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+        return filterQuotasForPod(resource, quotas);
+      }
+      // We plan on having other resources that scopes will affect
+      return quotas;
+    };
+
+    var humanizeQuotaResource = $filter('humanizeQuotaResource');
+    var humanizeKind = $filter('humanizeKind');
+
+    var getResourceLimitWarnings = function(resource, quota){
```
spadgett: `getResourceLimitAlerts`? Since some of these are errors and we return an alerts array.

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+          }
+          else {
+            details = "You can can still create " + humanizeKind(resource.kind) + " '" + resource.metadata.name + "' but no pods will be created until resources are freed.";
+          }
+          warnings.push({
+            type: resource.kind === 'Pod' ? 'error' : 'warning',
+            message: 'You are at your quota for ' + humanizeQuotaResource(type).toLowerCase() + ' on pods.',
+            details: details,
+            links: [{
+              href: "/project/" + quota.metadata.namespace + "/quota",
```
spadgett: Should this have the leading `/`?

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+              target: "_blank"
+            }]
+          });
+        }
+      });
+
+      return warnings;
+    };
+
+    var getQuotaWarnings = function(resources, quotas, clusterQuotas) {
```
spadgett: `getQuotaAlerts` ?

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+        var humanizedResource = APIService.kindToResource(resource.kind, true);
+        var humanizedKind = humanizeKind(resource.kind);
+        var quotaKey = "";
+        if (rgv.group) {
+          quotaKey = rgv.group + "/";
+        }
+        quotaKey += rgv.resource;
+        var warningsForQuota = function(quota) {
+          var q = quota.status.total || quota.status;
+          if(q.hard[quotaKey] !== undefined && q.hard[quotaKey] === q.used[quotaKey]) {
```
spadgett: `!isNil` here as well instead of `!== undefined`? For consistency with `getResourceLimitWarnings`.

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+            warnings.push({
+              type: 'error',
+              message: "You are at your quota of " + q.hard[quotaKey] + " " + (q.hard[quotaKey] === "1" ? humanizedKind : humanizedResource) +
+                        " in this project.",
+              details: "You will not be able to create the " + humanizedKind + " '" + resource.metadata.name + "'.",
+              links: [{
+                href: "/project/" + quota.metadata.namespace + "/quota",
```
spadgett: Remove leading `/`?

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+          }
+
+          warnings = warnings.concat(getResourceLimitWarnings(resource, quota));
+        };
+        _.each(filteredQuotas, function(quota) {
+          warningsForQuota(quota);
+        });
+        _.each(filteredClusterQuotas, function(quota) {
+          warningsForQuota(quota);
+        });
```
spadgett:
```js
_.each(filteredQuotas, warningsForQuota);
_.each(filteredClusterQuotas, warningsForQuota);
```

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+
+      return warnings;
+    };
+
+    var getLatestQuotaWarnings = function(resources, context) {
+      var result = $q.defer();
+      var quotaWarnings = [];
+      var quotas, clusterQuotas;
+      var remaining = 2;
+      function _checkDone() {
```
spadgett: Not for this PR, but I'd really like to switch `DataService.list` to return a promise. Then we can use `$q.all` here instead of `_checkDone`.

app/scripts/services/quota.js
```diff
@@ -0,0 +1,185 @@
...
+      });
+
+      DataService.list("appliedclusterresourcequotas", context, function(clusterQuotaData) {
+        clusterQuotas = clusterQuotaData.by("metadata.name");
+        Logger.log("cluster quotas", clusterQuotas);
+        remaining--;
+        _checkDone();
+      });
+
+      return result.promise;
```
spadgett: Will this promise ever get resolved on list errors?

app/scripts/services/quota.js
jwforres: I actually don't want to switch to $q.all because of the issue with it rejecting on the first rejection, I prefer we actually wait for all rejections to happen before we redirect.  I was actually planning on switching the $q.all thats in from yaml/json

app/scripts/services/quota.js
jwforres: Fixed

app/scripts/services/quota.js
jwforres: Fixed

app/scripts/directives/deployImage.js
```diff
@@ -201,6 +205,62 @@ angular.module("openshiftConsole")
...
+
+            modalInstance.result.then(function() {
+              createResources();
+            });
+          };
+
+          var showWarningsOrCreate = function(result){
+            // Now that all checks are completed, show any warnings if we need to
+            var quotaWarnings = result.quotaWarnings || [];
+            var errorAlerts = _.filter(quotaWarnings, {type: 'error'}) || [];
```
jwforres: Fixed

app/scripts/directives/deployImage.js
```diff
@@ -201,6 +205,62 @@ angular.module("openshiftConsole")
...
+                    message: "Problems were detected while checking your application configuration.",
+                    okButtonText: "Create Anyway",
+                    okButtonClass: "btn-danger",
+                    cancelButtonText: "Cancel"
+                  };
+                }
+              }
+            });
+
+            modalInstance.result.then(function() {
```
jwforres: Fixed

app/scripts/directives/deployImage.js
```diff
@@ -201,6 +205,62 @@ angular.module("openshiftConsole")
...
+
+            var nameTakenPromise = ApplicationGenerator.ifResourcesDontExist(generatedResources, $scope.project);
+            var checkQuotaPromise = QuotaService.getLatestQuotaWarnings(generatedResources, $scope.context);
+            // Don't want to wait for the name checks to finish before making the calls to quota
+            // so kick off the requests above and then chain the promises here
+            var setNameTaken = function(result) {
+              $scope.nameTaken = result.nameTaken;
+              return checkQuotaPromise;
+            };
+            nameTakenPromise.then(setNameTaken, setNameTaken).then(showWarningsOrCreate, showWarningsOrCreate);
```
jwforres: i tried that originally, but `finally` does not get the values from resolve/reject returned to it

app/scripts/controllers/create/createFromImage.js
```diff
@@ -297,9 +264,44 @@ angular.module("openshiftConsole")
...
+                  message: "Problems were detected while checking your application configuration.",
+                  okButtonText: "Create Anyway",
+                  okButtonClass: "btn-danger",
+                  cancelButtonText: "Cancel"
+                };
+              }
+            }
+          });
+
+          modalInstance.result.then(function() {
```
jwforres: Fixed

app/scripts/controllers/create/createFromImage.js
```diff
@@ -297,9 +264,44 @@ angular.module("openshiftConsole")
...
+            createResources();
+          });
+        };
+
+        var showWarningsOrCreate = function(result){
+          // Now that all checks are completed, show any warnings if we need to
+          var quotaWarnings = result.quotaWarnings || [];
+          var errorAlerts = _.filter(quotaWarnings, {type: 'error'}) || [];
```
jwforres: Fixed

app/scripts/controllers/create/createFromImage.js
```diff
@@ -316,15 +319,23 @@ angular.module("openshiftConsole")
...
+
+          var nameTakenPromise = ApplicationGenerator.ifResourcesDontExist(generatedResources, $scope.projectName);
+          var checkQuotaPromise = QuotaService.getLatestQuotaWarnings(generatedResources, context);
+          // Don't want to wait for the name checks to finish before making the calls to quota
+          // so kick off the requests above and then chain the promises here
+          var setNameTaken = function(result) {
+            $scope.nameTaken = result.nameTaken;
+            return checkQuotaPromise;
+          };
+          nameTakenPromise.then(setNameTaken, setNameTaken).then(showWarningsOrCreate, showWarningsOrCreate);
```
jwforres: as mentioned above, finally doesnt get the response value

app/scripts/filters/resources.js
```diff
@@ -1274,15 +1274,36 @@ angular.module('openshiftConsole')
...
       "NotTerminating": "Matches pods that do not have an active deadline.",
       "BestEffort": "Matches pods that have best effort quality of service.",
       "NotBestEffort": "Matches pods that do not have best effort quality of service."
     };
     return function(scope) {
-      return scopeMessages[scope];
+      return scopeMessages[scope] || sentenceCaseFilter(scope);
+    };
+  })
+  .filter('isBestEffortPod', function(isNilFilter, usageValueFilter) {
```
jwforres: I put it in a filter because it felt like something we would want to use in the Pod view, but that was before the discussion upstream about the calculated QoS tier being added to the Pod status.  I can move it, we can always add a wrapper filter if we decide we need it later.

app/scripts/filters/resources.js
jwforres: Its actually not the same thing.  The existing one is whether the Pod is currently being terminated, in other words a delete call was sent to it and its shutting down.  The new one has to do with what QoS the created Pods will be in and how that affects application of Quota scopes.  I think moving this into the QuotaService would help clarify it.

app/scripts/filters/resources.js
jwforres: Fixed

app/scripts/services/quota.js
jwforres: Fixed

app/scripts/services/quota.js
jwforres: Fixed

app/scripts/services/quota.js
jwforres: Fixed

app/scripts/services/quota.js
jwforres: Fixed

app/scripts/services/quota.js
jwforres: If we fail to list the quota, should we resolve or reject?  I can't decide, it's effectively rejecting with no warnings...  i guess at that point its up to the caller what to do with a rejection that contains no warnings.

app/scripts/filters/resources.js
jwforres: I put it in QuotaService since it relates to quota scopes, and it didn't make sense for isTerminatingPod to be in LimitsService. Would rather have them together.

app/scripts/filters/resources.js
jwforres: Moved it to QuotaService.

app/scripts/services/quota.js
spadgett: Maybe only reject the promise if there is an error with the list? Other cases we resolve the promise (either with warnings or an empty array).

jwforres:
That's true, even if there are warnings, we are resolving the promise of
returning warnings.  Seems reasonable.

app/scripts/directives/deployImage.js
spadgett: ok makes sense

jwforres:
@spadgett OK updated based on review feedback, also had to fix some stuff around the scope filtering, wasn't getting applied to things with pod templates

jwforres:
@spadgett added the support for create from template since that was a smallish change, so the only thing this doesn't cover is Import YAML / JSON

jwforres:
[merge]

spadgett:
```
A Jasmine spec timed out. Resetting the WebDriver Control Flow.
The last active task was: 
unknown
```

[merge]

jwforres:
@liggitt i think these failures are a result of your redirectURLs change, the AMIs are no longer set up to where the UI integration tests can run against the openshift server started using `systemctl start openshift`

jwforres:
@liggitt this is the spot in vagrant-openshift that sets up the openshift service for systemctl

liggitt:
testing a fix in https://github.com/openshift/origin-web-console/pull/607, can decide if we want it in the integration script or in vagrant

jwforres:
[merge]

PR #606, opened 2016-09-27
Remove incorrect error handling from AttachPVCController

### LINKED ISSUE #605: AttachPVCController incorrectly tries to handle error from DataService.list()
`DataService.list()` doesn't support an error handler currently. This controller is passing a callback function as the `opts` param to list.

https://github.com/spadgett/origin-web-console/blob/67a12b502bec69174656c0388761e1990bc5c4ff/app/scripts/controllers/attachPVC.js#L111-L123

I don't think this has any bad effects currently, but we should fix it.

/cc @jwforres

spadgett:
Fixes #605 

@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/attachPVC.js
+++ b/app/scripts/controllers/attachPVC.js
@@ -108,19 +108,12 @@ angular.module('openshiftConsole')
             }
           );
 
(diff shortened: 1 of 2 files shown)


jwforres:
[merge]

PR #613, opened 2016-09-30
Api token reset link in cli about page

rjstone:
I was using this today and was unable to find the link to reset the oauth token when I needed to. I suggested this change to the support rep and he suggested that I could submit a pull request for the change.

This seemed like the most appropriate way to insert the link. But I was unable to get a test environment working in a reasonable amount of time so I haven't tested this change.

(Sorry for the commit log slop. I made a mistake committing to the wrong branch and made a mess of the logs trying to fix it. If this is a problem, please let me know the best way to fix it. Only thing I can think of is to delete my fork, copy the files off, refork, and commit cleanly to the correct branch.)


rjstone:
Yep, didn't run grunt. Was trying to get this working under WSL and having long pathname problems. I will try alternate means.

liggitt:
Logging out and back in would be a better way to reset it. Simply requesting a new token would not invalidate the old one

rjstone:
If that's the case then why does https://api.preview.openshift.com/oauth/token/request even exist and why would I be advised to use it? Also, if this is how the cli API key works then there are some other issues:

1) It's not immediately obvious that the API token used for the CLI is the same oauth token used for the web login. I was expecting it to be a separate, persistent, random nonce similar to a REST API key. So documentation, and the same view in this pull request, needs to be updated to clarify that.

2) If it remains valid forever as long as someone hasn't explicitly logged out, and requesting a new one doesn't invalidate the old one, that would suggest that if someone leaves a web browser logged in somewhere and forgets about it then there could be a token floating around that can't be easily invalidated. (I mean that would be true anyway but when using as a CLI API key people are likely to store it in scripts, etc, where it's more likely to get leaked as people paste their code, etc.)

3) It sounds like perhaps this could be considered a bug in https://api.preview.openshift.com/oauth/token/request and that it should log you out when you request a new token.

In any case it seems like this warrants some discussion. Should I open an issue for it?

liggitt:
> If that's the case then why does https://api.preview.openshift.com/oauth/token/request even exist 

it exists as a redirect point for the CLI when a login method that requires a browser flow is configured (like google/github oauth, etc)

> If it remains valid forever as long as someone hasn't explicitly logged out, and requesting a new one doesn't invalidate the old one, that would suggest that if someone leaves a web browser logged in somewhere and forgets about it then there could be a token floating around that can't be easily invalidated.

it expires at an admin-configured interval, but if the intent here is to tell someone how to invalidate a token, then requesting a new one will not accomplish that

> It sounds like perhaps this could be considered a bug in https://api.preview.openshift.com/oauth/token/request and that it should log you out when you request a new token.

why wouldn't multiple valid tokens be ok?

rjstone:
Multiple valid tokens would be OK, but only if the UI suggests that you actually have multiple valid tokens.

Here is how I'd expect this to work in an environment with multiple valid tokens used as API keys:



Basically, for multiple valid tokens to be OK, I need to be able to:

1. See all currently valid tokens.
2. Invalidate any that I need to invalidate for whatever reason.
3. Generate new ones, preferably in a manner similar to the above with minimal hassle.

So the pull request as it stands is wrong/insufficient. But my questions would be, for the purpose of temporary workarounds and fixing this at least to the extent of avoiding any future confusion:

1. How do I retrieve a list of all currently valid API tokens?
2. How do I invalidate any particular token?
3. How do I generate a new token for something like an unattended script that needs a token for longer than a web ui session and that should be immune to being invalidated by web ui logins/logouts?

app/views/command-line.html
```diff
@@ -48,7 +48,7 @@ <h1 id="cli">Command Line Tools</h1>
                   <div ng-show="showSessionToken" class="alert alert-warning">
                     <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
                     <strong>A token is a form of a password.</strong>
-                    Do not share your API token.
+                    Do not share your API token. If your token may have been leaked or you otherwise want to reset it, please <a target="_blank" href="{{'token-request' | apiLink}}">Request a New API Token</a>.
```
rjstone: Turns out that this statement isn't true. Token's aren't invalidated by fetching a new one.

app/scripts/filters/util.js
```diff
@@ -195,6 +195,11 @@ angular.module('openshiftConsole')
       return Constants.HELP[type] || Constants.HELP["default"];
     };
   })
+  .filter('apiLink', function(Constants) {
+    return function(type) {
```
rjstone: Is this the right way to do this?

app/scripts/constants.js
```diff
@@ -28,6 +28,10 @@ window.OPENSHIFT_CONSTANTS = {
   CLI: {
     "Latest Release":          "https://github.com/openshift/origin/releases/latest"
   },
+  // Maps links to API UI URLs
+  API: {
+    "token-request": "https://api.preview.openshift.com/oauth/token/request"
```
rjstone: Is this the right place or should it have been thrown in the same object with the doc links above?

liggitt:
> How do I retrieve a list of all currently valid API tokens?

That isn't available to end users via the API yet

> How do I invalidate any particular token?

you can delete it via the oauthaccesstokens, or use something like `oc logout --token=<token>`

> How do I generate a new token for something like an unattended script that needs a token for longer than a web ui session and that should be immune to being invalidated by web ui logins/logouts?

Use a [service account token](https://docs.openshift.org/latest/dev_guide/service_accounts.html)

liggitt:
APIs to let users see/manage/create/revoke tokens are planned, but not available yet

coreydaley:
Maybe changing the text to "Request additional token" or something along those lines makes more sense.  "Request new token" leads me to believe that my old token has been revoked, as I have nowhere to view the list of tokens that are currently active, I would assume that only that one is active.

Using 'oc logout --token=.......' did seem to revoke the token, i was unable to use it again after that.  

'oc logout -h' produces the following output:
"An authentication token is stored in the config file after login - this command will delete
that token on the server, and then remove the token from the configuration file."

Which does point to the token being deleted/revoked, but being able to list the current active tokens does make sense here.

rjstone:
Ok, thanks. That helps.

>>How do I retrieve a list of all currently valid API tokens?

>That isn't available to end users via the API yet

Ok, so given that this is the case, and the token that I thought I had invalidated can no longer be invalidated using the method provided because I no longer have it, how do I guarantee that all oauth tokens are invalidated? Should an explicit logout do this or do I need to remove the openshift app from github temporarily? Or do I just need to logout from github itself? (Is the oauth token used by OS the exact one returned by the github oauth redirect or is there a second oauth layer?)

I don't have a way to verify that the lost token is no longer valid and I'm still a little confused about multiple tokens are associated with a web login using only one of those multiple tokens. I did read over [this doc](https://docs.openshift.org/latest/install_config/configuring_authentication.html#install-config-configuring-authentication) which suggests that origin oauth is a separate layer over GitHub authentication. So I get the feeling they are totally separate and perhaps nothing I do will invalidate the lost valid token.

*****

Also, I still don't know what to do with this pull request. I was looking to address the problem I encountered so others won't run into it. Normally I'd expect the part of the UI that provided the token to at least link to information about how to manage tokens but this doesn't appear to be the case right now, so:

1. How should the about command-line page be updated? Should it just link to documentation?
2. What should it link to and does that part of the documentation need an update?

So far I've found the following documentation related to authentication, in order of complexity from novice to API developer:

1. https://docs.openshift.com/online/getting_started/beyond_the_basics.html#btb-installing-the-openshift-cli (where I'd expect people to start. Currently points to the wrong about page.)
2. https://docs.openshift.org/latest/cli_reference/get_started_cli.html#basic-setup-and-login (really need a link to this from Getting Started)
3. https://docs.openshift.org/latest/dev_guide/authentication.html (a little more detail)
4. https://docs.openshift.org/latest/architecture/additional_concepts/authentication.html#architecture-additional-concepts-authentication
5. https://docs.openshift.org/latest/install_config/configuring_authentication.html#install-config-configuring-authentication

Comments?

liggitt:
> how do I guarantee that all oauth tokens are invalidated

there is not an end-user-facing API for that today, only an openshift cluster admin can do that currently. the token will also expire at the interval configured by the openshift admin.

> Should an explicit logout do this or do I need to remove the openshift app from github temporarily? Or do I just need to logout from github itself? (Is the oauth token used by OS the exact one returned by the github oauth redirect or is there a second oauth layer?)

it is not a github oauth token, the tokens used against OpenShift are minted by OpenShift and are unrelated to your github token.

app/scripts/constants.js
jwforres: API URLs would never be in this file.  Anything associated with building URLs to the API should go through services/data.js

app/scripts/filters/util.js
jwforres: having a filter that calls into one of our API related services might be OK, but it definitely should not be pulling things from the Constants

PR #615, opened 2016-10-03
Animation enhancement and revisions

sg00dwin:
Addition of an animate-slide class, giving affordance transition to specific elements on the overview and refactor of existing component-animations.less

The animate-slide applies brief movement to specific elements on the overview when they enter or exit so that surrounding elements don't abruptly jump in location. It is combined with a fade in/out opacity change that is slight faster, so that these transition are visually smoother.

Elements that have movement are builds, build-pipeline-expanded and service group alerts (pf-toast)

### DIFF
--- a/app/scripts/directives/alerts.js
+++ b/app/scripts/directives/alerts.js
@@ -19,6 +19,7 @@ angular.module('openshiftConsole')
         alerts: '=',
         // Filter function (optional).
         filter: '=?',
+        animateSlide: '=?',
         hideCloseButton: '=?',
         toast: '=?'
       },
--- a/app/views/_alerts.html
+++ b/app/views/_alerts.html
@@ -1,5 +1,5 @@
 <div ng-attr-row="{{toast}}"  ng-attr-wrap="{{toast}}">
-  <div ng-repeat="alert in (alerts | filterCollection : filter)" ng-if="!alert.hidden" class="alert-wrapper">
+  <div ng-repeat="(alertID, alert) in (alerts | filterCollection : filter) track by (alertID + (alert.message || alert.details))" ng-if="!alert.hidden" class="alert-wrapper animate-repeat" ng-class="{'animate-slide': animateSlide}">
     <div class="alert" ng-class="{
       'alert-danger': alert.type === 'error',
       'alert-warning': alert.type === 'warning',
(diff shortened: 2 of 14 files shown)


app/styles/_overview.less
```diff
@@ -128,7 +128,7 @@
   .service-group-header,
   .deployment-tile,
   .standalone-service .builds-block,
-  .service-group-triggers {
+  .service-group-triggers .build {
```
spadgett: Does this cause multiple shadows when there is more than one build or pipeline?

app/views/_alerts.html
```diff
@@ -1,5 +1,5 @@
-<div ng-attr-row="{{toast}}"  ng-attr-wrap="{{toast}}">
-  <div ng-repeat="alert in (alerts | filterCollection : filter)" ng-if="!alert.hidden" class="alert-wrapper">
+<div ng-attr-row="{{toast}}"  ng-attr-wrap="{{toast}}" style="overflow: hidden">
```
spadgett: Let's pull the style into a less file somewhere.

app/views/_alerts.html
```diff
@@ -1,5 +1,5 @@
...
+  <div ng-repeat="(alertID, alert) in (alerts | filterCollection : filter) track by (alertID + (alert.message || alert.details))" ng-if="!alert.hidden" class="alert-wrapper animate-repeat animate-slide" ng-class="{'animate-slide': has-slide}">
```
spadgett: You have `animate-slide` in both `class` and `ng-class` here.

app/scripts/directives/alerts.js
```diff
@@ -19,6 +19,7 @@ angular.module('openshiftConsole')
         alerts: '=',
         // Filter function (optional).
         filter: '=?',
+        hasSlide: '=?',
```
spadgett: `animateSlide` makes it more clear what this is.

app/views/_alerts.html
sg00dwin: This was removed

app/views/_alerts.html
sg00dwin: Removed the one from within `class`

sg00dwin:
@jwforres @spadgett PTAL

spadgett:
[merge]

PR #622, opened 2016-10-04
`make` should default to `build` goal

spadgett:
@stevekuznetsov PTAL


stevekuznetsov:
Sure, whatever use-cases you guys want to put in there is fine by me, I just need the targets so I can call stuff from my tool. LGTM

spadgett:
[merge]

PR #628, opened 2016-10-05
Update replica set page for replica sets owned by a deployment

spadgett:
https://trello.com/c/HGvCTI96

* Actions update the deployment instead of the replica set
* Show the deployment for the replica set in the breadcrumbs
* Fix problem showing events for a replication controller
* Update variables named `deployment` to `replicaSet` to avoid confusion about the types

@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/replicaSet.js
+++ b/app/scripts/controllers/replicaSet.js
@@ -23,6 +23,7 @@ angular.module('openshiftConsole')
                         keyValueEditorUtils,
                         kind) {
     var hasDC = false;
+    var annotation = $filter('annotation');
     var displayKind = $filter('humanizeKind')(kind);
     switch (kind) {
     case 'ReplicaSet':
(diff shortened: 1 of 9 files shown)


app/scripts/controllers/replicaSet.js
```diff
@@ -311,42 +383,41 @@ angular.module('openshiftConsole')
           }
         );
 
-        watches.push(DataService.watch($scope.resource, context, function(deployments, action, deployment) {
-          $scope.deployments = deployments.by("metadata.name");
+        watches.push(DataService.watch($scope.resource, context, function(replicaSets, action, replicaSet) {
+          $scope.replicaSets = replicaSets.by("metadata.name");
           $scope.emptyMessage = "No deployments to show";
-          $scope.deploymentsByDeploymentConfig = DeploymentsService.associateDeploymentsToDeploymentConfig($scope.deployments);
+          $scope.deploymentsByDeploymentConfig = DeploymentsService.associateDeploymentsToDeploymentConfig($scope.replicaSets);
```
jwforres: is this a wasted call if these are actually replica sets and not replication controllers?

spadgett:
@jwforres updated

jwforres:
[merge]

PR #630, opened 2016-10-05
Check quota on explicit request/limits in pod templates

jwforres:
Partially implements https://trello.com/c/mfeqgKvL

### DIFF
--- a/app/scripts/services/quota.js
+++ b/app/scripts/services/quota.js
@@ -68,6 +68,70 @@ angular.module("openshiftConsole")
     var humanizeQuotaResource = $filter('humanizeQuotaResource');
     var humanizeKind = $filter('humanizeKind');
 
+    var getQuotaResourceReachedAlert = function(quota, resource, type) {
+      var q = quota.status.total || quota.status;
+      if (usageValue(q.hard[type]) <= usageValue(q.used[type])) {
+        var details;
+        if (resource.kind === 'Pod') {
+          details = "You will not be able to create the " + humanizeKind(resource.kind) + " '" + resource.metadata.name + "'.";
+        }
+        else {
+          details = "You can can still create " + humanizeKind(resource.kind) + " '" + resource.metadata.name + "' but no pods will be created until resources are freed.";
+        }
+        return {
+          type: resource.kind === 'Pod' ? 'error' : 'warning',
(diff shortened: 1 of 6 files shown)


jwforres:
@spadgett PTAL, i'm not totally in love with the implementation, open to other ideas.  No matter what the idea is that we only check explicitly defined requests or limits, no defaulting behavior.  got agreement from @derekwaynecarr that this is the sanest approach

spadgett:
I'm fine with the approach.

Are we thinking about checking `spec.replicas`? Or the total for everything you're creating? It looks like we only check the resources one by one.

spadgett:
(I understand we need to draw the line somewhere.)

jwforres:
Yeah I'm making that choice intentionally, we are only warning you that you wont be able to have *any* pods created.  If you can get at least one, then we won't warn you, because you'll get the warnings on the overview that you are out of resources.

jwforres:
As far as checking the total across all pod template things you are trying to create, that gets even more complicated.  Since each quota may only apply to certain resources based on the scopes, we'd have to segment the totals by quota.

spadgett:
> Since each quota may only apply to certain resources based on the scopes, we'd have to segment the totals by quota.

OK you've convinced me :)

jwforres:
@spadgett ok pushed the dist files, any other comments or ready to merge ?

app/scripts/services/quota.js
```diff
@@ -92,24 +92,66 @@ angular.module("openshiftConsole")
...
+              var detail;
+              if (resource.kind === 'Pod') {
+                detail = "You may not be able to create the " + humanizeKind(resource.kind) + " '" + resource.metadata.name + "'.";
+              }
+              else {
+                detail = "You can can still create " + humanizeKind(resource.kind) + " '" + resource.metadata.name + "' but you may not have pods created until resources are freed.";
+              }
+              alerts.push({
+                type: 'warning',
+                message: 'You are close to your quota for ' + humanizeQuotaResource(type).toLowerCase() + " on pods.",
```
spadgett: Will this lowercase `CPU`? might want a case option on `humanizeQuotaResource` like we have for `humanizeKind`.

spadgett:
Should `pods` be in the list of types we check? Looks like the logic assumes a compute resource.

jwforres:
@spadgett updated

spadgett:
[merge]

PR #636, opened 2016-10-06
Show services with alternates as top-level services on overview

### LINKED ISSUE #631: Service can be hidden on overview if it's an alternate service to another dependent service
In some edge cases, a service can be hidden:

* Service A depends on service B
* Route B has primary service B and alternate service C
* Service C will not be displayed on the overview

/cc @jwforres @jorgemoralespou

spadgett:
Fixes #631
@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -59,16 +59,26 @@ angular.module('openshiftConsole')
     var isRecentDeployment = $filter('isRecentDeployment');
 
     var alternateServiceSet = {};
+    var servicesWithAlternates = {};
     var findAlternateServices = function() {
       // Map of service names and a truthy value if this is a alternate service for a route.
       alternateServiceSet = {};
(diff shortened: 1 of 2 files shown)


jwforres:
@spadgett what does this look like then if you have grouped either the primary or one of the alternates to another service

spadgett:
It shows up twice, which is not ideal. We already do this for services that have dependents, but are dependents themselves. I was planning on opening an issue about it because there are many cases where we show the same deployment twice on the overview, for instance, if it has more than one service.

Here is a screenshot:

screen shot 2016-10-06 at 11 06 10 am

jwforres:
Ok i think this is a reasonable change to backport to 3.3.1 since it at least gets rid of the loss of the service

I think we need to just have a larger discussion about linking in combination with A/B

jwforres:
[merge]

PR #638, opened 2016-10-06
Updating angular-ui-select version and fixing presentational bugs

### LINKED ISSUE #627: Select field on monitoring page collapses too narrow causing the text and down caret icon overlap
It looks like setting a min width on .ui-select-bootstrap would correct it.

screen shot 2016-10-05 at 1 36 02 pm

rhamilto:
Fixes #627 

screen shot 2016-10-06 at 2 00 26 pm

And maintains the fix that resulted in #627 without causing #627.

screen shot 2016-10-06 at 2 00 04 pm

@jwforres or @spadgett PTAL.

cc: @jhadvig

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -82,43 +82,6 @@
   }
 }
 
-/* Fix some styles with the angular-ui-select bootstrap theme that isn't right when combined with all of our styles */
-.ui-select-bootstrap {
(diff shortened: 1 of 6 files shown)


spadgett:
[merge]

jwforres:
[merge]

PR #640, opened 2016-10-06
Link to deployment on replica set environment tab

spadgett:
If a replica set has a deployment, link to the deployment for editing environment variables.

@jwforres PTAL


jwforres:
[merge]

PR #643, opened 2016-10-07
Add replica sets to monitoring page

spadgett:
https://trello.com/c/HGvCTI96

@jwforres PTAL. I moved the replica sets back under "Deployments" as one list with replication controllers.


jwforres:
hmm was originally thinking they should be merged into one list sorted by date, but I guess its not going to be that common for people to be using both RCs and RSes, and this is more efficient.

[merge]

PR #645, opened 2016-10-07
Use consistent close icon

jhadvig:
updating the close btn icon, from `fa-times` to `pficon-close` so we are consistent with the close btn icon that we use on other places, eg: `key-value-editor`  or `osc-source-secrets` directives ...


jhadvig:
@spadgett ptal

spadgett:
[merge]

PR #650, opened 2016-10-10
Added advanced option for selector labels

zherman0:
Add ability for users to add selector labels on the create form for Storage (PVC).

### DIFF
--- a/app/scripts/controllers/createPersistentVolumeClaim.js
+++ b/app/scripts/controllers/createPersistentVolumeClaim.js
@@ -8,7 +8,7 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('CreatePersistentVolumeClaimController', function ($filter, $routeParams, $scope, $window, ApplicationGenerator, DataService, Navigate, ProjectsService) {
+  .controller('CreatePersistentVolumeClaimController', function ($filter, $routeParams, $scope, $window, ApplicationGenerator, DataService, Navigate, ProjectsService,keyValueEditorUtils) {
     $scope.alerts = {};
     $scope.projectName = $routeParams.project;
     $scope.accessModes="ReadWriteOnce";
@@ -55,29 +55,31 @@ angular.module('openshiftConsole')
           }
         };
 
-      function generatePersistentVolumeClaim() {
-        var pvc = {
-          kind: "PersistentVolumeClaim",
-          apiVersion: "v1",
-          metadata: {
-            name: $scope.claim.name,
-            labels: {}
-          },
-          spec: {
-            resources: {
-              requests:{}
-
+         function generatePersistentVolumeClaim() {
+          var pvc = {
+            kind: "PersistentVolumeClaim",
+            apiVersion: "v1",
+            metadata: {
+              name: $scope.claim.name,
+              labels: {}
+            },
+            spec: {
+              resources: {
+                requests:{}
+              },
+              selector:{
+                matchLabels:{}
+              }
             }
-          }
-        };
+          };
 
-        pvc.spec.accessModes = [$scope.claim.accessModes || "ReadWriteOnce"] ;
-        var unit =  $scope.claim.unit || "Mi";
-        pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
+          pvc.spec.accessModes = [$scope.claim.accessModes || "ReadWriteOnce"] ;
+          var unit =  $scope.claim.unit || "Mi";
+          pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
+          pvc.spec.selector.matchLabels = keyValueEditorUtils.mapEntries( keyValueEditorUtils.compactEntries($scope.claim.selectedLabels) );
 
-        return pvc;
-      }
-
+          return pvc;
+        }
 
     }));
   });
--- a/app/scripts/directives/oscPersistentVolumeClaim.js
+++ b/app/scripts/directives/oscPersistentVolumeClaim.js
@@ -13,16 +13,17 @@ angular.module("openshiftConsole")
         scope.units = [{
           value: "Mi",
           label: "MiB"
-          }, {
+        }, {
           value: "Gi",
           label: "GiB"
-          }, {
+        }, {
           value: "Ti",
           label: "TiB"
-          }, {
+        }, {
           value: "Pi",
           label: "PiB"
-          }];
+        }];
+        scope.claim.selectedLabels = [];
         }
     };
   });
--- a/app/views/directives/osc-persistent-volume-claim.html
+++ b/app/views/directives/osc-persistent-volume-claim.html
@@ -91,5 +91,36 @@
       </div>
       </fieldset>
     </div>
+    <!--advanced options-->
+    <div>
+      <a href=""
+         ng-click="showAdvancedOptions = true"
+         ng-show="!showAdvancedOptions">Use label selectors in the storage request </a>
+      <a href=""
+         ng-click="showAdvancedOptions = false"
+         ng-show="showAdvancedOptions">Hide label selectors in the storage request</a>
+
+
+      <div ng-show="showAdvancedOptions" class="form-group" >
+        <fieldset class="compute-resource" >
+          <label for="claim-name" >Label Selector</label>
+          <div>
+            <span id="claim-label-help" class="help-block">
+              Enter a label and value to use with your requested storage.
+               <div class="learn-more-block" ng-class="{ 'gutter-bottom': showPodWarning }">
+                 <a href="{{'selector_label' | helpLink}}" target="_blank">Learn more <i class="fa fa-external-link" aria-hidden="true"> </i></a>
+               </div>
+            </span>
+          </div>
+          <key-value-editor
+            entries="claim.selectedLabels"
+            key-placeholder="label"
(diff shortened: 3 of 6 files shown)


spadgett:
cc @jwforres

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -55,29 +55,40 @@ angular.module('openshiftConsole')
...
+          pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
+          if ($scope.claim.selectedLabels.length > 0)
+          {
+            var arrayLength = $scope.claim.selectedLabels.length;
+            for (var ii = 0; ii < arrayLength; ii++) {
+              var label = $scope.claim.selectedLabels[ii];
+              var keyName = label["name"];
+              var keyValue = label["value"];
+              pvc.spec.selector.matchLabels[keyName] = keyValue;
+            }
```
spadgett: Suggest using lodash here, which is more concise. Should be able to replace the entire `if` block with

```js
_.each($scope.claim.selectedLabels, function(label) {
  pvc.spec.selector.matchLabels[label.name] = label.value;
});
```

app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -23,6 +23,7 @@ angular.module("openshiftConsole")
           value: "Pi",
           label: "PiB"
           }];
+        scope.claim.selectedLabels = [];
```
spadgett: Indentation

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -91,5 +91,33 @@
       </div>
       </fieldset>
     </div>
+    <!--advanced options-->
+    <div>
+      <a href=""
+         ng-click="showAdvancedOptions = true"
+         ng-show="!showAdvancedOptions">Show Advanced Options for Requested Storage</a>
```
spadgett: Since we're only showing label controls, let's just say that. Also we use sentence case for these links. Maybe

"Use label selectors in the storage request."

Probably want a "Learn more" link to this topic: https://docs.openshift.org/latest/install_config/persistent_storage/selector_label_binding.html

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -91,5 +91,33 @@
...
+      <a href=""
+         ng-click="showAdvancedOptions = false"
+         ng-show="showAdvancedOptions">Hide Advanced Options for Requested Storage</a>
+
+
+      <div ng-show="showAdvancedOptions" class="form-group" >
+        <fieldset class="compute-resource" >
+          <label for="claim-name" >Selector Label (optional)</label>
```
spadgett: Remove "(optional)" since we don't use that anywhere else in the console.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -91,5 +91,33 @@
...
+          <div>
+            <span id="claim-label-help" class="help-block">
+              Enter a Label and value to associate with your requested storage.
```
spadgett: Lower case "label"

app/views/directives/osc-persistent-volume-claim.html
spadgett: I'd probably say "Label Selector" as well.

app/views/directives/osc-persistent-volume-claim.html
spadgett: Maybe "to use" instead of "to associate"

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -91,5 +91,34 @@
...
+         ng-click="showAdvancedOptions = true"
+         ng-show="!showAdvancedOptions">Use label selectors in the storage request. </a>
+      <a href=""
+         ng-click="showAdvancedOptions = false"
+         ng-show="showAdvancedOptions">Hide label selectors in the storage request.</a>
+
+
+      <div ng-show="showAdvancedOptions" class="form-group" >
+        <fieldset class="compute-resource" >
+          <label for="claim-name" >Label selector</label>
```
spadgett: "Label Selector"

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -55,29 +55,36 @@ angular.module('openshiftConsole')
...
-        pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
-
-        return pvc;
-      }
+          };
 
+          pvc.spec.accessModes = [$scope.claim.accessModes || "ReadWriteOnce"] ;
+          var unit =  $scope.claim.unit || "Mi";
+          pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
+          if ($scope.claim.selectedLabels.length > 0)
```
spadgett: The length check is unnecessary. `_.each` handles empty arrays (and even `undefined` or `null`).

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -91,5 +91,34 @@
       </div>
       </fieldset>
     </div>
+    <!--advanced options-->
+    <div>
+      <a href=""
+         ng-click="showAdvancedOptions = true"
+         ng-show="!showAdvancedOptions">Use label selectors in the storage request. </a>
```
spadgett: Not sure if the trailing space after the period will be underlined. I'd prefer to take that out.

Also a nit, but we usually leave off the period in links like these. (I know I had that in my original comment.) Also in the link just below.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -91,5 +91,34 @@
...
+         ng-show="!showAdvancedOptions">Use label selectors in the storage request </a>
+      <a href=""
+         ng-click="showAdvancedOptions = false"
+         ng-show="showAdvancedOptions">Hide label selectors in the storage request</a>
+
+
+      <div ng-show="showAdvancedOptions" class="form-group" >
+        <fieldset class="compute-resource" >
+          <label for="claim-name" >Label Selector</label>
+          <a href="https://docs.openshift.org/latest/install_config/persistent_storage/selector_label_binding.html">Learn more</a>
```
spadgett: Sorry missed this on the last review, but the link needs to go in app/scripts/constants.js so it can be customized for the different versions. Then use the `helpLink` directive to get the link.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L11-L31

Here is an example:

https://github.com/openshift/origin-web-console/blob/e7a0c0a8f703d5429f70b78223abb31856a66670/app/views/set-limits.html#L21-L23

Make sure to have `target="_blank"` so it opens in a new window like our other help links.

spadgett: The learn more link should be part of `help-block` below.

zherman0:
@spadgett - how are things looking for this now?

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -91,5 +91,36 @@
...
+
+
+      <div ng-show="showAdvancedOptions" class="form-group" >
+        <fieldset class="compute-resource" >
+          <label for="claim-name" >Label Selector</label>
+          <div>
+            <span id="claim-label-help" class="help-block">
+              Enter a label and value to use with your requested storage.
+               <div class="learn-more-block" ng-class="{ 'gutter-bottom': showPodWarning }">
+                 <a href="{{'selector_label' | helpLink}}" target="_blank">Learn more <i class="fa fa-external-link" aria-hidden> </i></a>
```
spadgett: I believe this needs to be `aria-hidden="true"` on the `i` element instead of just `aria-hidden`.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -55,29 +55,34 @@ angular.module('openshiftConsole')
...
-
-        return pvc;
-      }
+          };
 
+          pvc.spec.accessModes = [$scope.claim.accessModes || "ReadWriteOnce"] ;
+          var unit =  $scope.claim.unit || "Mi";
+          pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
+          _.each($scope.claim.selectedLabels, function(label) {
+            if (label.name !== "" && label.value !== "") {
```
spadgett: @benjaminapetersen this shouldn't be necessary if @zherman0 calls `keyValueEditorUtils.compactEntries` correct?

benjaminapetersen: Right! That should be all that is needed to cleanup the list, there is also `.mapEntries()` to convert from list to object, I believe this will do all of that:
```javascript
pvc.spec.selector.matchLabels = keyValueEditorUtils.mapEntries( keyValueEditorUtils.compactEntries($scope.claim.selectedLabels) );
```

zherman0: Thanks @benjaminapetersen and @spadgett - although I would not say that is obvious code, it is pretty sweet.

benjaminapetersen: @zherman0 agree, I want to go back and make that a little bit more clean.

spadgett:
[merge]

PR #10, opened 2016-05-13
Updating Help > Documentation links to use helpLink filter

rhamilto:
and also improving markup semantics of help dropdown.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1318527

@jwforres, PTAL

PR #653, opened 2016-10-10
Sort replica set revisions as numbers

spadgett:
Sort revisions as numbers so `#10` is greater than `#9`.

@jwforres PTAL


jwforres:
[merge]

works for now, but we should talk about how we want to show the old history when replica sets are re-used

PR #655, opened 2016-10-10
Fix display of unserviced deployment configs

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1382895

The `$watch` wasn't getting updates for unserviced deployment configs. It worked on page refresh because the watch is always called once even for `undefined`, which set `activeReplicationController`. Missed the unserviced deployment config `ng-repeat` on the variable renames.

@jwforres PTAL


jwforres:
[merge]

PR #656, opened 2016-10-10
DC editor

jhadvig:
There is the DC editor work I've been working on.
The editor consists from three parts:
- Deployment Strategy Configuration
- Image Configuration
- Environment Variables

#### Deployment Strategy Configuration:
- serves for editing basic strategy parameters for each strategy:
  - Recreate - Timeout / Update Period / Interval / Max Unavailability / Max Surge
  - Rolling -  Timeout
  - Custom - Image name / Command / environment variables
- set `LifecycleHooks` for Recreate and Rolling strategy:
  - each hook might be for either:
      - Run a specific command in a new pod
      - Tag image if the deployment succeeds

#### Image Configuration:
- serves for setting manual or automatic trigger for each container in the podTemplate
- set of pullSecrets

#### Environment Variables:
Environment Variables for deployment

@spadgett @jwforres PTAL

### DIFF
--- a/app/scripts/controllers/edit/deploymentConfig.js
+++ b/app/scripts/controllers/edit/deploymentConfig.js
@@ -1,0 +1,298 @@
+'use strict';
+
+/**
+ * @ngdoc function
+ * @name openshiftConsole.controller:EditDeploymentConfigController
+ * @description
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('EditDeploymentConfigController', function ($scope, $routeParams, $uibModal, DataService, BreadcrumbsService, SecretsService, ProjectsService, $filter, ApplicationGenerator, Navigate, $location, AlertMessageService, SOURCE_URL_PATTERN, keyValueEditorUtils) {
+    $scope.projectName = $routeParams.project;
+    $scope.deploymentConfig = null;
+    $scope.alerts = {};
+    $scope.view = {
+      advancedStrategyOptions: false,
+      advancedImageOptions: false
+    };
+    $scope.breadcrumbs = BreadcrumbsService.getBreadcrumbs({
+      name: $routeParams.name,
+      kind: $routeParams.kind,
+      namespace: $routeParams.project,
+      subpage: 'Edit Deployment Config',
+      includeProject: true
+    });
+
+    $scope.deploymentConfigStrategyTypes = [
+      "Recreate",
+      "Rolling",
+      "Custom"
+    ];
+
+    AlertMessageService.getAlerts().forEach(function(alert) {
+      $scope.alerts[alert.name] = alert.data;
+    });
+    AlertMessageService.clearAlerts();
+    var watches = [];
+
+    var getParamsPropertyName = function(strategyType) {
+      switch (strategyType) {
+      case "Recreate":
+        return "recreateParams";
+      case "Rolling":
+        return "rollingParams";
+      case "Custom":
+        return "customParams";
+      default:
+        Logger.error('Unknown deployment strategy type: ' + strategyType);
+        return;
+      }
+    };
+
+    ProjectsService
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+        $scope.context = context;
+        DataService.get("deploymentconfigs", $routeParams.deploymentconfig, context).then(
+          // success
+          function(deploymentConfig) {
+            $scope.deploymentConfig = deploymentConfig;
+
+            $scope.breadcrumbs = BreadcrumbsService.getBreadcrumbs({
+              object: deploymentConfig,
+              project: project,
+              subpage: 'Edit',
+              includeProject: true
+            });
+
+            // Create map which will associate concatiner name to container's data(envVar, trigger and image which will be used on manual deployment) 
+            var mapContainerConfigByName = function(containers, triggers) {
+              var containerConfigByName = {};
+              var imageChangeTriggers = _.filter(triggers, {type: 'ImageChange'});
+              _.each(containers, function(container) {
+                var imageChangeTriggerForContainer = _.find(imageChangeTriggers, function(trigger) {
+                  return _.includes(trigger.imageChangeParams.containerNames, container.name);
+                });
+                var triggerData = {};
+                containerConfigByName[container.name] = {
+                  env: container.env || [],
+                  image: container.image,
+                  hasDeploymentTrigger: !_.isEmpty(imageChangeTriggerForContainer)
+                };
+                if (imageChangeTriggerForContainer) {
+                  var triggerFromData = imageChangeTriggerForContainer.imageChangeParams.from;
+                  var triggerImageNameParts = triggerFromData.name.split(':');
+                  triggerData = {
+                    data: imageChangeTriggerForContainer,
+                    istag: {namespace: triggerFromData.namespace || $scope.projectName, imageStream: triggerImageNameParts[0], tagObject: {tag: triggerImageNameParts[1]}}
+                  };
+                } else {
+                  triggerData = {
+                    istag: {namespace: "", imageStream: ""}
+                  };
+                }
+                _.set(containerConfigByName, [container.name, 'triggerData'], triggerData);
+              });
+              return containerConfigByName;
+            };
+
+            $scope.updatedDeploymentConfig = angular.copy($scope.deploymentConfig);
+            $scope.containerNames = _.map($scope.deploymentConfig.spec.template.spec.containers, 'name');
+            $scope.containerConfigByName = mapContainerConfigByName($scope.updatedDeploymentConfig.spec.template.spec.containers, $scope.updatedDeploymentConfig.spec.triggers);
+            $scope.pullSecrets = $scope.deploymentConfig.spec.template.spec.imagePullSecrets || [{name: ''}];
+            $scope.volumeNames = _.map($scope.deploymentConfig.spec.template.spec.volumes, 'name');
+            $scope.strategyData = angular.copy($scope.deploymentConfig.spec.strategy);
+            $scope.originalStrategy = $scope.strategyData.type;
+            $scope.strategyParamsPropertyName = getParamsPropertyName($scope.strategyData.type);
+
+            // If strategy is 'Custom' and no environment variables are present, initiliaze them.
+            if ($scope.strategyData.type === 'Custom' && !_.has($scope.strategyData, 'customParams.environment')) {
+              $scope.strategyData.customParams.environment = [];
+            }
+            
+            DataService.list("secrets", context, function(secrets) {
+              var secretsByType = SecretsService.groupSecretsByType(secrets);
+              // Add empty option to the image/source secrets
+              $scope.secretsByType = _.each(secretsByType, function(secretsArray) {
+                secretsArray.unshift("");
+              });
+            });
+
+            // If we found the item successfully, watch for changes on it
+            watches.push(DataService.watchObject("deploymentconfigs", $routeParams.deploymentconfig, context, function(deploymentConfig, action) {
+              if (action === 'MODIFIED') {
+                $scope.alerts["updated/deleted"] = {
+                  type: "warning",
+                  message: "This deployment configuration has changed since you started editing it. You'll need to copy any changes you've made and edit again."
+                };
+              }
+              if (action === "DELETED") {
+                $scope.alerts["updated/deleted"] = {
+                  type: "warning",
+                  message: "This deployment configuration has been deleted."
+                };
+                $scope.disableInputs = true;
+              }
+              $scope.deploymentConfig = deploymentConfig;
+            }));
+            $scope.loaded = true;
+          },
+          // failure
+          function(e) {
+            $scope.loaded = true;
+            $scope.alerts["load"] = {
+              type: "error",
+              message: "The deployment configuration details could not be loaded.",
+              details: $filter('getErrorDetails')(e)
+            };
+          }
+        );
+      })
+    );
+
+    // helper for detemining if strategy switch was done between Rolling <-> Recreate strategy
+    var isRollingRecreateSwitch = function() {
+      return ($scope.strategyData.type !== 'Custom' && $scope.originalStrategy !== 'Custom' && $scope.strategyData.type !== $scope.originalStrategy);
+    };
+
+    var promptToMoveParams = function(pickedStrategyParams) {
+      if (_.has($scope.strategyData, pickedStrategyParams)) {
+        return;
+      }
+      var modalInstance = $uibModal.open({
+        animation: true,
+        templateUrl: 'views/modals/confirm.html',
+        controller: 'ConfirmModalController',
+        resolve: {
+          modalConfig: function() {
+            return {
+              alerts: $scope.alerts,
+              message: "Some of your existing " + $scope.originalStrategy.toLowerCase() + " strategy parameters can be used for the " + $scope.strategyData.type.toLowerCase() + " strategy. Keep parameters?",
+              details: "The timeout parameter and any pre or post lifecycle hooks will be copied from " + $scope.originalStrategy.toLowerCase() + " strategy to " + $scope.strategyData.type.toLowerCase() + " strategy. After saving the changes, " + $scope.originalStrategy.toLowerCase() + " strategy parameters will be removed.",
+              okButtonText: "Yes",
+              okButtonClass: "btn-primary",
+              cancelButtonText: "No"
+            };
+          }
+        }
+      });
+      modalInstance.result.then(function () {
+        // Move parameters that belong to the origial strategy to the picked one.
+        $scope.strategyData[pickedStrategyParams] = $scope.strategyData[getParamsPropertyName($scope.originalStrategy)];
+        $scope.paramsMoved = getParamsPropertyName($scope.originalStrategy);
+      }, function() {
+        // Create empty parameters for the newly picked strategy
+        $scope.strategyData[pickedStrategyParams] = {};
+      });
+    };
+
+    $scope.strategyChanged = function() {
+      var pickedStrategyParams = getParamsPropertyName($scope.strategyData.type);
+      if (isRollingRecreateSwitch()) {
+        promptToMoveParams(pickedStrategyParams);
+      } else {
+        if (!_.has($scope.strategyData, pickedStrategyParams)) {
+          if ($scope.strategyData.type !== 'Custom') {
+            $scope.strategyData[pickedStrategyParams] = {};
+          } else {
+            $scope.strategyData[pickedStrategyParams] = {
+              image: "",
+              command: [],
+              environment: []
+            };
+          }
+        }
+      }
+      $scope.strategyParamsPropertyName = pickedStrategyParams;
+    };
+
+    var assembleImageChangeTrigger = function(containerName, ist, trigger) {
+      var istagObject = {
+        kind: "ImageStreamTag",
+        namespace: ist.namespace,
+        name: ist.imageStream + ':' + ist.tagObject.tag
+      };
+      if (trigger) {
+        trigger.imageChangeParams.from = istagObject;
+      } else {
+        trigger = {
+          type: "ImageChange",
+          imageChangeParams: {
+            automatic: true,
+            containerNames: [containerName],
+            from: istagObject
+          }
+        };
+      }
+      return trigger;
+    };
+
+    var updateTriggers = function() {
+      var updatedTriggers = _.filter($scope.updatedDeploymentConfig.spec.triggers, function(trigger) {return trigger.type !== 'ImageChange'});
+      _.each($scope.containerConfigByName, function(containerData, containerName) {
+        if (containerData.hasDeploymentTrigger) {
+          updatedTriggers.push(assembleImageChangeTrigger(containerName, containerData.triggerData.istag, containerData.triggerData.data));
+        } else {
+          var imageSpec = _.find($scope.updatedDeploymentConfig.spec.template.spec.containers, { name: containerName });
+          imageSpec.image = containerData.image;
+        }
+      });
+      return updatedTriggers;
+    };
+
+    $scope.save = function() {
+      $scope.disableInputs = true;
+
+      // Update env for each container
+      _.each($scope.containerConfigByName, function(containerData, containerName) {
+        var matchingContainer = _.find($scope.updatedDeploymentConfig.spec.template.spec.containers, { name: containerName });
+        matchingContainer.env = keyValueEditorUtils.compactEntries(containerData.env);
+      });
+
+      // Remove parameters of previously set strategy, if user moved 
+      if ($scope.paramsMoved && isRollingRecreateSwitch()) {
+        delete $scope.strategyData[getParamsPropertyName($scope.originalStrategy)];
+      }
+
+      if ($scope.strategyData.type !== 'Custom') {
+        _.each(['pre', 'mid', 'post'], function(hookType) {
+          if (_.has($scope.strategyData, [$scope.strategyParamsPropertyName, hookType, 'execNewPod', 'env'])) {
+            $scope.strategyData[$scope.strategyParamsPropertyName][hookType].execNewPod.env = keyValueEditorUtils.compactEntries($scope.strategyData[$scope.strategyParamsPropertyName][hookType].execNewPod.env);
+          }
+        });
+      } else if (_.has($scope, 'strategyData.customParams.environment')) {
+        $scope.strategyData.customParams.environment = keyValueEditorUtils.compactEntries($scope.strategyData.customParams.environment);
+      }
+      // Update image pull secrets 
+      $scope.updatedDeploymentConfig.spec.template.spec.imagePullSecrets = $scope.pullSecrets
+      $scope.updatedDeploymentConfig.spec.strategy = $scope.strategyData;
+      $scope.updatedDeploymentConfig.spec.triggers = updateTriggers();
+
+      DataService.update("deploymentconfigs", $scope.updatedDeploymentConfig.metadata.name, $scope.updatedDeploymentConfig, $scope.context).then(
+        function() {
+          AlertMessageService.addAlert({
+            name: $scope.updatedDeploymentConfig.metadata.name,
+            data: {
+              type: "success",
+              message: "Deployment config " + $scope.updatedDeploymentConfig.metadata.name + " was successfully updated."
+            }
+          });
+          var returnURL = Navigate.resourceURL($scope.updatedDeploymentConfig);
+          $location.url(returnURL);
+        },
+        function(result) {
+          $scope.disableInputs = false;
+          $scope.alerts["save"] = {
+            type: "error",
+            message: "An error occurred updating deployment config " + $scope.updatedDeploymentConfig.metadata.name + ".",
+            details: $filter('getErrorDetails')(result)
+          };
+        }
+      );
+    };
+
+    $scope.$on('$destroy', function(){
+      DataService.unwatchAll(watches);
+    });
+  });
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -503,17 +503,39 @@ label.checkbox {
 }
 
 .osc-secrets-form {
-  .picked-secret {
+  .basic-secrets {
+    .remove-secret {
+      display: inline-block;
+      margin-left: 7px;
+      vertical-align: 8px;
+    }
+    .secret-name {
+      width: 95%;
+      display: inline-block;
+    }
+  }
+  .advanced-secrets {
     margin-bottom: 5px;
+    .remove-secret {
+      display: inline-block;
+      margin-left: 7px;
+      vertical-align: 2px;
+    }
+    .directory {
+      width: 90%;
+      display: inline-block;
+    }
   }
-  .directory {
-    width:90%;
-    display: inline-block;
-  }
-  .remove-secret-row {
-    width: 10%;
-    margin-left: 7px;
-    display: inline;
+  @media (max-width: @screen-xs-max) {
+    .basic-secrets {
+      .secret-name {
+        width: 90%;
+        display: inline-block;
+      }
+    }
+    .remove-secret {
+      vertical-align: 2px;
+    }
   }
   .remove-btn {
     color: #333;
@@ -528,9 +550,15 @@ label.checkbox {
       text-decoration: none;
     }
   }
-  .ui-select-choices-row-inner {
-    height: 24px;
-    cursor: pointer;
+}
+
+.lifecycle-hook {
+  padding-bottom: 20px;
+  &:first-of-type {
+    padding-top: 20px;
+  }
+  .read-only-tag-image {
+    padding-bottom: 10px;
   }
 }
 
@@ -1132,7 +1160,7 @@ a.disabled-link {
     }
   }
   .checkbox {
-    margin: 10px;
+    margin: 10px 0;
   }
 
   .trigger-info {
--- a/app/styles/_select.less
+++ b/app/styles/_select.less
@@ -26,6 +26,10 @@
       color: @dropdown-link-active-color;
     }
   }
+  .ui-select-choices-row-inner {
+    min-height: 24px;
+    cursor: pointer;
+  }
   .ui-select-match-text, .ui-select-placeholder {
     font-weight: normal;
   }
@@ -41,6 +45,10 @@
     font-style: normal;
     right: 6px;
     top: 10px;
+    // Workaround for https://github.com/angular-ui/ui-select/issues/224
+    // A runtime error occurs when you click exactly on the down caret for some
+    // selects, such as the strategy type select in the deployment config editor.
+    pointer-events: none;
   }
   /* Fixes ui-select's input box extending to 100% page width if the page has been resized after page-load
   * https://github.com/angular-ui/ui-select/issues/1678
--- a/app/views/directives/lifecycle-hook.html
+++ b/app/views/directives/lifecycle-hook.html
@@ -1,0 +1,133 @@
+<ng-form name="editForm">
+  <div ng-switch="type">
+    <div class="help-block" ng-switch-when="pre">Pre hooks execute before the deployment begins.</div>
+    <div class="help-block" ng-switch-when="mid">Mid hooks execute after the previous deployment is scaled down to zero and before the first pod of the new deployment is created.</div>
+    <div class="help-block" ng-switch-when="post">Post hooks execute after the deployment strategy completes.</div>
+  </div>
+
+  <div class="gutter-top" ng-if="view.hookExists">
+    <fieldset ng-disabled="view.isDisabled">
+
+      <div class="form-group">
+        <label for="actionType" class="required">Lifecycle Action</label><br/>
+        <div class="radio">
+          <label class="radio-inline">
+            <input type="radio"
+                    name="{{type}}-action-newpod"
+                    ng-model="action.type"
+                    value="execNewPod"
+                    aria-describedby="action-help">
+            Run a specific command in a new pod
+          </label>
+          <label class="radio-inline">
+            <input type="radio"
+                  name="{{type}}-action-images"
+                  ng-model="action.type"
+                  value="tagImages"
+                  aria-describedby="action-help">
+              Tag image if the deployment succeeds
+          </label>
+        </div>
+        <div  id="action-help" class="help-block">
+          <span ng-if="action.type === 'execNewPod'">Runs a command in a new pod using the container from the deployment template. You can add additional environment variables and volumes.</span>
+          <span ng-if="action.type === 'tagImages'">Tags the current image as an image stream tag if the deployment succeeds.</span>
+          <a href="{{'new_pod_exec' | helpLink}}" aria-hidden="true" target="_blank"><span class="learn-more-inline">Learn more&nbsp;<i class="fa fa-external-link"></i></span></a>
+        </div>
+      </div>
+
+      <div ng-if="action.type === 'execNewPod'">
+        <div class="form-group">
+          <label class="required">Container Name</label>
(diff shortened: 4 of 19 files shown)


jhadvig:
`dist` will be added after review

app/scripts/directives/lifeCycleHook.js
```diff
@@ -0,0 +1,113 @@
...
+        ];
+        $scope.istagHook = {};
+        $scope.removedHookParams = {};
+
+        $scope.actionType = _.has($scope.hookParams, 'tagImages') ? "tagImages" : "execNewPod";
+
+        var setImageOptions = function(imageData) {
+          var istag = {};
+          if (!_.isEmpty(imageData)) {
+            istag = {namespace: imageData.namespace || $scope.namespace, imageStream: imageData.name.split(':')[0], tagObject: {tag: imageData.name.split(':')[1]}};
```
spadgett: Don't split twice the string twice. This is more readable with newlines.

```js
var imageNameParts;
if (!_.isEmpty(imageData)) {
  imageNameParts = imageData.name.split(':');
  istag = {
    namespace: imageData.namespace || $scope.namespace,
    imageStream: imageNameParts[0],
    tagObject: {
      tag: imageNameParts[1]
    }
  };
} else {
  istag = {
    namespace: $scope.namespace,
    imageStream: "",
    tagObject: {
      tag: ""
    }
  };
}
```

app/scripts/controllers/edit/deploymentConfig.js
```diff
@@ -0,0 +1,289 @@
...
+      return trigger;
+    };
+
+    var updateTriggers = function() {
+      var updatedTriggers = _.remove($scope.updatedDeploymentConfig.spec.triggers, function(trigger) {return trigger.type !== "ImageChange"});
+      _.each($scope.containersDataMap, function(containerData, containerName) {
+        if (containerData.hasDeploymentTrigger) {
+          updatedTriggers.push(assembleImageChangeTrigger(containerName, containerData.triggerData.istag, containerData.triggerData.data));
+        } else {
+          _.find($scope.updatedDeploymentConfig.spec.template.spec.containers, function(container) {return container.name === containerName}).image = containerData.image;
```
spadgett: Ah I see. I'd split this into two lines with a local var. Too easy to miss that you're setting the image value on the result.

app/scripts/controllers/edit/deploymentConfig.js
```diff
@@ -0,0 +1,289 @@
...
+            containerNames: [containerName],
+            from: istagObject
+          }
+        };
+      }
+      return trigger;
+    };
+
+    var updateTriggers = function() {
+      var updatedTriggers = _.remove($scope.updatedDeploymentConfig.spec.triggers, function(trigger) {return trigger.type !== "ImageChange"});
```
spadgett: But the code seems to be removing anything that's *not* an ICT?

jhadvig: It's kinds weird since but in fact it's vice versa... not sure if they have mistake in their docs but this will return array of non-ICT triggers

jhadvig: anyway.. I've updated went with the `_.filter` :+1:

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                              type="text"
+                              autocorrect="off"
+                              autocapitalize="off"
+                              spellcheck="false"
+                              aria-describedby="image-name-help"
+                              required>
+                          </div>
+                          <div class="help-block" id="image-name-help">Specifies a image which can carry out a deployment.</div>
+
+                          <div class="form-group gutter-top">
```
spadgett: Is `gutter-top` needed if the field above is inside a `form-group`?

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                            <edit-command args="strategyData.customParams.command"></edit-command>
+                          </div>
+                          <div class="form-group">
+                            <label>Environment Variables</label>
+                            <key-value-editor
+                              entries="strategyData.customParams.environment"
+                              key-validator="[a-zA-Z][a-zA-Z0-9_]*"
+                              key-validator-error-tooltip="A valid environment variable name is an alphanumeric (a-z and 0-9) string beginning with a letter that may contain underscores."
+                              add-row-link="Add environment variable"></key-value-editor>
+                            <div class="help-block">Specifies environment which will be given to the container for Image.</div>
```
spadgett: I'd just remove this help-block.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                  placeholder="600"
+                                  ng-pattern="/^\d+$/"
+                                  min="0"
+                                  select-on-focus
+                                  class="form-control"
+                                  aria-describedby="strategyTimeout">
+                              <span class="input-group-addon">seconds</span>
+                            </span>
+                            <div class="help-block" ng-attr-id="strategyTimeout">
+                              How long should deployment configuration retry to run the pod before times out.
```
spadgett: Maybe

"How long to wait for a pod to scale up before giving up."

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                      name="updatePeriod"
+                                      ng-model="strategyData[displayedParams].updatePeriodSeconds"
+                                      ng-pattern="/^\d+$/"
+                                      min="0"
+                                      select-on-focus
+                                      class="form-control"
+                                      aria-describedby="updatePeriod">
+                                  <span class="input-group-addon">seconds</span>
+                                </span>
+                                <div class="help-block" ng-attr-id="updatePeriod">
```
spadgett: `id="updatePeriod"`, you don't need `ng-attr-` here

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                  Time to wait between retrying to run individual pod.
```
spadgett: Is this correct? That sounds like timeout.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                </span>
+                                <div class="help-block" ng-attr-id="interval">
+                                  Time to wait between polling deployment status after running a pod.
+                                </div>
+                                <div ng-if="form.interval.$invalid && form.interval.$touched" class="has-error">
+                                  <div ng-if="form.interval.$error.number" class="help-block">
+                                    Must be a number.
+                                  </div>
+                                  <div ng-if="form.interval.$error.min" class="help-block">
+                                    Delay can't be negative.
```
spadgett: Interval can't be negative.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                </span>
+                                <div class="help-block" ng-attr-id="updatePeriod">
+                                  Time to wait between retrying to run individual pod.
+                                </div>
+                                <div ng-if="form.updatePeriod.$invalid && form.updatePeriod.$touched" class="has-error">
+                                  <div ng-if="form.updatePeriod.$error.number" class="help-block">
+                                    Must be a number.
+                                  </div>
+                                  <div ng-if="form.updatePeriod.$error.min" class="help-block">
+                                    Delay can't be negative.
```
spadgett: Update period can't be negative.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                            </span>
+                            <div class="help-block" ng-attr-id="strategyTimeout">
+                              How long should deployment configuration retry to run the pod before times out.
+                            </div>
+                            <div ng-if="form.strategyTimeout.$invalid && form.strategyTimeout.$touched" class="has-error">
+                              <div ng-if="form.strategyTimeout.$error.number" class="help-block">
+                                Must be a number.
+                              </div>
+                              <div ng-if="form.strategyTimeout.$error.min" class="help-block">
+                                Delay can't be negative.
```
spadgett: Timeout can't be negative.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+
+                            <div class="form-group">
+                              <label for="maxUnavailable">Maximal Unavailable</label>
+                              <div ng-class="{ 'has-error': form.maxUnavailable.$invalid && form.maxUnavailable.$touched }">
+                                <input id="maxUnavailable"
+                                    type="text"
+                                    placeholder="25%"
+                                    name="maxUnavailable"
+                                    ng-model="strategyData[displayedParams].maxUnavailable"
+                                    ng-pattern="/^\d+[%]{0,1}$/"
```
spadgett: Does this regex work?

`/^\d+%?$/`

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                    min="0"
+                                    select-on-focus
+                                    class="form-control"
+                                    aria-describedby="max-unavailable-help">
+                              </div>
+                              <div class="help-block" id="max-unavailable-help">
+                                Maximum number of pods that can be unavailable during the update. Can be set to either a percentage (e.g., 10%) or an absolute value (e.g., 2).
+                                <a href="{{'rolling_strategy' | helpLink}}" aria-hidden="true" target="_blank"><span class="learn-more-inline">Learn more&nbsp;<i class="fa fa-external-link"></i></span></a>
```
spadgett: Remove aria-hidden from the a element. Add aria-hidden to the i element.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                    ng-model="strategyData[displayedParams].maxSurge"
+                                    ng-pattern="/^\d+[%]{0,1}$/"
+                                    min="0"
+                                    select-on-focus
+                                    class="form-control"
+                                    aria-describedby="maxSurge">
+                              </div>
+                              <div class="help-block" ng-attr-id="maxSurge">
+                                Maximum number of pods that can be scheduled above the original number of pods. Can be set to either a percentage (e.g., 10%) or an absolute value (e.g., 2).
+                                <a href="{{'rolling_strategy' | helpLink}}" aria-hidden="true" target="_blank"><span class="learn-more-inline">Learn more&nbsp;<i class="fa fa-external-link"></i></span></a>
```
spadgett: Remove aria-hidden from the a element. Add aria-hidden to the i element.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                    aria-describedby="max-unavailable-help">
+                              </div>
+                              <div class="help-block" id="max-unavailable-help">
+                                Maximum number of pods that can be unavailable during the update. Can be set to either a percentage (e.g., 10%) or an absolute value (e.g., 2).
+                                <a href="{{'rolling_strategy' | helpLink}}" aria-hidden="true" target="_blank"><span class="learn-more-inline">Learn more&nbsp;<i class="fa fa-external-link"></i></span></a>
+                              </div>
+                              <div ng-if="form.maxUnavailable.$invalid && form.maxUnavailable.$touched" class="has-error">
+                                <span ng-if="form.maxUnavailable.$error.pattern && !form.maxUnavailable.$error.min" class="help-block">
+                                  Must be a whole number or percentage.
+                                </span>
```
spadgett: Do you need another error message for `$error.min`?

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                                  available-volumes='volumes'
+                                                  available-containers='containersName'
+                                                  namespace="projectName">
+                                </life-cycle-hook>
+                              </div>
+                            </div>
+                          </div>
+                        </div>
+                        <div class="gutter-top">
+                          <div ng-if="!view.advancedStrategyOptions">To set additional parameters or edit lifecycle hooks, switch to <a href="" ng-click="view.advancedStrategyOptions = true">advanced strategy configuration options.</a></div>
```
spadgett:
```
To set additional parameters or edit lifecycle hooks, view <a href="" ...>advanced options</a>.
```

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                          <a ng-if="view.advancedStrategyOptions" href="" ng-click="view.advancedStrategyOptions = false">Hide advanced strategy configuration options.</a>
```
spadgett: Hide advanced options.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                        </div>
+                      </dl>
+                    </div>
+
+
+                    <div class="section">
+                      <h3 class="with-divider">Image Configuration</h3>
```
spadgett: Maybe just 

```
<h3 class="with-divider">Images</h3>
```

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                      <dl class="dl-horizontal left">
+
+                        <div ng-repeat="(containerName, containerData) in containersDataMap">
+
+                          <div class="container-name">
+                            <h4>Container {{containerName}}</h4>
+                          </div>
+                          <div class="form-group" ng-if="containerData.hasDeploymentTrigger">
+                            <label class="required">Trigger image</label>
```
spadgett: Image Change Trigger

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                            <istag-select model="containerData.triggerData.istag" select-disabled="disableInputs" include-shared-namespace="true"></istag-select>
+                            <span class="help-block">New deployment will be automatically triggers once specified image is updated.</span>
```
spadgett: "New deployments will automatically run when the image stream tag is updated."

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                id="imageName"
+                                name="imageName"
+                                ng-model="containerData.image"
+                                type="text"
+                                autocorrect="off"
+                                autocapitalize="off"
+                                spellcheck="false"
+                                required>
+                            </div>
+                            <span class="help-block">New deployment can be triggered only manually.</span>
```
spadgett: I'd probably just remove this help block.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                          </div>
+
+                          <div class="checkbox form-group">
+                            <label>
+                              <input type="checkbox" ng-model="containerData.hasDeploymentTrigger">
+                              Automatically trigger new deployments when image is updated.
```
spadgett: "Automatically start new deployments when the image stream tag is updated."

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                                        secrets-by-type="secretsByType"
+                                        service-account-to-link="default"
+                                        alerts="alerts"
+                                        allow-multiple-secrets="true">
+                            </osc-secrets>
+                          </div>
+                        </div>
+
+                        <div class="gutter-top">
+                          <div ng-if="!view.advancedImageOptions">To set secrets for pulling your images from private image registries, switch to <a href="" ng-click="view.advancedImageOptions = true">advanced image configuration options.</a></div>
```
spadgett:
```
...view <a href="" ...>advanced options</a>.
```

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                          <a ng-if="view.advancedImageOptions" href="" ng-click="view.advancedImageOptions = false">Hide advanced image configuration options.</a>
```
spadgett: Hide advanced options.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+
+
+                    <div class="section">
+                      <h3 class="with-divider">Environment Variables<span class="help action-inline">
+                        <a href="">
+                        <i class="pficon pficon-help" data-toggle="tooltip" aria-hidden="true" data-original-title="Environment variables are used to configure and pass information to running containers.  These environment variables will be available during your container's runtime."></i>
+                        </a>
+                      </span></h3>
+
+                      <div ng-repeat="(containerName, containerData) in containersDataMap" class="">
```
spadgett: `class=""`?

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+                          <div ng-if="!view.advancedImageOptions">To set secrets for pulling your images from private image registries, switch to <a href="" ng-click="view.advancedImageOptions = true">advanced image configuration options.</a></div>
+                          <a ng-if="view.advancedImageOptions" href="" ng-click="view.advancedImageOptions = false">Hide advanced image configuration options.</a>
+                        </div>
+                      </dl>
+                    </div>
+
+
+                    <div class="section">
+                      <h3 class="with-divider">Environment Variables<span class="help action-inline">
+                        <a href="">
```
spadgett: I'd just remove the surround `a` element and keep the `i` as the tooltip.

app/scripts/directives/lifeCycleHook.js
```diff
@@ -0,0 +1,113 @@
...
+        };
+
+        $scope.$watch("istagHook.tagObject.tag", function() {
+          if (!_.has($scope.istagHook, ['tagObject', 'tag'])) {
+            return;
+          }
+          // Assamble image name when tag changes
+          _.set($scope.hookParams, ['tagImages', '0', 'to', 'kind'], 'ImageStreamTag');
+          _.set($scope.hookParams, ['tagImages', '0', 'to', 'namespace'], $scope.istagHook.namespace);
+          _.set($scope.hookParams, ['tagImages', '0', 'to', 'name'], $scope.istagHook.imageStream + ':' + $scope.istagHook.tagObject.tag);
```
jhadvig: I thought that we agreed that we will only edit the first IST (based on out blueJeans debate), since  I was told that most of the cases contain single IST in `tagImages`. But if you would like to I can add the logic that user can add/remove IST. Just let me know :)

spadgett: OK, but let's make sure we don't drop any others in the array. It might be better to show the values and make it read-only if there is more than one.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,380 @@
...
+              <form class="edit-form" name="form" novalidate ng-submit="save()">
+                <div class="row">
+                  <div class="col-lg-12">
+                    <div class="section">
+                      <h3>Deployment Strategy Configuration</h3>
+                      <dl class="dl-horizontal left">
+                        {{strategyParamsName}}
+                        <div class="form-group strategy-name">
+                          <label class="picker-label">Strategy Type</label>
+                          <ui-select ng-model="strategyData.type" search-enabled="false" ng-change="strategyChange(strategyData.type)">
```
jhadvig: yep it can

app/views/edit/deployment-config.html
jhadvig: should be right based on our api https://github.com/openshift/origin/blob/master/pkg/deploy/api/v1/types.go#L141

app/views/edit/deployment-config.html
jhadvig: it does

app/views/edit/deployment-config.html
jhadvig: since the input is text, because of the `%` sign that can occur in the input field, we can not check for `min`

spadgett:
Then you should have only one `ng-if` on the parent div if we only handle
one error.

Instead of

```html
<div ng-if="..." class="has-error">
  <span ng-if="..." class="help-block">...</span>
</div>
```

use

```html
<div ng-if="(all conditions here)" class="has-error">
  <span class="help-block">...</span>
</div>
```

jhadvig:
@spadgett I've update the PR. PTAL

spadgett:
Changing strategy type values several times, I hit this error:

```
Error: [ui.select:choices] Expected multiple .ui-select-choices-row but got '0'.
    at select.js:128
    at _ensureHighlightVisible (select.js:978)
    at select.js:420
    at angular.js:16414
    at completeOutstandingRequest (angular.js:4940)
    at angular.js:5328
```

spadgett:
`x` button is not vertically centered with the select here:

openshift_web_console

spadgett:
We don't need this many learn more links, particularly since they point to the same topic. I'd keep only the first.

openshift_web_console

spadgett:
The spacing seems off here. The help text should not be closer to the link below than the input it's describing.

openshift_web_console

spadgett:
We should use `&nbsp;` between "Learn more" and the icon instead of a normal space to keep the icon from wrapping on its own.

openshift_web_console

spadgett:
These need to be checkboxes instead of radio buttons because you can have both set.

openshift_web_console

"Run a specific command" should be checked by default for new hooks and validate that at least one of the checkboxes is selected.

spadgett:
We should warn the user if switching strategy will remove one of their hooks. For instance, switching from recreate to rolling will remove the mid hook. Or switching to custom will remove all hooks. But only if they have hooks defined.

spadgett:
These actions should be "Add another secret | Create new secret" to make it more obvious how they are different.

openshift_web_console

jhadvig:
@spadgett I've added commit(will squash afterwards) which is addressing most of yours comments.

> We should use &nbsp; between "Learn more" and the icon instead of a normal space to keep the icon from wrapping on its own.

I've already have the `&nbsp;` in the place for all the `Learn more` links. Not sure what could cause the issue :-/

> We should warn the user if switching strategy will remove one of their hooks. For instance, switching from recreate to rolling will remove the mid hook. Or switching to custom will remove all hooks. But only if they have hooks defined.

I've updated the dialog msg which is now saying:
"The timeout parameter and any pre or post lifecycle hooks will be copied from recreate strategy to rolling strategy. After saving the changes, recreate strategy parameters will be removed ."
This only applies for the Rolling <-> Recreate switch. We are not removing any params when switching to Custom strategy, which was part of the debate we had with @jwforres

spadgett:
But we want to warn them if a hook that exists will not longer run. That still applies to the custom strategy, which doesn't support the hooks.

app/scripts/controllers/edit/deploymentConfig.js
```diff
@@ -0,0 +1,297 @@
...
+    );
+
+    // helper for detemining if strategy switch was done between Rolling <-> Recreate strategy
+    var isRollingRecreateSwitch = function() {
+      return ($scope.strategyData.type !== 'Custom' && $scope.originalStrategy !== 'Custom' && $scope.strategyData.type !== $scope.originalStrategy);
+    };
+
+    $scope.strategyChanged = function() {
+      var pickedStrategyParams = getParamsPropertyName($scope.strategyData.type);
+      if (isRollingRecreateSwitch()) {
```
spadgett: I'd pull everything inside this `if` into a helper function, maybe `promptToMoveParams`. It'd be easier to follow the if/else logic in strategyChanged.

app/styles/_core.less
```diff
@@ -1132,7 +1159,7 @@ a.disabled-link {
     }
   }
   .checkbox {
-    margin: 10px;
+    margin: 10px 0px;
```
spadgett: Use `0` instead of `0px`. 0 values don't need a unit.

app/styles/_core.less
```diff
@@ -503,18 +503,34 @@ label.checkbox {
...
-  .picked-secret {
-    margin-bottom: 5px;
+  .remove-secret {
+    display: inline-block;
+    margin-left: 7px;
+    vertical-align: 0px;
   }
-  .directory {
-    width:90%;
+  .secret-name(@width: 90%) {
```
spadgett: I think I'd rather just have

```css
.secret-name {
  width: 90%;
  display: inline-block;
}
```

and later

```css
.basic-secrets {
  .secret-name {
    width: 95%;
  }
}
```

I don't think the mixin is necessary?

app/views/directives/lifecycle-hook.html
```diff
@@ -0,0 +1,151 @@
...
+          </ui-select>
+          <div class="help-block">
+            Use the image for this container as the source of the tag.
+          </div>
+        </div>
+
+        <div class="form-group">
+          <label class="required">Tag As</label>
+          <istag-select model="istagHook" allow-custom-tag="true" select-disabled="view.isDisabled"></istag-select>
+        </div> -->
```
spadgett: Remove commented out code?

app/views/directives/lifecycle-hook.html
```diff
@@ -0,0 +1,151 @@
...
+                Use the image for this container as the source of the tag.
+              </div>
+            </div>
+            <div class="form-group">
+              <label class="required">Tag As</label>
+              <istag-select model="istagHook" allow-custom-tag="true" select-disabled="view.isDisabled"></istag-select>
+            </div>
+            <p ng-if="hookParams.tagImages.length > 1">
+              <span class="pficon pficon-info" aria-hidden="true"></span>
+              More then one image tags are defined in this deployment configuration. Only the first one is editable, below is displayed the rest.
```
spadgett: It's weird that the first one only is editable if there is more than one. Either all image tags should be read-only (when there are multiple) or we should just try to support editing all of them.

app/styles/_core.less
```diff
@@ -503,18 +503,34 @@ label.checkbox {
 }
 
 .osc-secrets-form {
-  .picked-secret {
-    margin-bottom: 5px;
+  .remove-secret {
+    display: inline-block;
+    margin-left: 7px;
+    vertical-align: 0px;
```
spadgett: You and I are seeing very different behavior for `vertical-align` with this icon :(

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                                                    available-volumes='volumeNames'
+                                                    available-containers='containerNames'
+                                                    namespace="projectName">
+                                  </lifecycle-hook>
+                                </div>
+                              </div>
+                            </div>
+                          </div>
+                          <div class="gutter-top">
+                            <div ng-if="!view.advancedStrategyOptions">To set additional parameters or edit lifecycle hooks, view <a href="" ng-click="view.advancedStrategyOptions = true">advanced strategy options</a></div>
```
spadgett: Here I'd have the period after the link since the entire sentence is not an action link, but I'd leave the period off for the hide link.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                            <div class="checkbox form-group">
+                              <label>
+                                <input type="checkbox" ng-model="containerData.hasDeploymentTrigger">
+                                Automatically start new deployments when the image stream tag is updated.
+                              </label>
+                            </div>
+
+                          </div>
+
+                          <div ng-if="view.advancedImageOptions">
```
spadgett: Since there's only one field we're hiding and showing, I wonder if it's even worth having the advanced link. (I know that's different than what we told you.)

@jwforres opinion?

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                            <a ng-if="view.advancedImageOptions" href="" ng-click="view.advancedImageOptions = false">Hide advanced image options</a>
+                          </div>
+                        </dl>
+                      </div>
+
+
+                      <div class="section">
+                        <h3 class="with-divider">
+                          Environment Variables
+                          <span class="help action-inline"></span>
```
spadgett: Remove this empty `span`

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                        </h3>
+
+                        <div ng-repeat="(containerName, containerData) in containerConfigByName">
```
spadgett: nit: I'd probably call this `containerConfig` to match the map name.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                            add-row-link="Add environment variable"></key-value-editor>
+                        </div>
+                      </div>
+
+                      <div class="buttons gutter-top-bottom">
+                        <button
+                          type="submit"
+                          class="btn btn-primary btn-lg"
+                            ng-disabled="form.$invalid || form.$pristine || disableInputs">
+                            Save
```
spadgett: indentation

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                                  name="imageName"
+                                  ng-model="strategyData.customParams.image"
+                                  type="text"
+                                  autocorrect="off"
+                                  autocapitalize="off"
+                                  spellcheck="false"
+                                  aria-describedby="image-name-help"
+                                  required>
+                              </div>
+                              <div class="help-block" id="image-name-help">Specifies a image which can carry out a deployment.</div>
```
spadgett: "An image that can carry out the deployment."

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                            <ui-select ng-model="strategyData.type" search-enabled="false" ng-change="strategyChanged()">
+                              <ui-select-match>{{$select.selected}}</ui-select-match>
+                              <ui-select-choices repeat="strategyType in deploymentConfigStrategyTypes">
+                                {{strategyType}}
+                              </ui-select-choices>
+                            </ui-select>
+                            <div>
+                              <span ng-switch="strategyData.type">
+                                <span class="help-block" ng-switch-when="Recreate">
+                                  Recreate strategy has basic rollout behavior and supports lifecycle hooks for injecting code into the deployment process.
```
spadgett: I'd start each of these strategy descriptiong with "The".

"The recreate strategy has basic..."

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                                      name="maxUnavailable"
+                                      ng-model="strategyData[strategyParamsPropertyName].maxUnavailable"
+                                      ng-pattern="/^\d+%?$/"
+                                      select-on-focus
+                                      class="form-control"
+                                      aria-describedby="max-unavailable-help">
+                                </div>
+                                <div ng-if="form.maxUnavailable.$invalid && form.maxUnavailable.$touched && form.maxUnavailable.$error.pattern" class="has-error">
+                                  <span class="help-block">
+                                    Must be a non negative whole number or percentage.
```
spadgett: "non-negative"

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,373 @@
...
+                                      name="maxSurge"
+                                      ng-model="strategyData[strategyParamsPropertyName].maxSurge"
+                                      ng-pattern="/^\d+%?$/"
+                                      select-on-focus
+                                      class="form-control"
+                                      aria-describedby="maxSurge">
+                                </div>
+                                <div ng-if="form.maxSurge.$invalid && form.maxSurge.$touched && form.maxSurge.$error.pattern" class="has-error">
+                                  <span class="help-block">
+                                    Must be a non negative whole number or percentage.
```
spadgett: "non-negative"

spadgett:
@jhadvig Can you add the `pointer-events` line and comment to this rule in _select.less? It fixes the ui-select error for me.

```less
  .ui-select-toggle > .caret {
    font-size: (@font-size-base - 1);
    font-style: normal;
    right: 6px;
    top: 10px;
    // Workaround for https://github.com/angular-ui/ui-select/issues/224
    pointer-events: none;
  }
```

app/views/edit/deployment-config.html
jhadvig: imho I would keep it hidden. The idea of having "basic/advanced view" is to compress the default view to bare minimum, for both BC and DC editor

app/scripts/controllers/edit/deploymentConfig.js
```diff
@@ -157,39 +157,40 @@ angular.module('openshiftConsole')
       return ($scope.strategyData.type !== 'Custom' && $scope.originalStrategy !== 'Custom' && $scope.strategyData.type !== $scope.originalStrategy);
     };
 
+    var promptToMoveParams = function(pickedStrategyParams) {
+      if (!_.has($scope.strategyData, pickedStrategyParams)) {
```
spadgett: I prefer an early return rather than wrapping everything in an `if`. It's just a little easier to follow.

```js
if (_.has($scope.strategyData, pickedStrategyParams)) {
  return;
}

// Show modal...
```

app/scripts/controllers/edit/deploymentConfig.js
```diff
@@ -157,39 +157,40 @@ angular.module('openshiftConsole')
...
-          modalInstance.result.then(function () {
-            // Move parameters that belong to the origial strategy to the picked one.
-            $scope.strategyData[pickedStrategyParams] = $scope.strategyData[getParamsPropertyName($scope.originalStrategy)];
-            $scope.paramsMoved = getParamsPropertyName($scope.originalStrategy);
-          }, function() {
-            // Create empty parameters for the newly picked strategy
-            $scope.strategyData[pickedStrategyParams] = {};
-          });
-        }
+        promptToMoveParams(pickedStrategyParams);
```
spadgett: Thanks, I think this is a lot easier to follow.

app/styles/_core.less
```diff
@@ -503,34 +503,40 @@ label.checkbox {
...
-  }
-  .secret-name(@width: 90%) {
-    width: @width;
-    display: inline-block;
-  }
   .basic-secrets {
+    .remove-secret {
+      display: inline-block;
+      margin-left: 7px;
+      vertical-align: 8px;
```
spadgett: 👍 

glad we figured this out

app/styles/_select.less
```diff
@@ -45,6 +45,10 @@
     font-style: normal;
     right: 6px;
     top: 10px;
+    // Workaround for https://github.com/angular-ui/ui-select/issues/224
+    // A runtime error occurs when you click exactly on the down caret for some
+    // selects, such as the strategy type select in the deployment config editor.
+    pointer-events: none;
```
spadgett: 👍

app/views/directives/lifecycle-hook.html
```diff
@@ -115,13 +96,14 @@
...
               <istag-select model="istagHook" allow-custom-tag="true" select-disabled="view.isDisabled"></istag-select>
             </div>
-            <p ng-if="hookParams.tagImages.length > 1">
+          </div>
+
+          <div class="read-only-tag-image" ng-if="hookParams.tagImages.length > 1">
+            <p class="read-only-info" ng-if="$first">
               <span class="pficon pficon-info" aria-hidden="true"></span>
-              More then one image tags are defined in this deployment configuration. Only the first one is editable, below is displayed the rest.
+              More then one image tags are defined in this deployment configuration. To edit them, use deployment config YAML editor.
```
spadgett: "More than one image tag is defined. To change image tags, use the YAML editor."

app/views/directives/lifecycle-hook.html
```diff
@@ -115,13 +96,14 @@
...
             </p>
-          </div>
-          <div class="gutter-bottom read-only-tag-image" ng-if="!$first">
-            <span class="container-name">{{tagImage.containerName}}</span><i class="fa fa-long-arrow-right"></i><span class="image-tag">{{tagImage.to.namespace || namespace}}/{{tagImage.to.name}}</span>
+            <span class="container-name">{{tagImage.containerName}}</span>&rarr;<span class="image-tag">{{tagImage.to.namespace || namespace}}/{{tagImage.to.name}}</span>
```
spadgett: I'd remove the `container-name` and `image-tag` styles and just use this:

```html
{{tagImage.containerName}}&nbsp;&rarr;&nbsp;{{tagImage.to.namespace || namespace}}/{{tagImage.to.name}}
```

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,366 @@
...
+                                  id="imageName"
+                                  name="imageName"
+                                  ng-model="containerConfig.image"
+                                  type="text"
+                                  autocorrect="off"
+                                  autocapitalize="off"
+                                  spellcheck="false"
+                                  required>
+                              </div>
+                              <span class="help-block">New deployment can be triggered only manually.</span>
```
spadgett: This isn't really true since there could be config change triggers as well. I'd just remove the help here.

app/views/edit/deployment-config.html
```diff
@@ -0,0 +1,366 @@
...
+                            </div>
+
+                            <div class="checkbox form-group">
+                              <label>
+                                <input type="checkbox" ng-model="containerConfig.hasDeploymentTrigger">
+                                Automatically start new deployments when the image stream tag is updated.
+                              </label>
+                            </div>
```
spadgett: Forgot to ask you do this, but we should move this above the ICT and Image Name fields.

spadgett:
Just a few small requests. Will need to squash commits and add dist before we merge.

We should have a config change trigger toggle to the editor, but that can be a follow on.

spadgett:
Needs more top margin above the second container name:

openshift_web_console

spadgett:
When I clicked the "Automatically start new deployments when the image stream tag is updated." the dropdown prefilled an image stream that was invalid.

openshift_web_console

jhadvig:
@spadgett regarding your last comment. The problem is that we can not determine the Tag of the IST for the ICT from the imageSpec, in case switching from manual to automatic trigger, since the API doesn't support it yet. So I've skipped it and fill the istag directive with the namespace and imageStream, which we can get from the imageSpec. 
Other solution would be to only prefill the namespace, but that could be confusing cause the imageStream is defined.

spadgett:
>
> The problem is that we can not determine the Tag of the IST for the ICT
> from the imageSpec, in case switching from manual to automatic trigger,
> since the API doesn't support it yet.


It's not the tag. The image stream exists in a different namespace, so what
you prefilled is wrong.

jhadvig:
@spadgett I've updated the PR based on your comments + added `dist` + squashed
PTAL

spadgett:
[test]

spadgett:
[merge]

Issue #649, opened 2016-10-08
Monitor menu item off the screen


andrewklau:
![image](https://cloud.githubusercontent.com/assets/2239920/19213364/a45b7bda-8db6-11e6-8129-a787be572cb9.png)
Monitor item is off the screen, and can't scroll down to see what the button name is. 

The X1 carbon has a weird resolution...
![image](https://cloud.githubusercontent.com/assets/2239920/19213370/c675ca9a-8db6-11e6-8bac-025d22591b12.png)

PR #663, opened 2016-10-12
Make left nav height shorter when >600px

sg00dwin:
When the viewport height is less than 600px the icons are hidden and the left nav is reduced to 440px height. Also, the nav width is set at 145px up to the xlg screen width of 1600. 

Fixes https://github.com/openshift/origin-web-console/issues/649

9lk7vudbm5

### DIFF
--- a/app/styles/_sidebar.less
+++ b/app/styles/_sidebar.less
@@ -179,7 +179,7 @@
     .navbar-sidebar {
       border-right: 1px solid @sidebar-os-border-color;
       border-bottom: 0;
-      width: @sidebar-left-width-sm;
+      width: @sidebar-left-width-md;
       position: absolute;
       top:1px;
       right:0;
@@ -234,7 +234,7 @@
               font-size: 16px;
               position: absolute;
               right: 13px;
-              top: 37%;
+              top: 39%;
             }
             span {
               display: block;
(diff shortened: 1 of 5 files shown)


sg00dwin:
This doesn't address the issue of if there are more items added to the left nav. I batted around a few options with Rob. So along with the removal of the icons when >600 to make the nav scrollable the secondary menu can be tucked beneath it's primary nav item. Here's a quick and dirty inline edit showing it as such.

dgqb7ukqvj

sg00dwin:
@jwforres @spadgett PTAL
cc @rhamilto

spadgett:
> This doesn't address the issue of if there are more items added to the left nav.

If we're worried about this, I still think the simplest solution is to make the entire page scrollable rather than only a content div. Then the nav scrolls with the content. We can still have a sticky header. It might simply our markup and CSS and get rid of the log scrolling hacks.

sg00dwin:
Just so we are aware... going back to having a single scroll page scroll will reintroduce several issues that were addressed in the current structure of independently scrollable panels.

- Having a single scroll causes loss of current location/context, currently the left nav provides a quick reference to current location. How would this be addressed?
- Quick navigation between sections will be hindered on long pages requiring users to scroll up first.
- If we have a view, like on the previous overview, where a selected object in the main area shows related details in a separate column it will require some sort of "affix content to location" functionality.

sg00dwin:
@spadgett Do we want to include this pr as a short term fix?
cc @jwforres

app/styles/_sidebar.less
```diff
@@ -253,26 +253,31 @@
...
-          padding-bottom: @sidebar-os-padding-y + 10;
-          padding-top: @sidebar-os-padding-y + 10;
-          .fa-angle-right {
-            font-size: 18px;
-          }
-        }
+// Prevent left nav from extending off viewport when > 600px by hiding the icon
+// which reduces the nav height
+.sidebar-left .navbar-sidebar ul.nav-sidenav-primary > li > a {
+  @media (max-height: 600px) and (min-width: @screen-sm-min) {
```
rhamilto: Make max-height value a variable?

spadgett:
> Do we want to include this pr as a short term fix?

Sure

sg00dwin:
@spadgett @rhamilto updated pr.
created variable for max-height

spadgett:
[merge]

spadgett:
> Error: Timed out waiting for the WebDriver server

[merge]

PR #664, opened 2016-10-12
Improve display of metrics usage values on overview

spadgett:
Fix how small metrics values look on the overview (regression from 1.3).

* Show "0" instead of very small values like "0.0006 KiB/s"
* Only round to one significant digit if less than 1 ("0.02" instead of "0.023")

@jwforres PTAL


jwforres:
[merge]

PR #666, opened 2016-10-12
Share get, watch resolve fn in pod controller

benjaminapetersen:
@spadgett @jwforres 
re todo #129

### DIFF
--- a/app/scripts/controllers/pod.js
+++ b/app/scripts/controllers/pod.js
@@ -49,6 +49,7 @@ angular.module('openshiftConsole')
     $scope.selectedTab = {};
 
     var watches = [];
+    var requestContext = null;
 
     // Check if the metrics service is available so we know when to show the tab.
     MetricsService.isAvailable().then(function(available) {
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/pod.js
```diff
@@ -204,54 +205,47 @@ angular.module('openshiftConsole')
...
 
+    var podResolved = function(pod, action) {
+      $scope.loaded = true;
+      $scope.pod = pod;
+      var pods = {};
+      pods[pod.metadata.name] = pod;
+      setLogVars(pod);
+      setContainerVars();
+      ImageStreamResolver.fetchReferencedImageStreamImages(pods, $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, requestContext);
+      $scope.containerTerminals = makeTerminals();
```
jwforres: not sure this is a safe update, you've changed it so its now always calling makeTerminals on every update of the pod

app/scripts/controllers/pod.js
```diff
@@ -204,54 +205,47 @@ angular.module('openshiftConsole')
...
     };
 
+    var podResolved = function(pod, action) {
+      $scope.loaded = true;
+      $scope.pod = pod;
+      var pods = {};
+      pods[pod.metadata.name] = pod;
+      setLogVars(pod);
+      setContainerVars();
+      ImageStreamResolver.fetchReferencedImageStreamImages(pods, $scope.imagesByDockerReference, $scope.imageStreamImageRefByDockerReference, requestContext);
```
jwforres: I don't want to call this on an update, the image isn't going to change

benjaminapetersen: np, got this

app/scripts/controllers/pod.js
benjaminapetersen: good catch here as well, moved that out into just the `.get()`

app/scripts/controllers/pod.js
```diff
@@ -204,54 +205,51 @@ angular.module('openshiftConsole')
...
+      setContainerVars();
+      if (action === "DELETED") {
+        $scope.alerts["deleted"] = {
+          type: "warning",
+          message: "This pod has been deleted."
+        };
+      }
+      var pods = {};
+      pods[pod.metadata.name] = pod;
+      return pods;
```
jwforres: this is a little odd to me. there isnt really a reason to have this here when the object stuff is only being used for the image stream resolve

benjaminapetersen: Agree, is a little strange since its only for a fn arg.

benjaminapetersen: Updated this & put the `pods[pod.metadata.name] = pod;` back as it was originally. 
rebased

jwforres:
[merge]

PR #670, opened 2016-10-13
Show a warning on the overview when metrics connection fails

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1382728

openshift_web_console

@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -565,6 +565,33 @@ angular.module('openshiftConsole')
       MetricsService.isAvailable(true).then(function(available) {
         $scope.showMetrics = available;
       });
+
+      $scope.$on('metrics-connection-failed', function(e, data) {
+        var hidden = AlertMessageService.isAlertPermanentlyHidden('metrics-connection-failed');
+        if (hidden || $scope.alerts['metrics-connection-failed']) {
+          return;
+        }
+
+        $scope.alerts['metrics-connection-failed'] = {
+          type: 'warning',
+          message: 'An error occurred getting metrics.',
+          links: [{
+            href: data.url,
+            label: 'Open metrics URL',
+            target: '_blank'
+          }, {
+            href: '',
+            label: "Don't show me again",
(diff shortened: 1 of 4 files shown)


itewk:
Not to be nit picky but is it possible to make the warning message more specific? Like specifically say the SSL cert needs to be accepted for it to work. The way it is now if I see this I am going to assume something is wrong with my metrics deployment and I am not going to know just accepting the cert will fix all my problems.

spadgett:
@itewk I'm concerned about encouraging users to accept invalid SSL certificates. I specifically avoided that. You shouldn't need to manually accept the cert in a properly configured cluster.

Also we can't know with certainty that certificates are the problem. We get back status code 0, which can have other causes like not having network connectivity.

itewk:
@spadgett alright. is there at least a KB article somewhere? Right now its like this magic trick. You accept the cert and whoosh everything starts working. No indication that was what fixed it. And it seems to be a thing that anyone who has done this before "just knows" but if someone is installing this for the first time without outside help it is confusing.

spadgett:
> Right now its like this magic trick.

It's hard because it's something you should **not** do except in a development environment. There's a reason browsers show that warning. In a real cluster, the cluster admin should configure metrics with a valid certificate.

app/scripts/controllers/overview.js
```diff
@@ -565,6 +565,33 @@ angular.module('openshiftConsole')
...
+        var hidden = AlertMessageService.isAlertPermanentlyHidden('metrics-connection-failed');
+        if (hidden || $scope.alerts['metrics-connection-failed']) {
+          return;
+        }
+
+        $scope.alerts['metrics-connection-failed'] = {
+          type: 'warning',
+          message: 'An error occurred getting metrics.',
+          links: [{
+            href: data.url,
```
jwforres: Is this opening the particular request that failed, or is it going to the splash page?

jwforres:
[merge]

Issue #620, opened 2016-10-04
ENV variable manager sluggish with lots of variables


andrewklau:
We had a case where a pod had two containers, both with about 20+ ENV variables set. This ended up being very sluggish for the user when scrolling down and modifying these ENV variables.

Perhaps they could be split by a dropdown box similar to the terminal and logs. Although if an individual container still had a lot of ENV variables it would still also be quite slow.

jwforres:
we definitely shouldn't be sluggish at 40 env vars, its not that many.  1000 would have been a different story.  @benjaminapetersen consider this next up after the membership stuff is done

benjaminapetersen:
@andrewklau did you mean "sluggish" as in browser lag/response? I'm creating a few scenarios with a fairly absurd # of env vars and not experiencing that thus far.  However, I can definitely see that the user experience is less than idea at a certain point.

andrewklau:
@benjaminapetersen yeah browser response when scrolling it kept jolting as I scrolled up and down the list (also when using the browser search function)

spadgett:
@andrewklau Any chance there was a deployment in progress or something that could trigger many page updates?

andrewklau:
@spadgett there were a few deployments happening in the project, can't remember if it was for the one being modified. This occurred around the same time we were seeing #621 so possibly related

jwforres:
Ahhh OK, I can buy this causing a problem. So Ben you will need to set up a
pod crash loop example that has a bunch of env on it.

benjaminapetersen:
Right. I don't think the editor itself has this performance issue so much as we are allowing it to be used when things are in flux.   Thinking about how to solve -- is it reasonable to disable the editor in certain states?

spadgett:
@benjaminapetersen You might even use `$timeout` to simulate frequent digest loops and profile the KVE in Chrome to see where the time is spent.

spadgett:
We have no `track by` here, which is probably contributing to this.

spadgett:
Forgot the link...

https://github.com/spadgett/origin-web-console/blob/8c8ea04543ee0e3bb427c8af4120f960dfb0f4a2/app/views/directives/key-value-editor.html#L23

spadgett:
#2416 should at least partially fix this.

PR #674, opened 2016-10-13
Add select picker to env tab on pod page

benjaminapetersen:
Looking @ addressing #620 

screen shot 2016-10-13 at 4 44 22 pm

@jwforres @spadgett

### DIFF
--- a/app/scripts/controllers/pod.js
+++ b/app/scripts/controllers/pod.js
@@ -47,7 +47,14 @@ angular.module('openshiftConsole')
 
     // Must always be initialized so we can watch selectedTab
     $scope.selectedTab = {};
-
+    // we might want to start structuring this data more?
+    $scope.tabs = {
+      env: {
+        selected: {
+          container: null
+        }
+      }
+    };
     var watches = [];
 
     // Check if the metrics service is available so we know when to show the tab.
@@ -217,6 +224,7 @@ angular.module('openshiftConsole')
           function(pod) {
             $scope.loaded = true;
             $scope.pod = pod;
+            $scope.tabs.env.selected.container = pod.spec.containers[0].name;
             setLogVars(pod);
             setContainerVars();
             var pods = {};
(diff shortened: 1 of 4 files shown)


jwforres:
Is the "Container hello-openshift Environment Variables" something being passed into the kve? Otherwise i think I'd rather see Container [container dropdown] environment variables

spadgett:
It seems odd that the environment tab on the pod page would be different than deployment configs and replication controllers. But if we change the other pages, we'll have to think about editing environment variables. Not sure how well the dropdown works in those cases. If I add a variable and switch containers, will I lose my change? What if I have an invalid value in a field that's not visible?

This also doesn't fix any performance problems if there are a lot of variables for one container. I think we need to reproduce and profile the key value editor to find the bottleneck.

benjaminapetersen:
@jwforres agree, lemme check and get rid of that extra heading.
@spadgett yeah, I dont think this fixes all performance probs by any means, but I do think its more consistent w/other tabs where we use the drop down for containers.  Its at least easier to digest.  Def need to figure out the performance bottleneck issue as well.

spadgett:
>  but I do think its more consistent w/other tabs where we use the drop down for containers

Right, but it's an editor, too. So we need to solve the usability problems with using selects for editing. And we should look at updating the "Set Resource Limits" and "Add Health Checks" editors if we add the container select here.

benjaminapetersen:
Ie, decide if we want a "done" button for each one & alert the user if they select a different container but haven't finished the first?  We could let them edit both & submit, but I wonder if that is a more confusing interface.

spadgett:
> decide if we want a "done" button for each one

The problem is the first done click will usually trigger a new deployment for a deployment config. Then you'll get a bunch of conflict errors if you try to edit the second container's variables because of deployment status updates.

benjaminapetersen:
Hmm. So perhaps a msg indicating "currently editing, not saved", might even be smart to put the "done" button on top...?

benjaminapetersen:
Looking at this again. Ideally we would disable  the `save` on submit and enable when things have settled, but it isn't clear to me that we can really know if its truly settled.  

I'm looking at DCs now.  When I change an env var, I'll get ~10 updates via `DataService.watchObject("deploymentconfigs"...)`.

PR #679, opened 2016-10-14
Target deployment or deployment config for HPA

### LINKED ISSUE #675: Add autoscaler on the replica set page targets replica set, not deployment
If the replica set is owned by a deployment, the HPA should target the deployment instead.

/cc @jwforres

spadgett:
If looking at a replica set owned by a deployment or replication controller owned by a deployment config, use the owner for the HPA scale target.

Fixes #675
@jwforres PTAL


jwforres:
[merge]

PR #680, opened 2016-10-14
Build hook editor on build configuration page

MarkDeMaria:
-Multi-select tag for Command, Script, and Args
-Warning states:
--On page load, if both Command and Script are both present, forms are disabled and user told to use YAML
--If either Command or Script present, and user tries to add the type not present, Save button is disabled and has-error info-block informs user of error and told to use YAML

All Selected
image

Script & Command present at page load. Form disabled and notification present
image

Page loaded with either script or command. Entered the other. Save button disabled and message shown to user.
image

### DIFF
--- a/app/scripts/constants.js
+++ b/app/scripts/constants.js
@@ -51,6 +51,7 @@ window.OPENSHIFT_CONSTANTS = {
     "builds":                  "architecture/core_concepts/builds_and_image_streams.html#builds",
     "image-streams":           "architecture/core_concepts/builds_and_image_streams.html#image-streams",
     "storage":                 "architecture/additional_concepts/storage.html",
+    "build-hooks":             "dev_guide/builds.html#build-hooks",
     // default should remain last, add new links above
     "default":                 "welcome/index.html"
   },
--- a/app/scripts/controllers/edit/buildConfig.js
+++ b/app/scripts/controllers/edit/buildConfig.js
@@ -55,7 +55,8 @@ angular.module('openshiftConsole')
       "title": "Inline"
     }];
     $scope.view = {
-      advancedOptions: false
+      advancedOptions: false,
+      hasHooks: false
     };
     $scope.breadcrumbs = [
       {
@@ -117,6 +118,79 @@ angular.module('openshiftConsole')
       "SerialLatestOnly"
     ];
 
+    $scope.buildHookTypes = [{
+      id: "command",
+      label: "Command"
+    }, {
+      id: "script",
+      label: "Shell Script"
+    }, {
+      id: "args",
+      label: "Arguments to default image entry point"
+    }, {
+      id: "commandArgs",
+      label: "Command with arguments"
+    }, {
+      id: "scriptArgs",
+      label: "Shell script with arguments"
+    }];
+
+    $scope.buildHookSelection = {
+      type: {}
+    };
+
+    var getInitialBuildHookSelection = function() {
+      var hasArgs = !_.isEmpty(_.get($scope, 'buildConfig.spec.postCommit.args'));
+      var hasCommand = !_.isEmpty(_.get($scope, 'buildConfig.spec.postCommit.command'));
+      var hasScript = !!_.get($scope, 'buildConfig.spec.postCommit.script');
+      $scope.view.hasHooks = hasArgs || hasCommand || hasScript;
+
+      var id;
+
+      if (hasArgs && hasCommand) {
+        id = 'commandArgs';
+      } else if (hasArgs && hasScript) {
+        id = 'scriptArgs';
+      } else if (hasArgs) {
+        id = 'args';
+      } else if (hasScript) {
+        id = 'script';
+      } else {
+        id = 'command';
+      }
+
+      $scope.buildHookSelection.type = _.find($scope.buildHookTypes, { id: id });
+    };
+
+    var clearIncompatibleBuildHookFields = function() {
+      if (!$scope.view.hasHooks) {
+        delete $scope.updatedBuildConfig.spec.postCommit.command;
+        delete $scope.updatedBuildConfig.spec.postCommit.args;
+        delete $scope.updatedBuildConfig.spec.postCommit.script;
+      } else {
+        switch ($scope.buildHookSelection.type.id) {
+          case "script":
+            delete $scope.updatedBuildConfig.spec.postCommit.command;
+            delete $scope.updatedBuildConfig.spec.postCommit.args;
+            break;
+          case "command":
+            delete $scope.updatedBuildConfig.spec.postCommit.script;
+            delete $scope.updatedBuildConfig.spec.postCommit.args;
+            break;
+          case "args":
+            delete $scope.updatedBuildConfig.spec.postCommit.script;
+            delete $scope.updatedBuildConfig.spec.postCommit.command;
+            break;
+          case "scriptArgs":
+            delete $scope.updatedBuildConfig.spec.postCommit.command;
+            break;
+          case "commandArgs":
+            delete $scope.updatedBuildConfig.spec.postCommit.script;
+            break;
+        }
+      }
+    };
+
     AlertMessageService.getAlerts().forEach(function(alert) {
       $scope.alerts[alert.name] = alert.data;
     });
@@ -143,6 +217,8 @@ angular.module('openshiftConsole')
           // success
           function(buildConfig) {
             $scope.buildConfig = buildConfig;
+            getInitialBuildHookSelection();
+
             $scope.updatedBuildConfig = angular.copy($scope.buildConfig);
             $scope.buildStrategy = buildStrategy($scope.updatedBuildConfig);
             $scope.strategyType = $scope.buildConfig.spec.strategy.type;
@@ -180,7 +256,7 @@ angular.module('openshiftConsole')
               }
 
               if (imageOptions.type === "ImageStreamImage") {
-                isimage = (imageData.namespace || buildConfig.metadata.namespace) + "/" + imageData.name;               
+                isimage = (imageData.namespace || buildConfig.metadata.namespace) + "/" + imageData.name;
               } else {
                 isimage = "";
               }
@@ -414,6 +490,7 @@ angular.module('openshiftConsole')
 
     $scope.save = function() {
       $scope.disableInputs = true;
+      clearIncompatibleBuildHookFields();
       // Update Configuration
       buildStrategy($scope.updatedBuildConfig).forcePull = $scope.options.forcePull;
 
--- a/app/views/directives/_edit-command.html
+++ b/app/views/directives/_edit-command.html
@@ -1,5 +1,5 @@
 <ng-form name="form">
-  <p ng-hide="input.args.length"><em>No command set.</em></p>
+  <p ng-hide="input.args.length"><em>No {{type || 'command'}} set.</em></p>
   <p ng-show="input.args.length" as-sortable ng-model="input.args" class="command-args">
     <span ng-repeat="arg in input.args" as-sortable-item class="form-group">
       <span class="input-group">
@@ -25,15 +25,15 @@
         <span as-sortable-item-handle class="input-group-addon action-button drag-handle">
           <i class="fa fa-bars" aria-hidden="true"></i>
         </span>
-        <a href="" ng-click="removeArg($index)" class="input-group-addon action-button remove-arg" title="Remove Argument">
-          <span class="sr-only">Remove Argument</span>
+        <a href="" ng-click="removeArg($index)" class="input-group-addon action-button remove-arg" title="Remove Item">
+          <span class="sr-only">Remove Item</span>
           <i class="pficon pficon-close" aria-hidden="true"></i>
         </a>
       </span>
     </span>
   </p>
   <div class="form-group">
-    <label class="sr-only" ng-attr-for="{{id}}-add-arg">Add Argument</label>
+    <label class="sr-only" ng-attr-for="{{id}}-add-arg">Add {{type || 'Command'}}</label>
     <!-- Single-line entry -->
     <span ng-show="!multiline" class="input-group">
       <input type="text"
@@ -41,7 +41,7 @@
              name="nextArg"
              ng-attr-id="{{id}}-add-arg"
              on-enter="addArg()"
-             placeholder="Add argument"
+             placeholder="Add {{type || 'command'}}"
              class="form-control"
              autocorrect="off"
              autocapitalize="off"
@@ -64,7 +64,7 @@
                 name="nextArg"
                 rows="10"
                 ng-attr-id="{{id}}-add-arg"
-                placeholder="Add argument"
+                placeholder="Add {{type || 'command'}}"
                 class="form-control"
                 autocorrect="off"
                 autocapitalize="off"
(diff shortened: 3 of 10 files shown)


spadgett:
@MarkDeMaria I spoke to @jwforres, and we decided that it's better to have a ui-select with the valid options:

* Shell Script
* Command
* Arguments to Default Image Entry Point
* Shell Script with Arguments
* Command with Arguments

This prevents users from selecting an invalid combination.

MarkDeMaria:
As discussed, now uses UI-Select with each possible combination. Selecting save will overwrite the postCommit data with what is being shown on the page. For example, if saving on the Script+Args page while Commands has data, Commands will be wiped in favor on Script+Args, but Commands will still be accessible prior to saving is user switches back to the Commands drop-down.

Added bonus: Page will auto-select a dropdown option based on what is saved in Post Commit.

One problem that needs addressing is that the edit-command directive needs to be able to replace the phrase "No Command set" with "No Argument set." See screenshots below.

Drop Down
image

Script + Args
image

Command + Args
image

app/views/edit/build-config.html
```diff
@@ -536,7 +534,99 @@ <h3 class="with-divider">Run Policy
...
+                                  mode: 'yaml',
+                                  theme: 'eclipse',
+                                  onLoad: aceLoaded,
+                                  rendererOptions: {
+                                    fadeFoldWidgets: true,
+                                    showPrintMargin: false
+                                  }
+                                }"
+                                ng-model="buildConfig.spec.postCommit.script"
+                                class="ace-bordered ace-read-only ace-inline dockerfile-mode mar-top-md">
```
spadgett: Same command for command and command with arguments. Let's not repeat the markup unless necessary.

spadgett:
Can you add "Learn more"  link to this topic?

https://docs.openshift.org/latest/dev_guide/builds.html#build-hooks

app/views/edit/build-config.html
```diff
@@ -536,7 +534,99 @@ <h3 class="with-divider">Run Policy
...
+                              </div>
+                            </div>
+                            <div ng-show="buildHookSelection.type === 'Command'">
+                              <h4>Command</h4>
+                              <edit-command
+                                args="buildConfig.spec.postCommit.command"
+                                isRequired="true">
```
spadgett: I'm not sure setting `isRequired="true"` is correct since this is inside an `ng-show` block, not an `ng-if`. If I don't have a command and switch types, can I submit the form?

app/views/edit/build-config.html
```diff
@@ -536,7 +534,99 @@ <h3 class="with-divider">Run Policy
...
+                          </div>
+
+                          <fieldset>
+                            <div ng-show="buildHookSelection.type === 'Shell Script'">
+                              <h4>Script</h4>
+                              <div
+                                ui-ace="{
+                                  mode: 'yaml',
+                                  theme: 'eclipse',
+                                  onLoad: aceLoaded,
```
spadgett: `aceLoaded` is not defined in the controller, remove  `onLoad`

app/views/edit/build-config.html
```diff
@@ -536,7 +534,99 @@ <h3 class="with-divider">Run Policy
...
+                                isRequired="true">
+                              </edit-command>
+                            </div>
+                            <div ng-show="buildHookSelection.type === 'Shell Script with Arguments'">
+                              <h4>Shell Script with Arguments</h4>
+                              <div
+                                ui-ace="{
+                                  mode: 'yaml',
+                                  theme: 'eclipse',
+                                  onLoad: aceLoaded,
```
spadgett: `aceLoaded` is not defined in the controller, remove  `onLoad`

app/views/edit/build-config.html
```diff
@@ -536,7 +534,99 @@ <h3 class="with-divider">Run Policy
...
+                                  rendererOptions: {
+                                    fadeFoldWidgets: true,
+                                    showPrintMargin: false
+                                  }
+                                }"
+                                ng-model="buildConfig.spec.postCommit.script"
```
spadgett: All of these fields should use `updatedBuildConfig` in the model. You shouldn't mutate the original build config.

app/views/edit/build-config.html
```diff
@@ -536,7 +534,99 @@ <h3 class="with-divider">Run Policy
                         <div class="help-block" ng-switch-default>Builds triggered from this Build Configuration will run using the {{updatedBuildConfig.spec.runPolicy | sentenceCase}} policy.</div>
                       </div>
                     </div>
-                  
+
+                    <div class="section" ng-if="view.advancedOptions">
+                      <div ng-if="!(updatedBuildConfig | isJenkinsPipelineStrategy)" class="section">
```
spadgett: Maybe make this div and its parent one div

```html
<div ng-if="view.advancedOptions && !(updatedBuildConfig | isJenkinsPipelineStrategy)" class="section">
```

app/views/edit/build-config.html
spadgett: Yes, it shouldn't be there unless there is a controller `aceLoaded` method. (I don't think there is.)

MarkDeMaria:
@spadgett @jwforres 

Build Hook Checkbox unselected
image

Build Hook Checkbox selected
image

jwforres:
i think sentence case would look / read better in the dropdown "Shell
script with arguments"

MarkDeMaria:
Can do. Do you want sentence-case for all drop-down entries?

app/scripts/constants.js
```diff
@@ -51,6 +51,7 @@ window.OPENSHIFT_CONSTANTS = {
     "builds":                  "architecture/core_concepts/builds_and_image_streams.html#builds",
     "image-streams":           "architecture/core_concepts/builds_and_image_streams.html#image-streams",
     "storage":                 "architecture/additional_concepts/storage.html",
+    "build-hooks":             "https://docs.openshift.org/latest/dev_guide/builds.html#build-hooks",
```
spadgett: Make it a relative URL.

```js
"dev_guide/builds.html#build-hooks"
```

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -117,6 +118,76 @@ angular.module('openshiftConsole')
...
+    }, {
+      "id": "args",
+      "label": "Arguments to Default Image Entry Point"
+    }, {
+      "id": "scriptArgs",
+      "label": "Shell Script with Arguments"
+    }, {
+      "id": "commandArgs",
+      "label": "Command with Arguments"
+    }];
```
spadgett: Per @jwforres let's make these all sentence case

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -117,6 +118,76 @@ angular.module('openshiftConsole')
...
+
+    $scope.buildHookSelection = {
+      "type": {}
```
spadgett: The quotes are `type` aren't needed.

app/scripts/controllers/edit/buildConfig.js
spadgett: Nit: the quotes around `id` and `label` aren't needed.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -117,6 +118,76 @@ angular.module('openshiftConsole')
...
+    };
+
+    var getInitialBuildHookSelection = function() {
+      $scope.view.hasHooks = (_.size($scope.buildConfig.spec.postCommit.args) ||
+                              _.size($scope.buildConfig.spec.postCommit.command) ||
+                              $scope.buildConfig.spec.postCommit.script) ? true : false;
```
spadgett: `postCommit` is not guaranteed to be in the JSON, so need to handle that otherwise you could have a runtime error. Using `_.get` solves that.

```js
var hasArgs = !_.isEmpty(_.get($scope, 'buildConfig.spec.postCommit.args'));
var hasCommand = !_.isEmpty(_.get($scope, 'buildConfig.spec.postCommit.command'));
var hasScript = !!_.get($scope, 'buildConfig.spec.postCommit.command');
$scope.view.hasHooks = hasArgs || hasCommand || hasScript;
```

Then you can use the local vars below without checking size again.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -117,6 +118,76 @@ angular.module('openshiftConsole')
...
+        $scope.buildHookSelection.type.id = "scriptArgs";
+      } else if (_.size($scope.buildConfig.spec.postCommit.args)) {
+        $scope.buildHookSelection.type.id = "args";
+      } else if ($scope.buildConfig.spec.postCommit.script) {
+        $scope.buildHookSelection.type.id = "script";
+      } else {
+        $scope.buildHookSelection.type.id = "command";
+      }
+
+      $scope.buildHookSelection.type.label = _.find($scope.buildHookTypes, {'id': $scope.buildHookSelection.type.id}).label;
```
spadgett: Shouldn't need to set this. You can use `{{$select.selected.label}}` in the `ui-select` markup.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -117,6 +118,76 @@ angular.module('openshiftConsole')
...
+    };
+
+    var clearIncompatibleBuildHookFields = function(selectedBuildHookID) {
```
spadgett: I'd delete the incompatible fields instead of setting them to empty values.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -414,6 +487,7 @@ angular.module('openshiftConsole')
 
     $scope.save = function() {
       $scope.disableInputs = true;
+      clearIncompatibleBuildHookFields($scope.buildHookSelection.type.id);
```
spadgett: I'd check `$scope.buildHookSelection.type.id` directly in the `clearIncompatibleBuildHookFields` instead of passing it in.

app/views/edit/build-config.html
```diff
@@ -528,8 +524,72 @@ <h3 class="with-divider">Run Policy
                       </div>
                     </div>
 
-                    <div>
-                      <a href="" ng-click="view.advancedOptions = !view.advancedOptions" role="button">{{view.advancedOptions ? 'Hide' : 'Show'}} Advanced Options</a>
+                    <div ng-if="view.advancedOptions && !(updatedBuildConfig | isJenkinsPipelineStrategy)" class="section">
+                      <h3 class="with-divider">Build Hooks
+                        <span class="help action-inline">
+                          <a href="{{'build-hooks' | helpLink}}" aria-hidden="true" target="_blank"><span class="learn-more-inline">Learn more&nbsp;<i class="fa fa-external-link"></i></span></a>
```
spadgett: We've changed the "Learn More" links to use an uppercase `M`.

app/views/edit/build-config.html
```diff
@@ -528,8 +524,72 @@ <h3 class="with-divider">Run Policy
...
+                            </ui-select-choices>
+                          </ui-select>
+                        </div>
+
+                        <fieldset>
+                          <div ng-show="buildHookSelection.type.id === 'script' || buildHookSelection.type.id === 'scriptArgs'">
+                            <h4>Script</h4>
+                            <div
+                              ui-ace="{
+                                mode: 'sh',
```
spadgett: Need to add `mode-sh.js` to `ace-builds` in bower.json.

spadgett:
Need to update `edit-command` to customize the "No command" message and help text for this context.

MarkDeMaria:
@spadgett @jwforres All changes made with the exception of the edit-command directive alterations. Do you want that as a separate PR? And do you have any other categories (Command, Arg, Script, etc) you'd like supported for that directive?

spadgett:
I think it's small enough to add this PR. I'd make the label an optional attribute I can pass into the directive.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -117,6 +118,75 @@ angular.module('openshiftConsole')
...
+        $scope.buildHookSelection.type = _.find($scope.buildHookTypes, { id: 'commandArgs' });
+      } else if (hasArgs && hasScript) {
+        $scope.buildHookSelection.type = _.find($scope.buildHookTypes, { id: 'scriptArgs' });
+      } else if (hasArgs) {
+        $scope.buildHookSelection.type = _.find($scope.buildHookTypes, { id: 'args' });
+      } else if (hasScript) {
+        $scope.buildHookSelection.type = _.find($scope.buildHookTypes, { id: 'script' });
+      } else {
+        $scope.buildHookSelection.type = _.find($scope.buildHookTypes, { id: 'command' });
+      }
```
spadgett: Nit: I'd assign the id to a var and the only do the _.find in one place. Then avoids repeating the same code.

```js
var id;
if (hasArgs && hasCommand) {
  id = 'commandArgs';
} else if ...

$scope.buildHookSelection.type = _.find($scope.buildHookTypes, { id: id });

```

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -117,6 +118,75 @@ angular.module('openshiftConsole')
...
+            break;
+          case "commandArgs":
+            delete $scope.updatedBuildConfig.spec.postCommit.script;
+            break;
+        }
+      } else {
+        $scope.updatedBuildConfig.spec.postCommit.command = [];
+        $scope.updatedBuildConfig.spec.postCommit.args = [];
+        $scope.updatedBuildConfig.spec.postCommit.script = "";
+      }
```
spadgett: What you have here is fine, although I might handle the simple case first and return early. It makes the code a little easier to understand because it avoids a big `if` block.

```js
var clearIncompatibleBuildHookFields = function() {
  if (!$scope.view.hasHooks) {
    $scope.updatedBuildConfig.spec.postCommit.command = [];
    $scope.updatedBuildConfig.spec.postCommit.args = [];
    $scope.updatedBuildConfig.spec.postCommit.script = "";
    return;
  }
  
  switch ($scope.buildHookSelection.type.id) {
    // ...
  }
};
```

Probably should delete the keys for `!hasHooks` case as well instead of setting them to empty values.

MarkDeMaria:
@spadgett @jwforres Changes to edit-command directive made and bad dist file removed.

app/views/directives/_edit-command.html
```diff
@@ -25,23 +25,23 @@
...
+    <label class="sr-only" ng-attr-for="{{id}}-add-arg">Add {{type || 'Command'}}</label>
     <!-- Single-line entry -->
     <span ng-show="!multiline" class="input-group">
       <input type="text"
              ng-model="nextArg"
              name="nextArg"
              ng-attr-id="{{id}}-add-arg"
              on-enter="addArg()"
-             placeholder="Add argument"
+             placeholder="Add {{type || 'command'}}"
```
spadgett: Changing the default will affect other places where this is used, where it should be "Add argument"

app/views/directives/_edit-command.html
```diff
@@ -81,14 +81,19 @@
...
   </div>
   <div class="help-block">
-    Enter the command to run inside the container. The command is considered successful if its exit code is 0.
-    Drag and drop command arguments to reorder them.
+    <div ng-if="!type">
+      Enter the command to run inside the container. The command is considered successful if its exit code is 0.
+      Drag and drop command arguments to reorder them.
+    </div>
+    <div ng-if="type">
+      Enter the arguments that will be appended to the container's command. Drag and drop arguments to reorder them.
```
spadgett: This text is not necessarily always correct when the type is set. We should add a description scope value if we need to say something specific like this.

spadgett:
I'm going to go ahead and merge so that we make DCUT. The remaining comments are fairly minor, and we can fix them in a follow up PR. @jwforres FYI

[merge]

PR #682, opened 2016-10-17
[WIP] Add eslint, replace jshint

benjaminapetersen:
I was looking at this over the weekend.  It resurrects style guide ideas similar to [my original PR](https://github.com/openshift/origin/pull/3614) last year.  `Eslint` is `jshint` + a style guide enforcer like `jscs`.  It can replace `jshint` & do the work of `jscs` (`jscs` has been deprecated & team went to work on `eslint` to converge on one tool). 

I ran the code analysis tool that generates a config against current code (which gives  a config with most failing rules turned off so we can opt in to new rules as we choose).

An existing example using eslint is `angular.js` itself:
- Angular has `eslint` configs https://github.com/angular/angular.js
- Grunt task for `eslint` https://github.com/angular/angular.js/blob/master/Gruntfile.js#L118
- The `.editorconfig` file offers some minimal support

Benefits:
- Code to a style guide will improve our consistentency
- Code to a style guide may help us concentrate on improving logic and fixing bugs in code reviews

@jwforres @spadgett fire away ;)


jwforres:
lets table this discussion until at least next week

benjaminapetersen:
Yup, this kind of thing takes forever to land, I figured it would be a month of pondering if it survives.

docs/style-guides/js.md
```diff
@@ -0,0 +1,252 @@
...
+  // returns true if the user is logged in
+  isLoggedIn: function() {
+
+  }
+}
+
+```
+Avoid excessive comments as a code smell.  Favor meaningful variable names.  This probably doesn't need an example.
+
+Claim your todos and fixmes.  `FIXME (name):` to point out issues, `TODO (name):`
```
spadgett: > Claim your todos and fixmes

Why? I feel like the TODOs should be descriptive enough so it's not necessary. There's always git blame if you need to know who wrote something. I'd argue it's better to not have owners for todos.

docs/style-guides/js.md
```diff
@@ -0,0 +1,252 @@
...
+// bad
+var foo = function() {
+  // FIXME: it doesn't do anything
+  // TODO: it should do something.
+}
+
+// good
+var foo = function() {
+  // FIXME (bpeterse): it doesn't do anything
+  // TODO (bpeterse): it should do something
```
spadgett: Suggest an example of a descriptive TODO comment here

docs/style-guides/js.md
```diff
@@ -0,0 +1,252 @@
...
+}
+
+```
+
+
+### Whitespace
+
+Use 2 spaces for a tab
```
spadgett: Use spaces, not tabs. Each block is indented 2 spaces.

docs/style-guides/js.md
```diff
@@ -0,0 +1,252 @@
...
+## ESLint
+
+Run `grunt eslint` to check code for errors and style discrepancies.
+
+## Examples
+
+
+### Variables
+
+Group your variable names but use individual `var` statements:
```
spadgett: I don't have a strong opinion on this, but a lot of our current code doesn't do it. Unless we feel strongly about something, I think we should take the current style we're using as default. I will say it's something useful to declare the variable near where it's used. Would this apply to giant controller functions, too?

docs/style-guides/js.md
```diff
@@ -0,0 +1,252 @@
...
+
+
+### Comparison Operators
+
+Use `===` and `!==`.  Avoid `==` and `!=`.  This will avoid type coercion bugs:
+
+```javascript
+
+// bad
+if(foo == bar) {
```
spadgett: I personally prefer spaces after the `if`, and it's what most of the code uses today.

docs/style-guides/js.md
```diff
@@ -0,0 +1,252 @@
...
+
+```
+
+
+Don't comment obvious things:
+
+```javascript
+// bad
+var user = {
+  // returns true if the user is logged in
```
spadgett: Suggest a better example here.

Gruntfile.js
```diff
@@ -166,6 +166,14 @@ module.exports = function (grunt) {
       }
     },
 
+    // ESLint is a replacement for JSHint + includes style linting + is extensible
+    eslint: {
+        target: [
+          'Gruntfile.js',
+          '<%= yeoman.app %>/scripts/{,*/}*.js'
+        ]
+    },
```
spadgett: Will this cause build failures or only warnings if eslint fails?

rhamilto:
How do y'all feel about [PatternFly's HTML/CSS standards](http://codeguide.patternfly.org/)?

spadgett:
@rhamilto Yeah, I think they're good, and it'd save us from writing our own from scratch (and debating them forever).

We might want to add some things specific to Angular markup.

docs/style-guides/js.md
jwforres: I actually disagree with this one on multiple fronts. If you are just declaring a bunch of vars and not assigning them, I prefer a single var

```
var foo, bar, baz;
```
Variables that are assigned values should get their own var and their own line.

The second point I disagree with is grouping them. I think it really depends on the context.

docs/style-guides/js.md
jwforres: I agree with sam here, any TODOs that get left in after code review have already been reviewed to make sure its clear what is needed from the comment

docs/style-guides/js.md
benjaminapetersen: Heh, I think right now we really suffer from having no structure around this.  I'd prob like to suggest we pick something going forward & update old code to match when possible.

For the sake of furthering discussion:

@spadget  I think the "declare a variable near where it is used" might mean that entire block could be pulled out into a separate function.  Big functions seem to get harder and harder to maintain, so I definitely prefer asserting rules that help us work towards smaller fns, even if we cant perfectly adhere to it.

@jwforres I actually prefer that as well & did it in my early PRs here, but I think I was asked to turn my vars into individual var statements so I've done this ever since. :smile:

docs/style-guides/js.md
benjaminapetersen: I'd prob want us to be consistent with `if`, `for`, `while`, `switch`, etc with which every style we use.

docs/style-guides/js.md
benjaminapetersen: No strong opinions here, we could drop that.

jwforres:
agreed

docs/style-guides/js.md
spadgett: I suggest we go with @jwforres suggestion. Unless anyone disagrees, I'd add not to declare vars inside an `if` block since they're function scoped anyway.

benjaminapetersen:
Yeah, I don't care too much if we use these or throw them all out in favor of something like [PatternFly's HTML/CSS standards](http://codeguide.patternfly.org/).  I started with a fairly short list of things since we are currently not enforcing anything in particular & wanted to stir the conversation.  If we land on PF's html & CSS, I'd be perfectly fine with also adopting a JS style guide from someone like [AirB&B](https://github.com/airbnb/javascript), [idiomatic.js](https://github.com/rwaldron/idiomatic.js/), [jQuery](https://contribute.jquery.org/style-guide/js/), etc.

benjaminapetersen:
Todd Motto has a decent style guide for [Angular 1](https://github.com/toddmotto/angular-styleguide).  John Papa also published one.  I think they are widely adopted.

benjaminapetersen:
If nothing else we can fork and tweak. I think the only thing that they do that drives me crazy is declaring functions after `return` statements (what?!?!?! :confounded:).

docs/style-guides/js.md
```diff
@@ -0,0 +1,252 @@
...
+
+Wrap IIFE's with a set of parens to help show it's purpose
+
+```javascript
+// bad
+function() {
+  // do stuff
+}();
+
+// good
```
jawnsy: :+1: I think this might also improve performance: https://github.com/nolanlawson/optimize-js#why

benjaminapetersen: Nice, always good to get that!
Though my "bad" example is actually broken, should be `+function() {}();` or something to that effect.

Gruntfile.js
benjaminapetersen: Probably we want it to fail right?  Best way to enforce following the style.

spadgett: Just wondering how much existing code needs changes to get eslint 100% passing.

benjaminapetersen: I'd hope we could get a config that would only enforce new rules on newly changed files...

spadgett: I worry about having to fix a bunch of style warnings when I need to change one line to fix a bug. Particularly when back-porting fixes where you want to minimize changes. I'd prefer to either:

1. Bite the bullet and update all scripts upfront, then make eslint required.
2. Make eslint failures warnings and fix things where it makes sense as we're in the code.

Probably (2) if there are a lot of existing problems.

jwforres:
agreed

jawnsy:
Can you add something in here about formatting for `angular.module` declarations? For example:

[directives/buildPipeline.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/buildPipeline.js#L3
):

```js
  .directive('buildPipeline', function($filter, Logger) {
```

`angular.module` on same line?

[directives/actionChip.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/actionChip.js#L21):

```js
angular
  .module('openshiftConsole')
  .directive('actionChip', function() {
```

`.module` (or other method calls) on their own line, indented, below `angular`

[directives/deploymentMetrics.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/deploymentMetrics.js#L3
):

```js
  .directive('deploymentMetrics', function($interval,
                                           $parse,
                                           $timeout,
                                           $q,
                                           $rootScope,
                                           ChartsService,
                                           ConversionService,
                                           MetricsCharts,
                                           MetricsService,
                                           ModalsService) {
```

imports aligned to function call, to the right of directive declaration.

[directives/labels.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/labels.js#L3
):

```js
  .directive('labels', function($location, $timeout, LabelFilter) {
```

imports directly inline, not wrapped.

[controllers/attachPVC.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/attachPVC.js#L12):

```js
  .controller('AttachPVCController',
              function($filter,
                       $routeParams,
                       $scope,
                       $window,
                       APIService,
                       AuthorizationService,
                       BreadcrumbsService,
                       DataService,
                       Navigate,
                       ProjectsService,
                       StorageService) {
```

imports below the controller name, function declaration aligned to controller name

benjaminapetersen:
@jawnsy yeah, there will be a LOT of things I'm sure that follow a variety of styles.  I'd prefer this initial PR to land on a good foundation block, and we incrementally add to it over time.  Otherwise I don't think we will every land anything 😄

jawnsy:
@benjaminapetersen Fair point, I've opened #1121 to track this :ok_hand:

benjaminapetersen:
Cool cool

spadgett:
Suggest linking to the eslint rules for recommended styles that have them (like airbnb)

benjaminapetersen:
I think we probably would have to pick one (like airbnb), agree we like it, but then start a fresh rules file from scratch & gradually build up to it.  I'm guessing that is the only way we could reasonably adopt a format over time (and even then, hopefully it lets us do that as we update files rather than complain about every file....)

@jawnsy still thinking about your last comment.  Perhaps having a style guide that is ahead of our eslinter isn't a bad thing, so maybe we can include those changes.

Gruntfile.js
benjaminapetersen: Yeah I'm not sure there either.  Surely others have adopted a style guide on an already mature product, I'll do some research.

benjaminapetersen: @spadgett After some digging, I found that this is actually something you can configure in the `.eslint.yml` file, for example:

```yml
eqeqeq: error
no-alert: error
no-console: 'warn'
no-tabs: 'off'
other-pretend-rule:
  - 'can-support-list'
  - 'more-than-one-way-approved'
```

Running it:

```bash
origin-web-console/app/scripts/services/data.js
   100:64  warn  'current' is defined but never used         no-unused-vars
   167:59  warn  'statusText' is defined but never used      no-unused-vars
   201:59  warn  'statusText' is defined but never used      no-unused-vars
   270:59  warn  'statusText' is defined but never used      no-unused-vars
   370:9   warn  'force' is assigned a value but never used  no-unused-vars
   392:61  warn  'statusText' is defined but never used      no-unused-vars
   932:61  warn  'statusText' is defined but never used      no-unused-vars
   958:61  warn  'statusText' is defined but never used      no-unused-vars
   982:7   error  Unexpected console statement                no-console
  1068:76  warn  'event' is defined but never used           no-unused-vars
  1079:26  error  Expected '===' and instead saw '=='         eqeqeq
```

This could be nice if we want to initially flip-on a rule to "warn", and then eventually change it to "error" once we have let it settle a bit.

benjaminapetersen:
Links to examples of style guides, rules, etc:

Style Guides:
[airbnb](https://github.com/airbnb/javascript)
[jQuery](https://contribute.jquery.org/style-guide/)
[idiomatic.js](https://github.com/rwaldron/idiomatic.js)

Eslint Rules:
[airbnb base JS](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base/rules)
[airbnb extended JS - frameworks like react](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb/rules)
[jQuery](https://github.com/jquery/eslint-config-jquery)
[idiomatic.js](https://github.com/jamespamplin/eslint-config-idiomatic)

benjaminapetersen:
An alternative, I added the package `pre-commit` & a `script` to our `package.json`, which will only run `eslint` before a commit, and only on files changed, resulting in this:

screen shot 2017-02-06 at 11 36 41 am

I'm partial to this approach as it lets us opt in over time & only on files we touch, with the option of skipping `eslint` all together if we do the `git commit -n` flag (perhaps minor change on large file).

benjaminapetersen:
@sg00dwin shared a couple more interesting resources:
- [StandardJS style guide](http://standardjs.com/)
- [Prettier JS formatter](https://github.com/jlongster/prettier)

I'm intrigued by the formatter (per everyone talking about Go's in a generally favorable way).

jwforres:
Prettier sounds interesting, wasn't clear from the README whether it
supports ES6+. Don't want to add something to the stack if we may have to
rip it out later.

benjaminapetersen:
From what I read, it will support ES2017 (es7 right?), JSX, and Flow. Seems forward thinking + supporting offshoots like JSX.

benjaminapetersen:
More info via the [blog post](http://jlongster.com/A-Prettier-Formatter).

PR opened with an example, #1251.  I'm pretty favorable to the output.  Couple things i could nit, but I think it would be far more productive than trying to peacemeal our own styleguide (and avoid endless refactoring).

jwforres:
You are correct, I skimmed right over it :)

benjaminapetersen:
A downside could be that it is fairly new & "oh crap" bugs could surface.  I think we could mitigate that depending on how we implement tho. It could be either a grunt task that auto runs, or one that we run manually, or a pre-commit hook that we run w/warns but don't force.  

I'd prefer if we force it until we found a deal breaker though.

spadgett:
We've moved to eslint in the openshift/console repo.

/close

PR #684, opened 2016-10-17
Revert the headless testing changes

jwforres:


jwforres:
[test]

jwforres:
@stevekuznetsov this is ready to go if it ends up being needed
@benjaminapetersen fyi

stevekuznetsov:
I'm not sure the issue is with Karma. From `npm list` it looks like you have Selenium WebDriver 2.44, from [this page](https://www.joecolantonio.com/2014/10/29/selenium-webdriver-version-2-44-0-released/) it shows that version supporting Firefox 24, 31, 32 and 33. The CentOS/RHEL repositories don't have Firefox older then 38. We can't install a much newer version of Karma, anyway, since your `grunt-karma` integration can't support Karma v0.13 or higher. Karma is at v1.3 right now and you're on v0.12. 

Unclear what the way forward is. On a Fedora 24 system with Firefox 49.0 this seems to work fine. Even updating `npm` on the Fedora system to 3.10.3, without changing anything else, things work fine.

jwforres:
OK that's strange if it's working fine with the updated npm on Fedora.

spadgett:
After looking at this with @stevekuznetsov this morning, I don't think the problem is related to Selenium. The tests don't get that far.

For some reason Firefox won't start in Jenkins. We're looking at bumping the log level to get more detail.

jwforres:
We had the same exact symptom with it trying to start phantomjs though. If
you go back and look at the base-ami right after Dan took out the logic
that was attempting to downgrade npm.

stevekuznetsov:
The symptoms were similar but not exactly the same. On a CentOS VM that I provision with `npm 3.10.3` I am having no issues running the unit tests with Firefox. I am not trying on the RHEL systems that we provision with `vagrant-openshift` to determine what the issue is.

stevekuznetsov:
The problem ended up being that there were a couple async changes that we did not do correctly -- as we changed `grunt test` to use Firefox, we needed to also then update it to test headlessly. We updated the Makefile to correctly run the tests with `make test` but failed to realize that `vagrant-openshift` was doing an explicit `grunt test`. The update to the `vagrant-openshift` plugin merged and I anticipate seeing the `base_ami` job succeed now. 

@jwforres you had mentioned that you would allocate a developer to making a `vagrant-openshift` command that would simply run the `make` targets specified, instead of supporting things like `vagrant test-origin-console`. I agree that this is a good idea and once this occurs these issues will be easier to deal with.

stevekuznetsov:
`base_ami` passed the testing step as expected: https://ci.openshift.redhat.com/jenkins/job/base_ami/2439/console

jwforres:
Awesome, good to hear

benjaminapetersen:
Nice :sunglasses:

Issue #695, opened 2016-10-20
Travis tests always fail


spadgett:
https://travis-ci.org/openshift/origin-web-console/builds/169301439

```
Warning: Task "grunt" not found. Use --force to continue.
Aborted due to warnings.
```

Recent changes: #671, #686

/cc @jwforres

spadgett:
Fixed by #699

PR #699, opened 2016-10-21
Use Make targets in Travis script where appropriate

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

@jwforres @spaddgett any reason not to do the integration tests in Travis as well?

### DIFF
--- a/hack/install-deps.sh
+++ b/hack/install-deps.sh
@@ -27,14 +27,19 @@ function cmd() {
   exit $rc
 }
 
+# We don't need grunt and bower to be installed globally for the system,
+# so we can amend our path to look into the local node_modules for the
+# correct binaries.
+export PATH="${PATH}:$(pwd)/node_modules/bower/bin:$(pwd)/node_modules/grunt-cli/bin"
+
 # Install bower if needed
 if ! which bower > /dev/null 2>&1 ; then
-  cmd "sudo npm install -g bower"
+  cmd "npm install bower"
 fi
 
 # Install grunt if needed
 if ! which grunt > /dev/null 2>&1 ; then
-  cmd "sudo npm install -g grunt-cli"
+  cmd "npm install grunt-cli"
 fi
 
 cmd "npm install --unsafe-perm"
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,7 @@ cache:
   - node_modules
   - bower_components
 before_script:
-  - npm install -g grunt-cli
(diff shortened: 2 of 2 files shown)


stevekuznetsov:
I anticipate this fixing #695

spadgett:
```
[ERROR] Command 'sudo npm install -g bower' failed with rc 1, logs:
sudo: must be setuid root
make: *** [install] Error 1
```

jwforres:
i'm ok with this, we still tell people in the README to globally install them for development use, install-deps adding them was purely for CI environments

thanks @stevekuznetsov 

[merge]

hack/install-deps.sh
```diff
@@ -27,14 +27,19 @@ function cmd() {
   exit $rc
 }
 
+# We don't need grunt and bower to be installed globally for the system,
+# so we can amend our path to look into the local node_modules for the
+# correct binaries.
+export PATH="${PATH}:$(pwd)/node_modules/bower/bin:$(pwd)/node_modules/grunt-cli/bin"
```
jawnsy: I think this only works correctly if you invoke the script as: `hack/install-deps.sh` from the `origin-web-console` dir. It'd be preferable to detect the path to the currently-executing script and use the relative location `../node_modules`

Also, it's safer to move the export statements only within the `if` blocks (if the command isn't found, try to install it, then update the `PATH` so that it's available)

hack/install-deps.sh
```diff
@@ -27,14 +27,19 @@ function cmd() {
...
+
 # Install bower if needed
 if ! which bower > /dev/null 2>&1 ; then
```
stevekuznetsov: Will address in a separate PR.

hack/install-deps.sh
stevekuznetsov: > I think this only works correctly if you invoke the script as

Yep, thought this repo had the `cd` to the root dir as we did in Origin. Should fix that.

> Also, it's safer to move the export statements only within the if blocks (if the command isn't found, try to install it, then update the PATH so that it's available)

That would get the logic backwards -- those locations need to be in the `$PATH` for us to search for them in the first place

jwforres:
We would need a running api server to do the int tests there, didn't see a
need to slow down our Travis stuff to pull down and run the api server.
For us Travis is our quick validation that our checked in built dist is
correct and the spec tests are enough to validate we didn't completely
break the loading of our main angular module

PR #703, opened 2016-10-22
Pass opts to _startListOp from DataService.list

spadgett:
Allows `{ errorNotification: false }` to work for list.

@liggitt PTAL. I don't see any other `_startListOp` calls where this is a problem.


spadgett:
cc @jwforres

liggitt:
LGTM [merge]

PR #704, opened 2016-10-24
Bug 1387999 - Deployment strategy type should not be updated before saving strategy changes

jhadvig:
@spadgett PTAL

### DIFF
--- a/app/views/edit/deployment-config.html
+++ b/app/views/edit/deployment-config.html
@@ -15,7 +15,6 @@
             <div ng-if="loaded">
               <h1>
                 Edit Deployment Config {{deploymentConfig.metadata.name}}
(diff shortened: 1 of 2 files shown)


spadgett:
Should we just remove the strategy from the header? Since it's the first dropdown in the form, it's clear what the strategy is. If we don't show it twice, there's no confusion about when changes persist.

jhadvig:
@spadgett yeah,  wanted to be consistent with the header we have in BC editor. But its true that in BC editor that the only place we show the strategy is the header. 
WIll update

jhadvig:
@spadgett update :) PTAL

spadgett:
thanks @jhadvig 

[merge]

PR #707, opened 2016-10-24
Let c3.js pick y-axis values

spadgett:
c3.js has no API to unset an axis max value once set. Passing undefined
or null does not reset the value from when the chart was initialized.
Avoid this problem by letting c3.js pick the y-axis values in all cases.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1386708
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1386823

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/podMetrics.js
+++ b/app/scripts/directives/podMetrics.js
@@ -44,8 +44,6 @@ angular.module('openshiftConsole')
             chartPrefix: "memory-",
             convert: ConversionService.bytesToMiB,
             containerMetric: true,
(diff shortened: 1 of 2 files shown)


jwforres:
thats a shame, so this means we end up with spiky graphs for small fractional cpu usage again

spadgett:
It's less of a problem now that we don't round the values as aggressively before charting them.

spadgett:
And now that we use splines.

jwforres:
k, [merge]

PR #712, opened 2016-10-24
Update filter to immitate future annotation behavior

benjaminapetersen:
@jwforres @spadgett 

Re: https://github.com/openshift/origin/pull/11328
Fixes issue #712 

Since the above annotations didn't make DCUT, this PR updates the temp filter we are using to match the 8 roles agreed upon in discussion.

screen shot 2016-10-24 at 12 18 20 pm


jwforres:
[merge]

PR #715, opened 2016-10-24
Use pod UID instead of name to get pod metrics

spadgett:
Avoids a problem where a new pod with the same name will show metrics for a previously-deleted pod.

https://bugzilla.redhat.com/show_bug.cgi?id=1386838

(The problem with deployment name mentioned in the bug was previously fixed by #541.)

@jwforres PTAL


jwforres:
[merge]

PR #717, opened 2016-10-24
Use a monospace font for template messages

spadgett:
This preserves plain text formatting in the message. Also use alert styles so that the message text does not wrap below the icon.

openshift_web_console

@jwforres PTAL
@luciddreamz CC


jwforres:
[merge]

PR #718, opened 2016-10-24
Fix bug 1388043

benjaminapetersen:
@jwforres @spadgett 

Fix bugzilla #1388043 [link](https://bugzilla.redhat.com/show_bug.cgi?id=1388043) and #1388017 [link](https://bugzilla.redhat.com/show_bug.cgi?id=1388017)
Fixes overly aggressive escaping to ensure error messages are readable.  Example:

screen shot 2016-10-24 at 5 08 44 pm

### DIFF
--- a/app/scripts/controllers/membership.js
+++ b/app/scripts/controllers/membership.js
@@ -26,8 +26,11 @@ angular
       var allRoles = [];
 
       // NOTE: these could all be moved out into a strings service.
-      var messages = {
-        errorReason: _.template('Reason: "<%- httpErr %>"'),
+      var unescaped = {
+        errorReason: _.template('Reason: "<%= httpErr %>"')
+      };
+
+      var escaped = {
         notice: {
           yourLastRole: _.template('Removing the role "<%- roleName %>" may completely remove your ability to see this project.')
         },
@@ -88,16 +91,16 @@ angular
           .then(function() {
             resetForm();
             refreshRoleBindingList();
-            showAlert('rolebindingCreate', 'success', messages.update.subject.success({
+            showAlert('rolebindingCreate', 'success', escaped.update.subject.success({
               roleName: role.metadata.name,
-              subjectName: _.escape(newSubject.name)
+              subjectName: newSubject.name
             }));
           }, function(err) {
             resetForm();
-            showAlert('rolebindingCreateFail', 'error', messages.update.subject.error({
+            showAlert('rolebindingCreateFail', 'error', escaped.update.subject.error({
               roleName: role.metadata.name,
-              subjectName: _.escape(newSubject.name)
-            }), messages.errorReason({httpErr: $filter('getErrorDetails')(err)}));
+              subjectName: newSubject.name
+            }), unescaped.errorReason({httpErr: $filter('getErrorDetails')(err)}));
           });
       };
 
@@ -108,16 +111,16 @@ angular
           .then(function() {
             resetForm();
             refreshRoleBindingList();
-            showAlert('rolebindingUpdate', 'success', messages.update.subject.success({
+            showAlert('rolebindingUpdate', 'success', escaped.update.subject.success({
               roleName: rb.roleRef.name,
-              subjectName: _.escape(newSubject.name)
+              subjectName: newSubject.name
             }));
           }, function(err) {
             resetForm();
-            showAlert('rolebindingUpdateFail', 'error', messages.update.subject.error({
+            showAlert('rolebindingUpdateFail', 'error', escaped.update.subject.error({
               roleName: rb.roleRef.name,
-              subjectName: _.escape(newSubject.name)
-            }), messages.errorReason({httpErr: $filter('getErrorDetails')(err)}));
+              subjectName: newSubject.name
+            }), unescaped.errorReason({httpErr: $filter('getErrorDetails')(err)}));
           });
       };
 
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/membership.js
```diff
@@ -27,7 +27,7 @@ angular
 
       // NOTE: these could all be moved out into a strings service.
       var messages = {
-        errorReason: _.template('Reason: "<%- httpErr %>"'),
+        errorReason: _.template('Reason: "<%= httpErr %>"'),
```
spadgett: We shouldn't keep some plain text and some HTML markup messages in the same `messages` object. It's going to get too confusing to know when the message needs to be escaped. I'd create separate messages objects for alerts / html dialog messages. Or simply inline the messages.

benjaminapetersen: That seems reasonable. I def prefer not going back to inline, I'd rather keep making steps to getting the alert handling in a service, probably the text string templates in another service.

spadgett: At some point, we'll need a message catalog for translation. If we really want to pull the messages out of the controllers anyway, we should start using the format we'll eventually need for translation. But let's not worry about that for this PR.

benjaminapetersen: Here is the difference, for reference:

Escaping error messages from server:
screen shot 2016-10-25 at 9 50 43 am

Not escaping messages from server:
screen shot 2016-10-25 at 9 51 16 am

benjaminapetersen:
@spadgett updated, split into `escaped` and `unescaped` message sets.

app/scripts/controllers/membership.js
```diff
@@ -88,16 +91,16 @@ angular
           .then(function() {
             resetForm();
             refreshRoleBindingList();
-            showAlert('rolebindingCreate', 'success', messages.update.subject.success({
+            showAlert('rolebindingCreate', 'success', escaped.update.subject.success({
               roleName: role.metadata.name,
               subjectName: _.escape(newSubject.name)
```
spadgett: Won't this double escape `subjectName` (here and repeated below)?

spadgett: (You won't see this problem in the UI unless `subjectName` has special characters.)

benjaminapetersen: Ah, true, removing the `_.escape(` on these as redundant.
Though, even just with the template escape sequence `<<<bar` still appears as:

screen shot 2016-10-25 at 12 09 53 pm

spadgett: Right, we shouldn't HTML escape anything going into the alerts. Just messages passed as `detailsMarkup` to the dialog.

benjaminapetersen: Just to make sure I'm with you, alerts still get user input text, I'm just removing the `_.escape()` in favor of using `<%- ` in the templates, separating out the one template that processes server response with `<%=` (I could see this one being inline but would rather be consistent at this point).

spadgett: Any template that will displayed a plain text message in an alert should use `<%=`.

Any template that will display HTML content in a dialog should use `<%-` to escape the HTML. The only time you need to use this if you use `ng-bind-html` like in the dialog.

You shouldn't need to use `_.escape` anywhere if you use `<%-` for the HTML messages.

app/scripts/controllers/membership.js
```diff
@@ -88,16 +91,16 @@ angular
           .then(function() {
             resetForm();
             refreshRoleBindingList();
-            showAlert('rolebindingCreate', 'success', messages.update.subject.success({
+            showAlert('rolebindingCreate', 'success', escaped.update.subject.success({
```
spadgett: This still doesn't seem right. The alerts should not have any escaped messages. Only dialogs.

app/scripts/controllers/membership.js
```diff
@@ -26,8 +26,11 @@ angular
       var allRoles = [];
 
       // NOTE: these could all be moved out into a strings service.
-      var messages = {
-        errorReason: _.template('Reason: "<%- httpErr %>"'),
+      var unescaped = {
+        errorReason: _.template('Reason: "<%= httpErr %>"')
+      };
+
+      var escaped = {
```
spadgett: Maybe `messages` and `messagesWithMarkup`?

benjaminapetersen: Well, since the real issue here seems to be security, the word 'escaped' seems to be more descriptive if the specific problem we are trying to solve.  If I came back in a month I may wonder "why do I care which ones have markup?" and would have to dig. The word  'escaped' immediately tells me this is important for security.

app/scripts/controllers/membership.js
benjaminapetersen: The alerts can still have text from user input, not just the dialogs.

spadgett: > "Bob" was given the role "view". "Bob" needs to be escaped, correct?

No, Angular will escape any special characters you have in the string since the alerts use angular expressions.

https://github.com/openshift/origin-web-console/blob/master/app/views/_alerts.html#L22

We only need to escape values that we pass to `ng-bind-html` in the dialog.

benjaminapetersen: Ok I can update this.  Something about it still bothers me, looking at the docs:

`ngBindHTML`:

> Evaluates the expression and inserts the resulting HTML into the element in a secure way. By default, the resulting HTML content will be sanitized using the $sanitize service. 

`$sanitize`:

> Sanitizes an html string by stripping all potentially dangerous tokens.

spadgett: It sounds like they strip things like `<script>` tags from the content, which is good. But if you have a user name with special characters like `<` and `&`, I suspect it won't render correctly.

benjaminapetersen:
You can see all the places we are using templates in this screenshot:
screen shot 2016-10-31 at 1 36 05 pm
I think the latest including @spadgett's change (not squashed) is where we want this, should be good to go.

spadgett:
[merge]

PR #719, opened 2016-10-25
Bug 1388041 - Clear lifecycle hooks on strategy switch

jhadvig:
@spadgett PTAL
fyi had to remove the `inputChanged` from the `editCommand.js` since it wasn't detecting change of the values(when switching strategies that have `execNewPod` with a command set)

### DIFF
--- a/app/scripts/directives/lifecycleHook.js
+++ b/app/scripts/directives/lifecycleHook.js
@@ -17,7 +17,6 @@ angular.module("openshiftConsole")
         $scope.view = {
           isDisabled: false
         };
-        $scope.view.hookExists = !_.isEmpty($scope.hookParams);
 
         $scope.lifecycleHookFailurePolicyTypes = [
           "Abort",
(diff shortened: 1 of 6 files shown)


spadgett:
> had to remove the inputChanged from the editCommand.js since it wasn't detecting change of the values

@jhadvig You can't just remove this guard. It will cause infinite loops in some cases. The watch for `args` sets `input.args` and the watch for `input.args` sets `args`.

If I edit a health check with an existing command after pulling down this change, I can't type more than one character at a time for a command arg.

app/scripts/directives/lifecycleHook.js
```diff
@@ -58,64 +57,72 @@ angular.module("openshiftConsole")
...
           }
+        };
+
+        $scope.$watch("hookParams", function() {
+          paramsChange();
+        });
+
+        $scope.$watch("action.type", function() {
+          paramsChange();
         });
```
spadgett:
```js
$scope.$watchGroup(['hookParams', 'action.type'], paramsChange);
```

app/scripts/directives/lifecycleHook.js
```diff
@@ -58,64 +57,72 @@ angular.module("openshiftConsole")
...
-          if ($scope.action.type === 'execNewPod' && _.has($scope.hookParams, 'tagImages')) {
-            $scope.removedHookParams['tagImages'] = $scope.hookParams.tagImages;
-            delete $scope.hookParams.tagImages;
+        var paramsChange = function() {
+          if (!$scope.hookParams) {
+            return;
+          }
+          if ($scope.action.type === 'execNewPod') {
+            if (_.has($scope.hookParams, 'tagImages')) {
+              $scope.removedHookParams['tagImages'] = $scope.hookParams.tagImages;
```
spadgett:
```js
$scope.removedHookParams.tagImages = $scope.hookParams.tagImages;
```

app/scripts/directives/lifecycleHook.js
```diff
@@ -58,64 +57,72 @@ angular.module("openshiftConsole")
...
-        $scope.$watch("action.type", function() {
-          if ($scope.action.type === 'execNewPod' && _.has($scope.hookParams, 'tagImages')) {
-            $scope.removedHookParams['tagImages'] = $scope.hookParams.tagImages;
-            delete $scope.hookParams.tagImages;
+        var paramsChange = function() {
+          if (!$scope.hookParams) {
+            return;
+          }
+          if ($scope.action.type === 'execNewPod') {
+            if (_.has($scope.hookParams, 'tagImages')) {
```
spadgett: Note that this will evaluate to true if `tagImages` is set to `null`. Should it be this instead?

```js
if ($scope.hookParams.tagImages) { ... }
```

`hookParams` will always be defined because of your check above.

spadgett:
[merge]

PR #721, opened 2016-10-25
Bug 1388337 - Clear lifecycle hooks when changing strategy from Recreate to Rolling

jhadvig:
@spadgett talked to @mfojtik && @kargakis regarding this and we agreed that when  switching between rolling<->recreate strategy it doesnt make sense to preserve the old params.
PTAL

### DIFF
--- a/app/scripts/controllers/edit/deploymentConfig.js
+++ b/app/scripts/controllers/edit/deploymentConfig.js
@@ -183,7 +183,6 @@ angular.module('openshiftConsole')
       modalInstance.result.then(function () {
         // Move parameters that belong to the origial strategy to the picked one.
         $scope.strategyData[pickedStrategyParams] = $scope.strategyData[getParamsPropertyName($scope.originalStrategy)];
-        $scope.paramsMoved = getParamsPropertyName($scope.originalStrategy);
       }, function() {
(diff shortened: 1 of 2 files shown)


spadgett:
@jhadvig I'm unclear on what parameters should be moved when. Can you summarize the different cases?

jhadvig:
@spadgett so user can move parameters when switching between rolling and recreate strategy, since those two strategies have 3 common parameters (timeout, pre lifecycle hook, post lifecycle hook). If user switches from Rolling to Recreate (or vice versa) he will be prompt with dialog if the common params from the original strategy shall be moved to the selected one. If so we will copy them. But for both scenarios(1. moving the params to the selected strategy, 2. not moving params) we shall remove the parameters from the original strategy, since they are not necessary for the DC.
On the other hand if switching from Rolling || Recreate to Custom(or vice versa) it makes sense to keep the original strategy parameters.

spadgett:
OK, so the behavior moving the parameters is the same as before. The change is to delete the parameters from the previous strategy when making a rolling <-> recreate switch.

jhadvig:
@spadgett yes

spadgett:
[merge]

PR #723, opened 2016-10-25
Bug 1388428 - Get SA list even if the SA to which the secret be linked is defined

jhadvig:
Even if the `$scope.serviceAccountToLink` is defined(`builder` for the BC editor, `default` for the DC editor), we shall list the SAs so we can update the appropriate one, or have an alert which will inform that the SA which the secret should be link with is is not available.
The other approach could be to GET the `$scope.serviceAccountToLink` and if the request fails we can inform user user in the 'create secret' form that the linking with the SA wont be possible and only create the secret.
Anyway I think just informing the user that the created secret was not linked with the SA we determined for him(`builder` for the BC editor, `default` for the DC editor) is better since not linking the created secret with SA wont do any magic, unless there is a option enabled in the cluster, that will use those linked secrets automatically, which is disabled by default.
@spadgett PTAL

### DIFF
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -2,7 +2,7 @@
 
 angular.module("openshiftConsole")
 
-  .directive("createSecret", function(DataService, AuthorizationService) {
+  .directive("createSecret", function(DataService, AuthorizationService, $filter) {
     return {
       restrict: 'E',
       scope: {
@@ -13,7 +13,7 @@ angular.module("openshiftConsole")
         cancel: '&'
       },
       templateUrl: 'views/directives/create-secret.html',
-      link: function($scope, $filter) {
+      link: function($scope) {
         $scope.alerts = {};
 
         $scope.secretAuthTypeMap = {
@@ -79,7 +79,7 @@ angular.module("openshiftConsole")
         };
 
         // List SA only if $scope.serviceAccountToLink is not defined so user has to pick one.
-        if (!$scope.serviceAccountToLink && AuthorizationService.canI('serviceaccounts', 'list') && AuthorizationService.canI('serviceaccounts', 'update')) {
+        if (AuthorizationService.canI('serviceaccounts', 'list') && AuthorizationService.canI('serviceaccounts', 'update')) {
           DataService.list("serviceaccounts", $scope, function(result) {
             $scope.serviceAccounts = result.by('metadata.name');
             $scope.serviceAccountsNames = _.keys($scope.serviceAccounts);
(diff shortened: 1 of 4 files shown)


app/scripts/directives/createSecret.js
```diff
@@ -143,6 +143,17 @@ angular.module("openshiftConsole")
         };
 
         var linkSecretToServiceAccount = function(secret) {
+          if (!$scope.serviceAccountsNames.contains($scope.newSecret.pickedServiceAccountToLink)) {
+            var alert = {
+              name: 'createAndLink',
+              data: {
+                type: "success",
+                message: "Secret " + secret.metadata.name + " was created but couldn't be linked with service account " + $scope.newSecret.pickedServiceAccountToLink + "."
```
bparees: can we not break this into two steps (Create, Link) and then do a better job of reporting the result of each aspect?

jhadvig: @spadgett I like the idea silently fail if the user was not prompted(case of the BC/DC editor). If he was(create secret page) we can print success alert that the secret was created but also a warning the could not be linked.

@bparees we will need to have the linking action somewhere in the console eventually, but don't thing we have time to update it now, together with the TC's. Would rather stay with what we have and make it right.

bparees: i'm not suggesting it be two separate UI actions the user can take, i'm just suggesting you treat it as two steps internally and report error w/ finer granularity.

jhadvig: > Seems reasonable. These will be separate alert messages?

yes

> i'm not suggesting it be two separate UI actions the user can take, i'm just suggesting you treat it as two steps internally and report error w/ finer granularity.

yes I understand, but we dont have a good story how to model those steps so they will make sense. Not saying that we could not do it, just that we dont have time :'( But definitely worth for as a follow up ?!

jhadvig:
@spadgett updated... ptal

spadgett:
@jhadvig I'm still seeing symptoms of the bugzilla when linking fails. The secret is created, but you aren't directed back to the previous page and the alert message isn't displayed. You can force this to happen by changing the code to set an invalid property in the SA before the PUT.

jhadvig:
@spadgett updated :) Thanks for  noticing the additional issue :+1:

spadgett:
[merge]

spadgett:
@jhadvig LGTM, but you'll need to run grunt build again to fix the merge conflict.

jhadvig:
@spadgett updated

spadgett:
[merge]

jwforres:
@jhadvig this needs a rebase

jhadvig:
@jwforres @spadgett rebased

spadgett:
[merge]

PR #724, opened 2016-10-25
Truncate imageStreamName like we do for dockerImageReference.

sg00dwin:
- for @screen-*-min widths set a max-width
- max-width determined by minimum space needed, when
screen shot 2016-10-25 at 12 24 42 pm
screen shot 2016-10-25 at 12 24 07 pm
screen shot 2016-10-25 at 12 23 57 pm
screen shot 2016-10-25 at 12 23 49 pm
screen shot 2016-10-25 at 12 23 39 pm

 both imageStreamName and dockerImageReference fields need truncating
Fixes bug 1383623

### DIFF
--- a/app/styles/_mixins.less
+++ b/app/styles/_mixins.less
@@ -103,3 +103,24 @@
     display: inline-block !important;
   }
 }
+
+// Truncate or word-break long strings within tables
+// Wrap when < 768 & truncate when > 769
+.td-long-string {
+  @media (max-width: @screen-xs-max) {
+    .word-break();
+  }
+  @media (min-width: @screen-sm-min) {
+    max-width: 170px;
+    .text-overflow();
+  }
+  @media (min-width: @screen-md-min) {
+    max-width: 260px
+  }
+  @media (min-width: @screen-lg-min) {
+    max-width: 370px
+  }
+  @media (min-width: @screen-xlg-min) {
+    max-width: 540px
+  }
+}
--- a/app/views/browse/imagestream.html
+++ b/app/views/browse/imagestream.html
@@ -72,20 +72,20 @@
                     <td data-title="From">
                       <!-- this value can get long when its an ImageStreamImage. Use max width and class truncate to elide some
                            of the SHA1 for long values. They can still be copied even when elided. -->
-                      <div style="max-width: 400px;" class="truncate">
+
                         <span ng-if="!tag.spec.from"><em>pushed</em></span>
-                        <span ng-if="tag.spec.from">
-                          <span ng-if="!tag.spec.from._imageStreamName">
-                            {{tag.spec.from.name}}
-                          </span>
-                          <span ng-if="tag.spec.from._imageStreamName">
-                            <span ng-if="tag.spec.from._imageStreamName === imageStream.metadata.name">{{tag.spec.from._completeName}}</span>
-                            <span ng-if="tag.spec.from._imageStreamName !== imageStream.metadata.name" >
-                              <a ng-href="{{tag.spec.from._imageStreamName | navigateResourceURL : 'ImageStream' : (tag.spec.from.namespace || imageStream.metadata.namespace)}}"><span ng-if="tag.spec.from.namespace && tag.spec.from.namespace !== imageStream.metadata.namespace" >{{tag.spec.from.namespace}}/</span>{{tag.spec.from._imageStreamName}}</a>{{tag.spec.from._nameConnector}}{{tag.spec.from._idOrTag}}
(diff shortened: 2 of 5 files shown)


sg00dwin:
@jwforres @spadgett PTAL

app/views/browse/imagestream.html
```diff
@@ -72,20 +72,35 @@
...
-                      </div>
+                            <span ng-if="tag.spec.from._imageStreamName">
+                              <span ng-if="tag.spec.from._imageStreamName === imageStream.metadata.name">{{tag.spec.from._completeName}}</span>
+                              <span ng-if="tag.spec.from._imageStreamName !== imageStream.metadata.name" >
+                                <a ng-href="{{tag.spec.from._imageStreamName | navigateResourceURL : 'ImageStream' : (tag.spec.from.namespace || imageStream.metadata.namespace)}}"><span ng-if="tag.spec.from.namespace && tag.spec.from.namespace !== imageStream.metadata.namespace" >{{tag.spec.from.namespace}}/</span>{{tag.spec.from._imageStreamName}}</a>{{tag.spec.from._nameConnector}}{{tag.spec.from._idOrTag}}
+                              </span>
+                            </span>
+                          </div>
+                          <!-- Truncate when > 768 -->
+                          <div class="pull-spec truncate hidden-xs">
```
spadgett: If the markup is identical and only the classes change, can we use mixins and media queries to assign the styles for different screen widths? It's a little more logic than I'd like to repeat. I think this can be done entirely with less.

sg00dwin:
Updated pr to use a mixin that will handle wrapping or truncation based on size.

app/styles/_mixins.less
```diff
@@ -103,3 +103,23 @@
     display: inline-block !important;
   }
 }
+
+// Truncate or word-break long strings within tables
+.td-long-string-handler {
```
spadgett: Maybe just `td-long-string`?

app/views/browse/imagestream.html
```diff
@@ -72,20 +72,21 @@
...
-                        <span ng-if="tag.spec.from">
-                          <span ng-if="!tag.spec.from._imageStreamName">
-                            {{tag.spec.from.name}}
-                          </span>
-                          <span ng-if="tag.spec.from._imageStreamName">
-                            <span ng-if="tag.spec.from._imageStreamName === imageStream.metadata.name">{{tag.spec.from._completeName}}</span>
-                            <span ng-if="tag.spec.from._imageStreamName !== imageStream.metadata.name" >
-                              <a ng-href="{{tag.spec.from._imageStreamName | navigateResourceURL : 'ImageStream' : (tag.spec.from.namespace || imageStream.metadata.namespace)}}"><span ng-if="tag.spec.from.namespace && tag.spec.from.namespace !== imageStream.metadata.namespace" >{{tag.spec.from.namespace}}/</span>{{tag.spec.from._imageStreamName}}</a>{{tag.spec.from._nameConnector}}{{tag.spec.from._idOrTag}}
+                        <!-- Wrap when < 768 & truncate when > 769 -->
+                        <div ng-if="tag.spec.from" class="td-long-string-handler">
```
spadgett:
```html
<div ng-if="tag.spec.from" ng-attr-title="{{tag.spec.from.name}}" class="td-long-string-handler">
```

so you can hover to see the full value

app/views/browse/imagestream.html
```diff
@@ -72,20 +72,21 @@
...
                         <span ng-if="!tag.spec.from"><em>pushed</em></span>
-                        <span ng-if="tag.spec.from">
-                          <span ng-if="!tag.spec.from._imageStreamName">
-                            {{tag.spec.from.name}}
-                          </span>
-                          <span ng-if="tag.spec.from._imageStreamName">
-                            <span ng-if="tag.spec.from._imageStreamName === imageStream.metadata.name">{{tag.spec.from._completeName}}</span>
-                            <span ng-if="tag.spec.from._imageStreamName !== imageStream.metadata.name" >
-                              <a ng-href="{{tag.spec.from._imageStreamName | navigateResourceURL : 'ImageStream' : (tag.spec.from.namespace || imageStream.metadata.namespace)}}"><span ng-if="tag.spec.from.namespace && tag.spec.from.namespace !== imageStream.metadata.namespace" >{{tag.spec.from.namespace}}/</span>{{tag.spec.from._imageStreamName}}</a>{{tag.spec.from._nameConnector}}{{tag.spec.from._idOrTag}}
+                        <!-- Wrap when < 768 & truncate when > 769 -->
```
spadgett: Really this comment should be in the less file not here. It will get out of sync if the style changes.

sg00dwin:
@spadgett pr is updated

spadgett:
[merge]

PR #725, opened 2016-10-25
Open template description links and other `linky` links in a new window

spadgett:
This is consistent with how we handle help links and other external links in the web console.

https://code.angularjs.org/1.3.19/docs/api/ngSanitize/filter/linky

@jwforres PTAL
@luciddreamz FYI


jwforres:
[merge]

luciddreamz:
Makes sense and looks good.

PR #733, opened 2016-10-25
Don't allow scaling of in progress k8s deployments

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1383042

@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/replicaSet.js
+++ b/app/scripts/controllers/replicaSet.js
@@ -53,7 +53,6 @@ angular.module('openshiftConsole')
     $scope.replicaSet = null;
     $scope.deploymentConfig = null;
     $scope.deploymentConfigMissing = false;
-    $scope.deployments = {};
     $scope.imagesByDockerReference = {};
     $scope.builds = {};
     $scope.alerts = {};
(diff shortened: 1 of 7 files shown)


app/scripts/controllers/replicaSet.js
```diff
@@ -280,6 +281,33 @@ angular.module('openshiftConsole')
...
+              resource: 'replicasets'
+            }, context, function(replicaSets) {
+              var deploymentSelector = new LabelSelector($scope.deployment.spec.selector);
+              inProgressDeployment = false;
+
+              // See if there is more than one replica set that matches the
+              // deployment selector with active replicas.
+              var numActive = 0;
+              _.each(replicaSets.by('metadata.name'), function(replicaSet) {
+                if (!deploymentSelector.matches(replicaSet) || !replicaSet.status.replicas) {
```
jwforres: is matches the right check here?  shouldn't it be comparing the labelSelector of the deployment to the labelSelector of the replicaSet?  matches is going to checks the labels on the replicaSet, covers would check the labelSelector

spadgett:
@jwforres thanks, updated

jwforres:
@spadgett were you using matches anywhere else in the deployment/replicaset rollup that we should have been using covers instead?

jwforres:
actually i see you updated another one in your latest commit, thanks  [merge]

PR #735, opened 2016-10-26
Bug 1388446 - Refactor osc-secret and osc-source-secret style

jhadvig:
Fixing bug 1388446 by updating the `osc-secret.html` and `osc-source-secret.html` + styling. They have similar styling that we use in the key-value-editor now.
Wasn't sure if we want to create a common mixin function, since most of the attributes are the same, just width is changed based on the parent classes(`basic-secrets`, `advanced-secrets`).
@spadgett @rhamilto PTAL

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -464,56 +464,6 @@ label.checkbox {
   }
 }
 
-.osc-secrets-form {
-  .basic-secrets {
-    .remove-secret {
-      display: inline-block;
-      margin-left: 7px;
-      vertical-align: 8px;
-    }
-    .secret-name {
-      width: 95%;
-      display: inline-block;
-    }
-  }
-  .advanced-secrets {
-    margin-bottom: 5px;
-    .remove-secret {
-      display: inline-block;
-      margin-left: 7px;
-      vertical-align: 2px;
-    }
-    .directory {
-      width: 90%;
-      display: inline-block;
-    }
-  }
-  @media (max-width: @screen-xs-max) {
-    .basic-secrets {
-      .secret-name {
-        width: 90%;
-        display: inline-block;
-      }
-    }
-    .remove-secret {
-      vertical-align: 2px;
-    }
-  }
-  .remove-btn {
-    color: #333;
-    opacity: 0.65;
-    font-size: 15px;
-    vertical-align: middle;
-    &:hover {
-      opacity: 1;
-      text-decoration: none;
-    }
-    &:focus {
-      text-decoration: none;
-    }
-  }
-}
-
 .lifecycle-hook {
   padding-bottom: 20px;
   &:first-of-type {
@@ -984,30 +934,6 @@ a.disabled-link {
 // Disable numeric highlighting in YAML and Dockerfile modes because of bugs
 .editor.yaml-mode .ace_constant.ace_numeric, .ace_editor.dockerfile-mode .ace_constant.ace_numeric {
   color: inherit;
-}
-
-.create-secret-modal {
-  background-color: #F5F5F5;
-  .modal-footer{
-    margin-top: 0px
-  }
-  .modal-body {
-    padding: 0px 18px;
-  }
-}
-
-.secret-data {
-  max-width: 450px;
-  max-height: 150px;
-  @media (max-width: @screen-sm-max) {
-    // The `table-responsive` div adds a 100% border. Make sure the table is at
-    // least 100% to avoid a white block at some screen sizes.
-    max-width: 100%;
-  }
-}
-
-.create-secret-editor {
-  height: 150px;
 }
 
 .edit-yaml {
--- a/app/views/directives/osc-secrets.html
+++ b/app/views/directives/osc-secrets.html
@@ -1,10 +1,14 @@
 <ng-form name="secretsForm" class="osc-secrets-form">
-  <div ng-repeat="pickedSecret in pickedSecrets">
-    <div class="form-group">
-      <div class="row picked-secret">
-        <div class="col-lg-12">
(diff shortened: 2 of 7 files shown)


jhadvig:
PR rebased.. @spadgett PTAL

cc'ing @benjaminapetersen @rhamilto since you huys have been working on the key-value-editor and the styling updates are inspired by the styling in the editor

spadgett:
@rhamilto do you mind taking a look at this one?

jhadvig:
FYI: The issue was reported for FireFox

rhamilto:
>@rhamilto do you mind taking a look at this one?

Sure.

app/styles/_core.less
```diff
@@ -466,41 +466,77 @@ label.checkbox {
 
 .osc-secrets-form {
   .basic-secrets {
```
rhamilto: You could significantly DRY up this LESS like so:

```LESS
.osc-secrets-form {
  .advanced-secrets,
  .basic-secrets {
    .input-labels,
    .help-blocks {
      display: table;
      padding-right: 30px;
      position: relative;
      width: 100%;
      .input-label,
      .help-block {
        width: 100%;
        float: left;
        padding-right: 5px;
      }
    }
    .secret-row {
      display: table;
      padding-right: 30px;
      position: relative;
      width: 100%;
      padding-bottom: 5px;
      .secret-name {
        float: left;
        padding-right: 5px;
        width: 100%;
      }
      .remove-secret {
        position: absolute;
        right: 0;
        top: 0;
        width: 30px;
      }
    }
  }

  .advanced-secrets {
    .input-labels,
    .help-blocks {
      .input-label,
      .help-block {
        width: 50%;
      }
    }
    .secret-row {
      .secret-name,
      .destination-dir {
        width: 50%;
      }
    }
  }

  .remove-btn {
    color: #333;
    margin-left: 5px;
    opacity: 0.65;
    font-size: 15px;
    vertical-align: middle;
    &:hover {
      opacity: 1;
      text-decoration: none;
    }
    &:focus {
      text-decoration: none;
    }
  }
}
```

Note:  this is untested, so make sure you double check.

app/views/directives/osc-secrets.html
```diff
@@ -1,50 +1,41 @@
...
-              Secret for authentication when pulling images from a secured registry.
-              <a href="{{'pull_secret' | helpLink}}" target="_blank"><span class="learn-more-inline">Learn more&nbsp;<i class="fa fa-external-link" aria-hidden="true"></i></span></a>
-            </div>
-            <div class="help-block" ng-switch-when="push">
-              Secret for authentication when pushing images to a secured registry.
-              <a href="{{'pull_secret' | helpLink}}" target="_blank"><span class="learn-more-inline">Learn more&nbsp;<i class="fa fa-external-link" aria-hidden="true"></i></span></a>
-            </div>
+          <div class="remove-secret">
+            <a ng-click="removeSecret($index)" class="pficon pficon-close remove-btn" aria-hidden="true" href="" role="button"></a>
+            <a class="sr-only">Remove build secret</a>
```
rhamilto: To improve accessibility, this would be written like so:
```HTML
<a ng-click="removeSecret($index)"  href="" role="button" class="remove-btn">
  <span class="pficon pficon-close" aria-hidden="true"></span>
  <span class="sr-only">Remove build secret</span>
</a>
```

This hides the icon from screen readers, but keeps the descriptive text within the same element so the screen reader can properly describe the functionality of the link.

app/views/directives/osc-source-secrets.html
```diff
@@ -1,103 +1,92 @@
...
+                autocorrect="off"
+                autocapitalize="off"
+                spellcheck="false">
             </div>
             <div class="remove-secret">
-              <a ng-click="removeSecret($index)" href="" role="button">
-                <span class="pficon pficon-close remove-btn" aria-hidden="true"></span>
-                <span class="sr-only">Remove build secret</span>
-              </a>
+              <a ng-click="removeSecret($index)" class="pficon pficon-close remove-btn" aria-hidden="true" href="" role="button"></a>
```
rhamilto: Same comment about improving accessibility applies here, too.

app/views/directives/osc-source-secrets.html
```diff
@@ -1,103 +1,92 @@
...
+                type="text"
+                placeholder="/"
+                autocorrect="off"
+                autocapitalize="off"
+                spellcheck="false">
             </div>
             <div class="remove-secret">
-              <label class="sr-only">Remove Build Secret</label>
-              <a class="pficon pficon-close remove-btn" aria-label="Delete row" role="button" ng-click="removeSecret($index)" href=""></a>
+              <a ng-click="removeSecret($index)" class="pficon pficon-close remove-btn" aria-hidden="true" href="" role="button"></a>
```
rhamilto: Same comment about improving accessibility applies here, too.

jhadvig:
@rhamilto thanks for review. Comments addressed

app/styles/_core.less
benjaminapetersen: I'd tend to suggest moving it out of `_core.less` and into `_secrets.less` or something as well.

rhamilto: Agreed.  I debated making that suggestion in my review, but figured it'd previously been reviewed and okayed as is.  Since @benjaminapetersen suggested it, mind moving the secrets LESS to a new partial, @jhadvig?

jhadvig:
@rhamilto @benjaminapetersen  thanks for suggestion to create a standalone file. Definitely makes sense
PTAL

rhamilto:
LGTM.  @spadgett, your turn.

spadgett:
[merge]

PR #736, opened 2016-10-26
Update required field handling for source secrets

### LINKED ISSUE #734: Password field label missing class="required" on create secret page
The "Password or Token" field must be filled in before I can submit the form, but it's missing the required `*` on the label when creating a basic auth source secret.

spadgett:
* Show a required `*` on the password label when required
* Require ca.cert or .gitconfig content when the checkbox for those options is checked
* Change scope var names to add.gitconfig and add.cacert so the updated values can be read outside of the `ng-if` where they're set.
* Use `required` attribute on osc-file-input. `ng-required` doesn't work.

Fixes #734
@jhadvig PTAL

### DIFF
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -72,8 +72,11 @@ angular.module("openshiftConsole")
             pickedServiceAccountToLink: "",
           };
         }
-        $scope.addGitconfig = false;
-        $scope.addCaCert = false;
+
+        $scope.add = {
+          gitconfig: false,
+          cacert: false
+        };
 
         // List SA only if $scope.serviceAccountToLink is not defined so user has to pick one.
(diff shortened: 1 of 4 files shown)


spadgett:
@jwforres

spadgett:
Updated to fix a typo in the help text.

jhadvig:
@spadgett as we have spoken .. I'd rather name the `add.cacert` as `add.cacrt` to be more consistant with the naming we user internally. But would display ca.cert to the user

spadgett:
The file is usually called `ca.crt`, so I would show that to the user.

spadgett:
@jhadvig All the `id` attributes use `cacert` and also the model value `newSecret.data.cacert`. I'd rather not change it since it touches a lot of existing code.

jhadvig:
@spadgett yeah, true that. I wouldn't rather not change that as well

spadgett:
thanks @jhadvig 

[merge]

PR #737, opened 2016-10-26
Allow all scripts in hack/ to use local grunt and bower binaries

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

There are some hidden issues with finding the repo root using `dirname` (as you can see we have a more sophisticated means by which we do it in Origin's `hack/lib/init.sh`) but I do not know how complicated we need to make this ... the `dirname` approach should work in most standard environments.

### DIFF
--- a/hack/clean-deps.sh
+++ b/hack/clean-deps.sh
@@ -2,9 +2,11 @@
 
 set -e
 
-echo "Cleaning up bower_components and node_modules..."
-rm -rf bower_components/*
-rm -rf node_modules/*
+# We don't need bower to be installed globally for the system, so 
+# we can amend our path to look into the local node_modules for the
+# correct binaries.
+repo_root="$( dirname "${BASH_SOURCE}" )/.."
+export PATH="${PATH}:${repo_root}/node_modules/bower/bin"
 
 if which bower > /dev/null 2>&1 ; then
   # In case upstream components change things without incrementing versions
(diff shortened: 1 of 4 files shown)


hack/clean-deps.sh
```diff
@@ -6,10 +6,16 @@ echo "Cleaning up bower_components and node_modules..."
...
+# We don't need bower to be installed globally for the system, so 
+# we can amend our path to look into the local node_modules for the
+# correct binaries.
+repo_root="$( dirname "${BASH_SOURCE}" )/.."
+export PATH="${PATH}:${repo_root}/node_modules/bower/bin"
+
 if which bower > /dev/null 2>&1 ; then
   # In case upstream components change things without incrementing versions
   echo "Clearing bower cache..."
   bower cache clean --allow-root
```
spadgett: won't this fail since we removed node_modules above?

spadgett:
cc @danmcp

stevekuznetsov:
Also if this level of copy-pasta is too much I am happy to make a shared script.

/cc @spadgett @jwforres

hack/clean-deps.sh
stevekuznetsov: Unless you have it globally like I did, d'oh -- I uninstalled it now.

spadgett:
[merge]

PR #738, opened 2016-10-26
Wrap limit ranges section on quota page so it can be easily hidden with custom CSS

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1387440

### DIFF
--- a/app/views/quota.html
+++ b/app/views/quota.html
@@ -190,66 +190,68 @@
                   </div>
                 </div>
 
-                <h1>Limit Range</h1>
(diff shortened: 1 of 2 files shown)


jwforres:
@spadgett PTAL, the diff is primarily a whitespace change

jwforres:
@luciddreamz FYI

spadgett:
[merge]

jwforres:
@luciddreamz in the mean time you could do a hacky text content based css selector to hide that Limit Range header :-\  this fix is going into 3.4

Issue #732, opened 2016-10-25
Don't add app label when template has anything in it with the app label


jwforres:
This is based on discussions / changes that went into `oc new-app` already

PR #742, opened 2016-10-26
Disambiguate repeated app names on overview

### LINKED ISSUE #629: Overview app heading reused for all services in the same template
If a template contains more than one service and the services aren't grouped, each service section has the same app name. (`oc new-app` will add the `app` label to each service, which we display.) This is a little confusing.

The `app` labels are displayed in the web console, but aren't used for grouping, so they can be repeated.

/cc @jwforres @thoraxe

spadgett:
Show the primary service name beside the app name when repeated to
disambiguate it. The repeated app names can happen when deploying a
template with multiple services since the app label is applied to all
services.

Fixes #629

openshift_web_console

@jwforres opinions?

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -324,6 +324,13 @@ angular.module('openshiftConsole')
       $scope.childServicesByParent[parentName].push(child);
     };
 
+    var topLevelServicesByApp = {};
+
+    // Check if an app label is used more than once on a top-level service so we can disambiguate.
+    $scope.isDuplicateApp = function(app) {
+      return _.size(topLevelServicesByApp[app]) > 1;
+    };
+
     var groupServices = function() {
       if (!services || !routes) {
         return;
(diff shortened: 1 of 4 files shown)


jwforres:
seems reasonable, can you cc some of the folks who brought up the issue and get their feedback

spadgett:
> can you cc some of the folks who brought up the issue and get their feedback

cc @thoraxe

spadgett:
Note that this should be less common when we fix #732

spadgett:
Rebased.

jwforres:
@ajacobs21e thoughts on this?

ajacobs21e:
Sorry I'm a little confused. These are two different apps but both have the same app-label? @spadgett @jwforres

spadgett:
@ajacobs21e These are two different services with the same app label. This can happen when applying a template with two services that are not grouped together.

spadgett:
rebased

spadgett:
[merge]

PR #743, opened 2016-10-26
Sort overview services by app name, then service name

### LINKED ISSUE #731: Sort order of service groups on the overview could be improved
We should probably by sorting by app label and then by primary service name

spadgett:
Fixes #731

@jwforres PTAL


jwforres:
[merge]

PR #748, opened 2016-10-27
Bug 1388448 - Show warning when docker config file is not in JSON format

jhadvig:
Adding a warning that the input should be in a JSON format.
Screen:
screenshot-52

The blocking of the creation will be fixed as part of https://github.com/openshift/origin-web-console/pull/723

@spadgett PTAL

### DIFF
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -175,6 +175,19 @@ angular.module("openshiftConsole")
           });
         };
 
+        var updateEditorMode = _.debounce(function(){
+          try {
+            JSON.parse($scope.newSecret.data.dockerConfig);
+            $scope.invalidConfigFormat = false;
+          } catch (e) {
+            $scope.invalidConfigFormat = true;
+          }
+        }, 300, {
+          'leading': true
+        });
+
+        $scope.aceChanged = updateEditorMode;
+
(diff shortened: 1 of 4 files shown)


spadgett:
@jhadvig I know I said alert, but I think I would use the `has-warning` text below the field like

https://github.com/spadgett/origin-web-console/blob/cb93a4d8c41f25300694869a7bf14158831658d1/app/views/directives/osc-routing.html#L94-L98

We should show the message close to the input.

app/scripts/directives/createSecret.js
```diff
@@ -175,6 +175,23 @@ angular.module("openshiftConsole")
...
+            delete $scope.alerts.invalidFormat;
+          } catch (e) {
+            $scope.alerts["invalidFormat"] = {
+              type: "warning",
+              message: "Configuration file should be in JSON format."
+            };
+          }
+        }, 300, {
+          'leading': true,
+          'trailing': false
```
spadgett: With `trailing: false`, I don't think it will catch my last keystrokes, which could have fixed the validation failure.

jhadvig:
@spadgett updated
screenshot-53

fyi the `trailing: false` did catched the keystroke. But removed it just to be safe .. Thanks !

spadgett:
[merge]

PR #749, opened 2016-10-27
Bug fix 1389239 - Hang due to undefined object when no storage class …

zherman0:
Hang due to undefined object when no storage class is selected

### DIFF
--- a/app/scripts/controllers/createPersistentVolumeClaim.js
+++ b/app/scripts/controllers/createPersistentVolumeClaim.js
@@ -78,7 +78,7 @@ angular.module('openshiftConsole')
           var unit =  $scope.claim.unit || "Mi";
           pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
           pvc.spec.selector.matchLabels = keyValueEditorUtils.mapEntries( keyValueEditorUtils.compactEntries($scope.claim.selectedLabels) );
-          if ($scope.claim.storageClass !== null) {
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -78,7 +78,7 @@ angular.module('openshiftConsole')
           var unit =  $scope.claim.unit || "Mi";
           pvc.spec.resources.requests.storage = $scope.claim.amount + unit;
           pvc.spec.selector.matchLabels = keyValueEditorUtils.mapEntries( keyValueEditorUtils.compactEntries($scope.claim.selectedLabels) );
-          if ($scope.claim.storageClass !== null) {
+          if ( typeof($scope.claim.storageClass ) !== "undefined" && $scope.claim.storageClass  !== null ) {
```
spadgett: Can we simply check

`if ($scope.claim.storageClass) { ... }`

zherman0: @spadgett - that still does not seem to handle the undefined case which occurs when the user chooses no storage class

spadgett: When `$scope.claim.storageClass` is undefined, that `if` will evaluate to false.

zherman0:
@spadgett - On further testing, that does work for all cases.  Code updated. Thanks.

spadgett:
[merge]

PR #750, opened 2016-10-27
Bug fix 1383187 - container values were changed in controller but not…

zherman0:
Container values were changed in controller but not updated in html @spadgett


spadgett:
[merge]

PR #751, opened 2016-10-28
Updating PatternFly and Angular-PatternFly to v3.10.0

### LINKED ISSUE #438: Sort by popup tooltip on events covers dropdown menu and isn't needed
Suggest we remove the popover tooltip since "sort by" precedes the select field. If not removed, it's behavior should be changed to be positioned above so it doesn't overlay the dropdown menu.
@rhamilto @spadgett any objections?

screen shot 2016-08-22 at 10 16 38 am

rhamilto:
Resolves #438 

screen shot 2016-10-28 at 9 21 23 am

@jwforres or @spadgett, PTAL.

@benjaminapetersen, @sg00dwin, @jhadvig:  this PR contains dependency changes.  you'll need to hack/clean-deps.sh and hack/instal-deps.sh after it merges.


bower.json
```diff
@@ -51,11 +51,15 @@
...
   "appPath": "app",
   "resolutions": {
-    "lodash": "3.10.1",
     "angular-bootstrap": "0.14.3",
-    "moment": "2.10.6",
-    "matchHeight": "0.6.0",
-    "font-awesome": "4.6.1"
+    "font-awesome": "4.6.3",
+    "patternfly": "3.10.0",
+    "moment": "~2.14.1",
```
jwforres: Can this be an exact resolution of 2.14.2 ?

bower.json
```diff
@@ -51,11 +51,15 @@
...
+    "bootstrap-datepicker": "1.6.4",
+    "bootstrap-select": "1.10.0",
+    "bootstrap-touchspin": "~3.1.1",
```
jwforres: Can we resolve to a specific version

bower.json
```diff
@@ -51,11 +51,15 @@
...
+    "datatables-colreorder": "1.3.2",
+    "d3": "3.5.17"
```
jwforres: Why are we having to resolve some of these? We don't even use bootstrap-datepicker or bootstrap-touchspin. Some other dependency trying to pull in a diff version?

rhamilto: Yeah.  I was getting all sorts of conflicts when I updated.  I believe this is because PatternFly and Angular-PatternFly updated their dependencies.  I could only get resolution on these dependencies by including the resolutions stanza as included.  Am I missing something?

bower.json
rhamilto: Yeah.  Fixed.

bower.json
rhamilto: Apparently not.  If I remove the ~, bower asks me to resolve the version.  :(

rhamilto:
@jwforres, PTAL.  This now includes a new version of Angular-Key-Value-Editor that utilizes the same version of PatternFly (v3.10.0), which resolves the need for all the resolutions additions.

jwforres:
much better, thanks!  [merge]

PR #753, opened 2016-10-28
Fix selector comparisons that need to use covers checks

jwforres:


jwforres:
@spadgett PTAL

jwforres:
[merge]

PR #757, opened 2016-10-28
Fix set-limits link for HPA warning

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1389691

@jwforres PTAL, this is the only bad set-limits link I found


jwforres:
[merge]

PR #758, opened 2016-10-28
Adjustments so Safari maintains equal min-height on deployment tiles

sg00dwin:
Fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1389640

### DIFF
--- a/app/styles/_overview.less
+++ b/app/styles/_overview.less
@@ -10,7 +10,7 @@
 @two-column-min-width: 750px;
 @two-column-max-width: 749px;
 @tile-min-height:      256px;
-@tile-min-height-lg:   346px;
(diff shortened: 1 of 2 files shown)


sg00dwin:
@jwforres @spadgett  PTAL

spadgett:
@sg00dwin you'll need to call `hack/clean-assets.sh`, `hack/install-assets.js`, and `grunt build` again since a dependency changed.

jwforres:
if you want to save time, in this case a bower install should be sufficient

sg00dwin:
ok, is updated

sg00dwin:
screen shot 2016-10-28 at 5 14 02 pm

spadgett:
[merge]

PR #761, opened 2016-10-28
Setting popover content to wrap so that long strings don't overflow

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1389658

screen shot 2016-10-28 at 3 50 58 pm

@jwforres or @spadgett, PTAL.  We may want to revisit the inclusion of the entire certificate in the popover, but as we discussed on Friday, that behavior is consistent with the CLI.

### DIFF
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -455,9 +455,10 @@ code.probe-command {
 }
 
 .popover {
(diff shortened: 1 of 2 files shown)


jwforres:
[merge]

PR #762, opened 2016-10-31
Bug 1389685 - Support empty input of destinationDir on BC editor page

jhadvig:
Fixes two bugzillas:
- https://bugzilla.redhat.com/show_bug.cgi?id=1388812
- https://bugzilla.redhat.com/show_bug.cgi?id=1389685

Since the `destinationDir` && `mountPoint` parameters are not mandatory had to remove the ng-required attribute + adjust the logic for saving and adding the secrets.

@spadgett PTAL

### DIFF
--- a/app/scripts/controllers/edit/buildConfig.js
+++ b/app/scripts/controllers/edit/buildConfig.js
@@ -385,10 +385,12 @@ angular.module('openshiftConsole')
     };
 
     var updateSourceSecrets = function(object, pickedSecrets) {
-      var lastPickedSecret = _.head(pickedSecrets);
-      var property = $scope.strategyType === "Custom" ? "mountPath" : "destinationDir";
-      if (lastPickedSecret[property]) {
-        object.secrets = pickedSecrets;
+      var property = $scope.strategyType === "Custom" ? "secretSource" : "secret";
+      var nonEmptySecrets = _.filter(pickedSecrets, function(secret) {
+        return secret[property].name;
+      });
+      if (!_.isEmpty(nonEmptySecrets)) {
+        object.secrets = nonEmptySecrets;
       } else {
         delete object.secrets;
       }
--- a/app/views/directives/osc-source-secrets.html
+++ b/app/views/directives/osc-source-secrets.html
@@ -27,8 +27,7 @@
                   placeholder="/"
                   autocorrect="off"
                   autocapitalize="off"
-                  spellcheck="false"
-                  ng-required="pickedSecret.secret.name">
+                  spellcheck="false">
               </div>
             </div>
             <div class="remove-secret">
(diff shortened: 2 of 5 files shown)


app/scripts/controllers/edit/buildConfig.js
```diff
@@ -385,9 +385,12 @@ angular.module('openshiftConsole')
...
     var updateSourceSecrets = function(object, pickedSecrets) {
-      var lastPickedSecret = _.head(pickedSecrets);
-      var property = $scope.strategyType === "Custom" ? "mountPath" : "destinationDir";
-      if (lastPickedSecret[property]) {
+      var firstPickedSecret = _.head(pickedSecrets);
+      var property = $scope.strategyType === "Custom" ? "secretSource" : "secret";
+      if (firstPickedSecret[property].name) {
+        if (!_.last(pickedSecrets)[property].name) {
+          pickedSecrets.pop();
+        }
```
spadgett: I think I'd prefer something like

```js
var property = $scope.strategyType === "Custom" ? "secretSource" : "secret";
var secrets = _.filter(pickedSecrets, [property, 'name']);
if (!_.isEmpty(secrets)) {
  object.secrets = pickedSecrets;
}
```

app/views/directives/osc-source-secrets.html
```diff
@@ -5,7 +5,7 @@
         <div class="row advanced-secrets">
           <div class="col-lg-6">
             <label class="picker-label" ng-if="$first">Build Secret</label>
-            <ui-select ng-required="pickedSecret.destinationDir" ng-model="pickedSecret.secret.name">
+            <ui-select ng-model="pickedSecret.secret.name">
```
spadgett: Don't we still want to make the secret name required if a destination dir is entered? It doesn't make sense to entered a destination dir if you don't pick a secret.

app/views/directives/osc-source-secrets.html
```diff
@@ -59,9 +58,9 @@
         <div class="row advanced-secrets">
           <div class="col-lg-6">
             <label class="picker-label" ng-if="$first">Build Secret</label>
-            <ui-select ng-required="pickedSecret.mountPath !== ''" ng-model="pickedSecret.secretSource.name">
+            <ui-select ng-model="pickedSecret.secretSource.name">
```
spadgett: Don't we still want to make the secret name required if a mount path is entered?

jhadvig:
@spadgett comments addressed. PTAL :)

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -385,10 +385,10 @@ angular.module('openshiftConsole')
     };
 
     var updateSourceSecrets = function(object, pickedSecrets) {
-      var lastPickedSecret = _.head(pickedSecrets);
-      var property = $scope.strategyType === "Custom" ? "mountPath" : "destinationDir";
-      if (lastPickedSecret[property]) {
-        object.secrets = pickedSecrets;
+      var property = $scope.strategyType === "Custom" ? "secretSource" : "secret";
+      var nonEmptySecrets = _.filter(pickedSecrets, function(secret) {return secret[property].name});
```
spadgett: Missing semicolon.

```js
_.filter(pickedSecrets, function(secret) { return secret[property].name; });
```

Although I prefer this style

```js
_.filter(pickedSecrets, function(secret) {
  return secret[property].name;
});
```

Does the shorthand not work?

```js
_.filter(pickedSecrets, [property, 'name']);
```

jhadvig: no, it will return empty array

spadgett: ok thanks for checking

spadgett:
[merge]

PR #766, opened 2016-10-31
Allow filtering on project requester

### LINKED ISSUE #764: Project listing page hard for dedicated-admins
Dedicated customers have a few people who are marked as "dedicated admins" which gives them view access to ALL projects. These admins are also often end-users, ie they have their own personal projects and just shared ones. The searching is better but a little confusing. It would be good to be able to see who created the projects (or show only my projects) and/or filter by certain users.

Other key usability issue is being able to sort by creator or by creation date.

This would go a long way for those users who see many projects but want to see either their own or other team members.

/cc @spadgett what we chatted about

spadgett:
Show the project requester with the creation date in small text below
the project display name. Let users filter on the requester so they can
list the projects they created. Allow sorting by creator or creation date.

Fixes #764

screen shot 2016-10-31 at 9 45 42 am

@jwforres PTAL
@sspeiche CC

### DIFF
--- a/app/scripts/controllers/projects.js
+++ b/app/scripts/controllers/projects.js
@@ -33,7 +33,8 @@ angular.module('openshiftConsole')
     var filterFields = [
       'metadata.name',
       'metadata.annotations["openshift.io/display-name"]',
-      'metadata.annotations["openshift.io/description"]'
+      'metadata.annotations["openshift.io/description"]',
+      'metadata.annotations["openshift.io/requester"]'
     ];
 
     var filterProjects = function() {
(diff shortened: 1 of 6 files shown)


spadgett:
Updated to allow sorting by creator as well.

jwforres:
Seems reasonable to me, including @ajacobs21e

sspeiche:
Thanks @spadgett this is really useful

spadgett:
Fixed typos in commit message

spadgett:
[merge]

PR #767, opened 2016-10-31
Don't block creation on import yaml/json when we fail to do existence checks

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1389670


jwforres:
@spadgett PTAL

spadgett:
[merge]

Issue #669, opened 2016-10-13
Remove tech preview designation from pipelines


jwforres:

jwforres:
@bparees let us know when we are ok to do this

bparees:
probably when this goes in: https://github.com/openshift/origin/pull/11065

bparees:
(and that will go in after @gabemontero's oauth stuff has all landed in our jenkins image)

bparees:
you're ok to do this.

PR #768, opened 2016-10-31
Remove tech preview flag from pipeline pages

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/669

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -93,7 +93,6 @@
 .breadcrumb {
   font-size: @font-size-base - 1;
   padding-bottom: 0;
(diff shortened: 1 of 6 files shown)


jwforres:
@spadgett PTAL

spadgett:
Looks fine, but would you object to removing the breadcrumbs margin changed added here?

https://github.com/openshift/origin-web-console/pull/341/files#diff-e5367f9b3c5f479f2f08f2b050a22608R118

I think the breadcrumbs look better with the default margin.

jwforres:
updated with css change [merge]

PR #770, opened 2016-10-31
Bug fix 1389734- Double links due to logic check error

zherman0:
Double links on the deployment pages due to a logic error (forgot parens)


spadgett:
[merge]

PR #773, opened 2016-11-01
Bug 1388447 - On BC editor page, when 'Push To' is 'None', should disable 'Push Secret'

jhadvig:
@spadgett PTAL


spadgett:
[merge]

PR #774, opened 2016-11-01
Bug 1388884 - Specify only 'Source secret' on Create From Image page

jhadvig:
@spadgett thanks for the info in Bugzilla :+1: PTAL


spadgett:
[merge]

PR #776, opened 2016-11-01
Fix minification issue with overview metrics

spadgett:
The HTML minifier was changing attribute `compact="true"` to just `compact`, which caused the full metrics to be shown in the overview in the built binary.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1390499

@jwforres PTAL


jwforres:
[merge]

PR #777, opened 2016-11-01
Don't collapse boolean attributes in minified HTML

spadgett:
@jwforres Opinions on this change? I think it could avoid problems like 

https://bugzilla.redhat.com/show_bug.cgi?id=1390499

in the future.

https://github.com/kangax/html-minifier#options-quick-reference
http://perfectionkills.com/experimenting-with-html-minifier/#collapse_boolean_attributes

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -391,7 +391,7 @@ module.exports = function (grunt) {
           preserveLineBreaks: true,
           collapseWhitespace: true,
           conservativeCollapse: false,
-          collapseBooleanAttributes: true,
+          collapseBooleanAttributes: false,
           removeComments: true,
           removeCommentsFromCDATA: true,
           removeOptionalTags: false,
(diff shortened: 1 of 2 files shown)


jwforres:
have tried running with the change to see how big of a diff it creates for us?

jwforres:
have you tried running grunt build i mean

spadgett:
Added dist. It's not too large.

The changes from `spellcheck="false"` to just `spellcheck` by the minifier without this setting also seem wrong :(

jwforres:
honestly i think we should merge this

spadgett:
Rebased

benjaminapetersen:
Yeah, looks like thats probably a good idea...

jwforres:
[merge]

spadgett:
WebDriver flake [merge]

PR #781, opened 2016-11-01
Improve Membership columns with long words or text

benjaminapetersen:
Fix bug #1388340 
Improves word wrapping within container columns with or w/o hyphenation or natural breaks.

long user names:
screen shot 2016-11-01 at 12 08 27 pm 1

long service account names (namespace / name): 
screen shot 2016-11-01 at 12 15 55 pm

@jwforres @spadgett

### DIFF
--- a/app/styles/_membership.less
+++ b/app/styles/_membership.less
@@ -9,7 +9,7 @@
     }
   }
   .content-pane {
-    max-width: 1024px;
+    width: 100%;
     margin-top: 50px;
     .col-add-role {
       padding: 10px 0;
@@ -38,9 +38,13 @@
       }
     }
     .col-name {
-      min-width: 165px;
-      padding: 0 5px 0 0;
-      word-wrap: break-word;
+      &.half-width {
+        min-width: 50%;
+      }
+      .word-break-all;
+      min-width: 225px;
+      max-width: 30%;
+      padding: 0 5px 10px 0;
       input {
         max-width: 150px;
       }
@@ -81,8 +85,7 @@
   }
 }
 
-
-@media(min-width: @screen-sm-min) {
+@media(min-width: @screen-xs-min) {
   .membership {
     .content-pane {
       .col-add-role {
@@ -109,7 +112,7 @@
         }
       }
       .col-name {
-        min-width: 180px;
+        width: 30%;
         input {
           max-width: 175px;
         }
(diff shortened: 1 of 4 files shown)


app/styles/_membership.less
```diff
@@ -31,7 +31,7 @@
     .col-name {
       min-width: 165px;
       padding: 0 5px 10px 0;
-      word-wrap: break-word;
+      word-break: break-all;
```
jwforres: is there a reason you aren't using our word-break (i think is the name) mixin? it makes sure the best option is used for each browser

spadgett: Yeah, there are the `word-break` and `word-break-all` mixins. I'd prefer to use the former unless we really need `break-all`.

benjaminapetersen: @spadgett Yeah, `word-wrap: break-word;` doesn't get the job done with the non-hyphenated words.
@jwforres Mmm... I just forgot about it :)

Let me look @ the mixin quick and see if that does the job.

benjaminapetersen: With `word-wrap: break-word` you get this:
screen shot 2016-11-01 at 1 16 43 pm

benjaminapetersen: Updated w/mixin.

app/styles/_membership.less
```diff
@@ -100,7 +99,7 @@
         }
       }
       .col-name {
-        min-width: 180px;
+        width: 180px;
```
spadgett: Per our conversation, let's make the name width larger.

benjaminapetersen:
@sg00dwin happy to take any of your suggestions as well!

app/styles/_membership.less
benjaminapetersen: Also removing the max width on the content area.

benjaminapetersen:
@spadgett @sg00dwin here's where its at at this point:

50% for names in readonly mode:
screen shot 2016-11-01 at 3 21 12 pm
additional width in edit mode:
screen shot 2016-11-01 at 3 21 19 pm

tablet/mobile the same as before.

benjaminapetersen:
Additionally, this PR fixes an issue with column collapse in mobile IE:

before:
screen shot 2016-11-03 at 4 02 35 pm
after:
screen shot 2016-11-03 at 4 02 59 pm

benjaminapetersen:
rebased

app/styles/_membership.less
```diff
@@ -79,10 +80,12 @@
       }
     }
   }
+  .half-width {
+    min-width: 50% !important;
```
spadgett: Do we really need `!important`? Does setting `flex-basis` instead work?

benjaminapetersen: Its a class conditionally set:
```html
ng-class="{ 'half-width': !mode.edit }"
```
Setting `flex-basis: 50%` doesn't have the same effect.  There may be a better way to do it, but it is descriptive of my intent.  

Thinking on it...

benjaminapetersen: Updating to this:
```
.col-name {
      &.half-width {
        min-width: 50%;
      }
}
```

benjaminapetersen: that works, done

spadgett:
[merge]

PR #783, opened 2016-11-01
Don't add app label if anything in template already has app label

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/732


jwforres:
@spadgett PTAL

app/scripts/controllers/newfromtemplate.js
```diff
@@ -339,6 +339,18 @@ angular.module('openshiftConsole')
...
+        var shouldAddAppLabel = function() {
+          // If the template defines its own app label, we don't need to add one at all
+          if (_.get($scope.template, 'labels.app')) {
+            return false;
+          }
+
+          // Otherwise check if an object in the template has an app label defined
+          return _.every($scope.template.objects, function(obj) {
+            return !_.get(obj, "metadata.labels.app");
+          });
```
spadgett: Does this work?

```js
return _.some($scope.template.objects, "metadata.labels.app");
```

jwforres: needs a ! in front but yeah that works too, updating

jwforres:
updated [merge]

Issue #728, opened 2016-10-25
DataService list doesn't return a promise


jwforres:
DataService.list should return a promise to better handle error scenarios.

PR #786, opened 2016-11-01
Data service List method returns a promise, passing callback as a param now optional

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/728

### DIFF
--- a/app/scripts/services/data.js
+++ b/app/scripts/services/data.js
@@ -77,7 +77,7 @@ angular.module('openshiftConsole')
   }
 
   function DataService() {
-    this._listCallbacksMap = {};
+    this._listDeferredMap = {};
     this._watchCallbacksMap = {};
     this._watchObjectCallbacksMap = {};
     this._watchOperationMap = {};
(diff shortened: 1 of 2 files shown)


jwforres:
@spadgett PTAL

jwforres:
@benjaminapetersen @sub-mod FYI

benjaminapetersen:
Nice, looks good to me.  I wasn't super excited to volunteer to figure this one out. :smile:

jwforres:
i'd like to deprecate passing the callback in, but not till we get a chance to quit using it ourselves

jwforres:
@spadgett updated based on our discussion

benjaminapetersen:
Hmm. I'd be good with you adding a deprecation warning now & then we could make an issue w/a list of `DataService.list()` calls to fix (I'll even volunteer to do that part).

jwforres:
i would want to do a deprecation warning as a console.warning so extensions
writers are aware of it without turning on our Logger, and i don't want
that happening with our OOTB product calls, would rather wait till ours are
taken care of

benjaminapetersen:
Ah, that makes sense.

jwforres:
[merge]

Issue #780, opened 2016-11-01
Secrets page contains doubled border underneath h1


rhamilto:
<img width="1018" alt="screen shot 2016-11-01 at 10 26 25 am" src="https://cloud.githubusercontent.com/assets/895728/19893079/466d1a40-a01e-11e6-9c7c-5b7521929a45.PNG">

Related issue:  #779

Issue #779, opened 2016-11-01
Pipelines and Secrets pages have border underneath h1, but no filter


rhamilto:
I noticed the Pipelines and Secrets pages are outliers in that they have borders underneath the h1, but do not include the filter toolbar.  Comparable pages Quota and Membership do not have the border.

<img width="1018" alt="screen shot 2016-11-01 at 10 26 25 am" src="https://cloud.githubusercontent.com/assets/895728/19893079/466d1a40-a01e-11e6-9c7c-5b7521929a45.PNG">

<img width="1015" alt="screen shot 2016-11-01 at 10 31 31 am" src="https://cloud.githubusercontent.com/assets/895728/19893207/bab33b1e-a01e-11e6-95bd-87d02bc6f0d6.PNG">

sg00dwin:
I suggest all secondary level pages (any page that can be directly navigated to from the primary nav) have consistent header structure, and would have the header with the single border. And other pages that are tertiary (eg. /builds/pipelines/sample-pipeline) would not.

@rhamilto your thoughts?

rhamilto:
@sg00dwin, sounds good to me.  Consistency is key for me.

PR #787, opened 2016-11-01
Update pipelines, secrets, quota and membership page headers to be consistent

sg00dwin:
with the rest of secondary level pages.

- added `page-header` parent div for `h1`
- changed class name to more appropriately specify when it's used.

Fixes https://github.com/openshift/origin-web-console/issues/779
Fixes https://github.com/openshift/origin-web-console/issues/780

screen shot 2016-11-01 at 3 10 47 pm
screen shot 2016-11-01 at 3 07 40 pm
screen shot 2016-11-01 at 3 07 22 pm
screen shot 2016-11-01 at 3 07 14 pm

### DIFF
--- a/app/styles/_substructure.less
+++ b/app/styles/_substructure.less
@@ -80,7 +80,7 @@ html,body {
     }
   }
 }
-.header-light {
+.header-toolbar {
   background-color: @panel-light;
   border-bottom: 1px solid @page-header-border-color;
 }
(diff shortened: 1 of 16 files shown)


spadgett:
@sg00dwin did you mean to close this?

sg00dwin:
@spadgett yes, I noticed an inconsistency with the start pipeline button location compared to other pages. I would think it would be within the header aligned with the `h1` like on other pages that have actions. Would you agree?

spadgett:
@sg00dwin No, it's specific to that one build config. There can be multiple start pipeline buttons for each pipeline you have.

sg00dwin:
ok, i'll rebase the pr and open again

PR #788, opened 2016-11-01
Add additional check to addRoleToUser() to further insulate from duplication

benjaminapetersen:
Fixes bug #1388060 

The "select role" picker already filters out roles that a subject already owns.  This update is for the manual input box at the bottom of the list.  If a user enters an existing subject name & selects a role that that the subject already has an info alert will notify that the update is unnecessary.

screen shot 2016-11-01 at 4 29 41 pm

@jwforres @spadgett

### DIFF
--- a/app/scripts/controllers/membership.js
+++ b/app/scripts/controllers/membership.js
@@ -44,8 +44,9 @@ angular
         },
         update: {
           subject: {
-            success: _.template('The role "<%= roleName %>" was given to "<%= subjectName %>".'),
-            error: _.template('The role "<%= roleName %>" was not given to "<%= subjectName %>".')
+            success: _.template('The role "<%= roleName %>" was granted to "<%= subjectName %>".'),
+            error: _.template('The role "<%= roleName %>" could not be granted to "<%= subjectName %>".'),
+            exists: _.template('The role "<%= roleName %>" has already been granted to "<%= subjectName %>".')
           }
         },
         errorReason: _.template('Reason: "<%= httpErr %>"')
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/membership.js
```diff
@@ -45,7 +45,8 @@ angular
         update: {
           subject: {
             success: _.template('The role "<%= roleName %>" was given to "<%= subjectName %>".'),
-            error: _.template('The role "<%= roleName %>" was not given to "<%= subjectName %>".')
+            error: _.template('The role "<%= roleName %>" was not given to "<%= subjectName %>".'),
+            exists: _.template('The role "<%= roleName %>" has previously been given to "<%= subjectName %>".')
```
jwforres: hmm maybe

"has already been granted to"  OR  "is already granted to"

app/scripts/controllers/membership.js
```diff
@@ -45,7 +45,8 @@ angular
         update: {
           subject: {
             success: _.template('The role "<%= roleName %>" was given to "<%= subjectName %>".'),
-            error: _.template('The role "<%= roleName %>" was not given to "<%= subjectName %>".')
+            error: _.template('The role "<%= roleName %>" was not given to "<%= subjectName %>".'),
```
jwforres: not new, but could you change "was not" to "could not be" ?

I'd probably change "given" to "granted" on all of these.  You generally talk about granting someone access to something.  Other opinions  @spadgett @rhamilto  ?

benjaminapetersen: Assuming change "remove" to "revoke" as well?

benjaminapetersen: Ie, "Removing the role...."  to "Revoking the role..." ?

jwforres:
[merge]

PR #792, opened 2016-11-02
Don't request stacked metrics for container-level metrics

spadgett:
We were incorrectly using the stacked URL template for requesting memory and CPU usage for a container on the pod metrics tab, which caused pod-level metrics to be displayed rather than container-level metrics.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1387274

With this change, we are correctly displaying different values for the different containers.

screen shot 2016-11-02 at 9 46 33 am

screen shot 2016-11-02 at 9 46 27 am

@jwforres PTAL
@derekwaynecarr @sjenning CC


jwforres:
[merge]

PR #794, opened 2016-11-02
Adding middle truncation to warning popovers

rhamilto:
Additional fix for https://bugzilla.redhat.com/show_bug.cgi?id=1389658

Popover text exceeding 350 characters will get truncated in the middle.  I arbitrarily chose 350.  Should it be more/less?

screen shot 2016-11-02 at 11 47 17 am

@jwforres or @spadgett, PTAL

### DIFF
--- a/app/views/directives/_warnings-popover.html
+++ b/app/views/directives/_warnings-popover.html
@@ -1,6 +1,6 @@
 <span ng-if="content">
   <span
-    dynamic-content="{{content}}"
+    dynamic-content="{{content | middleEllipses:350:'...<br>...'}}"
     data-toggle="popover"
     data-trigger="hover"
     data-html="true"
--- a/app/scripts/filters/util.js
+++ b/app/scripts/filters/util.js
@@ -500,6 +500,35 @@ angular.module('openshiftConsole')
       return truncated;
     };
   })
+  .filter('middleEllipses', function() {
+    /* Adapted from https://github.com/jviotti/angular-middle-ellipses
+     * Usage:  {{ 'MyVeryLongString' | middleEllipses:5:' ... ' }}
+     */
+    return function(input, limit, ellipses) {
+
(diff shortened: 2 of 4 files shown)


app/views/directives/_warnings-popover.html
```diff
@@ -1,6 +1,6 @@
 <span ng-if="content">
   <span
-    dynamic-content="{{content}}"
+    dynamic-content="{{content | middleEllipses:350:'...<br>...'}}"
```
spadgett: Use HTML entity for the ellipsis?

```html
dynamic-content="{{content | middleEllipses:350:'&hellip;<br>&hellip;'}}"
```

spadgett:

app/views/directives/_warnings-popover.html
rhamilto: I agree!  Unfortunately, using the html entity caused the second ellipsis to take up its own line in the popover, so I settled for the less semantic three periods.  :(

rhamilto:
D'oh.  I had an outdated Bower dependency.  That's what I get for rushing to submit the PR before yoga.  Is fixed now.

app/views/directives/_warnings-popover.html
jwforres: Nevermind, looks like its handling it

rhamilto: Line 6 below makes this possible, no?  Bootstrap's popover documentation says, 

> Insert HTML into the popover. If false, jQuery's text method will be used to insert content into the DOM. Use text if you're worried about XSS attacks.

jwforres:
[merge]

app/views/directives/_warnings-popover.html
jwforres: Yes, something else we are doing is sanitizing the error message already.

PR #796, opened 2016-11-02
Update pipelines, secrets, quota and membership page headers to be consistent

sg00dwin:
with the rest of secondary level pages.

Fixes https://github.com/openshift/origin-web-console/issues/779
Fixes https://github.com/openshift/origin-web-console/issues/780

@spadgett PTAL

screen shot 2016-11-01 at 3 10 47 pm
screen shot 2016-11-01 at 3 07 40 pm
screen shot 2016-11-01 at 3 07 22 pm
screen shot 2016-11-01 at 3 07 14 pm

### DIFF
--- a/app/views/membership.html
+++ b/app/views/membership.html
@@ -5,30 +5,32 @@
     <div class="middle-container">
       <div class="middle-header">
         <div class="container-fluid">
-          <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
+          <div class="page-header page-header-bleed-right page-header-bleed-left">
+            <h1>
+              <a
+                class="pull-right btn btn-default"
+                href=""
+                ng-if="'rolebindings' | canI : 'update'"
+                ng-click="toggleEditMode()">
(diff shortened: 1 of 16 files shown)


app/views/membership.html
```diff
@@ -5,30 +5,32 @@
     <div class="middle-container">
       <div class="middle-header">
         <div class="container-fluid">
-          <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
```
spadgett: If you remove the breadcrumbs from the view, we should also take the breadcrumbs out of the controller.

app/views/quota.html
```diff
@@ -6,18 +6,19 @@
       <div class="middle-container">
         <div class="middle-header">
           <div class="container-fluid">
-            <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
```
spadgett: If you remove the breadcrumbs from the view, we should also take the breadcrumbs out of the controller.

spadgett:
[merge]

PR #797, opened 2016-11-02
Membership: support unlisted namespace for Service Account

benjaminapetersen:
- [Tc 5743](https://trello.com/c/365NeA8Y) allows user to type in an unlisted project in the project picker for service accounts

@jwforres @spadgett 

Aside, I'm using `refresh-delay="200"` to reduce the flicker as the list updates, we may want to do the same in the is-tag-select [directive](https://github.com/spadgett/origin-web-console/blob/bee47e4c56db4b4c623a7ec778f9adfcea4d4cdc/app/views/directives/istag-select.html).

### DIFF
--- a/app/scripts/controllers/membership.js
+++ b/app/scripts/controllers/membership.js
@@ -214,10 +214,18 @@ angular
 
 
       DataService.list('projects', {}, function(resp) {
+        var projects = _.map(resp.by('metadata.name'), function(project) {
+          return project.metadata.name;
+        });
         angular.extend($scope, {
-          projects: _.map(resp.by('metadata.name'), function(project) {
-            return project.metadata.name;
-          })
+          projects: projects,
+          refreshProjects: function(search) {
+            if(search && !_.includes($scope.projects, search)) {
+              $scope.projects = [search].concat(projects);
+            } else {
+              $scope.projects = projects;
+            }
+          }
         });
       });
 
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/membership.js
```diff
@@ -210,10 +210,21 @@ angular
...
-          projects: _.map(resp.by('metadata.name'), function(project) {
-            return project.metadata.name;
-          })
+          projects: projects,
+          refreshProjects: function(search) {
+            var copy;
+            if(search && !_.includes($scope.projects, search)) {
+              copy = projects.slice();
+              copy.unshift(search);
+              $scope.projects = copy;
```
spadgett: Does this work? I think it's a little more obvious:

```js
$scope.projects = [search].concat(projects);
```

app/views/membership.html
```diff
@@ -193,7 +193,9 @@
                             <span ng-bind="newBinding.namespace"></span>
                           </ui-select-match>
                           <ui-select-choices
-                            repeat="projectName in projects | filter: $select.search">
+                            repeat="projectName in projects track by $index | filter: $select.search"
```
spadgett: `track by $index` doesn't look right here and below, particularly since we're calling `unshift` to prepend things.

app/scripts/controllers/membership.js
benjaminapetersen: ah, much better, i agree.

app/views/membership.html
benjaminapetersen: Oop, true, habit.  removing.

benjaminapetersen:
rebased

spadgett:
[merge]

PR #798, opened 2016-11-02
Update refresh delay in istag-select

benjaminapetersen:
- Update the value to match setting on [membership page](https://github.com/openshift/origin-web-console/pull/797).

Eliminates the constant flicker, though there is a still re-render of the drop down menu.  

@jwforres @spadgett


spadgett:
[merge]

PR #805, opened 2016-11-02
Consistently use headline case for labels on resource pages

spadgett:
We sometimes use headline case and sometimes sentence case. We're even inconsistent on the same page at times. Use headline case for the labels always.

Patternfly guidelines:

http://www.patternfly.org/styles/terminology-and-wording/

The guidelines don't seem to deal with this case specifically, although for other labels like input labels they say to use headline. Whatever we pick, I'd like to be consistent :)

@jwforres PTAL
@rhamilto CC


jwforres:
[merge]

PR #810, opened 2016-11-02
Add trigger, but set automatic to false

### LINKED ISSUE #808: Removal of the image trigger from autodeploy shouldn't remove the trigger
autodeploy

Unselecting "New image is available" shouldn't remove the ICT from the DC. It should just set it to automatic=false. In that way user errors like https://bugzilla.redhat.com/show_bug.cgi?id=1383804 will be greatly reduced where a user creates a new DC, doesn't want to have automatic deployments, but when tries to deploy they get image pull errors because we cannot resolve their images. It would be great to have this fixed for 1.4.

spadgett:
If a user unchecks deploy automatically on new images, still add the image change trigger, but set automatic to false. This avoids image pull errors.

Fixes #808 

@jwforres PTAL
@jhadvig I think we should update the DC editor to handle this flag.
@kargakis CC

### DIFF
--- a/app/scripts/services/applicationGenerator.js
+++ b/app/scripts/services/applicationGenerator.js
@@ -220,23 +220,21 @@ angular.module("openshiftConsole")
           }
         }
       };
-      if(input.deploymentConfig.deployOnNewImage){
-        deploymentConfig.spec.triggers.push(
-          {
-            type: "ImageChange",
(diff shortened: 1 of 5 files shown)


jwforres:
changes LGTM but travis failed on the dist

spadgett:
[merge]

PR #813, opened 2016-11-03
Membership: temporarily eliminate default description tooltip for roles

benjaminapetersen:
Until descriptions merge: https://github.com/openshift/origin/pull/11328
The tooltip will still function if a cluster admin manually annotates roles.

@jwforres @spadgett


jwforres:
[merge]

PR #815, opened 2016-11-03
Fix inline learn more link in heading

benjaminapetersen:
Problem:
screen shot 2016-11-03 at 3 06 45 pm

Resolution:
screen shot 2016-11-03 at 3 07 04 pm

The question is, will we use this in the future & should it be made generic?
@rhamilto @sg00dwin may be interested in commenting on that.

@jwforres @spadgett

### DIFF
--- a/app/styles/_membership.less
+++ b/app/styles/_membership.less
@@ -1,4 +1,13 @@
 .membership {
+  // TODO: make generic?
+  h1 {
+    .learn-more-inline {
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
Fix left padding:
screen shot 2016-11-03 at 3 18 17 pm

jwforres:
[merge]

Issue #812, opened 2016-11-03
Awkward wrapping of env section tooltips in create flow on mobile


spadgett:
Tested at iPhone 7 screen resolution.

Creating from builder image, icon wraps by itself.

<img width="353" alt="glass" src="https://cloud.githubusercontent.com/assets/1167259/19967157/78bf3f9a-a1a5-11e6-9b3c-e0fffb2cfbd4.png">

---
Deploying image, help text closer to inputs than the header its describing.

<img width="367" alt="glass" src="https://cloud.githubusercontent.com/assets/1167259/19967250/eb20347c-a1a5-11e6-8f59-57e8ed5fb8fc.png">

PR #817, opened 2016-11-03
Introduce block class wrapper to control spacing consistency for form h2 headings

sg00dwin:
Add markup to keep help icon from wrapping awkwardly at mobile.
Misc: Add spacing class so create secret and create storage match
Misc: Change markup grid so that fromtemplate and [fromimage](https://github.com/openshift/origin-web-console/blob/master/app/views/create/fromimage.html#L20) forms are consistent widths

Fixes https://github.com/openshift/origin-web-console/issues/812

screen shot 2016-11-03 at 4 40 05 pm

screen shot 2016-11-03 at 4 36 17 pm

### DIFF
--- a/app/styles/_utils.less
+++ b/app/styles/_utils.less
@@ -16,3 +16,9 @@
 .hide-if-empty:empty {
   display: none !important;
 }
+
+// keep icon from wrapping early
+.appended-icon {
+  display:inline-block;
+  .nowrap;
+}
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -499,6 +499,11 @@ label.checkbox {
       &.right {
         font-size: @font-size-small;
         font-weight: normal;
+        .action-inline {
+          @media (max-width: @screen-xs-max) {
+            margin-left: 0;
+          }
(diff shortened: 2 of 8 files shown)


app/styles/_utils.less
```diff
@@ -16,3 +16,8 @@
 .hide-if-empty:empty {
   display: none !important;
 }
+
+// generic classes not included with twbs
+.inline-block {
+  display:inline-block;
+}
```
spadgett: Not sure this is worthwhile. How is it better than `style="display: inline-block;"`? It's basically an inline style that we've given a class name.

@rhamilto Thoughts?

rhamilto: I agree in that I think it's be better to create a consistent markup pattern with classes semantically describing the elements (separating content from presentation) rather than putting helper classes on elements as needed (putting presentational classes in the content).

sg00dwin: I changed it from a helper class to a semantic class `.appended-icon` that can be reused.

spadgett:
[merge]

PR #820, opened 2016-11-04
Quota error for controllers is wrong when out of quota for pods

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1386171


jwforres:
@spadgett PTAL

spadgett:
[merge]

PR #822, opened 2016-11-04
Bug Fix 1390826 - PVC does not bind

zherman0:
2 changes here to fix the binding, first changed references to 'beta' instead of 'alpha' which was causing most of the issues. Second, removed reference to label selectors if no label selector was entered instead of leaving empty object.


spadgett:
[merge]

Issue #803, opened 2016-11-02
Loading message too close to bottom border of Secrets page header


spadgett:
<img width="149" alt="screen shot 2016-11-02 at 3 46 42 pm" src="https://cloud.githubusercontent.com/assets/1167259/19944802/a3807ef2-a113-11e6-936c-eb48fed130ef.png">

See also #796

PR #823, opened 2016-11-04
Add spacing to 'Loading...' message on /secrets

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/803

screen shot 2016-11-04 at 3 03 27 pm

### DIFF
--- a/app/views/secrets.html
+++ b/app/views/secrets.html
@@ -17,7 +17,7 @@
       </div><!-- /middle-header-->
       <div class="middle-content">
         <div class="container-fluid">
-          <div ng-if="!loaded">Loading...</div>
+          <div ng-if="!loaded" class="mar-top-xl">Loading...</div>
           <div ng-if="loaded" class="row">
             <div class="col-md-12">
               <h3>Source Secrets</h3>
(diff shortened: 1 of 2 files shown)


sg00dwin:
@spadgett PTAL

sg00dwin:
@spadgett anything else needed for this?

app/styles/_ellipsis.less
```diff
@@ -133,3 +133,8 @@
     z-index: 1;
   }
 }
+
+// static loading... spacing
+.section-loading {
+  margin: 20px 0;
```
spadgett: Just wondering if we really should have a generic margin for the loading message. We usually try to position the loading message exactly where the content displays to avoid content jumping on the page as it loads. So the content would need the same margin, but `class="section-loading"` doesn't make sense on the content.

I'm thinking we should have a bottom margin as part of the header bottom border, then not have top margin on the content on any of those pages. That way we're always consistent.

I do like that we are adding standard classes for this stuff. The `mar-top-*` classes mean we frequently use different margins on different pages.

/cc @rhamilto

rhamilto: > We usually try to position the loading message exactly where the content displays to avoid content jumping on the page as it loads. 

Except when we use the other "Loading..." message style.  Is there a rationale for having both?

loading2

spadgett: I'd vote for the loading message style from the screenshot for all pages where it makes sense.

jwforres:
[merge]

PR #828, opened 2016-11-04
Adding .word-break() to card-pf-titles so unbroken strings don't

rhamilto:
make the card container expand.

Before:

screen shot 2016-11-04 at 3 54 41 pm

After:

screen shot 2016-11-04 at 3 54 53 pm

@spadgett, PTAL


spadgett:
[merge]

PR #829, opened 2016-11-04
Allow expandable list view headings to wrap instead of truncate

rhamilto:
So that the all of the heading is visible.  Plus a little LESS tidying.

Before:

screen shot 2016-11-04 at 4 09 15 pm

After:

screen shot 2016-11-04 at 4 09 24 pm

@spadgett, PTAL


spadgett:
[merge]

PR #830, opened 2016-11-07
Add project membership link to projects list

spadgett:
Rather than showing the CLI command to add users to the project, add an action directly on the projects list for viewing and editing membership now that it can be changed in the web console.

screen shot 2016-11-07 at 8 38 00 am

/cc @jwforres @benjaminapetersen @sspeiche

### DIFF
--- a/app/styles/_projects.less
+++ b/app/styles/_projects.less
@@ -15,6 +15,9 @@
   }
   .project-action-item {
     margin-left: 12px;
+    @media (min-width: @screen-sm-min) {
+      margin-left: 17px;
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


spadgett:
Thoughts on this change? I'm worried membership won't be discoverable buried in a menu.

sspeiche:
I agree with removing the prominent text pointing to the CLI. I think this change is very useful, since there will be a good % of the usage of project membership will be someone just jumping in to give someone access and it would be most likely they'd come at it from the project list -> membership. Seems natural. +1 for this change

jwforres:
What about when the project list is empty?

benjaminapetersen:
Yeah, agree, Membership under "Other Resources" is not intuitive, this definitely improves discoverability.

benjaminapetersen:
That cli command:
```
A project admin can add you to a role on a project by running the command oc policy add-role-to-user <role> developer -n <projectname>
```
I think it makes sense that we drop it.  No need to show it if you don't have any projects to begin with, and once you have a project the icon is pretty intuitive.

spadgett:
> What about when the project list is empty?

Here's how it looks for users who can't create projects:

screen shot 2016-11-07 at 8 59 33 am

jwforres:
[merge]

Issue #818, opened 2016-11-03
Hawtio-core.js logs random things


benjaminapetersen:
Not entirely sure why yet:
![screen shot 2016-11-03 at 4 40 57 pm](https://cloud.githubusercontent.com/assets/280512/19984360/5eef9f08-a1e4-11e6-87da-e5478e220e4a.png)

jwforres:
Found the problem, the deployment controller is using Logger but didnt add it to the dep list at the top.  So its actually using the Logger that comes with hawtio-core which just wraps console.log

jwforres:
we are actually doing this in a bunch of places @spadgett

benjaminapetersen:
Oh? So Logger from Hawtio is different than our Logger service...

PR #835, opened 2016-11-07
Fix references to Logger that are using the wrong one

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/818

### DIFF
--- a/app/scripts/controllers/create/nextSteps.js
+++ b/app/scripts/controllers/create/nextSteps.js
@@ -9,7 +9,7 @@
  * Controller of the openshiftConsole
  */
 angular.module("openshiftConsole")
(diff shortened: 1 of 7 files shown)


jwforres:
@spadgett PTAL

spadgett:
Wonder if there's a way to guard against this in the future. Is the other Logger coming from Hawtio?

jwforres:
@spadgett it is, we could call ours something different

jwforres:
[merge]

spadgett:
I think we should remove the `Logger` global from jshintrc:

https://github.com/spadgett/origin-web-console/blob/22cbe94128168282f06b4e52766f05170a77df92/.jshintrc#L30

PR #838, opened 2016-11-07
Allow project names to wrap and break as needed

### LINKED ISSUE #831: Long project name with no hyphens can overflow tile
screen shot 2016-11-07 at 8 47 41 am

/cc @jwforres @sg00dwin @rhamilto

rhamilto:
Fixes #831 

Rather than ellips-ing the project name at desktop resolutions and allowing it to wrap at mobile, how about we always allow it to wrap (wherever necessary)?  Breaking wherever necessary is the only way to resolve the bug.

screen shot 2016-11-07 at 10 17 37 am
screen shot 2016-11-07 at 10 17 51 am
screen shot 2016-11-07 at 10 18 00 am
screen shot 2016-11-07 at 10 18 20 am

@spadgett, PTAL

p.s., pfListView is a bag of hurt.  We should avoid using it for anything other than [the very specific context for which it was created](http://angular-patternfly.rhcloud.com/#/api/patternfly.views.directive:pfListView) because we're having to override a bunch of problems it introduces; we'd be better off just rolling our own solution.

### DIFF
--- a/app/styles/_projects.less
+++ b/app/styles/_projects.less
@@ -23,20 +23,29 @@
 
 .project-additional-info.list-view-pf-additional-info {
   @media (min-width: @screen-md-min) {
-    width: 60%;
+    width: 55%;
   }
 }
 
-.project-info.list-group-item {
-  border-bottom: 0;
-  border-color: #e0e0e0;
-  padding: 10px 20px;
+.project-info {
+  &.list-group-item {
+    border-bottom: 0;
+    border-color: #e0e0e0;
+    padding: 10px 20px;
+  }
+  .list-view-pf-description {
+    width: 100%;
+  }
+  .list-view-pf-main-info {
+    display: block;
+  }
+
 }
 
 .project-name-item.list-group-item-heading {
   @media (min-width: @screen-md-min) {
     white-space: normal;
-    width: 40%;
+    width: 45%;
   }
   .h1 {
     margin-bottom: 0;
@@ -46,7 +55,10 @@
       margin-left: 5px;
     }
     a {
+      display: block;
+      overflow: hidden;
       line-height: initial;
+      text-overflow: ellipsis;
     }
   }
   p {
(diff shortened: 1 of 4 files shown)


app/styles/_projects.less
```diff
@@ -43,7 +43,9 @@
       margin-left: 5px;
     }
     a {
+      display: block;
       line-height: initial;
+      .word-break-all();
```
spadgett: If you switch to `inline-block`, does `word-break()` work?

spadgett: I'd almost rather use `overflow: hidden` and truncate the text:

openshift_web_console_and_developer_tools_-_openshift_web_console_-_https___localhost_9000_

rhamilto:
So how about allowing lines to wrap, but truncating individual words that are too long to fit on a single line?

screen shot 2016-11-07 at 1 32 05 pm
screen shot 2016-11-07 at 1 32 15 pm
screen shot 2016-11-07 at 1 32 26 pm
screen shot 2016-11-07 at 1 32 42 pm

app/styles/_projects.less
benjaminapetersen: I'm def curious why that preference (not having a strong opinion myself, but impulse I'd prob tend to choose word-break-all).

spadgett:
> So how about allowing lines to wrap, but truncating individual words that are too long to fit on a single line?

+1, that is what I was thinking

spadgett:
Not new, but should the name column have more width and description less?

rhamilto:
Yeah.  Good catch!  There was a bug in those last two screenshots.  Fixed now.
screen shot 2016-11-07 at 1 58 36 pm
screen shot 2016-11-07 at 1 58 58 pm
screen shot 2016-11-07 at 1 58 43 pm

rhamilto:
One concession:  IE11 and Edge only ellips-ize the first word in the string.  

screen shot 2016-11-07 at 2 02 57 pm

 Are we ok with that?

rhamilto:
@jwforres or @spadgett, PTAL.

spadgett:
@jwforres opinion for 1.4?

jwforres:
yes since it blocks actions when the text overflows [merge]

spadgett:
WebDriver flake :(

[merge]

PR #839, opened 2016-11-07
add more detail to specific service types

juanvallejo:
Related Trello Card:
https://trello.com/c/5dZkZQAX/807-2-details-on-other-service-types-kubernetes-networking

Fixes https://github.com/openshift/origin/issues/5979

This patch adds more detail, such as **ingress points** and **external IPs**, to service types
that support these extra fields (LoadBalancer, etc).
 
add-more-detail-other-svc-types

@spadgett @jwforres

### DIFF
--- a/app/scripts/controllers/service.js
+++ b/app/scripts/controllers/service.js
@@ -14,6 +14,7 @@ angular.module('openshiftConsole')
                                              $filter) {
     $scope.projectName = $routeParams.project;
     $scope.service = null;
+    $scope.services = null;
     $scope.alerts = {};
     $scope.renderOptions = $scope.renderOptions || {};
     $scope.renderOptions.hideFilterWidget = true;
@@ -27,11 +28,60 @@ angular.module('openshiftConsole')
       }
     ];
 
+    $scope.podFailureReasons = {
+      "Pending": "This pod will not receive traffic until all of its containers have been created."
+    };
+
+    var allPods = {};
     var watches = [];
 
+    // receives routes for the current service and maps service ports to each route name
+    var getPortsByRoute = function() {
+      if(!$scope.service) {
+        return;
+      }
+
+      $scope.portsByRoute = {};
+
+      _.each($scope.service.spec.ports, function(port) {
+        var reachedByRoute = false;
+        if(port.nodePort) {
+          $scope.showNodePorts = true;
+        }
+
+        _.each($scope.routesForService, function(route) {
+          if(!route.spec.port || route.spec.port.targetPort === port.name ||
+            route.spec.port.targetPort === port.targetPort) {
+            $scope.portsByRoute[route.metadata.name] = $scope.portsByRoute[route.metadata.name] || [];
+            $scope.portsByRoute[route.metadata.name].push(port);
+            reachedByRoute = true;
+          }
+        });
+
+        if(!reachedByRoute) {
+          $scope.portsByRoute[''] = $scope.portsByRoute[''] || [];
+          $scope.portsByRoute[''].push(port);
+        }
+      });
+    };
+
+    // receive pods for the current service scope only when the service object is available
+    var getPodsForService = function() {
+      $scope.podsForService = {};
+      if (!$scope.service) {
+        return;
+      }
+
+      var ls = new LabelSelector($scope.service.spec.selector);
+      $scope.podsForService = ls.select(allPods);
+    };
     var serviceResolved = function(service, action) {
       $scope.loaded = true;
       $scope.service = service;
+
+      getPodsForService();
+      getPortsByRoute();
+
       if (action === "DELETED") {
         $scope.alerts["deleted"] = {
           type: "warning",
@@ -60,15 +110,41 @@ angular.module('openshiftConsole')
           }
         );
 
+        watches.push(DataService.watch("services", context, function(services) {
+          $scope.services = services.by("metadata.name");
+        }));
+
+        watches.push(DataService.watch("pods", context, function(pods) {
+          allPods = pods.by("metadata.name");
+          getPodsForService();
+        }));
+
+        watches.push(DataService.watch("endpoints", context, function(endpoints) {
+          $scope.podsWithEndpoints = {};
+          var svcEndpoint = endpoints.by("metadata.name")[$routeParams.service];
+          if (!svcEndpoint) {
+            return;
+          }
+
+          _.each(svcEndpoint.subsets, function(subset) {
+            _.each(subset.addresses, function(address) {
+              if (_.get(address, "targetRef.kind") === "Pod") {
+                $scope.podsWithEndpoints[address.targetRef.name] = true;
+              }
+            });
+          });
+        }));
+
         watches.push(DataService.watch("routes", context, function(routes) {
-          $scope.routesForService = [];
+          $scope.routesForService = {};
           angular.forEach(routes.by("metadata.name"), function(route) {
             if (route.spec.to.kind === "Service" &&
                 route.spec.to.name === $routeParams.service) {
-              $scope.routesForService.push(route);
+              $scope.routesForService[route.metadata.name] = route;
             }
           });
 
+          getPortsByRoute();
           Logger.log("routes (subscribe)", $scope.routesByService);
         }));
 
--- a/app/scripts/directives/resources.js
+++ b/app/scripts/directives/resources.js
@@ -165,9 +165,31 @@ angular.module('openshiftConsole')
       restrict: 'E',
       scope: {
         pods: '=',
+        // Optional active pods map to display whether or not pods have endpoints
+        activePods: '=?',
         // Optional empty message to display when there are no pods.
-        emptyMessage: '=?'
+        emptyMessage: '=?',
+        // Alternative header text to display in the 'Name' column.
+        customNameHeader: '=?',
+        // Optional map of explanations or warnings for each phase of a pod
+        podFailureReasons: '=?'
       },
       templateUrl: 'views/directives/pods-table.html'
+    };
+  })
+  .directive('trafficTable', function() {
+    return {
+      restrict: 'E',
+      scope: {
+        routes: '=',
+        services: '=',
+        portsByRoute: '=',
+        showNodePorts: '=?',
+        // Optional empty message to display when there are no pods.
+        emptyMessage: '=?',
+        // Alternative header text to display in the 'Name' column.
+        customNameHeader: '=?',
+      },
+      templateUrl: 'views/directives/traffic-table.html'
     };
   });
--- a/app/views/browse/service.html
+++ b/app/views/browse/service.html
@@ -62,50 +62,47 @@
                       <dd>{{service.spec.type}}</dd>
                       <dt>IP:</dt>
                       <dd>{{service.spec.clusterIP}}</dd>
+                      <dt>Hostname:</dt>
+                      <dd>
+                        {{service.metadata.name}}.{{service.metadata.namespace}}.svc
+                        <span data-toggle="popover" data-trigger="hover" data-content="{{'This address is only resolvable from within the cluster.'}}" style="cursor: help; padding-left: 5px;">
+                          <span class="pficon pficon-info" aria-hidden="true" data-toggle="tooltip" style="cursor: help;"></span>
+                        </span>
+                      </dd>
+                      <dt ng-if-start="service.spec.externalName">External Hostname:</dt>
+                      <dd ng-if-end>{{service.spec.externalName}}</dd>
                       <dt>Session affinity:</dt>
                       <dd>{{service.spec.sessionAffinity}}</dd>
-                      <dt ng-if="resource.status.loadBalancer.ingress.length">Ingress points</dt>
-                      <dd ng-if="resource.status.loadBalancer.ingress.length">
-                        <span ng-repeat="ingress in resource.status.loadBalancer.ingress"
+                      <dt ng-if-start="service.status.loadBalancer.ingress.length">Ingress Points:</dt>
+                      <dd ng-if-end>
+                        <span ng-repeat="ingress in service.status.loadBalancer.ingress"
                           >{{ingress.ip}}<span ng-if="!$last">, </span></span>
                       </dd>
-                      <dt>Routes:</dt>
-                      <dd>
-                        <span ng-if="!routesForService.length">
+                      <dt ng-if-start="service.spec.externalIPs.length">External IPs:</dt>
+                      <dd ng-if-end>
+                        <span ng-repeat="externalIP in service.spec.externalIPs"
+                          >{{externalIP}}<span ng-if="!$last">, </span></span>
+                      </dd>
+                      <dt ng-if-start="(routesForService | hashSize) == 0">Routes:</dt>
+                      <dd ng-if-end>
+                        <span>
                           <a ng-href="project/{{project.metadata.name}}/create-route?service={{service.metadata.name}}"  ng-if="'routes' | canI : 'create'">Create route</a>
-                        </span>
-                        <span ng-repeat="route in routesForService">
-                            <span ng-if="route | isWebRoute"><a ng-href="{{route | routeWebURL}}">{{route | routeLabel}}</a></span>
-                            <span ng-if="!(route | isWebRoute)">{{route | routeLabel}}</span>
-                            <span ng-show="!$last">, </span>
+                          <span ng-if="!('routes' | canI : 'create')"><em>None</em></span>
                         </span>
                       </dd>
                     </dl>
-                    <div ng-if="service.spec.ports.length" class="table-responsive">
-                      <table class="table table-bordered small service-table">
-                        <thead>
-                          <tr>
-                            <th>Node Port</th>
-                            <th role="presentation"></th>
-                            <th>Service Port</th>
-                            <th role="presentation"></th>
-                            <th>Target Port</th>
-                          </tr>
-                        </thead>
-                        <tbody>
-                          <tr ng-repeat="portMapping in service.spec.ports | orderBy:'port'">
-                            <td>
-                              <span ng-if="portMapping.nodePort">{{portMapping.nodePort}}</span>
-                              <span ng-if="!portMapping.nodePort" class="text-muted">none</span>
-                            </td>
-                            <td role="presentation" class="text-muted">&#8594;</td>
-                            <td>{{portMapping.port}}/{{portMapping.protocol}}
-                                <span ng-if="portMapping.name">({{portMapping.name}})</span></td>
-                            <td role="presentation" class="text-muted">&#8594;</td>
-                            <td>{{portMapping.targetPort}}</td>
-                          </tr>
-                        </tbody>
-                      </table>
+                    <h3>
+                      Traffic
+                    </h3>
+                    <div class="table-responsive">
+                      <traffic-table ports-by-route="portsByRoute" routes="routesForService" services="services" show-node-ports="showNodePorts" custom-name-header="'Route'"></traffic-table>
+                    </div>
+                    <div style="margin:-10px 0 10px 0;">
+                      Learn more about <a ng-href="{{'route-types' | helpLink}}" target="_blank">routes</a> and <a ng-href="{{'services' | helpLink}}" target="_blank">services</a>
+                    </div>
+                    <h3>Pods</h3>
+                    <div class="table-responsive">
+                      <pods-table pods="podsForService" active-pods="podsWithEndpoints" custom-name-header="'Pod'" pod-failure-reasons="podFailureReasons"></pods-table>
                     </div>
                     <annotations annotations="service.metadata.annotations"></annotations>
                   </div>
--- a/app/views/directives/pods-table.html
+++ b/app/views/directives/pods-table.html
@@ -1,19 +1,20 @@
 <table class="table table-bordered table-hover table-mobile">
   <thead>
     <tr>
-      <th>Name</th>
+      <th>{{customNameHeader || 'Name'}}</th>
       <th>Status</th>
       <th>Containers Ready</th>
       <th>Container Restarts</th>
       <th>Age</th>
+      <th ng-if="activePods">Receiving Traffic</th>
     </tr>
   </thead>
   <tbody ng-if="(pods | hashSize) == 0">
-    <tr><td colspan="5"><em>{{emptyMessage || 'No pods to show'}}</em></td></tr>
+    <tr><td colspan="6"><em>{{emptyMessage || 'No pods to show'}}</em></td></tr>
   </tbody>
   <tbody ng-repeat="pod in pods | orderObjectsByDate : true">
     <tr>
-      <td data-title="Name">
+      <td data-title="{{customNameHeader || 'Name'}}">
         <a href="{{pod | navigateResourceURL}}">{{pod.metadata.name}}</a>
         <span ng-if="pod | isDebugPod">
           <i class="fa fa-bug info-popover"
@@ -33,6 +34,18 @@
       <td data-title="Ready">{{pod | numContainersReady}}/{{pod.spec.containers.length}}</td>
       <td data-title="Restarts">{{pod | numContainerRestarts}}</td>
       <td data-title="Age"><span am-time-ago="pod.metadata.creationTimestamp" am-without-suffix="true"></span></td>
+      <td ng-if="activePods" data-title="Receiving Traffic">
+        <span ng-if="activePods[pod.metadata.name]">
+          <span class="fa fa-fw fa-check text-success" aria-hidden="true"></span>
+          <span class="sr-only">Yes</span>
+        </span>
+        <span ng-if="!activePods[pod.metadata.name]">
+           <span data-toggle="popover" data-trigger="hover" data-content="{{podFailureReasons[pod.status.phase] || 'This pod has no endpoints and is not accepting traffic.'}}" style="cursor: help;">
+            <span class="fa fa-fw fa-times text-danger" aria-hidden="true" data-toggle="tooltip" style="cursor: help;"></span>
+            <span class="sr-only">No</span>
+          </span>
+        </span>
+      </td>
     </tr>
   </tbody>
 </table>
--- a/app/views/directives/traffic-table.html
+++ b/app/views/directives/traffic-table.html
@@ -1,0 +1,78 @@
+<!-- Table combining routes and ports for a service -->
+<table class="table table-bordered table-hover table-mobile">
+  <thead>
+    <tr>
+      <th>{{customNameHeader || 'Route'}}<span ng-if="showNodePorts"> / Node Port</span></th>
+      <th role="presentation"></th>
+      <th>Service Port</th>
+      <th role="presentation"></th>
+      <th>Target Port</th>
+      <th>Hostname</th>
+      <th>TLS Termination</th>
+    </tr>
+  </thead>
+  <tbody ng-if="(portsByRoute | hashSize) == 0">
+    <tr><td colspan="7"><em>{{emptyMessage || 'No routes or ports to show'}}</em></td></tr>
+  </tbody>
+  <tbody ng-if="(portsByRoute | hashSize) > 0">
+    <tr ng-repeat-start="(routeName,ports) in portsByRoute" style="display:none;"></tr>
+    <tr ng-repeat="port in ports" ng-if="routeName !== ''">
+      <td data-title="{{customNameHeader || 'Route'}}{{ showNodePorts ? ' / Node Port' : '' }}">
+        <a href="{{routes[routeName] | navigateResourceURL}}">{{routes[routeName].metadata.name}}</a>
+        <route-warnings ng-if="routes[routeName].spec.to.kind !== 'Service' || services"
+                        route="routes[routeName]"
+                        service="services[routes[routeName].spec.to.name]">
+        </route-warnings>
+        <span ng-if="showNodePorts">
+          <span ng-if="port.nodePort"> / {{port.nodePort}}</span>
+        </span>
+      </td>
+      <td role="presentation" class="text-muted">&#8594;</td>
+      <td data-title="Service Port">
+        {{port.port}}/{{port.protocol}}
+        <span ng-if="port.name">({{port.name}})</span>
+      </td>
+      <td role="presentation" class="text-muted">&#8594;</td>
+      <td data-title="Target Port">
+        {{port.targetPort}}
+      </td>
+      <td data-title="Hostname">
+        <span ng-if="(routes[routeName] | isWebRoute)" class="word-break">
+          <a href="{{routes[routeName] | routeWebURL}}" target="_blank">{{routes[routeName] | routeLabel}}</a>
+        </span>
+        <span ng-if="!(routes[routeName] | isWebRoute)" class="word-break">
+          {{routes[routeName] | routeLabel}}
+        </span>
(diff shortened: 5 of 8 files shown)


app/views/browse/service.html
```diff
@@ -64,11 +64,16 @@
...
                       <dd>{{service.spec.sessionAffinity}}</dd>
-                      <dt ng-if="resource.status.loadBalancer.ingress.length">Ingress points</dt>
-                      <dd ng-if="resource.status.loadBalancer.ingress.length">
-                        <span ng-repeat="ingress in resource.status.loadBalancer.ingress"
+                      <dt ng-if="service.status.loadBalancer.ingress.length">Ingress points</dt>
+                      <dd ng-if="service.status.loadBalancer.ingress.length">
+                        <span ng-repeat="ingress in service.status.loadBalancer.ingress"
                           >{{ingress.ip}}<span ng-if="!$last">, </span></span>
                       </dd>
+                      <dt ng-if="service.spec.externalIPs.length">External IPs</dt>
```
juanvallejo: Would we want to add endpoints as well for each service?

app/views/browse/service.html
```diff
@@ -64,11 +64,16 @@
                       <dd>{{service.spec.clusterIP}}</dd>
                       <dt>Session affinity:</dt>
                       <dd>{{service.spec.sessionAffinity}}</dd>
-                      <dt ng-if="resource.status.loadBalancer.ingress.length">Ingress points</dt>
-                      <dd ng-if="resource.status.loadBalancer.ingress.length">
-                        <span ng-repeat="ingress in resource.status.loadBalancer.ingress"
+                      <dt ng-if="service.status.loadBalancer.ingress.length">Ingress points</dt>
```
jwforres: add the : character after the labels for consistency

app/views/browse/service.html
jwforres: does the CLI show them?

app/views/browse/service.html
juanvallejo: @spadgett 
> Do you mind also changing the labels to headline case like "Ingress Points:" while you're in the file?

Sure, will do

jwforres:
what does the output look like when there are endpoints

juanvallejo:
@jwforres 
> what does the output look like when there are endpoints

```
$ oc describe svc/idling-echo
Name:                   idling-echo
Namespace:              default
Labels:                 app=idling-echo
Selector:               app=idling-echo
Type:                   ClusterIP
IP:                     172.30.205.85
Port:                   tcp-echo        8675/TCP
Endpoints:              172.17.0.8:8675,172.17.0.9:8675
Port:                   udp-echo        3090/UDP
Endpoints:              172.17.0.8:3090,172.17.0.9:3090
Session Affinity:       None
No events.
```

jwforres:
which are just the IPs of the pods the service is currently sending traffic
to, correct?

I feel like a better approach in the console is to show the list of pods
the service should be sending traffic to and then part of that should be
confirmation of which ones are currently receiving traffic (which ones
actually have endpoints created)

juanvallejo:
@jwforres @spadgett https://github.com/openshift/origin-web-console/pull/839/commits/7e81227e660762f60a29aa4ef03a9a72c8251389 creates a routes table directive and adds it, along with the existing `pods table` directive, to the `service` page.

When a service has no routes to display, it reverts to the old behavior of showing `create route`:
```html
<span ng-if="!routesForService.length">
  <a ng-href="project/{{project.metadata.name}}/create-route?service={{service.metadata.name}}"  ng-if="'routes' | canI : 'create'">Create route</a>
</span>
```
service-page-no-routes

A service page with routes and pods would look like:
service-page-routes-and-pods

And a service page with pods that have no endpoints looks like this:
service-page-pods-with-warning

jwforres:
It feels like the tables need something to separate them, headers or something.  Right now it just feels like a wall of tables :)

jwforres:
I also don't think I would restrict the width on the Route and Pod tables, the only reason we did it on the ports table before was because it looked odd when it stretched the full width

app/scripts/controllers/service.js
```diff
@@ -60,6 +61,50 @@ angular.module('openshiftConsole')
           }
         );
 
+        watches.push(DataService.watch("services", context, function(services) {
+          $scope.services = services.by("metadata.name");
+        }));
+
+        watches.push(DataService.watch("pods", context, function(pods) {
+          var service = $scope.service;
+          if (!service) {
```
jwforres: dont request the service here. Instead call getPodsForService either way, and that function should just return if the service hasn't been loaded yet.  The serviceResolved method should also call getPodsForService, this way if the service loads after the pods the association will still happen.

app/scripts/controllers/service.js
```diff
@@ -60,6 +61,50 @@ angular.module('openshiftConsole')
...
+              return;
+          }
+          getPodsForService(service, pods);
+        }));
+
+        // receive pods for the current service scope only when the service object is available
+        function getPodsForService(service, pods) {
+          $scope.podsForService = [];
+          var ls = new LabelSelector(service.spec.selector);
+          angular.forEach(pods.by("metadata.name"), function(pod) {
```
jwforres: Instead of looping over the pods yourself, just do 

```
$scope.podsForService = ls.select(pods);
```

app/scripts/controllers/service.js
```diff
@@ -60,6 +61,50 @@ angular.module('openshiftConsole')
...
+            if (ls.matches(pod)) {
+              $scope.podsForService.push(pod);
+            }
+          });
+        }
+
+        watches.push(DataService.watch("endpoints", context, function(endpoints) {
+          $scope.podsWithEndpoints = {};
+          angular.forEach(endpoints.by("metadata.name"), function(endpoint) {
```
jwforres: endpoints.by("metadata.name") returns a map of objects indexed by the metadata name, so iterating over that map is wasting cycles, you can do a direct lookup in that map using $routeParams.service.

To follow our code style use _.each when looping over either arrays or hashes.

You don't need to do undefined checks when trying to find JSON subpaths if you use _.get, see the usage below to get targetRef.Kind

```
var svcEndpoint = endpoints.by("metadata.name")[$routeParams.service];
if (svcEndpoint) {
  _.each(svcEndpoint.subsets, function(subset) {
    _.each(subset.addresses, function(address) {
      if (_.get(address, "targetRef.Kind") === "Pod") {
        $scope.podsWithEndpoints[address.targetRef.name] = true;
      }
    });
  });
}
```

app/views/browse/service.html
```diff
@@ -64,23 +64,29 @@
...
-                      <dd ng-if="resource.status.loadBalancer.ingress.length">
-                        <span ng-repeat="ingress in resource.status.loadBalancer.ingress"
+                      <dt ng-if="service.status.loadBalancer.ingress.length">Ingress Points:</dt>
+                      <dd ng-if="service.status.loadBalancer.ingress.length">
+                        <span ng-repeat="ingress in service.status.loadBalancer.ingress"
                           >{{ingress.ip}}<span ng-if="!$last">, </span></span>
                       </dd>
-                      <dt>Routes:</dt>
-                      <dd>
+                      <dt ng-if="service.spec.externalIPs.length">External IPs:</dt>
```
jwforres: You should use ng-if-start and ng-if-end when you have consecutive sibling nodes checking the same thing.  We have some old code that didn't do this but any new code should.  Fix up the other instances in here like ingress above and routes below

ng-if-start="service.spec.externalIPs.length"

app/views/browse/service.html
```diff
@@ -64,23 +64,29 @@
...
+                      <dd ng-if="service.spec.externalIPs.length">
```
jwforres: ng-if-end

app/views/browse/service.html
```diff
@@ -64,23 +64,29 @@
...
-                            <span ng-if="!(route | isWebRoute)">{{route | routeLabel}}</span>
-                            <span ng-show="!$last">, </span>
-                        </span>
                       </dd>
                     </dl>
+                    <div ng-if="routesForService.length" class="table-responsive small service-table">
+                      <routes-table routes="routesForService" services="services" custom-name-header="routesForService[0].kind"></routes-table>
+                    </div>
+                    <div ng-if="podsForService.length" class="table-responsive small service-table">
+                      <pods-table pods="podsForService" active-pods="podsWithEndpoints" custom-name-header="podsForService[0].kind"></pods-table>
```
jwforres: switching to LabelSelector.select is going to return a map, I would just explicitly set custom-name-header to "Pod" and don't worry about trying to use the kind here

app/views/directives/pods-table.html
```diff
@@ -15,6 +15,10 @@
     <tr>
       <td data-title="Name">
```
jwforres: also need customNameHeader here, this is what shows up at mobile screen widths

app/views/directives/routes-table.html
```diff
@@ -0,0 +1,58 @@
...
+    <th>Target Port</th>
+    <th>TLS Termination</th>
+  </tr>
+</thead>
+<tbody ng-if="(routes | hashSize) == 0">
+  <tr><td colspan="5"><em>{{emptyMessage || 'No routes to show'}}</em></td></tr>
+</tbody>
+<tbody ng-repeat="route in routes | orderObjectsByDate : true">
+  <tr>
+    <td data-title="Name">
```
jwforres: customNameHeader

juanvallejo:
@jwforres Thanks for the feedback! Review comments addressed:

service-page-with-h3-headers

juanvallejo:
@jwforres Here is the pods table updated to display a `Receiving Traffic` column instead of the warning tooltip

service-page-with-receiving-traffic-column

juanvallejo:
@jwforres @spadgett Here is what the page would look like with all three tables of the same width, and a black-and-white check when a pod is receiving traffic:

service-page-with-check-and-full-port-table

spadgett:
I prefer the 100% width of the ports table, but we might need to center the `td` content so the arrows are centered. @jwforres ?

Suggest just a check or x for receiving traffic without the "yes/no" text.

jwforres:
@juanvallejo use the text-success class (green) on the check and the text-danger class (red) on the x

jwforres:
If I see I have pods that aren't receiving traffic I probably want to know why.  Is it worth having a help tip next to Receiving Traffic that suggests reasons your pod may not be getting any traffic yet?

@spadgett ?

jwforres:
I dunno I actually think i like the ports table shorter, it just doesn't have the content to be full width to me.

But I'm starting to wonder if maybe Routes and Pods should be in tabs instead

Just throwing ideas out, but what if we tied in the words Incoming and Outgoing somehow?  I feel like there isn't enough context here to someone that doesn't understand how the networking works.  How do they know that the Routes are saying what hostnames are pushing traffic to this service and how do they know that services send traffic to pods.  I feel like this is an opportunity for us to make this page better.

cc @ajacobs21e

juanvallejo:
@jwforres  
> Just throwing ideas out, but what if we tied in the words Incoming and Outgoing somehow? I feel like there isn't enough context here to someone that doesn't understand how the networking works. How do they know that the Routes are saying what hostnames are pushing traffic to this service and how do they know that services send traffic to pods.

Maybe we could have cards similar to the overview page to help display this? 
An overall card represents a service, with each of its routes displayed as sets (similar to how container information is displayed on the overview) that contain their hostname / other info with arrows pointing towards the pods donut (also similar to how they are displayed on the overview page, but without the "scale up / scale down" arrows), thoughts?

juanvallejo:
@jwforres @spadgett Page with tabs for each table

service-page-tabbed-tables-ports

service-page-tabbed-tables-routes

service-page-tabbed-tables-single-tab

spadgett:
I like the tabs now that I've seen it.

Should it be "Routes | Ports | Pods" since that is how the traffic flows?

Are tabs missing from the last screenshot because there are none to show? I'd probably keep the tab and have a message like no routes. Maybe we should put a number with the count in the tab title like we do on the membership page.

spadgett:
> If I see I have pods that aren't receiving traffic I probably want to know why. Is it worth having a help tip next to Receiving Traffic that suggests reasons your pod may not be getting any traffic yet?

It would be especially helpful if we can say exactly why, for instance when status is `ContainerCreating` or the container hasn't passed the readiness check. Then have a generic fallback explaining common reasons if we don't know?

juanvallejo:
@spadgett Thanks for the feedback. Here is the same page with each tab displaying the number of resources in each. If a tab has none to show, it displays a warning.

service-page-tabbed-tables-number

Empty table

service-page-with-tabbed-tables-missing-resources

Custom messages if pod not receiving traffic

service-page-custom-traffic-warning

ajacobs21e:
> I dunno I actually think i like the ports table shorter, it just doesn't have the content to be full width to me.

Agreed

> But I'm starting to wonder if maybe Routes and Pods should be in tabs instead

Tabs within the detail page or tabs between details and events? I am not sure if I like the tabs within tabs. How many rows would we expect each table to have?

> Just throwing ideas out, but what if we tied in the words Incoming and Outgoing somehow? I feel like there isn't enough context here to someone that doesn't understand how the networking works. How do they know that the Routes are saying what hostnames are pushing traffic to this service and how do they know that services send traffic to pods. I feel like this is an opportunity for us to make this page better.

As someone that doesn't fully understand networking, this could be helpful. Which word goes where? :)

@jwforres @juanvallejo

ajacobs21e:
How are pods, services, routes, and ports related? Maybe understanding those connections will help me suggest a design.

Would a user want to / need to see more than one table at a time to compare data and/or follow the path?

spadgett:
> How are pods, services, routes, and ports related? Maybe understanding those connections will help me suggest a design.

Traffic flows through the system from routes -> services -> pods. Services and pods can each have multiple ports. Each service port maps to a pod port, which is shown in the ports table.

ajacobs21e:
@spadgett how does a node port relate? 
> Traffic flows through the system from routes -> services -> pods. Services and pods can each have multiple ports. Each service port maps to a pod port, which is shown in the ports table.

Can a route point to more than one service port?

spadgett:
@ajacobs21e 

> how does a node port relate?

Node port is a special case that doesn't use routes.

https://docs.openshift.org/latest/dev_guide/getting_traffic_into_cluster.html#using-nodeport

> Can a route point to more than one service port?

No. A route can only point to one service port.

juanvallejo:
@spadgett @ajacobs21e 

Combined routes and ports table (added support for this in the `routes-table` directive)

service-page-no-tabs-combined-ports-routes

Pods table is now in a tab next to `Details`

service-page-pods-table-in-tab-next-to-details

spadgett:
If we pull the pods table into a separate (top-level) tab, we should probably do it for replication controllers and replica sets as well, which have a pods table. Probably should remove the number in that case as well since we don't have it for events.

Route / Node is not the right column header. It should be Node Port. I think it probably should be a separate column still, maybe hidden unless it's being used? @jwforres opinion?

Are the arrows really better in the table header than the table body? I find it easier to understand in the body.

juanvallejo:
@spadgett 
> Route / Node is not the right column header. It should be Node Port. I think it probably should be a separate column still, maybe hidden unless it's being used?

Added `Node Port` column and left the one with route info as the first one (although maybe we could just use the already-existing `Routes To` column and move the warnings there?)

service-page-with-node-port-included

`Node Port` column is hidden when no ports have a `nodePort` value.

service-page-nodeports-hidden

ajacobs21e:
With only two tables it's ok to leave both tables on the details page with headers "Traffic" and "Pods"

juanvallejo:
@spadgett @ajacobs21e @jwforres 

Thanks again for the feedback. Commit https://github.com/openshift/origin-web-console/pull/839/commits/c3f8c9852f0a7c14c2b7ce8b28570502709de454 makes a new `traffic-table` directive that combines routes and ports for a service. Pods are also shown in the same `Details` tab, rather than in their own tab.

services-page-dns-to-hostname-learn-more-link-traffic-table

A `Learn more` link has been added beside the `Traffic` header that links to `Route Types` [documentation](https://docs.openshift.org/latest/architecture/core_concepts/routes.html#route-types).

service-page-learn-more-link

Routes and Node Ports are shown in the same column. If there are no node ports, they are not shown at all.

service-page-with-route-and-nodeport


If a service has no routes, but does have ports, the same `traffic-table` directive is re-used, with any missing route information set to `none`.

service-page-with-ports-and-no-routes

service-page-no-routes-or-ports

jwforres:
I think part of what we were saying is that you should not even see "Node
Port" in the header if there are no node ports.

juanvallejo:
@jwforres 
> I think part of what we were saying is that you should not even see "Node
Port" in the header if there are no node ports.

Thanks, updated the last commit to hide `Node Port` from the header if there are none to show.

service-page-no-nodeport-header

app/views/browse/service.html
```diff
@@ -62,50 +62,48 @@
                       <dd>{{service.spec.type}}</dd>
                       <dt>IP:</dt>
                       <dd>{{service.spec.clusterIP}}</dd>
+                      <dt ng-if-start="service.spec.externalName">External Name</dt>
+                      <dd ng-if-end>{{service.spec.externalName}}</dd>
```
juanvallejo: @jwforres Added this field to display a service's external name (if any) for the new types of services. https://tcms-openshift.rhcloud.com/case/6047/

PTAL

juanvallejo: @jwforres Thanks, updated it to "External Hostname:", image below:

service-page-fields-external-name

juanvallejo:
@jwforres thanks for the feedback. The `traffic` and `pods` tables no longer have the `small` class name

services-page-bigger-tables-receiving-traffic-tweaks

The `Receiving traffic` icon is now an `x` when a pod is not receiving traffic

services-page-receiving-traffic-max

and there is now a footnote that links to the [routes](https://docs.openshift.org/latest/architecture/core_concepts/routes.html#route-types) docs, and [services](https://docs.openshift.org/latest/architecture/core_concepts/pods_and_services.html#services) docs.

jwforres:
the footnote should be a lot closer to the bottom of the table.  for the
icons are you using fa-fw class to make sure they are a fixed width? the x
and check looks like different sizes.

juanvallejo:
@jwforres added support for a "footnote" to the traffic table that decreases its bottom margin to bring it closer:

services-page-traffic-table-footer

Added the `fa-fw` class to both the `check` and the `x` icons in the pod's table, but I still think they seem to have different sizes:

services-page-pods-fw

app/views/directives/pods-table.html
```diff
@@ -39,11 +39,11 @@
       <td data-title="Age"><span am-time-ago="pod.metadata.creationTimestamp" am-without-suffix="true"></span></td>
       <td ng-if="activePods" data-title="Receiving Traffic">
         <span ng-if="activePods[pod.metadata.name]">
-          <status-icon status="'Succeeded'" disable-animation></status-icon>
+          <status-icon fixed-width="true" status="'Succeeded'" disable-animation></status-icon>
         </span>
         <span ng-if="!activePods[pod.metadata.name]">
            <span data-toggle="popover" data-trigger="hover" data-content="{{podFailureReasons[pod.status.phase] || 'This pod has no endpoints and is not accepting traffic.'}}" style="cursor: help; padding-left: 5px;">
```
jwforres: i think your problem is the padding-left that is on this span, thats what is offsetting this icon compared to the other

jwforres: i dont think the padding is needed, try just removing it

app/scripts/controllers/service.js
```diff
@@ -27,11 +29,48 @@ angular.module('openshiftConsole')
...
+    // receives routes for the current service and maps a service port to each route name
+    var getPortsForRoutes = function(service, routes) {
+      if (!service || !routes) {
+        return;
+      }
+
+      $scope.portsForRoutes = {};
+      var orderedPorts = $filter("orderBy")(service.spec.ports, "port");
+
+      for(var i = 0; i < routes.length; i++) {
```
jwforres: we don't use for loops, we use _.each for all iteration
```
_.each(routes, function(route) {

});
```

app/scripts/controllers/service.js
```diff
@@ -27,11 +29,48 @@ angular.module('openshiftConsole')
...
+        $scope.portsForRoutes[routes[i].metadata.name] = orderedPorts[i];
```
jwforres: This block doesn't seem right to me.  A route can specify what port it uses.  You might have multiple routes targeting the same ports on the service.  You might have some service ports that have no route associated with it.

app/scripts/controllers/service.js
```diff
@@ -27,11 +29,48 @@ angular.module('openshiftConsole')
...
+      }
+
+      var ls = new LabelSelector(service.spec.selector);
+      $scope.podsForService = ls.select(pods);
+    };
     var serviceResolved = function(service, action) {
       $scope.loaded = true;
       $scope.service = service;
+
+      getPodsForService(service, $scope.pods);
```
jwforres: service and pods are already on the scope, you don't need to pass them into the functions

app/scripts/controllers/service.js
```diff
@@ -60,6 +99,36 @@ angular.module('openshiftConsole')
           }
         );
 
+        watches.push(DataService.watch("services", context, function(services) {
+          $scope.services = services.by("metadata.name");
+        }));
+
+        watches.push(DataService.watch("pods", context, function(pods) {
+          $scope.pods = [];
```
jwforres: there shouldn't be any reason for converting the pod map to an array, you also probably have no reason to have all pods on the scope, you only care about the pods for the service from within the view.  You can use a local variable to the controller to keep track of the full set of pods.

app/scripts/controllers/service.js
```diff
@@ -60,6 +99,36 @@ angular.module('openshiftConsole')
...
+          angular.forEach(pods.by("metadata.name"), function(pod) {
+            $scope.pods.push(pod);
+          });
+          getPodsForService($scope.service, $scope.pods);
+        }));
+
+        watches.push(DataService.watch("endpoints", context, function(endpoints) {
+          $scope.podsWithEndpoints = {};
+          var svcEndpoint  = endpoints.by("metadata.name");
```
jwforres: i would probably just make this a one-liner instead of initializing a variable as a map and then turning it into a singleton
```
var svcEndpoint = endpoints.by("metadata.name")[$routeParams.service];
```

app/scripts/directives/resources.js
```diff
@@ -165,9 +165,46 @@ angular.module('openshiftConsole')
...
+        emptyMessage: '=?',
+        // Alternative header text to display in the 'Name' column.
+        customNameHeader: '=?',
+        // Optional map of explanations or warnings for each phase of a pod
+        podFailureReasons: '=?'
       },
       templateUrl: 'views/directives/pods-table.html'
     };
+  })
+  .directive('routesTable', function() {
```
jwforres: you aren't actually using the routesTable anywhere except the routes page right? i would just remove the directive and revert routes.html

app/views/browse/service.html
```diff
@@ -62,50 +62,53 @@
                       <dd>{{service.spec.type}}</dd>
                       <dt>IP:</dt>
                       <dd>{{service.spec.clusterIP}}</dd>
+                      <dt>Hostname:</dt>
+                      <dd>
+                        {{service.metadata.name + '.' + service.metadata.namespace + '.svc'}}
```
jwforres: Don't concatenate when it isn't necessary
```
{{service.metadata.name}}.{{service.metadata.namespace}}.svc
```

app/views/browse/service.html
```diff
@@ -62,50 +62,53 @@
...
                           >{{ingress.ip}}<span ng-if="!$last">, </span></span>
                       </dd>
-                      <dt>Routes:</dt>
-                      <dd>
+                      <dt ng-if-start="service.spec.externalIPs.length">External IPs:</dt>
+                      <dd ng-if-end>
+                        <span ng-repeat="externalIP in service.spec.externalIPs"
+                          >{{externalIP}}<span ng-if="!$last">, </span></span>
+                      </dd>
+                      <dt ng-if="!routesForService.length">Routes:</dt>
```
jwforres: switch to ng-if-start

app/views/browse/service.html
```diff
@@ -62,50 +62,53 @@
...
+                      <dd ng-if="!routesForService.length">
```
jwforres: ng-if-end

app/views/browse/service.html
```diff
@@ -62,50 +62,53 @@
...
                         <span ng-if="!routesForService.length">
```
jwforres: this if check is unnecessary, the one above it covers this

app/views/browse/service.html
```diff
@@ -62,50 +62,53 @@
...
                           <a ng-href="project/{{project.metadata.name}}/create-route?service={{service.metadata.name}}"  ng-if="'routes' | canI : 'create'">Create route</a>
```
jwforres: if user doesn't have access to create routes, this should say `<em>none</em>`

app/views/directives/traffic-table.html
```diff
@@ -0,0 +1,84 @@
+<!-- Table combining routes and ports for a service -->
+<table class="table table-bordered table-hover table-mobile" style="{{ hasFooter ? 'margin-bottom:5px;' : ''}}">
```
jwforres: when i saw hasFooter i expected the footer to actually be part of the directive, not to just change the styles of the table.  i would either make the footer part of the directive, or you can use negative margins on the footer.

app/views/directives/traffic-table.html
```diff
@@ -0,0 +1,84 @@
...
+<thead>
+  <tr>
+    <th ng-if="!showNodePorts">{{customNameHeader || 'Route'}}</th>
+    <th ng-if="showNodePorts">{{customNameHeader || 'Route'}} / Node Port</th>
```
jwforres: I'm not sure if this will screw up a screen reader making these different `<th>` elements, when the content can change dynamically, instead I would probably do

```
<th>{{customNameHeader || 'Route'}}<span ng-if="showNodePorts"> / Node Port</span></th>
```

app/views/directives/traffic-table.html
```diff
@@ -0,0 +1,84 @@
...
+    <th role="presentation"></th>
+    <th>Target Port</th>
+    <th>Hostname</th>
+    <th>TLS Termination</th>
+  </tr>
+</thead>
+<tbody ng-if="(routes | hashSize) == 0 && (ports | hashSize) == 0">
+  <tr><td colspan="7"><em>{{emptyMessage || 'No routes or ports to show'}}</em></td></tr>
+</tbody>
+<tbody ng-if="(routes | hashSize) > 0">
```
jwforres: the route / port iteration and association logic in here isn't right, lets sync up later

app/views/directives/traffic-table.html
```diff
@@ -0,0 +1,84 @@
...
+  <tr ng-repeat="route in routes | orderObjectsByDate : true">
+    <td data-title="{{ customNameHeader || 'Route' }}">
```
jwforres: this data-title needs to match the column header and include / Node Port when the other one does.

juanvallejo:
@jwforres thanks for your help and feedback, review comments addressed.

Traffic table now shows each route's information for every service port that it reaches.

service-page-all-ports-for-routes

Ports that have no routes linked to them are also displayed

service-page-ports-without-route

"No traffic" icon in the pod's table is now aligned properly

service-page-no-traffic-icon-aligned

app/scripts/controllers/service.js
```diff
@@ -27,11 +28,60 @@ angular.module('openshiftConsole')
...
+      if(!$scope.service) {
+        return;
+      }
+
+      $scope.portsByRoute = {};
+
+      _.each($scope.service.spec.ports, function(port) {
+        var reachedByRoute = false;
+        _.each($scope.routesForService, function(route) {
+          if(!route.spec.port || route.spec.port.targetPort === port.name ||
```
jwforres: @spadgett you are more familiar with the route / port stuff, can you review this section

app/scripts/controllers/service.js
```diff
@@ -27,11 +28,60 @@ angular.module('openshiftConsole')
...
+            route.spec.port.targetPort === port.targetPort) {
+            if(!$scope.portsByRoute[route.metadata.name]) {
```
jwforres: we usually do this if block as a one-liner

```
$scope.portsByRoute[route.metadata.name] = $scope.portsByRoute[route.metadata.name] || [];
```

app/scripts/controllers/service.js
```diff
@@ -27,11 +28,60 @@ angular.module('openshiftConsole')
...
+              $scope.portsByRoute[route.metadata.name] = [];
+            }
+            $scope.portsByRoute[route.metadata.name].push(port);
+            reachedByRoute = true;
+          }
+        });
+
+        if(!reachedByRoute) {
+          if(!$scope.portsByRoute['']) {
```
jwforres:
```
$scope.portsByRoute[''] = $scope.portsByRoute[''] || [];
```

app/scripts/controllers/service.js
```diff
@@ -60,15 +110,41 @@ angular.module('openshiftConsole')
...
+        }));
+
+        watches.push(DataService.watch("pods", context, function(pods) {
+          allPods = pods.by("metadata.name");
+          getPodsForService();
+        }));
+
+        watches.push(DataService.watch("endpoints", context, function(endpoints) {
+          $scope.podsWithEndpoints = {};
+          var svcEndpoint  = endpoints.by("metadata.name")[$routeParams.service];
```
jwforres: its a nit but you have an extra whitespace after svcEndpoint

app/views/directives/pods-table.html
```diff
@@ -33,6 +34,18 @@
       <td data-title="Ready">{{pod | numContainersReady}}/{{pod.spec.containers.length}}</td>
       <td data-title="Restarts">{{pod | numContainerRestarts}}</td>
       <td data-title="Age"><span am-time-ago="pod.metadata.creationTimestamp" am-without-suffix="true"></span></td>
+      <td ng-if="activePods" data-title="Receiving Traffic">
+        <span ng-if="activePods[pod.metadata.name]">
+          <span class="fa fa-check text-success" aria-hidden="true"></span>
```
jwforres: missing fa-fw

app/views/directives/traffic-table.html
```diff
@@ -0,0 +1,75 @@
...
+      </td>
+      <!-- Use shorter Termination title for table-mobile to avoid overlapping text. -->
+      <td data-title="Termination">
+        {{routes[routeName].spec.tls.termination}}
+        <span ng-if="!routes[routeName].spec.tls.termination">&nbsp;</span>
+      </td>
+    </tr>
+    <tr ng-repeat-end style="display:none;"></tr>
+    <tr ng-repeat="port in portsByRoute['']">
+      <td data-title="{{customNameHeader || 'Route'}}"><span class="text-muted">none</span></td>
```
jwforres: you are missing node port for this case

app/scripts/controllers/service.js
spadgett: It looks correct.

juanvallejo:
@jwforres Thanks, review comments addressed.

NodePort case is now handled when there are 0 or >0 routes

service-page-nodeport

service-page-nodeport-routes

jwforres:
@juanvallejo looks good now, can you squash your commits

juanvallejo:
@jwforres 
> looks good now, can you squash your commits

Thanks! Done

jwforres:
[merge]

PR #845, opened 2016-11-07
Resolves misc. card bugs

### LINKED ISSUE #834: Catalog word breaks awkward in some browsers
Because we use the `work-break-all`, some of the word breaks are awkward in IE and Firefox. For instance, "Ephemeral" is broken mid-word. I'd like to see if there is a fix for #828 without this side effect.

dashboard

/cc @jwforres @sg00dwin @rhamilto

rhamilto:
* instead of awkwardly wrapping titles (Firefox and IE), use an ellipses on strings that
don’t break
* word-wrap descriptions so that URLs don’t result in broken cards (Chrome)

Fixes #834

Before:

screen shot 2016-11-07 at 3 56 06 pm

screen shot 2016-11-07 at 3 57 03 pm

After:

screen shot 2016-11-07 at 3 55 28 pm

screen shot 2016-11-07 at 3 58 03 pm

@spadgett, PTAL.  cc: @sg00dwin

### DIFF
--- a/app/styles/_cards.less
+++ b/app/styles/_cards.less
@@ -36,17 +36,15 @@
   &:before {
     display: none;
   }
-  [class^="col"] {
+  [class^="col"], .card-pf {
     display: flex;
-  }
-  .card-pf {
-    display: flex;
-    flex-basis: 100%; // IE 11 overflow bug fix
     flex-direction: column;
+  }
+  .card-pf, .card-pf-body {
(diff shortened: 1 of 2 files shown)


spadgett:
@jwforres opinion for 1.4?

jwforres:
[merge] yes since our OOTB templates that have URLs are affected

PR #848, opened 2016-11-08
Remove top padding from no grouped services message

spadgett:
@jwforres After some testing, I opted to just remove the top padding. PTAL


jwforres:
[merge]

PR #849, opened 2016-11-08
Fix add storage links

spadgett:
* Use correct link for pod owned by a deployment config
* Update deployment when a replica set is owned by a deployment
  (in page link, the actions menu was already doing this)
* Update breadcrumb on add storage page to match action label

@jwforres


jwforres:
[merge]

PR #853, opened 2016-11-09
Add support for wildcard subdomains

spadgett:
Add web console support for creating wildcard routes. Show wildcard routes as `*.my.subdomain` on the overview and routes pages. Warn when the router has not admitted the wildcard policy.

openshift_web_console

screen shot 2016-11-09 at 3 11 06 pm

@jwforres @knobunc @rettori

### DIFF
--- a/app/scripts/constants.js
+++ b/app/scripts/constants.js
@@ -52,6 +52,10 @@ window.OPENSHIFT_CONSTANTS = {
   // true indicates that deployment metrics should be disabled on the web console overview
   DISABLE_OVERVIEW_METRICS: false,
 
+  // true indicates that none of the routers support wildcard subdomains and
+  // removes the option from the route creation form.
+  DISABLE_WILDCARD_ROUTES: true,
+
   // This blacklist hides certain kinds from the "Other Resources" page because they are unpersisted, disallowed for most end users, or not supported by openshift but exist in kubernetes
   AVAILABLE_KINDS_BLACKLIST: [
       // These are k8s kinds that are not supported in the current release of OpenShift
--- a/app/scripts/services/applicationGenerator.js
+++ b/app/scripts/services/applicationGenerator.js
@@ -126,12 +126,19 @@ angular.module("openshiftConsole")
           to: {
             kind: "Service",
             name: serviceName
-          }
+          },
+          wildcardPolicy: 'None'
         }
       };
 
-      if (input.routing.host) {
-        route.spec.host = input.routing.host;
+      var host = input.routing.host;
+      if (host) {
+        if (host.startsWith('*.')) {
+          route.spec.wildcardPolicy = 'Subdomain';
+          route.spec.host = 'wildcard' + host.substring(1);
+        } else {
+          route.spec.host = host;
+        }
       }
 
       if (input.routing.path) {
--- a/app/scripts/services/routes.js
+++ b/app/scripts/services/routes.js
@@ -63,6 +63,7 @@ angular.module("openshiftConsole")
     };
 
     var addIngressWarnings = function(route, warnings) {
+      var wildcardPolicy = _.get(route, 'spec.wildcardPolicy');
       angular.forEach(route.status.ingress, function(ingress) {
         var condition = _.find(ingress.conditions, { type: "Admitted", status: "False" });
         if (condition) {
@@ -71,6 +72,11 @@ angular.module("openshiftConsole")
             message += " Reason: " + (condition.message || condition.reason) + '.';
           }
           warnings.push(message);
+        }
+
+        // This message only displays with old router images that are not aware of `wildcardPolicy`.
+        if (!condition && wildcardPolicy === 'Subdomain' && ingress.wildcardPolicy !== wildcardPolicy) {
+          warnings.push('Router "' + ingress.routerName + '" does not support wildcard subdomains. Your route will only be available at host ' + ingress.host + '.');
         }
       });
     };
@@ -123,6 +129,12 @@ angular.module("openshiftConsole")
       return _.groupBy(routes, 'spec.to.name');
     };
 
+    // For host "foo.example.com" return "example.com"
+    var getSubdomain = function(route) {
+      var hostname = _.get(route, 'spec.host', '');
+      return hostname.replace(/^[a-z0-9]([-a-z0-9]*[a-z0-9])\./, '');
+    };
+
     return {
       // Gets warnings about a route.
       //
@@ -151,6 +163,7 @@ angular.module("openshiftConsole")
 
       getServicePortForRoute: getServicePortForRoute,
       getPreferredDisplayRoute: getPreferredDisplayRoute,
-      groupByService: groupByService
+      groupByService: groupByService,
+      getSubdomain: getSubdomain
     };
   });
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -449,12 +449,6 @@ label.checkbox {
   right: -3px;
 }
 
-.create-route-icon,
-.create-storage-icon {
-  padding-top: 0;
-  text-align: right;
-}
-
 .about,
 .command-line {
   .about-icon {
--- a/test/spec/services/applicationGeneratorSpec.js
+++ b/test/spec/services/applicationGeneratorSpec.js
@@ -186,6 +186,52 @@ describe("ApplicationGenerator", function(){
             name: "theServiceName"
           },
           host: "www.example.com",
+          wildcardPolicy: "None",
+          path: "/test",
+          port: {
+            targetPort: 'tcp-80'
+          },
+          tls: {
+            termination: "edge",
+            insecureEdgeTerminationPolicy: "Redirect",
+            certificate: "dummy-cert",
+            key: "dummy-key",
+            caCertificate: "dummy-ca-cert"
+          }
+        }
+      });
+    });
+
+    it("should generate a wildcard route when wildcard subdomains are set", function(){
+      // Add the same labels, annotations, and targetPort as application generator `generate()`
+      var routeInput = angular.copy(inputTemplate);
+      routeInput.labels.app = routeInput.name;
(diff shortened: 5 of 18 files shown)


spadgett:
openshift_web_console

app/scripts/constants.js
```diff
@@ -52,6 +52,10 @@ window.OPENSHIFT_CONSTANTS = {
   // true indicates that deployment metrics should be disabled on the web console overview
   DISABLE_OVERVIEW_METRICS: false,
 
+  // true indicates that none of the routers support wildcard subdomains and
+  // removes the option from the route creation form.
+  DISABLE_WILDCARD_SUBDOMAINS: false,
```
jwforres: @knobunc is it better to assume most routers will support this feature by default, or to assume most will not have it?

app/scripts/services/applicationGenerator.js
```diff
@@ -126,12 +126,18 @@ angular.module("openshiftConsole")
...
+          },
+          wildcardPolicy: input.routing.wildcardPolicy
         }
       };
 
-      if (input.routing.host) {
-        route.spec.host = input.routing.host;
+      var host = input.routing.host;
+      if (route.spec.wildcardPolicy === 'Subdomain') {
+        host = 'www.' + input.routing.subdomain;
```
jwforres: @knobunc are you ok with 'www' as the default we inject here under the covers

app/scripts/services/routes.js
```diff
@@ -123,6 +128,12 @@ angular.module("openshiftConsole")
       return _.groupBy(routes, 'spec.to.name');
     };
 
+    // For host "foo.example.com" return "example.com"
+    var getSubdomain = function(route) {
+      var hostname = _.get(route, 'spec.host', '');
+      return hostname.replace(/^[a-z0-9]([-a-z0-9]*[a-z0-9])./, '');
```
jwforres: shouldn't that be a `\.` and not just a `.` ?

app/styles/_core.less
```diff
@@ -449,12 +449,6 @@ label.checkbox {
   right: -3px;
 }
 
-.create-route-icon,
```
jwforres: guessing these werent being used anywhere?

spadgett: > guessing these werent being used anywhere?

Right

knobunc:
@ramr It would be helpful if you took a look over the UI for this change.  Thanks

app/scripts/services/applicationGenerator.js
knobunc: I think you need to let them set it, but then let them set the wildcard flag separately.

I realized that we will expose those names when a router doesn't support wildcards, so we really do have to expose the full names to the end user :-(

spadgett:
Updated based on feedback from design review. /cc @rhamilto @ajacobs21e 

openshift_web_console

If the user types `*.example.com`, we'll use `www.example.com` for host and set `wildcardPolicy: 'Subdomain'`.

@knobunc I understand this isn't ideal if the router doesn't support wildcards. We were wondering if the entire route shouldn't be rejected if the wildcard policy can't be honored. It seems odd that the route is still admitted without the wildcards.

app/scripts/constants.js
spadgett: > I vote that we keep it false for now.

Note that the setting disables web console support for wildcards. So leaving it false will *enable* the feature. Just confirming that's what you mean :)

ramr:
@spadgett if the router runs without wildcard support (`ROUTER_ALLOW_WILDCARD_ROUTES=false` or unset), then the route will get an error - the status would reflect that accordingly. This is what returns the error back: 
    https://github.com/openshift/origin/blob/master/pkg/cmd/infra/router/router.go#L140

ramr:
@spadgett  @knobunc  will review this in depth sometime later today.

spadgett:
@ramr This is the YAML I'm seeing. The route is admitted, but wildcard policy doesn't match?

```yaml
apiVersion: v1
kind: Route
metadata:
  name: myroute
  namespace: demo
  selfLink: /oapi/v1/namespaces/demo/routes/myroute
  uid: e76ac1c9-a778-11e6-b359-080027242396
  resourceVersion: '47863'
  creationTimestamp: '2016-11-10T19:07:20Z'
  labels:
    app: node-ex
spec:
  host: www.mydomain.com
  to:
    kind: Service
    name: node-ex
    weight: 100
  port:
    targetPort: 8080-tcp
  wildcardPolicy: Subdomain
status:
  ingress:
    - host: www.mydomain.com
      routerName: router
      conditions:
        - type: Admitted
          status: 'True'
          lastTransitionTime: '2016-11-10T19:07:20Z'
      wildcardPolicy: None
```

ramr:
@spadgett  looks like you have an older router image.  The old router would admit the route in (since it doesn't understand wildcards). There was a discussion on this on: https://github.com/openshift/origin/pull/11201/#issuecomment-255830602 

Anyway, the crux of that discussion is older routers would still admit in the route since they don't understand wildcards. That also allows routes to be backward compatible. For newer routers, since the feature is opt-in if you create a wildcard route and if the policy is not supported - it would reject it.

app/scripts/constants.js
ramr: ;^)   - btw, called `ROUTER_ALLOW_WILDCARD_ROUTES` (defaults to false) on the router side to keep you on your toes ... double negatives.

app/scripts/services/applicationGenerator.js
```diff
@@ -126,12 +126,19 @@ angular.module("openshiftConsole")
...
         }
       };
 
-      if (input.routing.host) {
-        route.spec.host = input.routing.host;
+      var host = input.routing.host;
+      if (host) {
+        if (host.startsWith('*.')) {
+          route.spec.wildcardPolicy = 'Subdomain';
+          route.spec.host = 'www' + host.substring(1);
```
ramr: Just a thought, might be better to use some name other than `www``. We do allow for multiple routes (wildcard + non)to coexist in a single namespace ... and www is more than likely to be a name used there for non-wildcard hosts.

test/spec/services/routesSpec.js
```diff
@@ -132,7 +133,25 @@ describe("RoutesService", function(){
...
+        routerName: 'foo',
+        conditions: [{
+          type: "Admitted",
+          status: "True",
+          lastTransitionTime: "2016-02-17T17:18:51Z"
+        }],
+        wildcardPolicy: 'None'
+      }];
+      var warnings = RoutesService.getRouteWarnings(route, serviceTemplate);
+      expect(warnings).toEqual(['Router "foo" does not support wildcard subdomains. Your route will only be available at host www.example.com.']);
```
ramr: The error message may be a bit different so you may want to check w/ older/newer router images.

test/spec/services/applicationGeneratorSpec.js
```diff
@@ -186,6 +186,52 @@ describe("ApplicationGenerator", function(){
             name: "theServiceName"
           },
           host: "www.example.com",
+          wildcardPolicy: "None",
```
spadgett: This is an object the web console is generating, so the value will always be set.

test/spec/services/routesSpec.js
spadgett: This particular error message is from the web console :)

app/scripts/services/routes.js
```diff
@@ -72,6 +73,10 @@ angular.module("openshiftConsole")
           }
           warnings.push(message);
         }
+
+        if (wildcardPolicy === 'Subdomain' && ingress.wildcardPolicy !== wildcardPolicy) {
+          warnings.push('Router "' + ingress.routerName + '" does not support wildcard subdomains. Your route will only be available at host ' + ingress.host + '.');
```
spadgett: Thanks, pulled the latest router image, and I see it now. It looks like we handle the case for both older and newer router images OK.

screen shot 2016-11-11 at 8 42 10 am

spadgett:
> did yesterday's technical discussion change the UI design?

@ajacobs21e No, my current working code is

https://github.com/openshift/origin-web-console/pull/853#issuecomment-259767144

which I think is what we agreed to during the design review?

> Does changing the second sentence to "Generated if not specified" keep that whole section on one line?

Depends entirely on screen width. At the widest widths, it doesn't wrap the text even without those changes. At narrower widths, it will wrap the the updated text. We generally use complete sentences in the help text, which is how I interpret the Patternfly guidelines.

https://www.patternfly.org/styles/terminology-and-wording

app/scripts/services/applicationGenerator.js
jwforres: @ramr are you thinking something more like "generated" or "placeholder"

jwforres: or maybe even "wildcard"

app/scripts/services/routes.js
```diff
@@ -72,6 +73,10 @@ angular.module("openshiftConsole")
           }
           warnings.push(message);
         }
+
+        if (wildcardPolicy === 'Subdomain' && ingress.wildcardPolicy !== wildcardPolicy) {
```
jwforres: Add a comment here that this will only show up for older routers that don't reject the route based on wildcard policy.

app/scripts/services/applicationGenerator.js
knobunc: I like wildcard!

spadgett:
@jwforres thanks, updated

jwforres:
@spadgett looks like your spec test failed in travis https://travis-ci.org/openshift/origin-web-console/builds/175844167#L1782

spadgett:
>  looks like your spec test failed in travis 

Oops, forgot to update the test for `www` -> `wildcard` prefix change

spadgett:
Fixed the test.

ajacobs21e:
thanks @spadgett. I'm happy with the UI.

jwforres:
[merge]

PR #854, opened 2016-11-09
Bug Fix 1390826 - PVC does not bind

zherman0:
PVC is not binding when using provisioner due to extra empty spec object being passed.  I have removed that and tweak some wording for the  form.  @spadgett

### DIFF
--- a/app/views/directives/osc-persistent-volume-claim.html
+++ b/app/views/directives/osc-persistent-volume-claim.html
@@ -30,6 +30,7 @@
                  name="storageClass"
                  ng-model="claim.storageClass">
             No Storage Class
+          <div class="help-block">No storage class will be assigned unless a default class has been assigned by the system administrator.</div>
         </label>
       </div>
     </div>
--- a/app/scripts/controllers/createPersistentVolumeClaim.js
+++ b/app/scripts/controllers/createPersistentVolumeClaim.js
@@ -67,28 +67,25 @@ angular.module('openshiftConsole')
           spec: {
             resources: {
               requests:{}
-            },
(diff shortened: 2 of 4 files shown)


spadgett:
@jwforres

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -29,7 +29,7 @@
           <input type="radio"
                  name="storageClass"
                  ng-model="claim.storageClass">
-            No Storage Class
+            No storage class or use default if set
```
jwforres: This reads a little weird to me, maybe just "Default storage class." and then add a help block underneath that says
"If your administrator has defined a default storage class then that will be used.  Otherwise no storage class will be selected."

@spadgett thoughts?

jwforres:
@zherman0 can you please address the review feedback

zherman0:
Hey, QA was complaining that the default storage class gets set if you
chose no storage class.  I like your idea of adding more info in the help
block. I will change that and redo Sam's suggestions later this evening. I
am currently flying.

On Nov 10, 2016 1:10 PM, "Jessica Forrester" <notifications@github.com>
wrote:

@zherman0 <https://github.com/zherman0> can you please address the review
feedback

—
You are receiving this because you were mentioned.

<https://github.com/openshift/origin-web-console/pull/854#issuecomment-259808336>,
or mute the thread

zherman0:
@spadgett - Can you take another look at this; Need to get this in today if possible.  Thanks.

spadgett:
[merge]

PR #855, opened 2016-11-10
Build config source fixes

spadgett:
* Show message when there are no editable source types in the editor
* Show the binary input as file in the browse page configuration tab
* Show the Dockerfile on the browse page if it exists, even when
  `spec.strategy.type` isn't `Docker`

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1393803

@jwforres PTAL


jwforres:
[merge]

you should probably explain the fix in the bug so they don't try to re-open it

PR #859, opened 2016-11-11
Fix env editors that were restricting underscore as the first char

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1394044


jwforres:
@spadgett PTAL, i swept all the kve instances

jwforres:
[merge]

PR #864, opened 2016-11-11
Add view events link to failed deployment alert

spadgett:
Frequently the cause of the failure is in the events, not the deployment logs. Add a link so that users are more likely to check there.

openshift_web_console

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1365525


jwforres:
LGTM for 1.5

On a related note, I feel like on an RC we should be showing events for the pods for that RC, or for a DC we should be showing events for the RCs and Pods underneath it. We already started doing something similar on the monitoring view, if you expand a deployment it highlights both the RCs events and the pod events.

jwforres:
[merge]

PR #871, opened 2016-11-14
Add ability to hide token in cli tools page

juanvallejo:
Related Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1388770
Fixes https://github.com/openshift/origin/issues/11975

cli-tools-page-with-hide-link

cc @spadgett @jwforres

### DIFF
--- a/app/views/command-line.html
+++ b/app/views/command-line.html
@@ -37,30 +37,28 @@
                   </p>
                   <p>
                     After downloading and installing it, you can start by logging in using<span ng-if="sessionToken"> this current session token</span>:
-                    <div class="code prettyprint ng-binding" ng-if="sessionToken">
-                      oc login {{loginBaseURL}} --token=<span ng-show="showSessionToken">{{sessionToken}}</span><a href="#" ng-click="toggleShowSessionToken()" ng-show="!showSessionToken">...click to show token...</a>
-                    </div>
+                    <copy-to-clipboard display-wide="true" clipboard-text="'oc login ' + loginBaseURL + ' --token=' + sessionToken" input-text="'oc login ' + loginBaseURL + ' --token=<hidden>'"></copy-to-clipboard>
                     <pre class="code prettyprint ng-binding" ng-if="!sessionToken">
                       oc login {{loginBaseURL}}
                     </pre>
                   </p>
 
-                  <div ng-show="showSessionToken" class="alert alert-warning">
+                  <div class="alert alert-warning">
                     <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
                     <strong>A token is a form of a password.</strong>
                     Do not share your API token.
                   </div>
 
                   <p>After you login to your account you will get a list of projects that you can switch between:
-                     <pre class="code prettyprint">oc project <i>project-name</i></pre>
+                    <copy-to-clipboard display-wide="true" clipboard-text="'oc project <project-name>'"></copy-to-clipboard>
                   </p>
 
                   <p>If you do not have any existing projects, you can create one:
-                     <pre class="code prettyprint">oc new-project <i>project-name</i></pre>
+                    <copy-to-clipboard display-wide="true" clipboard-text="'oc new-project <project-name>'"></copy-to-clipboard>
                   </p>
 
                   <p>To show a high level overview of the current project:
-                     <pre class="code prettyprint">oc status</pre>
+                    <copy-to-clipboard display-wide="true" clipboard-text="'oc status'"></copy-to-clipboard>
                   </p>
 
                   <p>For other information about the command line tools, check the <a target="_blank" href="{{'cli' | helpLink}}">CLI Reference</a> and <a target="_blank" href="{{'basic_cli_operations' | helpLink}}">Basic CLI Operations</a>.</p>
--- a/app/views/directives/_copy-to-clipboard.html
+++ b/app/views/directives/_copy-to-clipboard.html
@@ -1,7 +1,13 @@
-<div class="input-group copy-to-clipboard">
-  <input id="{{id}}" type="text" class="form-control" value="{{clipboardText}}" ng-disabled="isDisabled" ng-readonly="!isDisabled" select-on-focus>
+<div class="input-group copy-to-clipboard" ng-class="{'limit-width': !displayWide}">
+  <input id="{{id}}" type="text" class="form-control" value="{{inputText ? inputText : clipboardText}}" ng-disabled="isDisabled" ng-readonly="!isDisabled" select-on-focus>
   <span class="input-group-btn" ng-hide="hidden">
-    <a data-clipboard-target="#{{id}}"
+    <a ng-show="!inputText" data-clipboard-target="#{{id}}"
+            ng-disabled="isDisabled"
+            data-toggle="tooltip"
+            title="Copy to clipboard"
+            role="button"
+            class="btn btn-default"><i class="fa fa-clipboard"/></a>
+    <a ng-show="inputText" data-clipboard-text="{{clipboardText}}"
             ng-disabled="isDisabled"
             data-toggle="tooltip"
             title="Copy to clipboard"
--- a/app/scripts/directives/util.js
+++ b/app/scripts/directives/util.js
@@ -76,7 +76,9 @@ angular.module('openshiftConsole')
       restrict: 'E',
       scope: {
         clipboardText: "=",
-        isDisabled: "=?"
+        isDisabled: "=?",
+        displayWide: "=?",
+        inputText: "=?"
       },
       templateUrl: 'views/directives/_copy-to-clipboard.html',
       controller: function($scope) {
@@ -88,8 +90,13 @@ angular.module('openshiftConsole')
           return;
         }
 
-        var node = $('a', element);
-        var clipboard = new Clipboard( node.get(0) );
+        var nodeElem = $('a', element);
+        var node = nodeElem.get(0);
+        if ($scope.inputText) {
+          node = nodeElem.get(1);
+        }
+
+        var clipboard = new Clipboard( node );
         clipboard.on('success', function (e) {
           $(e.trigger)
             .attr('title', 'Copied!')
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -1205,7 +1205,7 @@ h1 small.meta, .build-config-summary .meta {
 }
 
 // don't allow copy-to-clipboard .input-groups get bigger than 600px so Github webhook secrets aren't revealed within the input-group
-copy-to-clipboard .input-group {
+copy-to-clipboard .input-group.limit-width {
   max-width: 300px;
 }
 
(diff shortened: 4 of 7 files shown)


jwforres:
@spadgett mind taking this one?

app/views/command-line.html
```diff
@@ -48,7 +48,7 @@ <h1 id="cli">Command Line Tools</h1>
                   <div ng-show="showSessionToken" class="alert alert-warning">
                     <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
                     <strong>A token is a form of a password.</strong>
-                    Do not share your API token.
+                    Do not share your API token. <a href="#" ng-click="toggleShowSessionToken()" ng-show="showSessionToken">Click to hide the token...</a>
```
spadgett: I'd probably just say "Hide token".

You don't need the `ng-show` here since it's on the parent `div`.

You only need `href=""`. (Do you mind fixing the reveal `href` above as well?)

spadgett:
@jwforres Opinion on a copy to clipboard link (without needing to reveal the token)? Right now I have to click show token and then manually select the command.

jwforres:
wouldn't be opposed to the idea, would need to see how it looks in context

juanvallejo:
@jwforres @spadgett 
cli-tools-page-copy-token-to-clipboard

cli-tools-page-copied-token-to-clipboard

spadgett:
@juanvallejo +1, except I would copy the full command. Something like...?

```html
<pre>oc login https://localhost:8443 --token=&lt;hidden&gt;</pre>
<a href="">Show token</a>
<span class="action-divider">|</span>
<a href="">Copy command to clipboard</a>
```

If we made that change, we'd probably always show the warning about tokens. And might want to avoid the copy to clipboard link for browsers that don't support it like iOS Safari.

juanvallejo:
@spadgett Sounds good, I'll update this.
Thanks for the feedback, addressed review comments on https://github.com/openshift/origin-web-console/pull/871#pullrequestreview-8488182 . I tried updating the reveal `href` to leave `ng-show` to its parent, however its sibling `span` has `ng-show` set to an opposite value

juanvallejo:
@spadgett 

cli-tools-page-showtoken-copycommand

> `Show token` is clicked

cli-tools-page-token-shown

> `Copy to clipboard` is clicked

cli-tools-page-copy-to-clipboard

app/scripts/controllers/commandLine.js
```diff
@@ -15,8 +15,36 @@ angular.module('openshiftConsole')
...
     $scope.sessionToken = AuthService.UserStore().getToken();
     $scope.showSessionToken = false;
+    $scope.copySessionTokenId = "showSessionTokenLink";
 
     $scope.toggleShowSessionToken = function() {
       $scope.showSessionToken = !$scope.showSessionToken;
     };
+
+    $scope.copySessionToken = function($scope) {
+      var node = $('#' + $scope.copySessionTokenId);
```
spadgett: I suggest we update the `copy-to-clipboard` directive to take in an element ID or create a new directive we could reuse (scripts/directives/util.js).

app/views/command-line.html
```diff
@@ -48,7 +53,7 @@ <h1 id="cli">Command Line Tools</h1>
                   <div ng-show="showSessionToken" class="alert alert-warning">
```
spadgett: Might want to show to show this always since I won't see the warning if I click copy to clipboard.

app/views/command-line.html
```diff
@@ -38,7 +38,12 @@ <h1 id="cli">Command Line Tools</h1>
                   <p>
                     After downloading and installing it, you can start by logging in using<span ng-if="sessionToken"> this current session token</span>:
                     <div class="code prettyprint ng-binding" ng-if="sessionToken">
-                      oc login {{loginBaseURL}} --token=<span ng-show="showSessionToken">{{sessionToken}}</span><a href="#" ng-click="toggleShowSessionToken()" ng-show="!showSessionToken">...click to show token...</a>
+                      oc login {{loginBaseURL}} --token=<span ng-show="showSessionToken">{{sessionToken}}</span><span ng-show="!showSessionToken">&lt;hidden&gt;</span>
+                      <span ng-show="!showSessionToken">
+                        <a href="#" ng-click="toggleShowSessionToken()">Show token</a>
+                        <span class="action-divider">|</span>
+                        <a href="#" id="{{copySessionTokenId}}" ng-click="copySessionToken(this)" title="Copy command to clipboard" data-toggle="tooltip">Copy command to clipboard</a>
```
spadgett: Hide this if it's iOS? There's an `IS_IOS` constant you can use in the controller to set a scope var.

app/views/command-line.html
```diff
@@ -48,7 +53,7 @@ <h1 id="cli">Command Line Tools</h1>
                   <div ng-show="showSessionToken" class="alert alert-warning">
                     <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
                     <strong>A token is a form of a password.</strong>
-                    Do not share your API token.
+                    Do not share your API token. <a href="#" ng-click="toggleShowSessionToken()">Hide token.</a>
```
spadgett: Might want to change the show link to a hide link and just have it toggle.

app/views/command-line.html
spadgett: How do the actions look if you move them below the `prettyprint` div?

juanvallejo:
@spadgett Review comments addressed, I'm now re-using the `<copy-to-clipboard` directive

cli-tools-page-with-new-copy-command

spadgett:
LGTM, @jwforres OK with this change?

spadgett:
@sspeiche FYI

jwforres:
yep, @spadgett were you going to do the code review on this

spadgett:
> were you going to do the code review on this

Yeah, code looks good. Thanks @juanvallejo

spadgett:
[merge]

spadgett:
hold, spotted one problem

app/views/command-line.html
```diff
@@ -38,14 +38,22 @@ <h1 id="cli">Command Line Tools</h1>
                   <p>
                     After downloading and installing it, you can start by logging in using<span ng-if="sessionToken"> this current session token</span>:
                     <div class="code prettyprint ng-binding" ng-if="sessionToken">
```
spadgett: @juanvallejo need an outer div with `ng-if="sessionToken"` surrounding both the `prettyprint` div and the action links below.

sspeiche:
Why would this "copy to clipboard" not be the same icon for webhooks ?

juanvallejo:
@spadgett Thanks, review comment addressed

@sspeiche 
> Why would this "copy to clipboard" not be the same icon for webhooks ?

Would that be the 
clip 
icon? I did not think it would fit with this page, as I have only seen it used when copying contents from an `<input` element

sspeiche:
It looks like something like this:

image

spadgett:
@sspeiche are you proposing to use the entire input field or just the icon?

sspeiche:
I'm proposing that everywhere in the web console that we "copy to clipboard" it doesn't do something that looks and feels that different.

spadgett:
I agree consistency is good, but we're being careful not to show the token value. Maybe the copy-to-clipboard input is not wide enough to reveal it. I'm not we should count on that, though.

jwforres:
if we are going to do that, we should just switch all of these code blocks on this page to copy to clipboard elements instead

@juanvallejo mind getting a screenshot of what that would look like?

juanvallejo:
@jwforres 
> if we are going to do that, we should just switch all of these code blocks on this page to copy to clipboard elements instead

Sure, I'll put an example together

juanvallejo:
@jwforres @spadgett @sspeiche 
cli-tools-page-inputs

sspeiche:
Fantastic, shouldn't the input fields be wider?

jwforres:
Making the input fields wider would cause the token to be visible, which is
bad.  We have them short because they are often obscuring things later in
the input, like with webhooks they are obscuring the secrets in the URL

jwforres:
That said, I'm fine with this change.  @spadgett ?

spadgett:
Does it make sense to copy commands with placeholder values like `<project-name>`? (Maybe?)

spadgett:
> Making the input fields wider would cause the token to be visible, which is bad.

We might be able to still show `token=<hidden>` in the input and copy a different value, although we'd need to add some additional logic to the `copy-to-clipboard` directive. I'd want to check how 100% width looks in other places we use it since it's a global style.

juanvallejo:
@spadgett I can add a `wide` option to the directive so that full width does not apply anywhere else. I'll put together an example with this and `<hidden>` values

juanvallejo:
@spadgett @jwforres @sspeiche 
cli-tools-page-wide-inputs

Clipboard contents from copying first input `oc login https://localhost:8443 --token=YGvpJBROMXDaY_uO8XZuPKRMh3TKycqGcK9IkBXnxzE`

juanvallejo:
@spadgett added commit with `copy-to-clipboard` changes, PTAL

app/views/directives/_copy-to-clipboard.html
```diff
@@ -1,11 +1,17 @@
-<div class="input-group copy-to-clipboard">
-  <input id="{{id}}" type="text" class="form-control" value="{{clipboardText}}" ng-disabled="isDisabled" ng-readonly="!isDisabled" select-on-focus>
+<div class="input-group copy-to-clipboard" style="{{displayWide ? 'max-width:100%;' : ''}}">
```
spadgett: Suggest something like

```
ng-class="{ 'limit-width': !displayWide }"
```

then update the CSS to key off that class.

https://github.com/spadgett/origin-web-console/blob/e01508561bb7d0c7f6c53dbedcf7dd42d37e0c52/app/styles/_core.less#L1208-L1210

app/views/directives/_copy-to-clipboard.html
```diff
@@ -1,11 +1,17 @@
...
+  <input id="{{id}}" type="text" class="form-control" value="{{inputText ? inputText : clipboardText}}" ng-disabled="isDisabled" ng-readonly="!isDisabled" select-on-focus>
   <span class="input-group-btn" ng-hide="hidden">
-    <a data-clipboard-target="#{{id}}"
+    <a ng-show="!inputText" data-clipboard-target="#{{id}}"
```
spadgett: If you use `ng-if` here and line 10, does it let you remove the `get(0)` / `get(1)` special case in the controller?

spadgett: Ah, OK. We could wrap the controller code in a `$timeout` to let the DOM render first, but what you have is probably better.

juanvallejo:
@spadgett thanks for the feedback, review comments addressed

spadgett:
thanks @juanvallejo 

LGTM, please squash, and I'll merge

juanvallejo:
@spadgett 
> LGTM, please squash, and I'll merge

Thanks, done!

spadgett:
[merge]

PR #877, opened 2016-11-15
Remove path from hostname in route view

juanvallejo:
Fixes: https://github.com/openshift/origin/issues/5994

This patch removes a route's path from its routeLabel, but leaves the
original path (hostname + path) in the route's link (if any).

route-page-with-hostname-and-path

@jwforres @spadgett

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -338,19 +338,19 @@ angular.module('openshiftConsole')
     };
   })
   .filter('routeWebURL', function(routeHostFilter){
-    return function(route, host){
+    return function(route, host, omitPath){
         var scheme = (route.spec.tls && route.spec.tls.tlsTerminationType !== "") ? "https" : "http";
         var url = scheme + "://" + (host || routeHostFilter(route));
-        if (route.spec.path) {
+        if (route.spec.path && !omitPath) {
             url += route.spec.path;
         }
         return url;
     };
   })
   .filter('routeLabel', function(RoutesService, routeHostFilter, routeWebURLFilter, isWebRouteFilter) {
-    return function(route, host) {
+    return function(route, host, omitPath) {
       if (isWebRouteFilter(route)) {
-        return routeWebURLFilter(route, host);
+        return routeWebURLFilter(route, host, omitPath);
       }
 
       var label = (host || routeHostFilter(route));
@@ -362,6 +362,10 @@ angular.module('openshiftConsole')
         label = '*.' + RoutesService.getSubdomain(route);
       }
 
+      if(omitPath) {
+        return label;
+      }
+      
       if (route.spec.path) {
         label += route.spec.path;
       }
--- a/app/views/browse/route.html
+++ b/app/views/browse/route.html
@@ -56,7 +56,7 @@
                   <dt>Hostname<span ng-if="route.status.ingress.length > 1">s</span>:</dt>
                   <dd>
                     <span ng-if="!route.status.ingress">
-                      {{route | routeLabel}}
+                      {{route | routeLabel : null : true}}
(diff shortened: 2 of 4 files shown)


app/scripts/filters/resources.js
```diff
@@ -349,18 +349,22 @@ angular.module('openshiftConsole')
         return url;
     };
   })
-  .filter('routeLabel', function(routeHostFilter, routeWebURLFilter, isWebRouteFilter) {
+  .filter('routeWebURLNoPath', function(routeHostFilter){
```
jwforres: instead of making a new function, i would change routeWebURL filter to take in an optional parameter that tells it not to include the path

app/scripts/filters/resources.js
```diff
@@ -349,18 +349,22 @@ angular.module('openshiftConsole')
...
+        var scheme = (route.spec.tls && route.spec.tls.tlsTerminationType !== "") ? "https" : "http";
+        var url = scheme + "://" + (host || routeHostFilter(route));
+        return url;
+    };
+  })
+  .filter('routeLabel', function(routeHostFilter, routeWebURLNoPathFilter, isWebRouteFilter) {
     return function(route, host) {
       if (isWebRouteFilter(route)) {
-        return routeWebURLFilter(route, host);
+        return routeWebURLNoPathFilter(route, host);
```
jwforres: this is going to affect everywhere that uses routeLabel today, I don't think that is what we want.  In this context we are showing you the path separately, so you don't need to see it, but in other places we don't and we may want to see the path included.

juanvallejo:
@jwforres thanks for the feedback, review comments addressed

app/scripts/filters/resources.js
```diff
@@ -364,6 +364,18 @@ angular.module('openshiftConsole')
       return label;
     };
   })
+  .filter('routeLabelWithNoPath', function(routeHostFilter, routeWebURLFilter, isWebRouteFilter) {
```
jwforres: same thing as before, i would just add omitPath as a param to routeLabel as well and just pass it the parameter through

app/views/browse/route.html
```diff
@@ -56,15 +56,15 @@
                   <dt>Hostname<span ng-if="route.status.ingress.length > 1">s</span>:</dt>
                   <dd>
                     <span ng-if="!route.status.ingress">
-                      {{route | routeLabel}}
+                      {{route | routeLabelWithNoPath}}
```
jwforres: Assuming you change routeLabel to take in omitPath then this would just become `route | routeLabel : true`

juanvallejo:
@jwforres Thanks, review comments addressed

jwforres:
looks good now, can you squash your commits

juanvallejo:
@jwforres 
> looks good now, can you squash your commits

Thanks for the review, done!

app/views/browse/route.html
```diff
@@ -56,15 +56,15 @@
...
+                      {{route | routeLabel : null : true}}
                       <span data-toggle="popover" data-trigger="hover" data-content="The route is not accepting traffic yet because it has not been admitted by a router." style="cursor: help; padding-left: 5px;">
                         <status-icon status="'Pending'" disable-animation></status-icon>
                         <span class="sr-only">Pending</span>                        
                       </span>
                     </span>  
                     <div ng-repeat="ingress in route.status.ingress">
                       <span ng-if="(route | isWebRoute)">
-                        <a href="{{route | routeWebURL : ingress.host}}" target="_blank">{{route | routeLabel : ingress.host}}</a>
+                        <a href="{{route | routeWebURL : ingress.host}}" target="_blank">{{route | routeLabel : ingress.host : true}}</a>
```
jwforres: sorry this isn't new but i just noticed it, this should be using ng-href and not href, mind fixing that here?

juanvallejo: Done!

jwforres:
thanks! [merge]

PR #884, opened 2016-11-16
Adding "Run on OpenShift" functionality

rhamilto:
Proposed image stream query string params:

* imageStream (required)
* imageTag (required)
* namespace (optional, but if not specified, defaults to 'openshift')
* name (optional)
* sourceURI (optional)
* sourceRef (optional)
* contextDir (optional)

Proposed template query string params:

* template (required)
* templateParamsMap (optional)
* namespace (optional, but if not specified, defaults to 'openshift')

Testing URLs:

create?name=nodejs-edited&imageStream=nodejs&imageTag=4&sourceURI=https:%2F%2Fgithub.com%2Fopenshift%2Fnodejs-ex.git-edited&sourceRef=master-edited&contextDir=%2F-edited

create?template=nodejs-mongodb-example&templateParamsMap=%7B"SOURCE_REPOSITORY_URL":"https:%2F%2Fgithub.com%2Fopenshift%2Fnodejs-ex.git-edited"%7D

Screenshots:

User has existing project(s) and can create new project(s)

screen shot 2016-11-17 at 8 36 27 am
screen shot 2016-11-17 at 8 36 39 am

User has no project(s), but can create new project(s)

screen shot 2016-11-17 at 8 55 55 am

User has existing project(s), but cannot create new project(s)

screen shot 2016-11-17 at 8 56 28 am

User has no existing project(s) nor can create new project(s)

screen shot 2016-11-17 at 8 56 56 am
Wondering if the messaging at the bottom is obvious enough?  This shouldn't be a common case, but still...

@jwforres or @spadgett, PTAL.


spadgett:
For query parameter names, I think we should align with the API fields where we can. Proposing

* `contextDir` instead of `contextDirectory`
* `sourceURI` (API field is `source.git.uri`)
* `sourceRef` (API field is `source.git.ref`)
* `app` instead of `appName` (matching the label key)

Can't decide between `sourceURI` and `gitURI`.

We might talk about the best way to handle template parameters. JSON is reasonable, but it's a bit hard to build that URL by hand. (I know I suggested JSON before :)

rhamilto:
Great suggestions, @spadgett.  I've implemented your proposed changes.  I favor sourceURI as it seems more future proof.  

@benjaminapetersen and I debated the JSON template params, but decided to go with it since it not only keeps the code simpler, but it also seems to simpler with regard to how to construct the query string param since you're overriding template-defined JSON objects with query string-ified JSON objects (URL-building complications aside, which really isn't that bad as the browser will do the bracket escaping for you).  

Thoughts?

benjaminapetersen:
Agree on making the params match, that makes sense.
I'd be up for exploring alternates to the JSON syntax in the url, `jQuery.param()` works ok, might need to tweak it: http://codepen.io/benjaminapetersen/pen/NbbMjJ?editors=0010 (it will turn a nested object into `[object Object]`).

app/scripts/constants.js
```diff
@@ -369,5 +369,7 @@ window.OPENSHIFT_CONSTANTS = {
         }
       ]
     }
-  ]
+  ],
+  // 'openshift' should always be included
+  CREATE_FROM_URL_WHITELIST: ['openshift','run-on-openshift']
```
spadgett: I think we should leave this just `['openshift']` by default.

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,205 @@
...
+      }
+    };
+
+    if (!(_.includes(namespaceWhitelist, createDetails.namespace))) {
+      alertInvalidNamespace(createDetails.namespace);
+    } else {
+      if (createDetails.imageStream && createDetails.template) {
+        showInvalidResource();
+      } else if (!(createDetails.imageStream) && !(createDetails.template)) {
+        alertResourceRequired();
```
spadgett: We typically use `Navigate.toErrorPage()` for these types of errors (invalid or conflicting route params)

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,205 @@
...
+        getResources();
+      }
+    }
+
+    angular.extend($scope, {
+      createDetails: createDetails,
+      createWithProject: function(project) {
+        $location
+          .url('/project/' + (project ||  $scope.selected.project.metadata.name) + '/create/' + ($routeParams.imageStream ? 'fromimage' : 'fromtemplate'))
+          .search(createDetails);
```
spadgett: We have this URL built in a few places. Probably want to consolidate it in Navigate service (and have filters call the service when needed). See

https://github.com/spadgett/origin-web-console/blob/2c15578e8963210ce91e795ff515f9c845883940/app/scripts/services/navigate.js#L72-L85

https://github.com/spadgett/origin-web-console/blob/2c15578e8963210ce91e795ff515f9c845883940/app/scripts/filters/resources.js#L619-L644

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,205 @@
...
+    $scope.projects = {};
+    $scope.canCreateProjectProject = undefined;
+
+    DataService.list("projects", $scope, function(items) {
+      $scope.loaded = true;
+      $scope.projects = items.by("metadata.name");
+      $scope.noProjects = (_.isEmpty($scope.projects));
+    });
+
+    // copied from app/scripts/controllers/projects.js
```
spadgett: Move into `ProjectsService` so we have the logic in one place?

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,205 @@
...
+
+    var messages = {
+      inaccessibleNamespace: _.template('You do not have access to the namespace "<%= namespace %>".'),
+      invalidImageStream: _.template('The requested image stream "<%= imageStream %>" is not valid.'),
+      invalidImageTag: _.template('The requested image stream tag "<%= tag %>" is not valid.'),
+      invalidNamespace: _.template('Resources from the namespace "<%= namespace %>" are not permitted.'),
+      invalidResource:'Image streams and templates cannot be combined.',
+      invalidTemplate: _.template('The requested template "<%= template %>" is not valid.'),
+      resourceRequired: 'A valid image stream or template is required.'
+    };
```
spadgett: Some of these are templates, and some are just strings. I'm not sure I like mixing them in the same object because you have no way to know which is which where they're called. Is this really better than just inlining the messages? Then there's no confusion.

app/scripts/controllers/newfromtemplate.js
```diff
@@ -31,8 +31,7 @@ angular.module('openshiftConsole')
     keyValueEditorUtils,
     Constants) {
 
-
-    var name = $routeParams.name;
+    var name = $routeParams.template;
```
spadgett: Need to update at least Navigate service as well since we've changed a query parameter name.

app/scripts/controllers/newfromtemplate.js
```diff
@@ -355,6 +354,13 @@ angular.module('openshiftConsole')
             $scope.parameterDisplayNames[parameter.name] = parameter.displayName || parameter.name;
           });
 
+          _.each($scope.template.parameters, function(parameter) {
+            var found = _.find(JSON.parse($routeParams.templateParamsMap), function(param){
```
spadgett: Need to catch the exception on JSON parse failures. This is not JSON we're generating, so it might have syntax errors.

app/scripts/controllers/newfromtemplate.js
```diff
@@ -355,6 +354,13 @@ angular.module('openshiftConsole')
...
+              return param.name === parameter.name;
+            });
```
spadgett: Can use the shorthand (where `params` is the parsed JSON)

```js
_.find(params, { name: parameter.name });
```

app/scripts/directives/createProject.js
```diff
@@ -0,0 +1,53 @@
...
+        alerts: '=',
+        submitButtonLabel: '@',
+        redirectAction: '&'
+      },
+      templateUrl: 'views/directives/_create-project-form.html',
+      controller: function($scope, $filter, $location, DataService) {
+        if(!($scope.submitButtonLabel)) {
+          $scope.submitButtonLabel = 'Create';
+        }
+        $scope.createProject = function() {
```
spadgett: We typically make the directive just the editor that takes in the object and updates it, then let the controller call create and handle errors.

app/views/create-from-url.html
```diff
@@ -0,0 +1,76 @@
...
+                    </div>
+                    <div ng-if="!noProjects && canCreateProject">
+                      <uib-tabset>
+                        <uib-tab>
+                          <uib-tab-heading>Choose Existing Project</uib-tab-heading>
+                          <ui-select ng-model="selected.project">
+                            <ui-select-match placeholder="Project name">
+                              {{$select.selected.metadata.name}}
+                            </ui-select-match>
+                            <ui-select-choices repeat="project in projects | toArray | filter : { metadata: { name: $select.search } } | orderBy : 'metadata.name'">
```
spadgett: Better to convert to an array / sort once in the controller.

spadgett: Should we match keywords on display name as well?

app/views/create-from-url.html
```diff
@@ -0,0 +1,76 @@
...
+                      <h2>Create a New Project</h2>
+                      <create-project alerts="alerts" submit-button-label="Next" redirect-action="createWithProject"></create-project>
+                    </div>
+                    <div ng-if="!noProjects && !canCreateProject">
+                      <h2>Choose a Project</h2>
+                      <ui-select ng-model="selected.project">
+                        <ui-select-match placeholder="Project name">
+                          {{$select.selected.metadata.name}}
+                        </ui-select-match>
+                        <ui-select-choices repeat="project in projects | toArray | filter : { metadata: { name: $select.search } } | orderBy : 'metadata.name'">
```
spadgett: Same comment about filter and sorting as line 43.

app/scripts/controllers/newfromtemplate.js
spadgett: Any reason to not make the JSON a map? For example,

```json
{
  "SOURCE_URL_REPOSITORY": "http://github.com/openshift/origin-ruby-example"
}
```

app/scripts/controllers/createFromURL.js
spadgett: @rhamilto So the idea is that I can tweak the URL in the location bar until I get it right? Makes sense.

Is the rest of the page blank? How does that look?

rhamilto: > Is the rest of the page blank? How does that look?

Yep.  It looks like a blank page.  I wasn't sure what else to include since it's hard to know what the user is trying to do.  Thoughts?

spadgett: Maybe some text with help building the URL? Or a link to the docs when they're written?

app/scripts/controllers/newfromtemplate.js
rhamilto: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/files/examples/v1.4/quickstart-templates/nodejs-mongodb.json#L364  If we do that, then users can only override name and value key values.  If we leave it as is, they can override any key values they choose.  Do we want to restrict to only name and value values?

spadgett: > Do we want to restrict to only name and value values?

Yeah, I don't think they should be able to change anything other than the value for an existing parameter. We should probably show an error if they give the name that does not match a param in the template.

jwforres:
Yes we should use display name when picking a project here.  You'll need
the same logic we use for the main project dropdown, since you have to
handle duplicate display names.

rhamilto:
Code review to dos:
- [X] set whitelist to only 'openshift'
- [X] make error messages inline strings
- [ ] add link to documentation on /create
- [ ] consolidate create URL in Navigate service
- [ ] move projectrequests in to ProjectsService to eliminate duplication of code
- [X] update Navigate service with renamed $routeParams.template
- [X] catch the exception on JSON parse failures
- [X] make the JSON a map
- [X] fix display and sort of project names in the dropdown
- [ ] utilize display-name in project dropdown if set

rhamilto:
@spadgett, I believe @benjaminapetersen and I addressed your comments, but a few additional TODOs remain.   So we're not ready for merge yet, but you're welcome to review the requested changes.

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,191 @@
...
+                $scope.imageStreamTag = imageStreamTag;
+                $scope.validationPassed = true;
+                $scope.resource = imageStreamTag;
+                createDetails.displayName = $filter('displayName')(imageStreamTag);
+              }, function(e){
+                alertInvalidImageTag(createDetails.imageTag, e);
+              });
+          }, function(e) {
+            if(e.status === 403) {
+              alertInaccessibleNamespace(createDetails.namespace, e);
```
spadgett: You might have access to the namespace, but not the image stream. I'd change the message to say specifically that you don't have access to the image stream or template.

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,191 @@
...
+        details: "Reason: " + $filter('getErrorDetails')(e)
+      };
+    };
+
+    var alertInvalidNamespace = function(namespace) {
+      $scope.alerts.invalidNamespace = {
+        type: "error",
+        message: namespace ?
+                  "Resources from the namespace \"" + namespace + "\" are not permitted." :
+                  "A valid namespace is required."
```
spadgett: Let's open a follow-on issue to add a doc link here for whitelisting namespaces.

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,191 @@
...
+        type: "error",
+        message: "You do not have access to the namespace \"" + namespace + "\".",
+        details: "Reason: " + $filter('getErrorDetails')(e)
+      };
+    };
+
+    var alertInvalidImageStream = function(imageStream, e) {
+      $scope.alerts.invalidImageStream = {
+        type: "error",
+        message: "The requested image stream \"" + imageStream + "\" is not valid.",
```
spadgett: Suggest a more general message saying "Could not load image stream ..." instead of saying "not valid." It might be valid, but we couldn't load it for another reason. Same for templates.

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,191 @@
...
+    AlertMessageService.clearAlerts();
+
+    $scope.alerts = {};
+    $scope.selected = {};
+
+    var alertInaccessibleNamespace = function(namespace, e) {
+      $scope.alerts.inaccessibleNamespace = {
+        type: "error",
+        message: "You do not have access to the namespace \"" + namespace + "\".",
+        details: "Reason: " + $filter('getErrorDetails')(e)
```
spadgett: `getErrorDetails` can in some cases return an empty string, so you could have "Reason: " with nothing following. We should either fix the filter or leaving off "Reason:".

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,191 @@
...
+          message: "The template parameters are not valid."
+        };
+      }
+    };
+
+    var namespaceWhitelist = window.OPENSHIFT_CONSTANTS.CREATE_FROM_URL_WHITELIST;
+    var whiteListedCreateDetailsKeys = ['namespace', 'app', 'imageStream', 'imageTag', 'sourceURI', 'sourceRef', 'contextDir', 'template', 'templateParamsMap'];
+    var createDetails = _.pick($routeParams, function(value, key) {
+      // routeParams without a value (e.g., ?app&) return true, which results in "true" displaying in the UI
+      return _.contains(whiteListedCreateDetailsKeys, key) && (value && value !== true);
```
spadgett: Maybe

```js
return _.contains(whiteListedCreateDetailsKeys, key) && _.isString(value);
```

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    var getTemplateParamsMap = function () {
+      try {
+        return JSON.parse($routeParams.templateParamsMap);
+      }
+      catch (e) {
+        $scope.alerts.invalidTemplateParams = {
+          type: "error",
+          message: "The template parameters are not valid."
```
spadgett: We should say specifically that it's not valid JSON and ideally show the syntax error details from `e` if present.

app/scripts/controllers/newfromtemplate.js
```diff
@@ -80,6 +80,18 @@ angular.module('openshiftConsole')
...
     var outputImage = $parse('spec.output.to');
 
+    var getValidTemplateParamsMap = function () {
+      try {
+        return JSON.parse($routeParams.templateParamsMap);
+      }
+      catch (e) {
+        $scope.alertsTop.invalidTemplateParams = {
+          type: "error",
+          message: "The template parameters are not valid."
```
spadgett: Same comment as above about JSON parsing.

app/scripts/controllers/newfromtemplate.js
```diff
@@ -355,6 +367,15 @@ angular.module('openshiftConsole')
             $scope.parameterDisplayNames[parameter.name] = parameter.displayName || parameter.name;
           });
 
+          if($routeParams.templateParamsMap) {
+            var templateParams = getValidTemplateParamsMap();
+            _.each($scope.template.parameters, function(parameter) {
+              if (templateParams[parameter.name]) {
+                parameter.value = templateParams[parameter.name];
```
spadgett: Did we switch back to array for this controller? I think it would be better to use a map everywhere to be consistent if possible.

app/views/newfromtemplate.html
```diff
@@ -16,6 +16,7 @@
                   {{ emptyMessage }}
                 </div>
                 <div class="osc-form" ng-show="template">
+                  <alerts alerts="alertsTop" hide-close-button="true"></alerts>
```
spadgett: I understand why you added this, but not sure we should have multiple alerts directives on this page. @jwforres Opinion? Were the alerts moved down for the quota changes?

jwforres: yeah they were moved for the quota change because you would submit the form and not see the errors if they were scrolled off the top of the page

spadgett: This feels like a general problem with our alerts we might try and solve. Assuming it can happen with other forms, too.

app/scripts/controllers/createFromURL.js
spadgett: (I realize we have this problem in many places in existing code.)

app/scripts/controllers/createFromURL.js
rhamilto: That doesn't work.  Presumably because the boolean gets converted to a string?

app/scripts/controllers/newfromtemplate.js
rhamilto: Not sure I (we) understand the question.  $scope.template.parameters is the existing array that was already in use on the page.

rhamilto:
Blocked by https://github.com/openshift/origin/issues/12044

app/views/newfromtemplate.html
benjaminapetersen: We talked about moving to [toasts](http://www.patternfly.org/pattern-library/communication/toast-notifications/#/api) at one point, which I think would work as they are fixed position & lay on top of the screen.  Maybe we consider again?

benjaminapetersen:
So it still seems like we could do a better job with a visual cue implying that the user will be installing a significant piece of software.  I mentioned a tile before, though I know there were qualms.  A rough side by side:

run-on-tile-side-by-side

I can roll with it if we generally think the plain text approach is sufficient, but wanted to bring it up one more time.

test/integration/create-from-url/createFromURLTemplate.spec.js
```diff
@@ -6,70 +6,126 @@ var page = require('./createFromURL.js');
...
+    h.afterAllTeardown();
+  });
+
+  describe('using an image stream and image tag supplied as query string params', function() {
+    var qs = '?imageStream=nodejs&imageTag=4';
+    it('should display details about the the image', function() {
+      page.visit(qs);
+      expect(element(by.css('h1')).getText()).toEqual('Node.js 4');
+    });
+    it('should create a project to add to', function(){
```
benjaminapetersen: A quick thoughts here. an `it()` should have an `expect()` inside as an assertion of something, otherwise its not quite a test.  After the `tab.click()` we can probably `browser.wait()` then check that the url contains the project name?

test/integration/create-from-url/createFromURLTemplate.spec.js
```diff
@@ -6,70 +6,126 @@ var page = require('./createFromURL.js');
...
+
+  describe('using a template supplied as a query string param', function() {
+    var qs = '?template=nodejs-mongodb-example';
+    var uri = 'project/' + project['name'] + 'create/fromtemplate' + qs;
+    it('should display details about the template', function() {
+      page.visit(qs);
+      expect(element(by.css('h1')).getText()).toEqual('Node.js + MongoDB (Ephemeral)');
+    });
+    it('should use an existing project', function(){
+      // if the tabs are not on the page, create a new project instead
```
benjaminapetersen: Same, we should `expect()` in here so that the test will pass or fail.

test/integration/helpers.js
```diff
@@ -92,3 +96,31 @@ exports.waitFor = function(item, timeout, msg) {
   msg = msg || '';
   return browser.wait(item, timeout, msg);
 };
+
```
benjaminapetersen: I like that we can reuse this.  Maybe we should do a `helpers/` directory & put this in `helpers/projectHelpers.js`

test/integration/create-from-url/createFromURLTemplate.spec.js
rhamilto: Yeah.  The expects are going to be confirmation of the query string params in the corresponding fields.  I just haven't gotten there yet.  :)

rhamilto:
@spadgett, I think @benjaminapetersen have successfully added sufficient e2e tests, so I believe this is ready for final review.

rhamilto:
> So it still seems like we could do a better job with a visual cue implying that the user will be installing a significant piece of software. I mentioned a tile before, though I know there were qualms. A rough side by side:

For me, the tile disassociates the image stream or template description from the project selection.  They no longer feel like they're part of the same thing.

app/scripts/constants.js
```diff
@@ -377,5 +377,7 @@ window.OPENSHIFT_CONSTANTS = {
         }
       ]
     }
-  ]
+  ],
+  // 'openshift' should always be included
+  CREATE_FROM_URL_WHITELIST: ['openshift']
```
spadgett: I'd move this above `PROJET_NAVIGATION` because it's a bit easy to miss below the two long constants.

Suggest adding more comments describing what it is. I'd like the file to be as self-documenting as possible since we expect admins to customize these values.

app/scripts/controllers/createFromURL.js
```diff
@@ -0,0 +1,183 @@
...
+    var whiteListedCreateDetailsKeys = ['namespace', 'app', 'imageStream', 'imageTag', 'sourceURI', 'sourceRef', 'contextDir', 'template', 'templateParamsMap'];
+    var createDetails = _.pick($routeParams, function(value, key) {
+      // routeParams without a value (e.g., ?app&) return true, which results in "true" displaying in the UI
+      return _.contains(whiteListedCreateDetailsKeys, key) && _.isString(value);
+    });
+    // if no namespace is specified, set it to 'openshift'
+    createDetails.namespace = createDetails.namespace || 'openshift';
+
+    var validateApp = function (app) {
+      return /^[a-z]([-a-z0-9]*[a-z0-9])?$/.test(app);
```
spadgett: Should also check it's not longer than 63 characters.

app/scripts/services/projects.js
```diff
@@ -64,6 +65,9 @@ angular.module('openshiftConsole')
           update: function(projectName, data) {
             return DataService
                     .update('projects', projectName, cleanEditableAnnotations(data), {projectName: projectName}, {errorNotification: false});
+          },
+          canCreate: function() {
+              return DataService.get("projectrequests", null, {}, { errorNotification: false});
```
spadgett: nit: indentation

app/views/create-from-url.html
```diff
@@ -0,0 +1,76 @@
...
+                      <h2>Create a New Project</h2>
+                      <create-project alerts="alerts" submit-button-label="Next" redirect-action="createWithProject"></create-project>
+                    </div>
+                    <div ng-if="!noProjects && !canCreateProject">
+                      <h2>Choose a Project</h2>
+                      <ui-select ng-model="selected.project">
+                        <ui-select-match placeholder="Project name">
+                          {{$select.selected | uniqueDisplayName : projects}}
+                        </ui-select-match>
+                        <ui-select-choices repeat="project in projects | filter : { metadata: { name: $select.search }}">
```
spadgett: Need to filter on display name as well if that's what we're showing.

app/views/create-from-url.html
```diff
@@ -0,0 +1,76 @@
...
+                    </div>
+                    <div ng-if="!noProjects && canCreateProject">
+                      <uib-tabset>
+                        <uib-tab>
+                          <uib-tab-heading>Choose Existing Project</uib-tab-heading>
+                          <ui-select ng-model="selected.project">
+                            <ui-select-match placeholder="Project name">
+                              {{$select.selected | uniqueDisplayName : projects}}
+                            </ui-select-match>
+                            <ui-select-choices repeat="project in projects | filter : { metadata: { name: $select.search }}">
```
spadgett: Same comment: Need to filter on display name as well.

rhamilto:
@spadgett, comments addressed.

benjaminapetersen:
Btw, tests are in line with [the PageObjects PR](https://github.com/openshift/origin/pull/7124) I opened a good while ago, hoping to more easily iterate after this merges.

app/scripts/controllers/createFromURL.js
spadgett: I'm thinking we shouldn't change this to `name` so it's not confused with the `app` label. It's also consistent with `oc new-app` and the label we uses in the web console.

```
--name='': Set name to use for generated application artifacts
```

spadgett:
@benjaminapetersen @rhamilto thank you for writing a lot of tests on this

benjaminapetersen:
The "public api" of this PR is currently the following:
```javascript
['namespace', 'app', 'imageStream', 'imageTag', 'sourceURI', 'sourceRef', 'contextDir', 'template', 'templateParamsMap'];
```
Just confirming we are good with that.  The only one I was not 100% on was `templateParamsMap`, if it should be shortened to `templateParams` or something similar.

spadgett:
We might change `app` to `name` for consistency with the CLI and field label, @jwforres ?

jwforres:
name makes sense

rhamilto:
I've renamed the query string param "app" to "name".

spadgett:
[merge]

spadgett:
@benjaminapetersen @rhamilto some test failures in jenkins

rhamilto:
@spadgett and @benjaminapetersen, I removed the tests commit from the PR and moved it [to a different branch](https://github.com/rhamilto/origin-web-console/tree/run-on-e2e) so the feature commit can merge.

spadgett:
[merge]

PR #887, opened 2016-11-17
Bug 1333162 Improve descriptions for quota scopes

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1333162

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -1233,10 +1233,10 @@ angular.module('openshiftConsole')
   })
   .filter('scopeDetails', function(sentenceCaseFilter) {
     var scopeMessages = {
-      "Terminating": "Matches pods that have an active deadline.",
-      "NotTerminating": "Matches pods that do not have an active deadline.",
-      "BestEffort": "Matches pods that have best effort quality of service.",
-      "NotBestEffort": "Matches pods that do not have best effort quality of service."
+      "Terminating": "Affects pods that have an active deadline. These pods usually include builds, deployers, and jobs.",
+      "NotTerminating": "Affects pods that do not have an active deadline. These pods usually include your applications.",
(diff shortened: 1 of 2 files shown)


jwforres:
@spadgett PTAL

spadgett:
cc @derekwaynecarr 

We originally used the same wording as the CLI. Do we want to update the CLI as well?

jwforres:
some of these are specific to openshift, like the one that references builds, so if did update the CLI we would probably need to have different text in kubectl vs oc, or just change the text in oc

jwforres:
cc @fabianofranz @juanvallejo for the CLI discussion

juanvallejo:
I think these two updates would be fine for both openshift and kubernetes
```
"BestEffort": "Affects pods that do not have resource limits set. These pods have a best effort quality of service.",
"NotBestEffort": "Affects pods that have at least one resource limit set. These pods do not have a best effort quality of service."
```

Not sure if the concept of "applications" exists upstream, if this is the case, maybe we could list more detail on what consists of an application?
```
"NotTerminating": "Affects pods that do not have an active deadline. These pods usually include your applications.",
```

@fabianofranz WDYT?

jwforres:
I don't think the strings necessarily need to match 100% between the CLI and the console, but @juanvallejo can you open an issue for the CLI in origin to consider changing the CLI strings.

@spadgett can we merge this?

spadgett:
[merge]

juanvallejo:
@jwforres 
> I don't think the strings necessarily need to match 100% between the CLI and the console, but @juanvallejo can you open an issue for the CLI in origin to consider changing the CLI strings.

Sounds good, opened issue here https://github.com/openshift/origin/issues/11985

jwforres:
had to abort previous job, it was stuck [merge]

PR #889, opened 2016-11-17
Show environment variables coming from the builder image in the Environment tab

jwforres:
Fixes https://github.com/openshift/origin/issues/11388
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1357107

### DIFF
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -11,7 +11,9 @@ angular.module('openshiftConsole')
                                                  $filter,
                                                  $routeParams,
                                                  AlertMessageService,
+                                                 APIService,
                                                  BuildsService,
+                                                 ImagesService,
                                                  DataService,
                                                  LabelFilter,
                                                  ProjectsService,
@@ -23,6 +25,7 @@ angular.module('openshiftConsole')
     $scope.alerts = {};
     $scope.breadcrumbs = [];
     $scope.forms = {};
+    $scope.expand = {imageEnv: false};
 
     if ($routeParams.isPipeline) {
       $scope.breadcrumbs.push({
@@ -99,6 +102,8 @@ angular.module('openshiftConsole')
       $scope.forms.bcEnvVars.$setPristine();
     };
 
+    var lastLoadedBuildFromImageKey;
+
     var buildConfigResolved = function(buildConfig, action) {
       $scope.loaded = true;
       $scope.buildConfig = buildConfig;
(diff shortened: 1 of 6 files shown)


jwforres:
bc-img-env-collapsed
bc-img-env-expanded

jwforres:
Implemented for BCs but it is only supported for ImageStreamTag or ImageStreamImage. I hesitate to support this for DockerImage because it is going to put extra load onto the image import API.  I didn't do this for individual builds for the same reason because they have already resolved to a particular DockerImage reference.

@bparees @spadgett PTAL

bparees:
pretty.

jwforres:
We could do the same thing for DCs that have an ICT

app/scripts/controllers/buildConfig.js
```diff
@@ -110,6 +114,28 @@ angular.module('openshiftConsole')
...
         });
       }
+      var buildFrom = _.get(buildStrategy(buildConfig), 'from', {});
+      // We don't want to reload the image every time the BC updates, only load again if the from changes
+      var buildFromImageKey = buildFrom.kind + "/" + buildFrom.name + "/" + (buildFrom.namespace || $scope.projectName);
+      if (lastLoadedBuildFromImageKey !== buildFromImageKey) {
+        if (_.includes(["ImageStreamTag", "ImageStreamImage"], buildFrom.kind)) {
+          lastLoadedBuildFromImageKey = buildFromImageKey;
+          DataService.get(APIService.kindToResource(buildFrom.kind), buildFrom.name, {namespace: buildFrom.namespace || $scope.projectName}).then(function(imageStreamImage){
+            $scope.BCEnvVarsFromImage = _.map(_.get(imageStreamImage, 'image.dockerImageMetadata.Config.Env'),
```
spadgett: Should be able to use `$filter('imageEnv')`

https://github.com/spadgett/origin-web-console/blob/ca4be237df6935819e795eeb2962e985230876e1/app/scripts/filters/resources.js#L287-L298

app/views/browse/build-config.html
```diff
@@ -365,6 +365,21 @@
                   <uib-tab heading="Environment" active="selectedTab.environment" ng-if="buildConfig && !(buildConfig | isJenkinsPipelineStrategy)">
                     <uib-tab-heading>Environment</uib-tab-heading>
                     <h3>Environment Variables</h3>
+                    <div style="margin-top: -5px;" ng-if="BCEnvVarsFromImage.length">
+                      The builder image has additional environment variables defined.  Variables defined below will overwrite any from the image with the same name.
+                      <a href="" ng-click="expand.imageEnv = true" ng-if="!expand.imageEnv">View image environment variables</a>
```
spadgett: "Show image environment variables"? To be consistent with other show/hide links

app/views/browse/build-config.html
```diff
@@ -365,6 +365,21 @@
                   <uib-tab heading="Environment" active="selectedTab.environment" ng-if="buildConfig && !(buildConfig | isJenkinsPipelineStrategy)">
                     <uib-tab-heading>Environment</uib-tab-heading>
                     <h3>Environment Variables</h3>
+                    <div style="margin-top: -5px;" ng-if="BCEnvVarsFromImage.length">
```
spadgett: `class="mar-top-sm"`?

spadgett:
> We could do the same thing for DCs that have an ICT

I think we have the image stream image already to fill in the pod template details.

app/scripts/controllers/buildConfig.js
spadgett: Nevermind, that's looking for a specific key. We might generalize it though or add another filter

app/views/browse/build-config.html
jwforres: its negative margin, i dont think we have a class for applying negative margin?

jwforres:
@spadgett as discussed this will just cover the BC env vars, will work on another PR for DCs

app/scripts/services/images.js
```diff
@@ -213,10 +213,23 @@ angular.module("openshiftConsole")
       return resources;
     };
 
+    var getEnvironment = function(imageStreamImage) {
+      return _.map(_.get(imageStreamImage, 'image.dockerImageMetadata.Config.Env'),
+              function(entry) {
+                var pair = entry.split('=');
+                return {
+                  name: _.head(pair),
+                  value:_.last(pair)
```
spadgett: Not a new problem, but this won't correctly handle env var values with a `=`. Maybe

```js
var index = entry.indexOf('=');
return {
  name: entry.substring(0, index),
  value: entry.substring(index + 1)
};
```

app/views/browse/build-config.html
spadgett: You're right

jwforres:
Yes we have an open / bug issue for that. I can look at fixing as part of
this

jwforres:
@spadgett updated to handle env var values with =

spadgett:
[merge]

PR #893, opened 2016-11-18
Add test for membership service & roleBindings service

benjaminapetersen:
@jwforres @spadgett tests. woohoo.


jwforres:
[merge]

benjaminapetersen:
Cool. the tests for `MembershipService` aren't great, honestly, they don't declare intent well  (compare a json blob with another blob, no fun to debug), but I think the `RolebindingsService` is better.  Will prob revise the first ones.

PR #894, opened 2016-11-18
Update karma to run both firefox & chrome tests

benjaminapetersen:
Updating Karma to run both tests in both Firefox & Chrome concurrently:
- Firefox is much more valuable since we all tend to use Chrome
- Chrome, however, has much better error messages that make you less likely to throw yourself out the window while writing tests.

screen shot 2016-11-18 at 12 54 38 pm

@jwforres @spadgett

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -576,7 +576,15 @@ module.exports = function (grunt) {
     karma: {
       unit: {
         configFile: 'test/karma.conf.js',
-        singleRun: true
+        singleRun: true,
+        // default in karma.conf.js is Firefox, however, Chrome has much better
+        // error messages when writing tests.  Call like this:
+        // grunt test
+        // grunt test --browsers=Chrome
+        // grunt test --browsers=Chrome,Firefox,Safari (be sure karma-<browser_name>-launcher is installed)
+        browsers: grunt.option('browsers') ?
+                    grunt.option('browsers').split(',') :
+                    ['Firefox']
       }
     },
 
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "jasmine-spec-reporter": "1.1.2",
     "jshint-stylish": "0.2.0",
     "karma": "0.12.23",
+    "karma-chrome-launcher": "^2.0.0",
     "karma-coverage": "0.2.6",
     "karma-firefox-launcher": "^1.0.0",
     "karma-jasmine": "0.1.5",
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -79,10 +79,9 @@ module.exports = function(config) {
     // - Safari (only Mac)
     // - PhantomJS
     // - IE (only Windows)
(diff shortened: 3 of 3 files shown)


jwforres:
This has to be optional.  From my testing a few months ago Chrome does not work well inside a container, which means its probably not going to work well in Travis either.  Our tests have to be runnable inside a container for when we move to testing all of our stuff on openshift itself

jwforres:
With most of us developing in Chrome anyway, I'm not sure we get much benefit from running the spec tests in Chrome.

benjaminapetersen:
- Right, let me see if I can make it optional
- Agree, I'd rather always run them in Firefox or any non-chrome.  During dev though, Firefox's error reporting is inferior, switching to Chrome is much easier to debug (only reason I did this).

benjaminapetersen:
how about this?  can now set the browser else default to FF:
```bash
browsers=Chrome grunt test
```

jwforres:
you should do that with a grunt option not an environment variable for consistency with how we do all of our options

benjaminapetersen:
tried that, grunt doesn't exist in the `karma.conf.js` file. Wasn't sure if I should import `grunt` into that file & use `grunt.options('browser')`.  Thoughts?

jwforres:
looks like you can override anything in the config file in the Gruntfile https://github.com/karma-runner/grunt-karma#heres-an-example-that-puts-the-config-in-the-gruntfile

benjaminapetersen:
Aha, I tried this in `karma.conf.js`:
```javascript
    karma: {
      unit: {
        configFile: 'test/karma.conf.js',
        singleRun: true,
        browsers: grunt.options('browsers') ? [grunt.options('browsers')] : ['Firefox']
      }
    },
```
But it doesn't work cuz no `grunt`.  Started to import, then realized I should prob do this in `gruntfile.js` instead:
```javascript
    karma: {
      unit: {
        configFile: 'test/karma.conf.js',
        singleRun: true,
        // defaulting to Firefox for running tests, but recommend Chrome for writing
        // tests due to better error messages:
        //      grunt test --browsers=Chrome
        browsers: grunt.option('browsers') ? [grunt.option('browsers')] : ['Firefox']
      }
    }
```
Seems better?

jwforres:
Yeah generally looks like the right direction to me.
But that doesn't look like it would work correctly if multiple browsers
were specified, if you are only supporting 1 browser running then make the
flag browser (not plural)

benjaminapetersen:
:smile:
yeah, have been trying to figure out multiple, not quite there yet.

jwforres:
i'd probably just go with comma separated and do a string split if the
option exists

benjaminapetersen:
Nice, that works, no quiry 'multiple options for `grunt.options()`' needed.
```bash
grunt test
grunt test --browsers=Chrome
grunt test --browsers=Chrome,Firefox
```

jwforres:
[merge]

PR #895, opened 2016-11-18
Updating annotations directive to always display Show/Hide Annotation…

rhamilto:
…s link

and adding message “There are no annotations on this resource.” if
there are no annotations.

Resolves https://github.com/openshift/origin/issues/7643

Always show link
screen shot 2016-11-18 at 1 53 34 pm

If no annotations, display message
screen shot 2016-11-18 at 1 53 39 pm

If annotations, show annotations
screen shot 2016-11-18 at 1 53 44 pm

@jwforres, PTAL


spadgett:
Should we have the message instead of the "Show annotations" link when there are none?

jwforres:
Personally I thought having the message just hanging out would be more
strange than it appearing

rhamilto:
Why are we hiding annotations anyway?

spadgett:
> Why are we hiding annotations anyway?

They're usually system fields, sometimes with large values like JSON blobs

jwforres:
[merge]

PR #898, opened 2016-11-18
Show 3 catalog columns instead of 4

spadgett:
I think this improves the layout for large descriptions or categories with only a few items. We might even truncate the text sooner.

Before:

screen shot 2016-11-18 at 2 45 26 pm

After:

screen shot 2016-11-18 at 2 44 15 pm

@rhamilto @jwforres opinions?


rhamilto:
Do the xpass categories look really insane with 3 tiles per row?

spadgett:
> Do the xpass categories look really insane with 3 tiles per row?

I actually think they look better.

jwforres:
[merge]

PR #918, opened 2016-11-22
Removing hover opacity change for pod-donut defocused states and …

sg00dwin:
…turning on expand behavior. And center tooltip correctly

Fixes https://github.com/openshift/origin/issues/9253

pod-donut

### DIFF
--- a/app/scripts/directives/podDonut.js
+++ b/app/scripts/directives/podDonut.js
@@ -49,7 +49,7 @@ angular.module('openshiftConsole')
           type: "donut",
           bindto: '#' + $scope.chartId,
           donut: {
-            expand: false,
+            expand: true,
             label: {
               show: false
             },
@@ -74,16 +74,12 @@ angular.module('openshiftConsole')
 
                 // Disable the tooltip for empty donuts.
                 if (id === "Empty") {
-                  return undefined;
+                  return "No pods exist";
                 }
 
                 // Show the count rather than a percentage.
                 return value;
               }
-            },
-            position: function() {
-              // Position in the top-left to avoid problems with tooltip text wrapping.
-              return { top: 0, left: 0 };
             }
           },
           transition: {
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -281,12 +281,28 @@ code.probe-command {
 }
 
 .pod-donut {
+  .c3-defocused {
+    // Removing hover opacity change which caused the "c3-Running" state to look very similar to "c3-Not-Ready" state
+    // https://github.com/openshift/origin/issues/9253
+    opacity: 1 !important;
+  }
+  .c3-tooltip-container {
+    top: -27px !important;
+    left: 50% !important;
+    transform: translateX(-50%);
+    // prevent text wrap
+    white-space: nowrap;
+  }
   // Give an empty pod donut chart a light gray outline.
   path.c3-arc-Empty {
     stroke: #d1d1d1;
     // Override the style set in JavaScript since you can't click or hover
     // on the empty chart.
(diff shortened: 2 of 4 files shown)


app/scripts/directives/podDonut.js
```diff
@@ -49,7 +49,7 @@ angular.module('openshiftConsole')
           type: "donut",
           bindto: '#' + $scope.chartId,
           donut: {
-            expand: false,
+            expand: true,
```
spadgett: This was disabled when we added empty donuts when there are no pods. Can you double check this case?

https://github.com/openshift/origin/pull/5464#discussion-diff-43297113

sg00dwin: yeah, I assumed we wanted to switch behaviors (remove opacity change and add expand). Maybe we can disable the expand for empty donuts?

sg00dwin:
@spadgett reverted expand to false

pod-donut

spadgett:
@sg00dwin Yeah I don't see a way in the API to toggle `expand` after the chart is created for empty charts. It's why we just disabled it entirely.

@jwforres Thoughts on this change?

jwforres:
I do prefer it without the opacity change. But it feels more interactive with the expansion.  How weird does it actually look for the empty donut to expand?

spadgett:
> How weird does it actually look for the empty donut to expand?

It's not awful, but there's no tooltip so a little weird. I just pulses on mouse over.

app/styles/_components.less
```diff
@@ -281,6 +281,19 @@ code.probe-command {
...
+    opacity: 1 !important;
+  }
+  .c3-tooltip-container {
+    // centered above donut
+    left: 50% !important;
+    top: -10px !important;
+    transform: translate(-50%, -50%);
+    // prevent text wrap
+    white-space: nowrap;
+  }
```
spadgett: Can you remove the following lines with this change (and the trailing comma on the line above)?

https://github.com/spadgett/origin-web-console/blob/339a32eff93ba3a1298717fe2e9f82f55ea0d884/app/scripts/directives/podDonut.js#L84-L87

sg00dwin:
urzbgo1sz6

app/styles/_components.less
spadgett: Or maybe we can set just `white-space: nowarp` and let c3 position the tooltip.

spadgett:
We could add a custom tooltip when there are no pods. Or disable the expand in CSS for empty styles (not sure if possible).

jwforres:
Yeah i considered a custom tooltip as well. Something simple like "No pods
exist".   If disabling the expand from CSS for the empty state is possible,
I would prefer that option.

sg00dwin:
I looked closer at the expand to see if there's a way to disable it, using css, for empty pods and found that it's the actual `svg` path coordinates that are being transformed. So css wouldn't effect it.

@spadgett updated pr

sg00dwin:
@ajacobs21e 
Here's the [screenshot gif](https://cloud.githubusercontent.com/assets/1874151/20571334/fd5b1d44-b174-11e6-8e2f-4cdc3ddc505e.gif) showing the hover expand over empty pod issue.

ajacobs21e:
@sg00dwin are you able to add in a tooltip for an empty donut?

ajacobs21e:
Custom tooltip with "No Pods" or "Zero Pods" ? @spadgett

sg00dwin:
@spadgett, @jwforres 
Updated the pr with 'No pods exist' tooltip for empty state.

no-pods-tooltip

spadgett:
[merge]

PR #919, opened 2016-11-22
[WIP] Convert some pod warnings to errors

juanvallejo:
Fixes https://github.com/openshift/origin/issues/7733

This patch displays an "error" highlight on the pod donut if at least
a single container is crash-looping, or has failed.

overview-pafge-with-failed-pod

pods-table-with-failed-pod

@jwforres @spadgett

### DIFF
--- a/app/scripts/directives/podDonut.js
+++ b/app/scripts/directives/podDonut.js
@@ -5,7 +5,7 @@ angular.module('openshiftConsole')
                                   hashSizeFilter,
                                   isPullingImageFilter,
                                   isTerminatingFilter,
-                                  isTroubledPodFilter,
+                                  podWarningsFilter,
                                   numContainersReadyFilter,
                                   Logger,
                                   ChartsService) {
@@ -153,7 +153,10 @@ angular.module('openshiftConsole')
             return 'Terminating';
           }
 
-          if (isTroubledPodFilter(pod)) {
+          var warnings = podWarningsFilter(pod);
+          if (_.some(warnings), { severity: 'error' }) {
+            return 'Failed';
+          } else if (warnings.length) {
             return 'Warning';
           }
 
--- a/app/views/directives/_warnings-popover.html
+++ b/app/views/directives/_warnings-popover.html
@@ -4,7 +4,8 @@
     data-toggle="popover"
     data-trigger="hover"
     data-html="true"
-    class="pficon pficon-warning-triangle-o warnings-popover"
+    class="pficon warnings-popover"
+    ng-class="{'pficon-warning-triangle-o': !hasError, 'pficon-error-circle-o': hasError}"
     aria-hidden="true">
   </span>
   <span class="sr-only">{{content}}</span>
--- a/app/scripts/directives/popups.js
+++ b/app/scripts/directives/popups.js
@@ -85,6 +85,9 @@ angular.module('openshiftConsole')
           if (content) {
             content += '<br>';
           }
+          if (warnings[i].severity === "error") {
+            $scope.hasError = true;
+          }
           content += warnings[i].message;
         }
         $scope.content = content;
--- a/app/scripts/directives/statusIcon.js
+++ b/app/scripts/directives/statusIcon.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module('openshiftConsole')
-  .directive('statusIcon', [
+  .directive('statusIcon',
     function() {
       return {
         restrict: 'E',
@@ -16,4 +16,4 @@ angular.module('openshiftConsole')
         }
       };
     }
-  ]);
+  );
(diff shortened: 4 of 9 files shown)


app/scripts/directives/podDonut.js
```diff
@@ -153,6 +154,13 @@ angular.module('openshiftConsole')
             return 'Terminating';
           }
 
+          // check for 'Failed' phase before checking if pod isTroubled
+          // as isFailingPodFilter shares some of the same container statuses
+          // as isTroubledPodFilter
+          if(isFailingPodFilter(pod)) {
```
juanvallejo: Though this is the easiest way to do this, I was not sure if it was better to have `isTroubledPodFilter` *not* check for the same container errors that `isFailingPodFilter` checks for, and update conditional statements in places like [pods-table.html](https://github.com/openshift/origin-web-console/blob/master/app/views/directives/pods-table.html#L18), for example, to display a `tooltip` with the container error if `isTroubledPod || isFailingPod`

spadgett:
@juanvallejo If we change the donut color to a red error, we should update the alert severity everywhere. For instance, the alert above the service on the overview should use a red x instead of warning yellow.

We might need to change `podWarnings` to something like `podProblems` or `podConditions` and add in a severity. Then we can show the right icon wherever we display them.

juanvallejo:
@spadgett +1 to adding a `severity` to pod warnings. 

pods-page-with-errors-tooltip

monitoring-page-with-error

I think `podWarnings` still fits, even with the added severity to indicate errors. I have pushed my changes for the above images https://github.com/openshift/origin-web-console/pull/919/commits/7ca80adfd64d8df72bf4018a93c16c469df7cd6a though I can go ahead and change this filter name if you still agree that `podWarnings` should be renamed to something more general like `podConditions`.
```

**Comment by spadgett** [comment]
👍 It feels better to call it out as an error when your container is crash looping.

Is it weird to have the `x` icon twice? @jwforres

**Comment by spadgett** [comment]
Wondering if there is a better way for us to show the errors, although most things I've tried don't look great. Maybe we should replace the status-icon with the warnings icon / tooltip if there are warnings?

**Comment by spadgett** [comment]
If we do that, we'd want to use the same red x.

**Comment by juanvallejo** [comment]
@spadgett 

Replaced the status red `x` with a warning icon and all non-error severity messages. (Although all of the warning/error messages are still shown in the first tooltip, should these be filtered to only errors if an error exists?)

pod-with-error-and-non-error-warning

**Comment by juanvallejo** [comment]
@spadgett @jwforres Updated error severity to show the old status icon and no icon or tooltip beside the pod name.

pods-page-old-status-icon

The pod page displays the error tooltip if any errors exist

pod-page-with-error-tooltip

The monitoring page shows the error status icon like it did before

monitoring-page-old-status-icon

Overview page 

overview-page-error

File: app/scripts/directives/podDonut.js [hunk 2]
```diff
@@ -153,7 +154,16 @@ angular.module('openshiftConsole')
             return 'Terminating';
           }
 
+          var warnings = podWarningsFilter(pod);
           if (isTroubledPodFilter(pod)) {
```
spadgett: Shouldn't need to call `isTroubledPodFilter` any longer if you have the warnings. If warnings is not empty, it's troubled.

app/scripts/directives/podDonut.js
```diff
@@ -153,7 +154,16 @@ angular.module('openshiftConsole')
...
+            // if at least one warning has an "error" severity
+            // return a "Failed" pod phase
+            for (var i = 0; i < warnings.length; i++) {
+              if(warnings[i].severity === "error") {
+                return 'Failed';
+              }
+            }
```
spadgett: This is good JS code, but we use lodash to make working with collections this easier.

```js
var warnings = podWarningsFilter(pod);
if (_.some(warnings, { severity: 'error' }) {
  return 'Failed';
}
```

https://lodash.com/docs/3.10.1

app/views/directives/_warnings-popover.html
```diff
@@ -4,7 +4,7 @@
     data-toggle="popover"
     data-trigger="hover"
     data-html="true"
-    class="pficon pficon-warning-triangle-o warnings-popover"
+    ng-class="{pficon: true, 'pficon-warning-triangle-o': !hasError, 'pficon-error-circle-o': hasError, 'warnings-popover': true}"
```
spadgett: I'd leave the regular class attribute for the ones we always have

```
class="pficon warnings-popover"
ng-class="{ 'pficon-warning-triangle-o': !hasError, 'pficon-error-circle-o': hasError }"
```

juanvallejo:
@spadgett Thanks for the feedback, review comments addressed

spadgett:
LGTM, please squash, and I'll merge.

juanvallejo:
@spadgett 
> LGTM, please squash, and I'll merge.

Thanks! done

spadgett:
[merge]

PR #924, opened 2016-11-23
Improve MembershipService tests

benjaminapetersen:
My first round of tests for `MembershipService` were rather terrible.  This is a bit better.
@jwforres @spadgett

### DIFF
--- a/test/spec/services/membership/membershipSpec.js
+++ b/test/spec/services/membership/membershipSpec.js
@@ -60,6 +60,11 @@ describe('MembershipService', function() {
           kind: 'RoleBinding', metadata: { name: 'view' },
           roleRef: { name: 'view' },
           subjects: [{kind: 'User', name: 'jill'}]
+        },
+        systemImagePuller: {
+          kind: 'RoleBinding', metadata: { name: 'system:image-puller' },
+          roleRef: { name: 'system:image-puller' },
+          subjects: [{kind: 'ServiceAccount', name: 'foo', namespace: 'fake-project'}]
         }
       };
     });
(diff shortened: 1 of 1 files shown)


test/spec/services/membership/membershipSpec.js
```diff
@@ -169,19 +186,55 @@ describe('MembershipService', function() {
   });
 
   describe('#getSubjectKinds', function() {
-    // No test needed, this returns static data
+    it('should return the kind list in the preferred ourder', function() {
+      var kinds = MembershipService.getSubjectKinds();
```
jwforres: If this getSubjectKinds is returning a map, there is no guarantee of iteration order. If you are expecting stable order that method probably needs to change.  If its returning an array and not a map I'm not sure why the next line is needed.

benjaminapetersen: Oop, my test is actually wrong, this function isn't supposed to care about sort order.  The 'mapRolebindingsForUI' handles that.

benjaminapetersen:
& rebased

test/spec/services/membership/membershipSpec.js
```diff
@@ -169,19 +186,64 @@ describe('MembershipService', function() {
   });
 
   describe('#getSubjectKinds', function() {
-    // No test needed, this returns static data
+    it('should return the kind list in the preferred ourder', function() {
```
jwforres: typo, order

benjaminapetersen: :+1:

benjaminapetersen: fixed

benjaminapetersen:
@jwforres let me know if you think anything else needs to happen here.  I think these are definitely better than the previous round for testing membership.

jwforres:
[merge]

PR #929, opened 2016-11-28
Show warning tooltip when DC/RS/RC is missing

jhadvig:
Inform user that the DC that belongs to the displayed deployment has been deleted.

In case of RS/RC inform that the RS/RC no longer exists.

@jwforres PTAL

### DIFF
--- a/app/scripts/controllers/replicaSet.js
+++ b/app/scripts/controllers/replicaSet.js
@@ -205,8 +205,17 @@ angular.module('openshiftConsole')
             errorNotification: false
           }).then(
             // success
-            function(deploymentConfig) {
-              $scope.deploymentConfig = deploymentConfig;
+            function() {
+              watches.push(DataService.watchObject("deploymentconfigs", annotation($scope.replicaSet, 'deploymentConfig'), context, function(deploymentConfig, action) {
+                if (action === "DELETED") {
+                  $scope.alerts["deleted"] = {
+                    type: "warning",
+                    message: "This deployment configuration has been deleted."
+                  };
+                  $scope.deploymentConfigMissing = true;
+                }
+                $scope.deploymentConfig = deploymentConfig;
+              }));
             },
             // failure
             function(e) {
@@ -363,6 +372,7 @@ angular.module('openshiftConsole')
                   type: "warning",
                   message: "This " + displayKind + " has been deleted."
                 };
+                $scope.deploymentMissing = true;
               }
               $scope.replicaSet = replicaSet;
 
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/replicaSet.js
```diff
@@ -205,8 +205,17 @@ angular.module('openshiftConsole')
             errorNotification: false
           }).then(
             // success
-            function(deploymentConfig) {
-              $scope.deploymentConfig = deploymentConfig;
+            function() {
+              watches.push(DataService.watchObject("deploymentconfigs", annotation($scope.replicaSet, 'deploymentConfig'), context, function(deploymentConfig, action) {
```
jwforres: I don't feel like its worth adding a watch on deploymentconfigs just to watch if the DC might get deleted, its an edge case.  It would be different if we were already watching it for something else

app/views/browse/deployment-config.html
```diff
@@ -84,7 +84,7 @@
                     </li>
                   </ul>
                 </div>
-                <!-- <span ng-if="deploymentConfig.status.details.message" class="pficon pficon-warning-triangle-o" style="cursor: help;" data-toggle="popover" data-trigger="hover" dynamic-content="{{deploymentConfig.status.details.message}}"></span> -->
+                <span ng-if="deploymentConfigMissing" class="pficon pficon-warning-triangle-o" data-toggle="tooltip" aria-hidden="true" data-original-title="This deployment configuration no longer exists"></span>
```
jwforres: Are we not already showing a message somewhere that the DC does not exist?  I don't think I would treat this the same way as the deployment having a missing deployment config.  The fact that the page you are viewing is for an object which no longer exists is a bigger issue.

app/views/browse/replica-set.html
```diff
@@ -13,12 +13,11 @@
...
                 {{replicaSet.metadata.name}}
                 <span
-                  ng-if="deploymentConfigMissing"
+                  ng-if="deploymentMissing || deploymentConfigMissing"
                   class="pficon pficon-warning-triangle-o"
-                  style="cursor: help; vertical-align: middle;"
                   data-toggle="tooltip"
                   data-trigger="hover"
-                  title="The deployment's deployment config is missing."
+                  data-original-title="{{deploymentConfigMissing ? 'The deployment configuration for this deployment no longer exists.' : 'This ' + (kind | humanizeKind ) + ' is missing.'}}"
```
jwforres: Same comment as above, if the page you are viewing is for something that does not exist, thats a bigger deal than a little warning icon next to the name.

Issue #904, opened 2016-11-21
Align Table Columns When Possible


stevekuznetsov:
Unclear if this is technically difficult to do, but on pages with lots of tables, it would be nice to have the columns aligned when possible, for example this looks a little jumbled:

![tables](https://cloud.githubusercontent.com/assets/7328370/20489550/02cdf3c0-afd9-11e6-9f6f-082383598dc4.png)

spadgett:
We could use fixed column widths.

spadgett:
cc @rhamilto

PR #933, opened 2016-11-28
Set column widths so columns align across multiple tables per page

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/904

table-col-sizes

### DIFF
--- a/app/styles/_tables.less
+++ b/app/styles/_tables.less
@@ -97,6 +97,9 @@
     border-top: 2px solid @table-border-color;
     // so that .word-break() will work
     table-layout: fixed;
+    colgroup, col {
+      display: none;
+    }
     thead, tbody, th, td, tr {
       display: block;
     }
@@ -195,5 +198,15 @@
 .config-map-table {
   .key, .truncated-content {
     font-family: @font-family-monospace;
+  }
+}
+
+.table-word-break-all td {
+  // Break long non-breaking strings by default
+  .word-break-all;
+  // Override specific cells
+  &.word-break-normal {
+    word-break: normal;
+    overflow-wrap: normal;
   }
 }
--- a/app/views/secrets.html
+++ b/app/views/secrets.html
@@ -21,7 +21,11 @@
           <div ng-if="loaded" class="row">
             <div class="col-md-12">
               <h2>Source Secrets</h2>
-              <table class="table table-bordered table-hover table-mobile secrets-table">
+              <table class="table table-bordered table-hover table-mobile secrets-table table-word-break-all">
+                <colgroup>
+									<col class="col-sm-5">
+									<col class="col-sm-5">
+								</colgroup>
                 <thead>
                   <tr>
                     <th>Name</th>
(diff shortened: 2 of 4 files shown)


app/styles/_tables.less
```diff
@@ -197,3 +197,22 @@
...
+		@media (min-width: 992px) { width: 42%;}
+		@media (min-width: 1200px) { width: 44%;}
+	}
+	&:nth-child(2) {
+		.word-break-all;
+		@media (min-width: 768px) { width: 40%;}
+		@media (min-width: 992px) { width: 42%;}
+		@media (min-width: 1200px) { width: 44%;}
+	}
+}
```
spadgett: I'd rather give the columns class names and set the styles based on class name. nth child can break if we ever change the table contents, and it won't be obvious editing the view.

```css
.name-col, .type-col {
  .word-break-all;
  @media (min-width: 768px) { width: 40%;}
  @media (min-width: 992px) { width: 42%;}
  @media (min-width: 1200px) { width: 44%;}
}
```

spadgett: By the way, you could use `:nth-child(-n+2)` to match the first two columns, but I still prefer the approach of giving the columns class names.

rhamilto: Rather than hard-coding these widths and breakpoints to a specific class, is it conceivable to utilize the existing bootstrap column widths and <col> elements on the table?  See http://stackoverflow.com/a/28228661

sg00dwin:
@spadgett @rhamilto 
Switch to `colgroup` `col` and bootstrap grid to control column widths (not used at mobile). And set word-break-all (FF and IE need `word-break: break-all`) for non-breaking strings to wrap. And reset to normal for 'Created' column.

table-col-sizes

app/styles/_tables.less
```diff
@@ -197,3 +200,15 @@
     font-family: @font-family-monospace;
   }
 }
+
+.secrets-table {
```
rhamilto: Should this be more generic so it could be used on other tables?

sg00dwin:
... now generic ftw!

app/styles/_tables.less
```diff
@@ -197,3 +200,15 @@
     font-family: @font-family-monospace;
   }
 }
+
+.table-word-break-all {
+  td {
```
rhamilto: Nit pick.  Unnecessary nesting since there is only one rule.

sg00dwin:
@rhamilto @spadgett updated

app/styles/_core.less
```diff
@@ -737,7 +737,7 @@ a.subtle-link {
   .task-expanded-details {
     margin-top: 20px;
     .alert {
-      margin-bottom: 0;
+      margin-bottom: 7px;
```
rhamilto: Did you intend for this change to be part of this PR?

sg00dwin: no doh!

rhamilto:
LGTM.  Ready for merge, @spadgett.

spadgett:
[merge]

PR #937, opened 2016-11-28
Change order of resources on monitoring page

### LINKED ISSUE #931: Reverse the order of sections on the monitoring page
image

It seems like most people would be going to the monitoring page to see running pods.  While it's nice to see builds and deployments here (ignoring that deployments show you the running pods after the deploy has happened), I would think they should show up below pods on the page.  Flipping the order of pods and builds seems appropriate.

spadgett:
Show pods, then deployments, then builds

Fixes #931

@jwforres @danmcp

### DIFF
--- a/app/scripts/controllers/monitoring.js
+++ b/app/scripts/controllers/monitoring.js
@@ -35,22 +35,21 @@ angular.module('openshiftConsole')
 
     $scope.kinds = [
       {
-        kind: "Pods"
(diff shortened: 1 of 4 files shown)


spadgett:
@jwforres Diff is not as bad as it looks... I just moved the sections around.

jwforres:
[merge]

PR #940, opened 2016-11-28
Change "View History" to "View Pipeline Runs"

### LINKED ISSUE #936: View History is confusing on the pipeline page
screenshot from 2016-11-28 14-27-03

It's confusing because:
- It takes you to the summary page and not a history page.
- It shows you the currently running pipelines and not just history
- It does the same thing as the named link about it.  The rest of the UI doesn't have the view history link.

I think it either needs to be renamed to something like:
- View Pipeline Runs
- View Summary

Or at least be consistent in the rest of the UI's terminology.  Or it should be removed.

spadgett:
Also change the edit label to "Edit Pipeline" for pipeline builds.
This is consistent with the pipelines page.

Fixes #936
@jwforres @danmcp


jwforres:
[merge]

PR #947, opened 2016-11-29
Add in-context Jenkinsfile help

spadgett:
Show information about the Jenkinsfile in a modal for pipeline build configs and builds. We give a help link builds that use Jenkinsfile path as well.

https://trello.com/c/nEjWsvyL

openshift_web_console

openshift_web_console

Show the examples below the editor when editing a pipeline so that you can copy and paste.

openshift_web_console

openshift_web_console

@jwforres @bparees @gabemontero @danmcp @sspeiche

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -270,6 +270,7 @@
         <script src="scripts/controllers/modals/confirmReplaceModal.js"></script>
         <script src="scripts/controllers/modals/processTemplateModal.js"></script>
         <script src="scripts/controllers/modals/linkService.js"></script>
+        <script src="scripts/controllers/modals/jenkinsfileExamplesModal.js"></script>
         <script src="scripts/controllers/about.js"></script>
         <script src="scripts/controllers/commandLine.js"></script>
         <script src="scripts/controllers/createPersistentVolumeClaim.js"></script>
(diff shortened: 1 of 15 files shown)


danmcp:
These are awesome.  What about the case where the jenkins file isn't in the template?

spadgett:
> What about the case where the jenkins file isn't in the template?

I added a link to open the same help dialog when using a Jenkinsfile path instead of an embedded Jenkinsfile.

openshift_web_console

jwforres:
changes LGTM, will give the others time for feedback

bparees:
lgtm

jwforres:
[merge]

sspeiche:
Very helpful, help icon not enough to launch this?

spadgett:
> help icon not enough to launch this

We could do that, although we've only used the help icon for tooltips elsewhere.

@jwforres @ajacobs21e opinion?

jwforres:
yeah we usually use ? for tooltips, in this particular case i think its
nice to see the "What's a Jenkinsfile?"

i'm not against using the ? though,  @ajacobs21e ?

ajacobs21e:
(?) should be for popovers only. 

There isn't anything in PatternFly now about what to call these types of links. UX is kind of split on whether or not we should use questions in a UI - I personally would prefer to not use a question for a link. Other options are "Learn More" or "Learn more about Jenkinsfiles". However since this hasn't been decided in PF yet and there's no consistency across products either option is fine for now.

ajacobs21e:
@spadgett @jwforres After talking with UX and thinking some more, learn more isn't quite right either. 

"Show examples" is clear what you're going to see after clicking the link.
" > Show examples" for the edit case since the examples show in an expansion. 

We could use "Show jenkinsfile examples" but that is getting kind of long and is probably redundant if it's with the jenkinsfile section.

PR #950, opened 2016-11-29
Create new config maps and secrets when adding config files

spadgett:
Let users create a new config map or secret directly from the "add config files" page.

https://trello.com/c/gUzfL5OA

openshift_web_console

@jwforres PTAL


spadgett:
Changed link casing to "Change secret"

app/views/add-config-volume.html
```diff
@@ -22,18 +22,7 @@
                 </div>
 
                 <div ng-if="targetObject && configMaps && secrets">
-                  <div ng-if="!configMaps.length && !secrets.length" class="empty-state-message empty-state-full-page">
```
jwforres: Should we keep this for the case where you aren't authorized to create secrets / config maps?

spadgett:
@jwforres updated

jwforres:
[merge]

PR #954, opened 2016-11-29
Fix "What's a Jenkinsfile" link positioning on build page

spadgett:
@jwforres PTAL, the link positioning isn't right for builds. This matches the build config page.


jwforres:
[merge]

PR #956, opened 2016-11-30
Add `canI` checks to editors

spadgett:
Check if the user can update or create resources in our editor pages
before trying to make the change. We remove the actions that take you to
these pages if you don't have authority, but someone could send you a
direct link or your authority might have changed since you loaded the
previous page.

@jwforres PTAL


jwforres:
[merge]

PR #958, opened 2016-11-30
Show a "Start Pipeline" button on overview

### LINKED ISSUE #928: Remove 'Start Deployment' on overview from DCs that are part of a pipeline
When a deployment config is part of a pipeline, we shouldn't show the 'Start Deployment' button when there are no deployments.  Instead we should probably say something like "This deployment is part of the pipeline **foo**"

start-deployment

spadgett:
When a deployment config has no deployments and the deployment config is
part of a pipeline, link to the pipeline build config and show a "Start
Pipeline" button in place of the "Start Deployment" button.

openshift_web_console

Fixes #928

@jwforres PTAL
@danmcp FYI

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -401,6 +401,7 @@ angular.module('openshiftConsole')
     };
 
     var buildConfigForBuild = $filter('buildConfigForBuild');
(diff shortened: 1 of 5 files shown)


jwforres:
[merge]

PR #959, opened 2016-11-30
Fix loading message position on create secret page

spadgett:
The loading message had no top margin.

@jwforres PTAL


jwforres:
[merge]

PR #960, opened 2016-11-30
CLI and About menu items should have same ordering on mobile

spadgett:
Currently they have a different order at mobile vs full screen.

@jwforres PTAL

### DIFF
--- a/app/scripts/extensions/nav/dropdownMobile.js
+++ b/app/scripts/extensions/nav/dropdownMobile.js
@@ -17,8 +17,8 @@ angular.module('openshiftConsole')
           type: 'dom',
           node: [
             '<li>',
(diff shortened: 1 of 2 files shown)


app/scripts/extensions/nav/helpDropdown.js
```diff
@@ -17,6 +17,10 @@ angular.module('openshiftConsole')
           },
           {
             type: 'dom',
+            node: '<li class="divider"></li>'
```
jwforres: keep in mind in online we add menu items to the bottom of this menu, is that going to look weird with this divider added

spadgett:
@jwforres updated

jwforres:
[merge]

PR #963, opened 2016-12-01
Improve display of secrets

spadgett:
* Decode secret values that don't have non-printable characters
* Provide a copy-to-clipboard button for secret values when revealed
* Improve styling of revealed secrets

openshift_web_console

@jwforres PTAL
@smarterclayton FYI


jwforres:
[merge]

PR #966, opened 2016-12-01
Display Lifecycle hooks on the DC browse page

jhadvig:
I've created 3 previews on how we could display lifecycle hooks on the DC browse page. All three previews have similar structure, they only differ by the section where the particular hook description is  displayed and the position of the action that the hook is doing, which is either right from the hook type, or under the hook type (`text-muted` class)

1. As part of the **Configuration** section
screenshot-67
screenshot-1

2. Separate section in the `Details` tab
screenshot-68
screenshot
Here I was not sure where to put the section so I've put it under the `Triggers` section

3. A separate `Lifecycle Hooks` tab since as you can see on the previous screens the `Details` tab can get pretty messy whe it contains a lot of things.
screenshot-69

IMHO I like the last option with the new tab, since it makes things more clean&clear. But I also understand that the hooks are more or less part of a "Configuration". On the other hand I kinda like the idea to split the some sections of the `Details` tab into a standalone tabs... eg: Autoscaling or Triggers ?!?

@jwforres @spadgett please let me know what what do you think. Please dont review the code for now ..was just prototyping. Thanks :)


spadgett:
I like having the separate tab. Shorten it to just "Hooks"?

I'm not sure about using the pod template styling for hooks. We currently use it as a distinct visual style for pod templates. Maybe something closer to the styling we use for dc details.

Otherwise looks good!

jhadvig:
yeah regarding similar styling that we have for the pod template.. I've reused it since it's using the concept of separating the details per container and thats basically what the lifecycle hooks do.

jwforres:
Honestly I prefer option 2.  I think if we were going to split stuff into another tab, we should really take a look at everything we have on that Details tab and come up with a logical split.  Hooks as its own tab doesn't make sense to me.  I'd expect a split more like (not necessarily this order):

How am I deploying:
- Strategy and params
- Hooks
- Scaling / Autoscaling
- Triggers

What am I deploying:
- Pod template details
- Volumes

jhadvig:
@spadgett @jwforres  here is preview of the styling we have in dc details in a separate tab:
screenshot-2
screenshot-3

Here is preview of the styling we have in dc details in the Details tab:
screenshot-4
screenshot-5

jhadvig:
I liked the the pod-template like styling more, but dont have any major objection against the styling in the last comment as well (exepct https://github.com/openshift/origin-web-console/pull/966#issuecomment-264213424)

jhadvig:
I've updated the PR based on the conversation with @spadgett ..
Here is the screen on the how the Hooks will be displayed:
screenshot-6

PTAL

app/scripts/filters/resources.js
```diff
@@ -1012,6 +1012,14 @@ angular.module('openshiftConsole')
       return 'PAUSED_PENDING_INPUT' === stage.status;
     };
   })
+  .filter('strategyNameToParams', function () {
+    return function(strategyName) {
+      if (!strategyName) {
+        return null;
+      }
+      return strategyName.toLowerCase() + 'Params';
```
spadgett: We shouldn't assume this will always work. I'd rather have a switch statement for all the strategies we know about.

app/views/browse/deployment-config.html
```diff
@@ -129,17 +129,24 @@
...
+
+                            <div ng-if="deploymentConfig.spec.strategy.rollingParams || deploymentConfig.spec.strategy.recreateParams">
                               <dt>Timeout:</dt>
-                              <dd>{{deploymentConfig.spec.strategy.rollingParams.timeoutSeconds}} sec</dd>
-                              <dt>Max Unavailable:</dt>
-                              <dd>{{deploymentConfig.spec.strategy.rollingParams.maxUnavailable}}</dd>
-                              <dt>Max Surge:</dt>
-                              <dd>{{deploymentConfig.spec.strategy.rollingParams.maxSurge}}</dd>
+                              <dd>{{deploymentConfig.spec.strategy[strategyParams].timeoutSeconds}} sec</dd>
+                              <div ng-if="deploymentConfig.spec.strategy.rollingParams">
```
spadgett: While you're in the code, can you change this to...?

```html
<dt ng-if-start="deploymentConfig.spec.strategy.rollingParams">Update Period:</dt>
<dd>{{deploymentConfig.spec.strategy.rollingParams.updatePeriodSeconds}} sec</dd>
...
<dt>Max Surge:</dt>
<dd ng-if-end>{{deploymentConfig.spec.strategy.rollingParams.maxSurge}}</dd>
```

Then you don't have the unnecessary div in the middle of the `dl`.

app/views/browse/deployment-config.html
```diff
@@ -198,6 +205,27 @@
                           </div>
                         </div>
 
+                        <div class="col-lg-6" ng-if="deploymentConfig.spec.strategy.type !== 'Custom'">
+                          <h3>
+                            Hooks
+                            <a href="{{'lifecycle_hooks' | helpLink}}" style="display: inline-block;" target="_blank">
```
spadgett: Let's not use inline styles. Does `class="learn-more-inline"` work here?

app/views/browse/deployment-config.html
```diff
@@ -198,6 +205,27 @@
...
+                              <span class="learn-more-block">Learn more <i class="fa fa-external-link"> </i></span>
```
spadgett: Should be using `learn-more-inline` not `learn-more-block`.

Can you try

```html
<span class="learn-more-inline">
  <a ng-href="{{'lifecycle_hooks' | helpLink}}" target="_blank">Learn more&nbsp;<i class="fa fa-external-link" aria-hidden="true"></i></a>
</span>
```

app/scripts/directives/lifecycleHook.js
```diff
@@ -130,4 +130,15 @@ angular.module("openshiftConsole")
...
       }
     };
+  })
+  .directive("lifecycleHook", function() {
+    return {
+      restrict: 'E',
+      scope: {
+        hookParams: "=",
+        namespace: "=",
+        type: "@"
```
spadgett: I think this is better as

```js
scope: {
  deploymentConfig: '=',
  type: '@'  // "pre", "mid", or "post"
}
```

Should be able to determine strategy from the deployment config spec, correct? Then you don't need to pass namespace because it's in the deployment config metadata.

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,56 @@
...
+  </span>
+</h4>
+<dl class="dl-horizontal left">
+  <div ng-if="hookParams.execNewPod">
+    <h5 class="container-name">Container {{hookParams.execNewPod.containerName}}</h5>
+    <dl class="dl-horizontal left">
+      <div class="is-item-description">
+        <dt>Command:</dt>
+        <dd>
+          <code class="probe-command">
```
spadgett: I'm glad we're reusing these styles, but we should generalize the class name if it's not only used for health checks.

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,56 @@
...
-              <label class="required">Tag As</label>
-              <istag-select model="istagHook" allow-custom-tag="true" select-disabled="view.isDisabled"></istag-select>
+<h4>
+  {{type}} Hook
+  <small class="text-muted">{{hookParams.execNewPod ? "Runs a command in a new pod based on the specified container" : "Tags image on deployment success"}}</small>
+  <span ng-switch="hookParams.failurePolicy">
+    <small ng-switch-when="Ignore">(Continue with deployment on failure)</small>
+    <small ng-switch-when="Abort">(Abort deployment on failure)</small>
+    <small ng-switch-when="Retry">(Continue with deployment on failure)</small>
+  </span>
```
spadgett: I don't think we should style this as help text since it's actual data in the hook and is not always the same value. I'd say we call it out specifically as failure policy below the header.

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,56 @@
...
+            <span ng-repeat="arg in hookParams.execNewPod.command">
+              <truncate-long-text content="arg" limit="80" newline-limit="1" expandable="false" use-word-boundary="false"></truncate-long-text>
+            </span>
+          </code>
+        </dd>
+        <div ng-if="hookParams.execNewPod.env">
+          <dt>Environment Variables:</dt>
+          <dd>
+            <div ng-repeat="env in hookParams.execNewPod.env">
+              {{env.name}}&#8201;&#8594;&#8201;{{env.value}}
```
spadgett: I think `{{env.name}}={{env.value}}` is more natural for environment variables.

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,56 @@
...
+    <small ng-switch-when="Ignore">(Continue with deployment on failure)</small>
+    <small ng-switch-when="Abort">(Abort deployment on failure)</small>
+    <small ng-switch-when="Retry">(Continue with deployment on failure)</small>
+  </span>
+</h4>
+<dl class="dl-horizontal left">
+  <div ng-if="hookParams.execNewPod">
+    <h5 class="container-name">Container {{hookParams.execNewPod.containerName}}</h5>
+    <dl class="dl-horizontal left">
+      <div class="is-item-description">
```
spadgett: I don't see where this CSS class is defined.

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,56 @@
...
-  <span>
-    <a href="" role="button" ng-if="!hookParams" ng-click="addHook()">Add {{type}} lifecycle hook</a>
-    <a href="" role="button" ng-if="hookParams" ng-click="removeHook()">Remove {{type}} lifecycle hook</a>
-  </span>
-</ng-form>
+  <div ng-if="hookParams.tagImages">
+    <div ng-repeat="tagImage in hookParams.tagImages">
+    <h5 class="container-name">Container {{tagImage.containerName}}</h5>
+      <dl class="dl-horizontal left">
+        <div class="is-item-description">
```
spadgett: I don't see where this CSS class is defined.

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,56 @@
...
-              <div class="help-block">
-                Use the image for this container as the source of the tag.
-              </div>
-            </div>
-            <div class="form-group">
-              <label class="required">Tag As</label>
-              <istag-select model="istagHook" allow-custom-tag="true" select-disabled="view.isDisabled"></istag-select>
+<h4>
+  {{type}} Hook
+  <small class="text-muted">{{hookParams.execNewPod ? "Runs a command in a new pod based on the specified container" : "Tags image on deployment success"}}</small>
```
spadgett: Same comment as failure policy: I don't think we should style this as help text because it's a configuration value that changes. I'd rather describe the pre/mid/post hook type in help text, then show `execNewPod` and `failurePolicy` clearly as configuration below the header.

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,56 @@
...
-  </span>
-</ng-form>
+  <div ng-if="hookParams.tagImages">
+    <div ng-repeat="tagImage in hookParams.tagImages">
+    <h5 class="container-name">Container {{tagImage.containerName}}</h5>
+      <dl class="dl-horizontal left">
+        <div class="is-item-description">
+          <dt>Tag As:</dt>
+          <dd>
+            <a ng-href="{{tagImage.to.name | navigateResourceURL : 'ImageStreamTag' : namespace}}">
```
spadgett: Do you need to check first for `tagImage.to.namespace`? Then fall back `namespace` if not defined. (If `tagImage.to.namespace && tagImage.to.namespace !== namespace`, it should not be a link.)

jhadvig:
@spadgett I've updated the PR based on your comments.

Here is the screen of the Hooks in the DC browse page:
screenshot-8

Wasnt really sure how to properly name the `execNewPod` and `tagImages` actions, so I named it `Action` and put its description next to it.
Also added a tooltip to the `failurePolicy` to explain what the value does.

PTAL

app/scripts/directives/lifecycleHook.js
```diff
@@ -130,4 +130,17 @@ angular.module("openshiftConsole")
...
+  .directive("lifecycleHook", function($filter) {
+    return {
+      restrict: 'E',
+      scope: {
+        deploymentConfig: '=',
+        type: '@'  // "pre", "mid", or "post"
+      },
+      templateUrl: 'views/directives/lifecycle-hook.html',
+      controller: function($scope) {
+        $scope.strategyParams = $filter('deploymentStrategyNameToParams')($scope.deploymentConfig.spec.strategy.type);
```
spadgett: You should watch the type in case it changes.

```js
$scope.$watch('deploymentConfig.spec.strategy.type', function(type) {
  $scope.strategyParams = $filter('deploymentStrategyNameToParams')(type);
});
```

app/scripts/filters/resources.js
```diff
@@ -1012,6 +1012,20 @@ angular.module('openshiftConsole')
       return 'PAUSED_PENDING_INPUT' === stage.status;
     };
   })
+  .filter('deploymentStrategyNameToParams', function () {
```
spadgett: I'd rather the filter give me the strategy params instead of the property name.

```js
.filter('strategyParams', function() {
  return function(deploymentConfig) {
    var strategy = _.get(deploymentConfig, 'spec.strategy.type');
    switch (strategy) {
    case 'Recreate':
      return _.get(deploymentConfig, 'spec.strategy.recreateParams', {});
    // ...
    default:
      return null;  
    }
  };
});
```

Then in the directive `link` function, you can get the strategy params with

```js
$scope.$watch('deploymentConfig', function(deploymentConfig) {
  $scope.strategyParams = $filter('strategyParams')(deploymentConfig);
});
```

(Should have made that comment on first review, sorry.)

app/styles/_components.less
```diff
@@ -188,7 +188,7 @@
   }
 }
 
-code.probe-command {
+code.command {
```
spadgett: 👍

app/views/browse/deployment-config.html
```diff
@@ -198,6 +198,27 @@
                           </div>
                         </div>
 
+                        <div class="col-lg-6" ng-if="deploymentConfig.spec.strategy.type !== 'Custom'">
+                          <h3>
+                            Hooks
+                            <span class="learn-more-inline">
+                              <a ng-href="{{'lifecycle_hooks' | helpLink}}" target="_blank">Learn more&nbsp;<i class="fa fa-external-link" aria-hidden="true"></i></a>
```
spadgett: Should be "Learn More" (uppercase M) now. Changed the doc links in #978.

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,59 @@
...
-    <div class="help-block" ng-switch-when="pre">Pre hooks execute before the deployment begins.</div>
-    <div class="help-block" ng-switch-when="mid">Mid hooks execute after the previous deployment is scaled down to zero and before the first pod of the new deployment is created.</div>
-    <div class="help-block" ng-switch-when="post">Post hooks execute after the deployment strategy completes.</div>
+<h4>
+  {{type | upperFirst}} Hook
+</h4>
+<span ng-switch="type">
+  <small class="text-muted" ng-switch-when="pre">Pre hooks execute before the deployment begins.</small>
+  <small class="text-muted" ng-switch-when="mid">Mid hooks execute after the previous deployment is scaled down to zero and before the first pod of the new deployment is created.</small>
+  <small class="text-muted" ng-switch-when="post">Post hooks execute after the deployment strategy completes.</small>
```
spadgett: Instead of `small`, let's use the `help-block` style. (Margin looked off in the screenshot, but I think `help-block` will fix that.)

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,59 @@
...
+</span>
+<dl class="dl-horizontal left">
+  <dt>Action:</dt>
+  <dd>{{deploymentConfig.spec.strategy[strategyParams][type].execNewPod ? "Runs a command in a new pod based on the specified container" : "Tags image on deployment success"}}</dd>
```
spadgett: I'd probably shorten these to "Run a command" and "Tag the image"

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,59 @@
...
-          <div class="help-block" ng-switch-when="Abort">Fail the deployment if the hook fails.</div>
-          <div class="help-block" ng-switch-when="Ignore">Ignore hook failures and continue the deployment.</div>
-        </div>
-      </div>
-    </fieldset>
+  <div ng-if="deploymentConfig.spec.strategy[strategyParams][type].tagImages">
+    <div ng-repeat="tagImage in deploymentConfig.spec.strategy[strategyParams][type].tagImages">
+    <h5 class="container-name">Container {{tagImage.containerName}}</h5>
+      <dl class="dl-horizontal left">
+        <dt>Tag As:</dt>
```
spadgett: If there is only ever one item in this `dl`, I'd just put it in a `div`.

```html
<div>
  Tag image as <a href=...>myproject/image:latest</a>
</div>
```

jhadvig:
@spadgett comments addressed

jhadvig:
Adding screen:

screenshot-12

app/views/directives/lifecycle-hook.html
```diff
@@ -1,133 +1,58 @@
...
-  <div ng-switch="type">
-    <div class="help-block" ng-switch-when="pre">Pre hooks execute before the deployment begins.</div>
-    <div class="help-block" ng-switch-when="mid">Mid hooks execute after the previous deployment is scaled down to zero and before the first pod of the new deployment is created.</div>
-    <div class="help-block" ng-switch-when="post">Post hooks execute after the deployment strategy completes.</div>
+<h4>
+  {{type | upperFirst}} Hook
+  <span ng-switch="type">
+    <small ng-switch-when="pre">&ndash; runs before the deployment begins.</small>
+    <small ng-switch-when="mid">&ndash; runs after the previous deployment is scaled down to zero and before the first pod of the new deployment is created.</small>
+    <small ng-switch-when="post">&ndash; runs after the deployment strategy completes.</small>
```
spadgett: nit: remove the period from the end of these messages

spadgett:
Instead of truncate long text, does `class="truncate"` work? Since it's a single line, I prefer to use CSS. Then it will truncate exactly at the edge of the screen.

spadgett:
If we switch to `class="truncate"`, set a `title` attribute with the full value as well.

jhadvig:
@spadgett updated :)
added the `dist` as well
PTAL

spadgett:
[merge]

PR #967, opened 2016-12-01
Fix error displaying catalog in openshift namespace

spadgett:
Fixes https://github.com/openshift/origin/issues/12094

@jwforres PTAL. We were getting a "duplicates in a repeater" error because multiple templates with the same UID were in the `ng-repeat` (each template twice).


jwforres:
[merge]

PR #974, opened 2016-12-02
Bug 1400775 - Disable the Create button when Configuration File field didn't pass the validation check

jhadvig:
@spadgett PTAL

### DIFF
--- a/app/views/directives/create-secret.html
+++ b/app/views/directives/create-secret.html
@@ -116,7 +116,7 @@
             fadeFoldWidgets: true,
             showPrintMargin: false
           }
-        }" ng-model="newSecret.data.cacert" class="create-secret-editor ace-bordered" id="cacert-editor"></div>
+        }" ng-model="newSecret.data.cacert" class="create-secret-editor ace-bordered" id="cacert-editor" required></div>
       </div>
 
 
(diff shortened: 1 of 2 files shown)


app/views/directives/create-secret.html
```diff
@@ -292,7 +292,7 @@
   <div class="buttons gutter-top-bottom">
     <button class="btn btn-lg btn-primary"
             type="button"
-            ng-disabled="secretForm.$invalid || secretForm.$pristine"
+            ng-disabled="secretForm.$invalid || secretForm.$pristine || invalidConfigFormat"
```
spadgett: We should be calling `$setValidity` in the controller to mark the field invalid, and then just check `$invalid` here as before. For instance, give the input `name="dockerConfig"`, then in the controller (untested)

```js
        var updateEditorMode = _.debounce(function(){
          try {
            JSON.parse($scope.newSecret.data.dockerConfig);
            $scope.$apply(function() {
              $scope.secretForm.dockerConfig.$setValidity('validJSON', true);
            });
          } catch (e) {
            $scope.$apply(function() {
              $scope.secretForm.dockerConfig.$setValidity('validJSON', false);
            });
          }
        }, 300, {
          'leading': true
        });
```

Use `$scope.$apply` because the function is debounced.

In the view, show the error with

```html
        <div class="has-warning" ng-if="secertForm.dockerConfig.$error.validJSON">
          <span class="help-block">
            Configuration file should be valid JSON.
          </span>
        </div>
```

spadgett:
I noticed a type here in the view:

```html
        <osc-file-input
          if="dockercfg-file-input"
          model="newSecret.data.dockerConfig"
          drop-zone-id="docker-config"
          help-text="Upload a .dockercfg or .docker/config.json file"
          show-values="false"
          required="true"></osc-file-input>
```

`if` should be `id`, correct?

jhadvig:
Indeed it should.. will update right away

spadgett:
[merge]

PR #976, opened 2016-12-02
Show correct image stream name on deploy image

### LINKED ISSUE #973: Deploying an ImageStream Tag displays the wrong ImageStream name
When deploying an image through **Add to Project** &rarr; **Deploy Image** &rarr; **Image Stream Tag**, if the `ImageStream` name is different from the application name, wrong `ImageStream` name is displayed since it assumes the `ImageStream` and application names are the same. In the following example the `ImageStream` being deployed is `myapp-dev:preprod` and the application name is `myapp-preprod`.  In the middle of the screen, the image to be deployed is displayed as `myapp-preprod:preprod` which does not exist.

deploy-is

spadgett:
If I deploy from an image stream image and customize the app name, don't
change the image stream name in the summary header.

openshift_web_console

Fixes #973
@jwforres PTAL

### DIFF
--- a/app/views/directives/deploy-image.html
+++ b/app/views/directives/deploy-image.html
@@ -73,7 +73,8 @@
         administrator.
       </div>
       <h2>
-        {{app.name}}<span ng-if="import.tag">:{{import.tag}}</span>
+        <span ng-if="mode === 'dockerImage'">{{import.name}}</span>
+        <span ng-if="mode === 'istag'">{{istag.imageStream}}<span ng-if="import.tag">:{{import.tag}}</span></span>
         <small>
           <span ng-if="import.result.ref.registry">from {{import.result.ref.registry}},</span>
           <span am-time-ago="import.image.dockerImageMetadata.Created"></span>,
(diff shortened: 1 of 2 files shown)


app/views/directives/deploy-image.html
```diff
@@ -73,7 +73,8 @@
         administrator.
       </div>
       <h2>
-        {{app.name}}<span ng-if="import.tag">:{{import.tag}}</span>
+        <span ng-if="mode === 'dockerImage'">{{app.name}}<span ng-if="import.tag">:{{import.tag}}</span></span>
```
jwforres: hmm does it make sense to be changing that header at all? isn't that header the name and metadata of the image?

spadgett: Yeah we can change it for the other case. The difference is that we will actually create an image stream tag with that value in your project for `mode === 'dockerImage'`

jwforres: Do we have a line that tells you we will create an image stream?  That seems like the right thing vs changing the name of the image that we loaded.

spadgett: Yes, I'll update the PR

spadgett:
@jwforres another look?

jwforres:
[merge]

PR #977, opened 2016-12-02
Warn when Git URL is not an absolute URL

spadgett:
Warn about some Git URLs that the API server will allow, but won't actually work for builds.

openshift_web_console

https://bugzilla.redhat.com/show_bug.cgi?id=1225420

@jwforres PTAL
@bparees CC


jwforres:
[merge]

PR #979, opened 2016-12-02
Prevent veritcal scrollbars on the edit yaml page

spadgett:
Adjust editor height to prevent the scrollbars from showing at certain window heights.

@jwforres


jwforres:
[merge]

PR #981, opened 2016-12-02
[WIP] improve success msg on dc and bc env tab

juanvallejo:
Related Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1368869

This patch improves the success message a user receives when editing
environment variables for a deploymentconfig and a buildconfig on the
key-value editor under the `environment` tab.

I was hoping to begin improving these messages with this patch, however
it would be ideal to keep receiving feedback on what further improvements
can be done. I am hoping this PR can at least start discussion around this
kind of improvement.

For now, this patch appends the total number of diffs a user performed on
the key-value editor (counting addition / deletion of fields)

dc-page-improved-success-msg

@spadgett @jwforres @openshift/ui-review

### DIFF
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -75,18 +75,35 @@ angular.module('openshiftConsole')
     };
 
     $scope.saveEnvVars = function() {
+      $scope.envVarDiffAmount = 0;
       $scope.envVars = _.filter($scope.envVars, 'name');
       buildStrategy($scope.updatedBuildConfig).env = keyValueEditorUtils.compactEntries(angular.copy($scope.envVars));
+
+      // count number of updated environment vars
+      var updatedEnvs = buildStrategy($scope.updatedBuildConfig).env;
+      var originalEnvs = $scope.originalEnvVars;
+      if (originalEnvs.length !== updatedEnvs.length) {
+        $scope.envVarDiffAmount = Math.abs(originalEnvs.length - updatedEnvs.length);
+      }
+      for(var i = 0; i < originalEnvs.length; i++) {
+       if(updatedEnvs[i] &&
+         (originalEnvs[i].name !== updatedEnvs[i].name ||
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/buildConfig.js
```diff
@@ -75,18 +75,35 @@ angular.module('openshiftConsole')
     };
 
     $scope.saveEnvVars = function() {
+      $scope.envVarDiffAmount = 0;
       $scope.envVars = _.filter($scope.envVars, 'name');
       buildStrategy($scope.updatedBuildConfig).env = keyValueEditorUtils.compactEntries(angular.copy($scope.envVars));
+
```
benjaminapetersen: I like the count, looks helpful.  We have an [environment service](https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/environment.js) now, it might be nice to push the logic up into `EnvironmentService.diff()` or `EnvironmentService.diffCount()`.  

Prob a nit, but I like `_.each()` over `for` loops for readability: 
```JavaScript
_.each(originalEnvs, function(env, i) {
 if(updatedEnvs[i] &&
   env.name !== updatedEnvs[i].name &&
   env.value !== updatedEnvs[i].value) {
   $scope.envVarDiffAmount++;
 }
});     
```
Though `_.reduce` might be a nice way to contain all the pieces:
```JavaScript
      $scope.envVarDiffAmount = _.reduce(originalEnvs, function(count, env, i) {
        if(updatedEnvs[i] &&
          env.name !== updatedEnvs[i].name &&
          env.value !== updatedEnvs[i].value) {
          count++;
        }
        return count;
      }, 0); 
```
We also now have `env.valueFrom` that we are handling in a lot of places.  Might need to add something for that as well.

jwforres:
Closing this PR since the issue is being resolved a different way by switching to toasts https://github.com/openshift/origin-web-console/pull/1804

PR #983, opened 2016-12-05
Show pod metrics donut to right of sparkline

### LINKED ISSUE #912: Millicore Doughnut Seems Out-of-Place on Metrics Overview
In the metrics overview, the CPU millicore doughnut takes up a huge amount of screen real-estate and breaks up the flow of the page:
cores

spadgett:
Place the pod metrics donut to the right of the sparkline at wider screen widths to make better use of the screen space. At mobile, still place the donut above the sparkline.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1387286
Fixes #912 

screen shot 2016-12-05 at 8 29 49 am

@jwforres PTAL
@smarterclayton @stevekuznetsov @jupierce CC

### DIFF
--- a/app/scripts/directives/podMetrics.js
+++ b/app/scripts/directives/podMetrics.js
@@ -1,7 +1,8 @@
 'use strict';
 
 angular.module('openshiftConsole')
-  .directive('podMetrics', function($interval,
+  .directive('podMetrics', function($filter,
+                                    $interval,
                                     $parse,
                                     $timeout,
                                     $q,
@@ -17,6 +18,7 @@ angular.module('openshiftConsole')
         sparklineWidth: '=?',
         sparklineHeight: '=?',
         includedMetrics: '=?', // defaults to ["cpu", "memory", "network"]
+        stackDonut: '=?', // Keep donut on top of sparkline (e.g. on the monitoring page)
         alerts: '=?'
       },
       templateUrl: 'views/directives/pod-metrics.html',
@@ -124,12 +126,14 @@ angular.module('openshiftConsole')
         // Show last hour by default.
         scope.options.timeRange = _.head(scope.options.rangeOptions);
 
+        var upperFirst = $filter('upperFirst');
         var createDonutConfig = function(metric) {
           var chartID = '#' + metric.chartPrefix + scope.uniqueID + '-donut';
           return {
             bindto: chartID,
             onrendered: function() {
-              ChartsService.updateDonutCenterText(chartID, metric.datasets[0].used, metric.units);
+              ChartsService.updateDonutCenterText(chartID, metric.datasets[0].used,
+                                                  upperFirst(metric.units) + " Used");
             },
             donut: {
               label: {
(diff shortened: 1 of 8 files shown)


jwforres:
what does it look like with the network metrics that don't have a donut, i assume the sparklines are all the same width?

spadgett:
@jwforres 

screen shot 2016-12-05 at 12 23 30 pm

jwforres:
hmm the horizontal line makes network look like a part of the same section as cpu

spadgett:
> the horizontal line makes network look like a part of the same section as cpu

Let me see what I can do.

spadgett:
@jwforres How about this?

screen shot 2016-12-05 at 12 47 08 pm

spadgett:
Hold, need to make some updates for the monitoring page, which has the sidebar

spadgett:
Updated for the monitoring page

jwforres:
Thats definitely better although i think i would reduce the spacing between Network and the bar, probably half as much.  Its definitely larger than the space between the numbers and the bars on the other two.

jwforres:
I can't decide but I wonder if it would look better to move the number information directly to the right of the header, like CPU - 500[etc]

and then match the header font size and number font size

spadgett:
We do show small "Container Metrics" or "Pod Metrics" text by the header if there is more than one container, although we could move that.

spadgett:
> I can't decide but I wonder if it would look better to move the number information directly to the right of the header, like CPU - 500[etc]

openshift_web_console

spadgett:
Or if we just bump the font size for the metric label

openshift_web_console

jwforres:
yeah inline at the same font size it looks weirdly likes its part of the header, dont like it :-\

I can't decide on the larger header size, feels a little overwhelming, @ajacobs21e opinion?

stevekuznetsov:
Maybe the "X available" could be on top of the doughnut and "in use" could come after the "x millicores" inside of the doughnut?

stevekuznetsov:
As both are only set up if quota is on, right?

spadgett:
Here is the Patternfly style for reference:

http://www.patternfly.org/pattern-library/cards/utilization-trend-card/#/api

jwforres:
Eh, maybe just stick with the smaller header then.

spadgett:
Tightened the spacing under the Network header

screen shot 2016-12-05 at 2 34 08 pm

jwforres:
Do the horizontal lines going all the way across look strange when none of the donuts are there?

spadgett:
> Do the horizontal lines going all the way across look strange when none of the donuts are there?

I only include them when there's at least one donut. And the sparklines are 100% width when there are no donuts.

screen shot 2016-12-05 at 2 43 10 pm

spadgett:
If we always want the lines, this is how it looks:

screen shot 2016-12-05 at 2 45 47 pm

jwforres:
im ok with either of those, though i think i prefer with the lines

spadgett:
> i think i prefer with the lines

Let me add them to deployment metrics in that case to be consistent

spadgett:
Added "Used" inside the donut.

screen shot 2016-12-05 at 3 04 32 pm

spadgett:
Updated, @jwforres PTAL

jwforres:
[merge]

PR #984, opened 2016-12-05
Redirect to overview when updating Template in fromFile

jhadvig:
This is fixes the case when user updates the Template in the fromFile page and he is redirected to the `fromTemplate` page. By setting the template process option to `false` user will be redirected to the overview page.

@jwforres PTAL

### DIFF
--- a/app/scripts/directives/fromFile.js
+++ b/app/scripts/directives/fromFile.js
@@ -158,7 +158,12 @@ angular.module("openshiftConsole")
               openTemplateProcessModal();
             // Else if any resources already exist
             } else if (!_.isEmpty($scope.updateResources)) {
-              confirmReplace();
+              $scope.updateTemplate = $scope.updateResources.length === 1 && $scope.updateResources[0].kind === "Template";
+              if ($scope.updateTemplate) {
+                openTemplateProcessModal();
+              } else {
+                confirmReplace();
+              }
             } else {
               QuotaService.getLatestQuotaAlerts($scope.createResources, $scope.context).then(showWarningsOrCreate);
             }
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -900,6 +900,10 @@ a.disabled-link {
   p {
     font-size: @font-size-large + 1;
   }
+  .help-block {
+    margin-top: 0px;
+    margin-bottom: 10px;
+  }
 }
 
 // Disable numeric highlighting in YAML and Dockerfile modes because of bugs
--- a/app/views/modals/process-template.html
+++ b/app/views/modals/process-template.html
@@ -1,6 +1,6 @@
 <div class="modal-resource-action">
   <div class="modal-body">
-    <h1>Add Template</h1>
+    <h1>{{updateTemplate ? "Update" : "Add"}} Template</h1>
     <p>What would you like to do?</p>
 
     <div>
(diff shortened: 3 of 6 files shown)


app/scripts/directives/fromFile.js
```diff
@@ -157,6 +157,10 @@ angular.module("openshiftConsole")
               openTemplateProcessModal();
             // Else if any resources already exist
             } else if (!_.isEmpty($scope.updateResources)) {
+              // If the updated resource is a Template, set the process option to false
+              if ($scope.updateResources.length === 1 && $scope.updateResources[0].kind === "Template") {
+                $scope.templateOptions.process = false;
```
jwforres: this doesn't seem right.  if I checked both the Save and Process options then it should ask me to confirm the update and then go to the from template page

jhadvig: When you are updating an existing Template we dont show the dialog in which user can pick if he wants to process or save(or both) the Template, only the one in which user has to confirm that the new one will replace the old one. After he hits `Replace` he should be redirected to the overview and not to the fromTemplate page as it is now.

jhadvig: @jwforres thoughts ?

jwforres: I think someone might get confused by that behavior. If we give you the option to process a template when you first upload it, it seems like you should get the option on an update too.

jwforres: @spadgett thoughts?

spadgett: I think the best experience would be to show the same dialog, but ask if they want to update the template or process it (instead of create and process). At least if there aren't other items in the YAML. @jhadvig how hard would that be?

jhadvig:
@jwforres @spadgett I've update the PR based on the your comments, so now when user provides a template which already exists he will get a dialog similar to the save/process one, but instead of save he will have update. Here is the screen:
screenshot-13

PTAL

jwforres:
I'd probably make the help text something like "This will overwrite the current version of the template."

jhadvig:
@jwforres comment adressed :)

jwforres:
@jhadvig you need to rebase

jhadvig:
@jwforres rebased

app/scripts/directives/fromFile.js
```diff
@@ -249,7 +255,7 @@ angular.module("openshiftConsole")
           var path;
           if ($scope.resourceKind === "Template" && $scope.templateOptions.process && !$scope.errorOccured) {
             var namespace = ($scope.templateOptions.add || $scope.updateResources.length > 0) ? $scope.projectName : "";
-            path = Navigate.createFromTemplateURL(resource, $scope.projectName, {namespace: namespace});
+            path = Navigate.createFromTemplateURL($scope.resourceList[0], $scope.projectName, {namespace: namespace});
```
jwforres: is this change necessary or did this get messed up in the merge conflict / rebase.  robb had just fixed an issue related to this call

jhadvig: yep. sorry for that.. didnt noticed there was this change

jhadvig:
@jwforres updated

jwforres:
[merge]

jwforres:
@jhadvig was there a bug/issue associated with this fix

PR #985, opened 2016-12-05
Let users define labels when creating routes

spadgett:
Fixes https://github.com/openshift/origin/issues/6095

openshift_web_console

@jwforres PTAL

### DIFF
--- a/app/views/create-route.html
+++ b/app/views/create-route.html
@@ -18,7 +18,7 @@
                   <div class="help-block">
                     Routing is a way to make your application publicly visible.
                   </div>
-                  <form name="createRouteForm" class="mar-top-xl">
+                  <form name="createRouteForm" class="mar-top-xl osc-form">
                     <div ng-if="!services">Loading...</div>
                     <div ng-if="services">
                       <fieldset ng-disabled="disableInputs">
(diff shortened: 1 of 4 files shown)


jwforres:
I can't decide on the system labels here, I see the argument for it, but at the same time we don't let you remove the system labels.

spadgett:
I don't disagree, but I think it's a problem in general with how we approach system labels. What if I don't want an `app` label when I deploy my template for instance?

jwforres:
i think the difference that bothers me here is that in other cases they are really system labels, in this case its not just copying the system labels like the app label, its copying all the labels from the service, some of which may have been user defined

spadgett:
I could prefill them as user-defined labels, but it gets tricky if the user changes services in the dropdown. Or maybe a "Copy from Service" link?

jwforres:
copy from service link seems reasonable, then its easy without being forced into it

spadgett:
Updated with a "Copy Service Labels" link.

app/views/create-route.html
```diff
@@ -27,6 +27,13 @@
                             services="services"
                             show-name-input="true">
                         </osc-routing>
+                        <label-editor
+                            labels="labels"
+                            expand="true"
+                            can-toggle="false"
+                            help-text="Labels for this route.">
+                        </label-editor>
+                        <a href="" ng-click="copyServiceLabels()">Copy service labels</a>
```
jwforres: headline case

spadgett: Right above this link is "Add label" (at least until #978 merges)

spadgett: Let's merge the other one first, and I'll rebase this one. About to push updates for #978

spadgett:
Updated to headline case.

[merge]

PR #986, opened 2016-12-05
Use headline case for replication controller details labels

spadgett:
@jwforres PTAL, must have missed this one in #805


jwforres:
[merge]

Issue #925, opened 2016-11-24
Pipeline should more stages per row


siamaksade:
Pipeline stages are currently set to 25% width and therefore show only 4 stages per row regardless of how wide the screen/browser is. Most pipelines are going to have more than 4 stages.

The width should be smaller or preferably set dynamically based on the number of stages in order to fit as many stages as possible on the same row.

<img width="1393" alt="pipeline" src="https://cloud.githubusercontent.com/assets/1174934/20592151/5ca56b9e-b22c-11e6-8c1c-d762307b4233.png">

I have changed the width using the "Inspect" tool in Chrome to show how many actually fit within the same width:

<img width="1392" alt="pipeline-adjusted" src="https://cloud.githubusercontent.com/assets/1174934/20592225/95b02320-b22c-11e6-8115-f1580dc88322.png">

# Version
OpenShift Master:
v3.3.1.3
Kubernetes Master:
v1.3.0+52492b4

PR #987, opened 2016-12-05
Adjust pipeline-stage widths to accommodate more stages based on browser widths

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/925

Changes to allow space for more stages to fit inline before wrapping. 

**Current number of stages within single row before wrapping.**

- less than 992px    - 2 stages
- 992px to 1199px  - 3 stages
- more than 1200px - 4 stages

**Proposed number of stages within single row before wrapping.**

- less than 992px - 2 stages
- 992px to 1199px - 4 stages
- 1200px to 1399px - 5 stages
- 1400px to 1599px - 6 stages
- more than 1600px - 7 stages

Stage widths are percentages. The least number of character spaces before the ellipsis is 20 and a `title attribute` is added to stage name and shown on hover.

**4 stages when > 992px**
screen shot 2016-12-02 at 4 38 40 pm

**5 stages when > 1200px**
screen shot 2016-12-02 at 4 38 04 pm

**6 stages when > 1400px**
screen shot 2016-12-02 at 4 37 28 pm

**7 stages when > 1600px**
screen shot 2016-12-02 at 4 37 08 pm

### DIFF
--- a/app/styles/_pipeline.less
+++ b/app/styles/_pipeline.less
@@ -261,7 +261,7 @@
   .pipeline .pipeline-stage {
     padding-right: (@pipeline-padding * 5.2);
     padding-bottom: (@pipeline-padding + 5);
-    width: 50%;
+    width: (100% / 2);
     &:before {
       bottom: auto;
       content: '\2192';
@@ -286,7 +286,7 @@
   .build-summary {
     border-bottom-width: 0;
     border-right: 1px solid @build-pipeline-border-color;
-    .flex(@columns: 0 0 175px);
+    .flex(@columns: 0 0 125px);
     .flex-direction(@direction: column);
     .justify-content(@justify: center);
   }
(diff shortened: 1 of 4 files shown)


sg00dwin:
@jwforres @spadgett @rhamilto 
thoughts on these pipeline stage width changes

spadgett:
I think it's better.

> less than 992px - 2 stages

Do we still collapse down to one column with vertical arrows at mobile?

app/styles/_pipeline.less
```diff
@@ -145,8 +145,11 @@
       text-align: center;
     }
     // hide arrow on last stage
-    &:last-child:before {
-      display: none;
+    &:last-child {
```
rhamilto: Do you feel strongly about this change?  I prefer to avoid unnecessary nesting...

app/styles/_pipeline.less
```diff
@@ -261,7 +264,7 @@
   .pipeline .pipeline-stage {
     padding-right: (@pipeline-padding * 5.2);
     padding-bottom: (@pipeline-padding + 5);
-    width: 50%;
+    width: calc(100% / 2);
```
rhamilto: Curious:  Why use calc() over letting the Less do the math?

app/styles/_pipeline.less
sg00dwin: that actually wasn't meant to be included.

app/styles/_pipeline.less
rhamilto: In this instance, I prefer to just let the Less do the math since the addition of calc doesn't gain us anything, but it's not a big deal either way.

sg00dwin:
@spadgett 
yes it still switches to a single column in mobile. 
PR is updated and ready for merge.

app/styles/_pipeline.less
spadgett: I'd prefer to use less for the math

spadgett:
[merge]

spadgett:
> Error: Timed out waiting for the WebDriver server at http://172.18.1.112:60902/wd/hub

[merge]

PR #990, opened 2016-12-06
Compact customParams envVars when changing from Custom to Rolling/Recreate deployment strategy

jhadvig:
@spadgett PTAL


spadgett:
[merge]

PR #995, opened 2016-12-06
Show the same current usage value regardless of time range

spadgett:
Make a separate request for the last minute of usage when showing the
pod metrics donut. This will show consistent values for current usage
regardless of the selected time range.

Fixes https://github.com/openshift/origin/issues/5471

@jwforres PTAL


jwforres:
[merge]

spadgett:
[merge]

PR #996, opened 2016-12-06
Add build failure reasons to monitoring and overview pages

jwforres:


jwforres:
@spadgett missed this two pages, i couldn't decide between the long or short form on the overview

app/views/monitoring.html
```diff
@@ -303,7 +303,9 @@
                               <small>created <span am-time-ago="build.metadata.creationTimestamp"></span></small>
                             </div>
                             <div class="list-group-item-text">
-                              <build-status build="build"></build-status>
```
spadgett: Should we delete the build-status directive if it's not used? Or update it to show the reason or message instead?

jwforres:
Realized this change was removing the duration completely from the monitoring page, i'm updating build-status directive for when we still need status + duration, but using the longer form duration on this page there isn't really enough space for it when the events sidebar is expanded.

wrapping

spadgett:
Yeah I think we need to use "1m 48s" format, maybe small / text-muted?

jwforres:
i dunno about changing the font style, in some cases its read as part of a statement like with Cancelled

jwforres:
updated PTAL

spadgett:
[merge]

Issue #946, opened 2016-11-29
Add learn more to more pages (particularly the resources section)


danmcp:
Related to:

https://github.com/openshift/origin/issues/7664

![image](https://cloud.githubusercontent.com/assets/492994/20713563/becddaf2-b616-11e6-999c-c2b4f837febb.png)

Membership has a learn more link to the docs but the other resources do not.  I think it would be helpful on most every page in the UI.

PR #998, opened 2016-12-07
Adding learn more links to primary and secondary pages.

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/946

Moved .page-header styles,  along with several other misc rules, from _core.less to _typeography.less

Also fixes https://github.com/openshift/origin/issues/7664


sg00dwin:
@jwforres PTAL

spadgett:
We've switch to "Learn More" with an uppercase M

app/views/browse/config-maps.html
```diff
@@ -10,7 +10,13 @@
             <div class="pull-right" ng-if="project && ('configmaps' | canI : 'create')">
               <a ng-href="project/{{project.metadata.name}}/create-config-map" class="btn btn-default">Create Config Map</a>
             </div>
-            <h1>Config Maps</h1>
+            <h1>Config Maps
```
spadgett: Formatting nit: When there are other elements inside the header, I find it much easier to read if the header text is on the next line...

```html
<h1>
  Config Maps
  <span>...</span>
</h1>
```

app/views/browse/config-maps.html
```diff
@@ -10,7 +10,13 @@
...
+              <span class="page-header-link">
+                <a ng-href="{{'config-maps' | helpLink}}" target="_blank">
+                  Learn more <i class="fa fa-external-link" aria-hidden="true"></i>
```
spadgett: Probably should use a non-breaking space between the text and icon so that the icon doesn't ever wrap by itself. It could break on some longer headers at mobile.

spadgett:
[merge]

app/views/browse/config-maps.html
sg00dwin: I set the wrapping span to `whitespace: no-wrap` and `inline-block` so it wouldn't wrap separately.

sg00dwin:
@spadgett updated based on your feedback.

spadgett:
[merge]

PR #999, opened 2016-12-07
If there is only 1 pod, navigate directly to it from the donut

rhamilto:
Fixes https://github.com/openshift/origin/issues/7947

@jwforres or @spadgett , PTAL


jwforres:
LGTM but you are getting a diff on vendor.js, you need to run hack/install-dep.sh

rhamilto:
D'oh.  Should be fixed now, @jwforres.

spadgett:
[merge]

PR #1000, opened 2016-12-07
Show the container command and args in the detailed pod template

jwforres:
Fixes https://github.com/openshift/origin/issues/8736


jwforres:
Args with no command:

args-no-cmd

Command with args, or just command:

cmd-and-args


@spadgett PTAL

spadgett:
[merge]

PR #1007, opened 2016-12-08
Fixing bug where creating from a template file was busted

### LINKED ISSUE #1005: Create from template can throw an error
Using the `nodejs-mongodb.json` template & pasting it in the `Import YAML / JSON` field, I get a console error w/o any feedback on the page:

screen shot 2016-12-07 at 5 07 33 pm

@jwforres @spadgett

rhamilto:
Fixes #1005 

@benjaminapetersen, @spadgett, @jwforres, PTAL

### DIFF
--- a/app/scripts/directives/fromFile.js
+++ b/app/scripts/directives/fromFile.js
@@ -91,10 +91,11 @@ angular.module("openshiftConsole")
           }
         };
 
+        var resource;
+
         $scope.create = function() {
           $scope.alerts = {};
           delete $scope.error;
-          var resource;
 
           // Trying to auto-detect what format the input is in. Since parsing JSON throws only SyntexError
           // exception if the string to parse is not valid JSON, it is tried first and then the YAML parser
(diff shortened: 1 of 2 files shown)


app/scripts/directives/fromFile.js
```diff
@@ -118,6 +118,7 @@ angular.module("openshiftConsole")
             return;
           }
 
+          $scope.resource = resource;
```
spadgett: Shouldn't need to set this as a scope var, can just use `resource` below when you call `createFromTemplateURL` below

spadgett:
@rhamilto can you sweep the code for other places where this might be a problem (templates or image streams)

app/scripts/directives/fromFile.js
rhamilto: Resource is scoped to only this function.  How do I get it out of this function?

rhamilto:
Ok.  I think I got it this time.  Thanks, @jwforres and @spadgett.

> @rhamilto can you sweep the code for other places where this might be a problem (templates or image streams)

Done.  We should be good now.

spadgett:
[merge]

Issue #1009, opened 2016-12-08
Container command should truncate sooner


spadgett:
![screen shot 2016-12-08 at 2 08 04 pm](https://cloud.githubusercontent.com/assets/1167259/21024930/5f2f6eae-bd54-11e6-84ff-7f3f983597f9.png)

PR #1010, opened 2016-12-08
Truncate container commands and probes for the entire content not each arg

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/1009

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -1323,7 +1323,7 @@ angular.module('openshiftConsole')
     };
 
     return function(container, image) {
-      if (!container || !image) {
+      if (!container) {
         return null;
       }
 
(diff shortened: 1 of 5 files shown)


app/scripts/filters/resources.js
```diff
@@ -1322,8 +1322,8 @@ angular.module('openshiftConsole')
       return cmd;
     };
 
-    return function(container, image) {
-      if (!container || !image) {
+    return function(container, image, returnArgsIfNoCmd) {
```
spadgett: Should `returnArgsIfNoCmd` just always be true? We should update debug terminal to set that, and that's the only other place the filter is used..

jwforres: Originally I wasn't adding `<image-entrypoint>` when we didn't have the image and it was only returning the args so I didn't want it to change the behavior on debug terminal. But yeah now I think I can drop the arg and just always have that behavior.

jwforres:
@spadgett updated

spadgett:
[merge]

PR #1012, opened 2016-12-08
Revolving issue where secondary nav appears below page dropdown

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1366090

screen shot 2016-12-08 at 4 54 35 pm

@jwforres or @spadgett, PTAL

### DIFF
--- a/app/styles/_sidebar.less
+++ b/app/styles/_sidebar.less
@@ -199,6 +199,7 @@
         position: absolute;
         padding: 10px;
         top: -2px;
+        z-index: @zindex-dropdown + 1;
       }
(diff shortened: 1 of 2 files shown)


jwforres:
This is better than before, but I wonder why the ui-select drop down isn't closing. I suspect something is stopping propagation of the click event.

spadgett:
[merge]

rhamilto:
@jwforres, this is a known bug with ui-select.  See https://github.com/angular-ui/ui-select/issues/228.

PR #1013, opened 2016-12-09
Fix tile overlap when a service has multiple pods

### LINKED ISSUE #1011: Pods that share the same service, but have no `deploymentconfig` set show up in separate cards
Pods that are selected by the same service, but have no `deploymentconfig` set, show up in separate cards which overflow into the second column of the overview page.

screenshot from 2016-12-08 16-23-01

##### Version
oc v1.3.0-rc1+db098d5-1612-dirty
kubernetes v1.4.0+776c994
features: Basic-Auth

Server https://10.13.137.149:8443
openshift v1.3.0-rc1+a38151f-1683-dirty
kubernetes v1.4.0+776c994

##### Steps To Reproduce
1. `oc process -f examples/zookeeper/template.json | oc apply -f - `
2. Take a look at the overview page in the web console

##### Current Result
Three separate pod cards for one service

spadgett:
Remove redundant flex style for `.primary-service .deployment-block` which was overriding the `display: block` style for `.service-multiple-targets` because it was more specific. The `.deployment-block` already has the flex style set here

https://github.com/openshift/origin-web-console/blob/10c6ca8502411aaecd08360233749493c2915fc8/app/styles/_overview.less#L316-L317

See https://github.com/openshift/origin-web-console/issues/1011#issuecomment-266011901, fixes #1011 

Note that this stacks the tiles, which was the original intent, although not ideal. If you have a lot of pods, it won't scale that well. We should try to do something better in https://trello.com/c/QxbqJsoK

screen shot 2016-12-09 at 8 38 00 am

### DIFF
--- a/app/styles/_overview.less
+++ b/app/styles/_overview.less
@@ -160,11 +160,6 @@
     }
     .service-group-body {
       margin-top: 0;
-      .overview-services .primary-service {
-        .deployment-block {
-          .flex-display(@display: flex);
-        }
-      }
     }
     .alert-wrapper .alert {
       margin-bottom: @service-group-vertical-margin;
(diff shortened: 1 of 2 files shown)


jwforres:
[merge]

PR #1016, opened 2016-12-09
Fixes "wobble" or "jitter" of spinning .status-icons

rhamilto:
was most noticeable in Firefox for Linux

Before: http://recordit.co/NzMKf14SW5
After: http://recordit.co/IYypdfreyn

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1365301

@jwforres or @spadgett, PTAL

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -1165,10 +1165,15 @@ copy-to-clipboard .input-group.limit-width {
   justify-content: space-between;
 }
 
-
 // add an option for a single-column display at screen resolutions 0-479px
 @media (max-width: @screen-xxs-max) {
   .col-xxs-12 {
     width: 100%;
   }
+}
+
+// fixes wobble that occurs when status-icons utilize fa-spin (see https://github.com/FortAwesome/Font-Awesome/issues/671#issuecomment-249591495)
+.status-icon > .fa-spin {
+  font-size: 14px;
+  line-height: normal;
 }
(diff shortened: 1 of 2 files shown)


app/styles/_core.less
```diff
@@ -1161,10 +1161,19 @@ copy-to-clipboard .input-group.limit-width {
...
     width: 100%;
   }
 }
+
+// fixes wobble that occurs when status-icons utilize fa-spin
+// since font awesome icons are created on a 14px grid, setting the font-size to 14px improves anti-aliasing
+// line-height: normal fixes the wobble (see https://github.com/FortAwesome/Font-Awesome/issues/671#issuecomment-249591495 )
+// margin-right puts a bit of space between the icon and the text that follows, which is necessary given the larger icon size
+.status-icon > .fa-spin {
+  font-size: 14px;
```
jwforres: any reason not to make all the status icons 14px? not just when fa-spin is there

rhamilto: Yeah.  It turns out we're using status-icons at several different font sizes depending on the context, so making all status-icons 14px creates several new bugs.  Scoping this to only .fa-spin fixes this issue, but doesn't create new ones.  

Additionally, I've [opened a new issue](https://github.com/openshift/origin-web-console/issues/1023) to address some of the inconsistencies with status-icons I discovered when investigating.

jwforres:
[merge]

PR #1017, opened 2016-12-09
Preserve line breaks in project descriptions

rhamilto:
Resolves https://github.com/openshift/origin/issues/12194

Limits line breaks to 10 for the truncated display (does not apply to the highlighted text display).  Also adds word-break for long, unbroken strings so they do not break the layout.

screen shot 2016-12-09 at 1 55 28 pm
screen shot 2016-12-09 at 1 55 35 pm
screen shot 2016-12-09 at 1 55 45 pm
screen shot 2016-12-09 at 1 56 43 pm

@jwforres or @spadgett, PTAL

### DIFF
--- a/app/styles/_projects.less
+++ b/app/styles/_projects.less
@@ -31,6 +31,14 @@
   }
 }
 
+.project-description {
+  .highlighted-content,
+  .truncated-content {
+    white-space: pre-line;
+    .word-break();
+  }
+}
+
 .project-info {
   &.list-group-item {
     border-bottom: 0;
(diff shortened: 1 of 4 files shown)


jwforres:
so in other places we chose not to show all text when highlighting because it made things jump around

rhamilto:
The show all text when highlighting is not a change to this PR.  The existing word limit was/is 1000 characters.  So I assumed that since we're already showing a max of 1000 characters (which seems like a whole lotta characters for a description), we probably don't won't to limit the line breaks, either.  Otherwise, you could end up with a situation where the highlighting isn't visible.  Yea/nay?

jwforres:
ah k thats fine then, maybe i misremembered the result of our previous discussion on showing more when filtering.

jwforres:
[merge]

PR #1022, opened 2016-12-12
Show CPU usage as cores rather than millicores

spadgett:
Cores is a more universally understood unit than millicores.

Related to https://github.com/openshift/origin/issues/7768

screen shot 2016-12-12 at 9 33 58 am

screen shot 2016-12-12 at 9 34 28 am

screen shot 2016-12-12 at 9 34 42 am

### DIFF
--- a/app/scripts/services/metricsCharts.js
+++ b/app/scripts/services/metricsCharts.js
@@ -11,6 +11,8 @@ angular.module("openshiftConsole")
       switch (metricID) {
       case 'memory/usage':
         return _.round(ConversionService.bytesToMiB(point.value), 2);
+      case 'cpu/usage_rate':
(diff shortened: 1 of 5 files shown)


app/scripts/services/metricsCharts.js
```diff
@@ -80,11 +82,6 @@ angular.module("openshiftConsole")
                 left: 0,
                 top: 20,
                 bottom: 0
-              },
-              tick: {
-                format: function(value) {
-                  return d3.round(value, 3);
-                }
```
spadgett: Since we're letting c3.js pick the tick values, this is no longer necessary. c3.js picks round numbers. It can also cause problems for very low CPU usage where a tick is repeated.

jwforres:
[merge]

PR #1024, opened 2016-12-12
Replace select tag with ui-select - part 1

jhadvig:
This is the first part of work which is moving all our `select` tags to `ui-select`
So far it contains the :

- fromimage page
- create route page
- pod browse page -> metrics tab

I've added the `input-id` attribute as suggested and also the `aria-describedby` in case the help text is below the select-box.

@spadgett PTAL

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -1022,6 +1022,20 @@ angular.module('openshiftConsole')
       }
     };
   })
+  .filter('humanizeTLSTermination', function () {
+    return function(termination) {
+      switch (termination) {
+        case 'edge':
+          return 'Edge';
+        case 'passthrough':
+          return 'Passthrough';
+        case 'reencrypt':
+          return 'Re-encrypt';
+        default:
+          return termination;
+      }
+    };
+  })
   .filter('humanizeKind', function (startCaseFilter) {
     // Changes "ReplicationController" to "replication controller".
     // If useTitleCase, returns "Replication Controller".
--- a/app/views/browse/routes.html
+++ b/app/views/browse/routes.html
@@ -83,7 +83,7 @@
                     </td>
                     <!-- Use shorter Termination title for table-mobile to avoid overlapping text. -->
                     <td data-title="Termination">
-                      {{route.spec.tls.termination}}
+                      {{route.spec.tls.termination | humanizeTLSTermination}}
                       <span ng-if="!route.spec.tls.termination">&nbsp;</span>
                     </td>
                   </tr>
--- a/app/views/create/fromimage.html
+++ b/app/views/create/fromimage.html
@@ -298,12 +298,12 @@
                               >
                               <div class="form-group">
                                 <label for="scale-type">Strategy</label>
-                                <select ng-model="scaling.autoscale"
-                                        ng-options="option.value as option.label for option in scaling.autoscaleOptions"
-                                        id="scale-type"
-                                        class="form-control"
-                                        aria-describedby="scale-type-help">
-                                </select>
+                                <ui-select ng-model="scaling.autoscale" input-id="scale-type" search-enabled="false" aria-describedby="scale-type-help">
+                                  <ui-select-match>{{$select.selected.label}}</ui-select-match>
+                                  <ui-select-choices repeat="option.value as option in scaling.autoscaleOptions">
+                                    {{option.label}}
+                                  </ui-select-choices>
+                                </ui-select>
                                 <div class="help-block" id="scale-type-help">
                                   Scale replicas manually or automatically based on CPU usage.
                                 </div>
--- a/app/views/directives/osc-routing-service.html
+++ b/app/views/directives/osc-routing-service.html
@@ -4,14 +4,12 @@
       <label for="{{id}}-service-select" class="required">
         Service
       </label>
-      <select
-          id="{{id}}-service-select"
-          ng-model="model.service"
-          ng-options="service as service.metadata.name for service in services"
-          required
-          class="form-control"
-          ng-attr-aria-describedby="{{id}}-service-help">
-      </select>
+      <ui-select ng-model="model.service" input-id="{{id}}-service-select" aria-describedby="{{id}}-service-help" required>
+        <ui-select-match>{{$select.selected.metadata.name}}</ui-select-match>
+        <ui-select-choices repeat="service in (services | filter : {metadata: { name: $select.search }}) track by (service | uid)">
+           <div ng-bind-html="service.metadata.name | highlight : $select.search"></div>
+        </ui-select-choices>
+      </ui-select>
       <div>
         <span ng-attr-id="{{id}}-service-help" class="help-block">
           <span ng-if="!isAlternate">Service to route to.</span>
--- a/app/views/directives/osc-routing.html
+++ b/app/views/directives/osc-routing.html
@@ -120,14 +120,12 @@
     <!-- Target Port -->
     <div ng-if="route.portOptions.length" class="form-group">
       <label for="routeTargetPort">Target Port</label>
-      <select
-          id="routeTargetPort"
-          ng-if="route.portOptions.length"
-          ng-model="route.targetPort"
-          ng-options="portOption.port as portOption.label for portOption in route.portOptions"
-          class="form-control"
-          aria-describedby="target-port-help">
-      </select>
+      <ui-select ng-if="route.portOptions.length" ng-model="route.targetPort" input-id="routeTargetPort" search-enabled="false" aria-describedby="target-port-help">
+        <ui-select-match>{{$select.selected.label}}</ui-select-match>
+        <ui-select-choices repeat="portOption.port as portOption in route.portOptions">
+          {{portOption.label}}
+        </ui-select-choices>
+      </ui-select>
       <div>
         <span id="target-port-help" class="help-block">
           Target port for traffic.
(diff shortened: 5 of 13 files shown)


app/views/directives/osc-routing.html
```diff
@@ -170,29 +168,27 @@
...
       <div class="form-group">
-        <label>TLS Termination</label>
-        <select ng-model="route.tls.termination" class="form-control">
-          <option value="edge">Edge</option>
-          <option value="passthrough">Passthrough</option>
-          <option value="reencrypt">Re-encrypt</option>
-        </select>
+        <label for="tlsTermination">TLS Termination</label>
+        <ui-select ng-model="route.tls.termination" input-id="tlsTermination" search-enabled="false">
+          <ui-select-match>{{$select.selected | startCase}}</ui-select-match>
```
spadgett: This will give a different value than before for `reencrypt`. 

We should be able  create an array of object with `id` and `label` attributes and then bind the value as the `id`. That way we can have future values that won't work for `startCase` filter and will let us translate the labels to different languages eventually. See

https://github.com/angular-ui/ui-select/wiki/ui-select-choices#example-binding-single-property

app/views/directives/osc-routing.html
```diff
@@ -170,29 +168,27 @@
...
-            class="form-control"
-            aria-describedby="route-insecure-policy-help">
-          <option value="">None</option>
-          <option value="Allow">Allow</option>
-          <option value="Redirect">Redirect</option>
-        </select>
+        <label for="insecureTraffic">Insecure Traffic</label>
+        <ui-select ng-model="route.tls.insecureEdgeTerminationPolicy" ng-disabled="route.tls.termination !== 'edge'" input-id="insecureTraffic" aria-describedby="route-insecure-policy-help">
+          <ui-select-match>{{$select.selected.label}}</ui-select-match>
+          <ui-select-choices repeat="option.value as option in [{value: '', label:'None'}, {value: 'Allow', label: 'Allow'}, {value: 'Redirect', label: 'Redirect'}]">
```
spadgett: :+1: but I suggest defining the options as a scope var in the controller

app/views/directives/osc-routing.html
```diff
@@ -170,29 +168,27 @@
...
-            ng-model="route.tls.insecureEdgeTerminationPolicy"
-            ng-disabled="route.tls.termination !== 'edge'"
-            class="form-control"
-            aria-describedby="route-insecure-policy-help">
-          <option value="">None</option>
-          <option value="Allow">Allow</option>
-          <option value="Redirect">Redirect</option>
-        </select>
+        <label for="insecureTraffic">Insecure Traffic</label>
+        <ui-select ng-model="route.tls.insecureEdgeTerminationPolicy" ng-disabled="route.tls.termination !== 'edge'" input-id="insecureTraffic" aria-describedby="route-insecure-policy-help">
```
spadgett: `search-enabled="false"`

app/views/directives/osc-routing-service.html
```diff
@@ -4,14 +4,12 @@
...
-          id="{{id}}-service-select"
-          ng-model="model.service"
-          ng-options="service as service.metadata.name for service in services"
-          required
-          class="form-control"
-          ng-attr-aria-describedby="{{id}}-service-help">
-      </select>
+      <ui-select ng-model="model.service" input-id="{{id}}-service-select" aria-describedby="{{id}}-service-help" required>
+        <ui-select-match>{{$select.selected.metadata.name}}</ui-select-match>
+        <ui-select-choices repeat="service as service in services | filter : $select.search">
```
spadgett: Probably worth adding `track by (service | uid)`

app/views/directives/pod-metrics.html
```diff
@@ -7,21 +7,24 @@
...
     </div>
     <div class="form-group">
       <label for="timeRange">Time Range:</label>
-      <select id="timeRange"
-              ng-model="options.timeRange"
-              ng-options="range.label for range in options.rangeOptions"
-              ng-disabled="metricsError">
-      </select>
+      <div class="select-range">
+        <ui-select  ng-model="options.timeRange" search-enabled="false" input-id="timeRange">
```
spadgett: nit extra whitespace before `ng-model`

spadgett:
We should improve how the metrics selects stack and align at mobile:

openshift_web_console

spadgett:
Does Bootstrap inline form work for the metrics page? Then might be able to remove the vertical-align changes you had to make.

http://getbootstrap.com/css/#forms-inline

app/views/create/fromimage.html
```diff
@@ -298,12 +298,12 @@
                               >
                               <div class="form-group">
                                 <label for="scale-type">Strategy</label>
-                                <select ng-model="scaling.autoscale"
-                                        ng-options="option.value as option.label for option in scaling.autoscaleOptions"
-                                        id="scale-type"
-                                        class="form-control"
-                                        aria-describedby="scale-type-help">
-                                </select>
+                                <ui-select ng-model="scaling.autoscale" input-id="scale-type" aria-describedby="scale-type-help">
```
spadgett: I think the `aria-describedby` attributes need to go on the `form-group` element. I don't see them in the markup after the control is rendered.

jhadvig: Hmm.. I can see the `aria-describedby` attribute on the top `div` elements that renders from the `ui-select` tag.

jhadvig:
I've tried the inline form but unfortunately its not working with `ui-select`

jhadvig:
@spadgett comments addressed. Here is a screen from the mobile version of pod-metrics tab:

screenshot-17

Also check these comments: 
- https://github.com/openshift/origin-web-console/pull/1024#issuecomment-266713760
- https://github.com/openshift/origin-web-console/pull/1024#discussion_r92141603

PTAL

app/views/create/fromimage.html
spadgett: > I can see the aria-describedby attribute on the top div elements that renders from the ui-select tag.

OK

app/scripts/filters/resources.js
```diff
@@ -1022,6 +1022,20 @@ angular.module('openshiftConsole')
...
+    return function(string) {
+      switch (string) {
+        case 'edge':
+          return 'Edge';
+        case 'passthrough':
+          return 'Passthrough';
+        case 'reencrypt':
+          return 'Re-encrypt';
+        default:
+          return null;
```
spadgett: Fall back to returning the raw value:

```js
return string;
```

That's consistent with our other humanize filters and lets us use the filter without ever having to check for a null value.

app/scripts/filters/resources.js
```diff
@@ -1022,6 +1022,20 @@ angular.module('openshiftConsole')
       }
     };
   })
+  .filter('humanizeTLSTermination', function () {
+    return function(string) {
```
spadgett: Call the arg something more specific like `termination` rather than `string`

app/views/browse/routes.html
```diff
@@ -83,7 +83,7 @@
                     </td>
                     <!-- Use shorter Termination title for table-mobile to avoid overlapping text. -->
                     <td data-title="Termination">
-                      {{route.spec.tls.termination}}
+                      {{route.spec.tls.termination | humanizeTLSTermination}}
```
spadgett: Can you humanize this as well on the service page (which shows routes in a table) and the route details page?

app/scripts/filters/resources.js
```diff
@@ -1022,6 +1022,20 @@ angular.module('openshiftConsole')
       }
     };
   })
+  .filter('humanizeTLSTermination', function () {
```
spadgett: 👍

app/views/directives/osc-routing.html
```diff
@@ -170,29 +168,27 @@
...
-            ng-model="route.tls.insecureEdgeTerminationPolicy"
-            ng-disabled="route.tls.termination !== 'edge'"
-            class="form-control"
-            aria-describedby="route-insecure-policy-help">
-          <option value="">None</option>
-          <option value="Allow">Allow</option>
-          <option value="Redirect">Redirect</option>
-        </select>
+        <label for="insecureTraffic">Insecure Traffic</label>
+        <ui-select ng-model="route.tls.insecureEdgeTerminationPolicy" ng-disabled="route.tls.termination !== 'edge'" input-id="insecureTraffic" aria-describedby="route-insecure-policy-help" search-enabled="false">
```
spadgett: It looks like nothing is being preselected:

openshift_web_console

It should default to `None`.

jhadvig:
@spadgett updated... Thanks for the comments :)
added `dist` as well

app/views/directives/osc-routing.html
```diff
@@ -170,29 +168,27 @@
...
-            ng-model="route.tls.insecureEdgeTerminationPolicy"
-            ng-disabled="route.tls.termination !== 'edge'"
-            class="form-control"
-            aria-describedby="route-insecure-policy-help">
-          <option value="">None</option>
-          <option value="Allow">Allow</option>
-          <option value="Redirect">Redirect</option>
-        </select>
+        <label for="insecureTraffic">Insecure Traffic</label>
+        <ui-select ng-model="route.tls.insecureEdgeTerminationPolicy" ng-init="route.tls.insecureEdgeTerminationPolicy = insecureTrafficOptions[0]" ng-disabled="route.tls.termination !== 'edge'" input-id="insecureTraffic" aria-describedby="route-insecure-policy-help" search-enabled="false">
```
spadgett: This `ng-init` will change the value if I edit an existing route. I suggest you initialize it in the controller, but only if not set.

jhadvig:
@spadgett updated. Thanks !

app/views/directives/osc-routing-service.html
```diff
@@ -4,14 +4,12 @@
...
-          id="{{id}}-service-select"
-          ng-model="model.service"
-          ng-options="service as service.metadata.name for service in services"
-          required
-          class="form-control"
-          ng-attr-aria-describedby="{{id}}-service-help">
-      </select>
+      <ui-select ng-model="model.service" input-id="{{id}}-service-select" aria-describedby="{{id}}-service-help" required>
+        <ui-select-match>{{$select.selected.metadata.name}}</ui-select-match>
+        <ui-select-choices repeat="service in services track by (service | uid) | filter : $select.search">
```
spadgett: Sorry missed this before, but `track by` should come last, after `filter`

jhadvig:
@spadgett done :)

spadgett:
thanks @jhadvig 

[merge]

PR #1027, opened 2016-12-12
Show help about compute units

spadgett:
Add a link to resource limit forms and metrics pages that explain compute units like millicores and MiB. The link opens this dialog:

screen shot 2016-12-12 at 3 10 34 pm

cc @derekwaynecarr 

Fixes https://github.com/openshift/origin/issues/7768

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -275,6 +275,7 @@
         <script src="scripts/controllers/modals/processTemplateModal.js"></script>
         <script src="scripts/controllers/modals/linkService.js"></script>
         <script src="scripts/controllers/modals/jenkinsfileExamplesModal.js"></script>
+        <script src="scripts/controllers/modals/aboutComputeUnitsModal.js"></script>
         <script src="scripts/controllers/about.js"></script>
         <script src="scripts/controllers/commandLine.js"></script>
         <script src="scripts/controllers/createPersistentVolumeClaim.js"></script>
--- a/app/scripts/controllers/modals/aboutComputeUnitsModal.js
+++ b/app/scripts/controllers/modals/aboutComputeUnitsModal.js
@@ -1,0 +1,8 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .controller('AboutComputeUnitsModalController', function($scope, $uibModalInstance) {
+    $scope.ok = function() {
+      $uibModalInstance.close('ok');
+    };
+  });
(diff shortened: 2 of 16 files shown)


spadgett:
openshift_web_console

app/views/create/fromimage.html
```diff
@@ -365,6 +365,9 @@
                               expand="true"
                               can-toggle="false"
                               >
+                              <div class="learn-more-block">
+                                <a href="" ng-click="showComputeUnitsHelp()">What are millicores and MiB?</a>
+                              </div>
```
spadgett: @jwforres Different text here because there is already an "About Resource Limits" link for the form section. Let me know if you have a better idea for where to put it.

openshift_web_console

jwforres:
@spadgett in your first screenshot, what if "CPU and memory" was a link instead? or maybe the order of the two links just needs to be reversed, i think it would make more sense to have the simple About Computer Resources and then the Learn More ->

spadgett:
@jwforres I think I like changing the order. Or I could even remove the learn more link since it's in the dialog.

spadgett:
Pushed a change to switch the order for now

jwforres:
for the resource limits editor, I think I would rather have a ? link next to the units dropdown than have the full "What are millicores and MiB?"  Its tough since ? is usually reserved for the popups...

spadgett:
Maybe in the help text?

```html
The minimum amount of CPU the container is guaranteed. Can be specified in
<a href="">millicores or cores</a>.
```

jwforres:
@spadgett yep i like that idea

spadgett:
@jwforres Not sure about it now that I've tried it. It's odd since it's repeated twice for CPU and twice for memory, and they all bring up the same dialog.

screen shot 2016-12-13 at 1 28 25 pm

jwforres:
What if you took off the sentence you added and instead made "CPU" the link, i think that still works with the content of the dialog.

spadgett:
screen shot 2016-12-13 at 3 58 24 pm

jwforres:
Yeah i'm fine with this, how do you feel about it?

spadgett:
With input from @rhamilto 👍 

screen shot 2016-12-14 at 9 16 05 am

jwforres:
screenshot LGTM, let me know when the changes are pushed

spadgett:
@jwforres changes pushed

jwforres:
[merge]

spadgett:
Resolved dist conflict

[merge]

PR #12, opened 2016-05-16
Bug 1334630 - update Template's objects, params and labels

jhadvig:
@spadgett PTAL

PR #1032, opened 2016-12-13
Make deployment and build config pages more consistent

### LINKED ISSUE #880: Inconsistencies between build and deploy pages
Perhaps a minor thing, though it is interesting that the build configuration page is structured as:

* Summary
* Configuration
* Environment
* (doesn't include events)

But the deployment configuration page is structured as:

* Details
  * (includes configuration)
* Environment
* Events

Seems like it would make sense to follow the same pattern from both. For example, if I'm looking for **Triggers** I need to look in different sections.

spadgett:
Closes #880

cc @sspeiche @jwforres @ajacobs21e 

screen shot 2016-12-13 at 12 26 08 pm

### DIFF
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -75,8 +75,8 @@
             <div class="row" ng-if="loaded">
               <div class="col-md-12" ng-class="{ 'hide-tabs' : !buildConfig }">
                 <uib-tabset>
-                  <uib-tab active="selectedTab.summary">
-                    <uib-tab-heading>Summary</uib-tab-heading>
+                  <uib-tab active="selectedTab.history">
+                    <uib-tab-heading>History</uib-tab-heading>
 
                     <!-- If unfilteredBuilds is undefined, the builds are still loading. -->
                     <div ng-if="!unfilteredBuilds" class="gutter-bottom">Loading...</div>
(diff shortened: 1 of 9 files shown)


jwforres:
This does give us a little space for doing a more detailed status there, like the metrics for the current deployment and / or the pod donut? doesnt necessarily need to happen in this PR just throwing the idea out

spadgett:
@jwforres PTAL, ready for review if you're OK with the change. I think it's an improvement.

app/views/browse/build-config.html
```diff
@@ -75,8 +75,8 @@
             <div class="row" ng-if="loaded">
               <div class="col-md-12" ng-class="{ 'hide-tabs' : !buildConfig }">
                 <uib-tabset>
-                  <uib-tab active="selectedTab.summary">
-                    <uib-tab-heading>Summary</uib-tab-heading>
+                  <uib-tab active="selectedTab.builds">
```
jwforres: this will invalidate bookmarked URLs from 3.4, can you just validate if someone comes in with tab=summary that it correctly resets the url to tab=builds

app/views/browse/build-config.html
```diff
@@ -133,10 +133,10 @@
...
                     <div ng-if="builds && (builds | hashSize) > 0" class="build-config-summary">
                       <!-- Show the latest build status first. -->
-                      <div class="h3" style="margin-bottom: 0;">
-                        <span class="latest-build-status">
-                          <status-icon status="latestBuild.status.phase" style="margin-right: 5px;"></status-icon>
-                          Latest build
+                      <div class="h3">
+                        <span class="last-status">
+                          <status-icon status="latestBuild.status.phase"></status-icon>
+                          Last build
```
jwforres: hmm i actually dont care for this change, i dont like "Last" in the context of a running build

app/views/browse/deployment.html
```diff
@@ -75,25 +75,62 @@
...
+                rollouts from running until the deployment is resumed.
+                <span ng-if="!updatingPausedState && ({ group: 'extensions', resource: 'deployments' } | canI : 'update')">
+                  <a href="" ng-click="setPaused(false)" role="button">Resume deployment</a>
+                </span>
+              </div>
               <uib-tabset>
-                <uib-tab active="selectedTab.details">
-                  <uib-tab-heading>Details</uib-tab-heading>
+                <uib-tab active="selectedTab.replicaSets">
+                  <uib-tab-heading>Replica Sets</uib-tab-heading>
```
jwforres: how would you feel about "History" instead.  I don't really like the tab being called Replica Sets

spadgett: The command is also `oc rollout history`, so I think "History" is better

app/views/browse/build-config.html
spadgett: Yeah I was trying to avoid "Latest deployment" since that has a specific meaning (and not always most recent). I changed it here to be consistent, but I can change it back.

app/views/browse/build-config.html
spadgett: The link works and shows the first tab, but it doesn't update the param in the URL. I can change that if you like.

app/views/browse/build-config.html
spadgett: Maybe I'll just take out "Last" and say, "Build #12 is running."

jwforres:
sure that works

jwforres:
nah, as long as the link works I'm ok with it. I suspect it is just falling
back to the first tab as a default.

spadgett:
@jwforres thanks, updated

spadgett:
@jwforres updated to use "History" as tab header

jwforres:
[merge]

PR #1034, opened 2016-12-13
Bug 1404417 - Don't show negative durations, they are always due to client clock skew

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1404417


jwforres:
This actually fixes two problems, it prevents showing negative duration, but it also fixes the fact that a zero duration was being returned as empty string, when really we wanted it to be 0s

spadgett:
[merge]

PR #1035, opened 2016-12-13
Improving display of <status-icon>s

### LINKED ISSUE #1023: Alignment and spacing around <status-icon>s is inconsistent
In working on #1016, I noticed sometimes &lt;status-icon&gt;s have margin-right.  Some times they don't.  Some times they use .fa-fw.  Some times they don't.  The icon size differs based on the context (which is ok), so make sure an fixes for the aforementioned aren't negatively impacted by the size.

rhamilto:
Fixes #1023, improving spacing and alignment issues with the status-icons directive.  

Note:  status-icons always get a fixed width now so they're consistent everywhere given the nature of the directive to change the icon based upon a dynamically changing status.  As a result, I removed the fixedWidth option from the directive.

Before:

screen shot 2016-12-14 at 10 36 00 am

fkycukdcng

After:

screen shot 2016-12-14 at 10 31 22 am

x4omimoxt8

@jwforres or @spadgett, PTAL

### DIFF
--- a/app/scripts/directives/statusIcon.js
+++ b/app/scripts/directives/statusIcon.js
@@ -8,8 +8,7 @@ angular.module('openshiftConsole')
         templateUrl: 'views/directives/_status-icon.html',
         scope: {
           status: '=',
-          disableAnimation: "@",
-          fixedWidth: "=?"
+          disableAnimation: "@"
         },
         link: function($scope, $elem, $attrs) {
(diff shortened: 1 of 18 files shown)


spadgett:
Looks like you removed `disable-animation` in a few places. Just checking if intentional.

spadgett:
[merge]

Issue #319, opened 2016-08-08
Pod page should link back to deployment and deployment config


spadgett:
There's no link to the RC or deployment that owns the pod on the pod page. This would be convenient for debugging. For instance, I was recently looking at an error in a pod log and needed to change an env var in the deployment config. It's a lot of clicks to get there.

PR #1039, opened 2016-12-14
[WIP] Show deployment in breadcrumbs for pod

spadgett:
When the pod is part of a deployment, show the deployment config and deployment number as links in the breadcrumbs.

See #319

screen_shot_2016-12-14_at_1_22_42_pm

### DIFF
--- a/app/scripts/controllers/pod.js
+++ b/app/scripts/controllers/pod.js
@@ -13,6 +13,7 @@ angular.module('openshiftConsole')
                                          $timeout,
                                          $uibModal,
                                          Logger,
+                                         BreadcrumbsService,
                                          DataService,
                                          ImageStreamResolver,
                                          MetricsService,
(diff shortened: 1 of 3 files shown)


spadgett:
@jwforres Opinion on this change? I can't decide.

It's nice for navigation, but it's not obvious that you're looking at a pod. The type in the breadcrumbs was the only clue. I feel like that's a general problem we should solve, though.

spadgett:
Might be better just to link to the deployment config and deployment somewhere in the page content

jwforres:
If I had navigated to the page from the deployment page I could see it, but if I navigated there from the pods page it would be weird that my breadcrumb didn't reflect that.  I say keep it how we have it and your other PR at least gets us the easy link back.

spadgett:
Fair enough, closing for now

PR #1040, opened 2016-12-14
Let users save logs

### LINKED ISSUE #356: Need a way to copy full log text
For logs with thousands of lines, it's very annoying to select the text manually in the web console. It would be nice to provide a link that either copies the content or at least selects it for the user.

spadgett:
Adds a "Save" link to the log viewer that saves the log contents as a file. Closes #356

https://trello.com/c/OaF7o3U3

Saves just the log text for logs that have links or colors. Does not add in the line numbers.

openshift_web_console

Uses library https://github.com/eligrey/FileSaver.js which is very small (< 200 lines of code). See browser support here:

  https://github.com/eligrey/FileSaver.js#supported-browsers

Tested with Chrome, Firefox, Safari, IE, and iOS. On iOS, it opens a new window with the plain log text:

screen shot 2016-12-14 at 2 22 54 pm

cc @benjaminapetersen

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -36,6 +36,7 @@
     "Clipboard": false,
     "$": false,
     "_" : false,
-    "URITemplate": false
+    "URITemplate": false,
+    "saveAs": false
   }
 }
--- a/app/index.html
+++ b/app/index.html
@@ -153,6 +153,7 @@
     <script src="bower_components/js-yaml/dist/js-yaml.js"></script>
     <script src="bower_components/angular-moment/angular-moment.js"></script>
     <script src="bower_components/angular-utf8-base64/angular-utf8-base64.js"></script>
+    <script src="bower_components/file-saver/FileSaver.js"></script>
     <!-- endbower -->
     <!-- endbuild -->
 
@@ -209,6 +210,7 @@
         <script src="scripts/services/labels.js"></script>
         <script src="scripts/services/catalog.js"></script>
         <script src="scripts/services/modals.js"></script>
+        <script src="scripts/services/cliHelp.js"></script>
         <script src="scripts/controllers/projects.js"></script>
         <script src="scripts/controllers/pods.js"></script>
         <script src="scripts/controllers/pod.js"></script>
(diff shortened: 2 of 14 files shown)


jwforres:
I really hesitate to call this Save if it's only copying the part of the log that we have loaded

spadgett:
We could warn if it's only part of the log (and give the CLI command)

benjaminapetersen:
Nice to get this feature back with such a small lib!

spadgett:
Something like this:

screen shot 2016-12-14 at 4 30 03 pm

jwforres:
Sure but the command should be the full width copy to clipboard

spadgett:
> Sure but the command should be the full width copy to clipboard

Yeah, needs a custom dialog to use the directive. Working on it.

spadgett:
@jwforres Updated. Note I added some tests for `CLIHelp` (bottom of diff).

Here is the current dialog:

screen shot 2016-12-15 at 8 59 06 am

jwforres:
does the command line tools link open in a new tab? i would prefer that
over it dismissing the dialog

spadgett:
> does the command line tools link open in a new tab?

Yes

jwforres:
What happens if we have 5000 really long log lines and we hit the 500MiB limit in some of the browsers?  Or does the log viewer already fall over at that point :)

spadgett:
We limit the log to 10MiB

https://github.com/spadgett/origin-web-console/blob/316d74fe26de8b01c7d77d52bf77e8ad980df815/app/scripts/directives/logViewer.js#L299

jwforres:
Can you add a comment above that line that if we ever increase that limit that we can't go over the Blob limit

jwforres:
otherwise this LGTM

spadgett:
Thanks @jwforres, added the comment about `limitBytes`.

[merge]

spadgett:
Fixed typo in test string

[merge]

spadgett:
[merge]

Issue #1031, opened 2016-12-13
Improve service traffic table on mobile


spadgett:
Two issues:

* It's not clear where each row ends in the mobile layout. They run together.
* The arrows should be hidden or changed to down arrows like we do for pipelines.

Screenshot:

![screen shot 2016-12-13 at 10 20 00 am](https://cloud.githubusercontent.com/assets/1167259/21145880/cdffb27c-c11d-11e6-8fe3-76daf7d70f32.png)

spadgett:
Also seems to be an extra table cell at the bottom.

juanvallejo:
@spadgett 
> It's not clear where each row ends in the mobile layout. They run together.

Hm, this seems to be happening because the [traffic-table directive](https://github.com/openshift/origin-web-console/blob/master/app/views/directives/traffic-table.html#L18) has the `ng-repeat` attribute on the `<tr>` tag rather than on the `<tbody>`. [pods-table](https://github.com/openshift/origin-web-console/blob/master/app/views/directives/pods-table.html#L15) and `app/views/browse/routes.html` both have `ng-repeat` on the `<tbody>`.

cc @jwforres

PR #1042, opened 2016-12-14
[WIP] improve service traffic table on mobile

juanvallejo:
Fixes https://github.com/openshift/origin-web-console/issues/1031

**todo**
- [ ] fix double border (top) in traffic table (depends on https://github.com/openshift/origin-web-console/pull/1045)
- [ ] move ng-repeat from `<tbody>` to `<tr>` (fixed by https://github.com/openshift/origin-web-console/pull/1045)
- [x] Add 2px border between rows on mobile
- [x] Arrows change to down arrows on mobile
- [x] Fix extra space on bottom of table

table top

service-page-tbody-loop

table bottom

service-page-traffic-table-loop-tbody-bottom

cc @spadgett


app/views/directives/traffic-table.html
```diff
@@ -14,9 +14,9 @@
   <tbody ng-if="(portsByRoute | hashSize) == 0">
     <tr><td colspan="7"><em>{{emptyMessage || 'No routes or ports to show'}}</em></td></tr>
   </tbody>
-  <tbody ng-if="(portsByRoute | hashSize) > 0">
-    <tr ng-repeat-start="(routeName,ports) in portsByRoute" style="display:none;"></tr>
-    <tr ng-repeat="port in ports" ng-if="routeName !== ''">
+  <tbody ng-if="(portsByRoute | hashSize) > 0" ng-repeat-start="(routeName,ports) in portsByRoute" style="display:none;"></tbody>
+  <tbody ng-repeat="port in ports" ng-if="routeName !== ''">
```
juanvallejo: @spadgett The reason why the double border occurs at the top of the table, is because I am "nesting" two `ng-repeat`s using sibling `tbody` elements. Although the first one is hidden with `display: none`, it is causing the second one to have a `2px` border when it otherwise would not due to it normally being the [first `tbody` element](https://github.com/openshift/origin-web-console/blob/master/app/styles/_tables.less#L123).

juanvallejo: A solution would be to add the `2px` top border to `tbody > tr + tr` instead. That way, we could keep the `ng-repeat` in the `tr` element, and update the `pods-table` directive and `routes.html` to do this as well.

cc @jwforres

spadgett: I think we have the 2px border on other responsive tables. Can you check? If so, let's open a separate issue to fix them all.

cc @rhamilto @sg00dwin

juanvallejo: @spadgett 
> I think we have the 2px border on other responsive tables. Can you check? If so, let's open a separate issue to fix them all.

It seems we do have a 2px border on the `pods-table` as well as the routes table but it appears to come from the table's top border rather than the `tbody`. In the `traffic-table`, the table's top border is adding `2px`, and an additional `2px` are getting added by the first visible `tbody` due to the hidden `tbody` right before it.
 
pods-table-mobile

I can go ahead and open an issue to address this by repeating the `tr` tag rather than the `tbody`, and apply this change to the pods and routes table as well.

app/styles/_core.less
```diff
@@ -1209,3 +1209,12 @@ copy-to-clipboard .input-group.limit-width {
     width: 100%;
   }
 }
+
+// swaps direction of arrow in table cell
+td[role="presentation"].arrow:after {
```
spadgett: Probably belongs in _tables.less

juanvallejo:
@spadgett Thanks, review comment addressed.

> How does it look if we use text-center on the arrows?

Because the table cell is a bit wide, they look offset from the rest of the content if I align them center.
We could add some padding to the left, but I am not sure if this would be a consistent solution for all cases.

service-page-arrows-center-mobile

juanvallejo:
Related PR: https://github.com/openshift/origin-web-console/pull/1045

spadgett:
[merge]

Issue #1038, opened 2016-12-14
Unable to create fromFile List without metadata field 


jhadvig:
When trying to create resources via List in the fromFile tab, on Add to project page, an alert saying `Resource is missing metadata field.` is shown and I'm unable to create those resources.

![screenshot-19](https://cloud.githubusercontent.com/assets/1668218/21193121/64f4b9ee-c22b-11e6-87d1-d696dbc64e02.png)

PR #1043, opened 2016-12-15
Create List without metadata field on fromFile page

jhadvig:
@spadgett PTAL

closes https://github.com/openshift/origin-web-console/issues/1038

### DIFF
--- a/app/scripts/directives/fromFile.js
+++ b/app/scripts/directives/fromFile.js
@@ -114,15 +114,17 @@ angular.module("openshiftConsole")
             }
           }
 
-          // Top level resource field check.
-          if (!validateFields(resource)) {
+          if (!isKindValid(resource)) {
             return;
           }
 
           $scope.resourceKind = resource.kind;
+          $scope.resourceKind.endsWith("List") ? $scope.isList = true : $scope.isList = false;
 
-          if ($scope.resourceKind.endsWith("List")) {
-            $scope.isList = true;
+          if (!isMetadataValid(resource)) {
+            return;
+          }
+          if ($scope.isList) {
             $scope.resourceList = resource.items;
             $scope.resourceName = '';
           } else {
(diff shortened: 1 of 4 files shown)


app/scripts/directives/fromFile.js
```diff
@@ -178,24 +178,25 @@ angular.module("openshiftConsole")
...
-              message: "Resource name is missing in metadata field."
-            };
-            return false;
-          }
-          if (item.metadata.namespace && item.metadata.namespace !== $scope.projectName) {
-            $scope.error = {
-              message: item.kind + " " + item.metadata.name + " can't be created in project " + item.metadata.namespace + ". Can't create resource in different projects."
-            };
-            return false;
+          if (!item.kind.endsWith("List")) {
```
spadgett: I'd rather return early here.

```js
if (item.kind.endsWith("List")) {
  // Nothing else to check for lists.
  return true;
}
```

Instead of having a large `if` block. It's easier to reason about this way.

jhadvig:
@spadgett update

app/scripts/directives/fromFile.js
```diff
@@ -178,14 +179,16 @@ angular.module("openshiftConsole")
             };
             return false;
           }
+          if (item.kind.endsWith("List")) {
```
spadgett: Is it possible to check `$scope.isList` here, or has that not been updated yet?

jhadvig: `$scope.isList` is set after the validation, but we can update the code so we set the variable before it.

jhadvig:
@spadgett updated... I've spit the validation into two function:
- validating kind
- validating metadata fields

and between them the `$scope.isList` var will be set.

app/scripts/directives/fromFile.js
```diff
@@ -170,22 +172,29 @@ angular.module("openshiftConsole")
           });
         };
 
-        // Takes item that will be inspect for kind, metadata fields and if the item is meant to be created in current namespace
-        function validateFields(item) {
+        // Takes item that will be inspect kind field.
+        function validateKind(item) {
```
spadgett: Looks good. Just one nit on the name: I think `isKindValid` and `isMetadataValid` are better names for these functions. The current name sounds like "Should I validate the metadata?" rather than "Is the metadata valid?"

spadgett:
LGTM 👍  except for the one comment

jhadvig:
@spadgett comment addressed :) thanks

spadgett:
[merge]

PR #1045, opened 2016-12-15
Improve accessibility: use ng-repeat on <tr> instead of <tbody>

juanvallejo:
This patch repeats table rows rather than table bodies for any list of
data displayed in a table.

It also resolves the issue of a double top-border on tables when there are
nested ng-repeat loops. (In these cases, a double top border was
appearing due to the need for a sibling <tr> element to contain the
outer loop. This extra element caused the css rule "tbody > tr + tr" to
apply a top-border to the first visible <tr> element as well, making the
table's already existing top-border redundant).

deployments-table

@spadgett @jwforres @rhamilto @sg00dwin

### DIFF
--- a/app/styles/_tables.less
+++ b/app/styles/_tables.less
@@ -92,7 +92,7 @@
 
 @media (max-width: @screen-xs-max) {
   .table-mobile {
-    border-top: 2px solid @table-border-color;
+    border-top-width: 0;
     // so that .word-break() will work
     table-layout: fixed;
     colgroup, col {
@@ -117,10 +117,8 @@
       }
     }
     > tbody {
-      + tbody {
-        border-top-width: 2px;
-      }
       > tr {
+        border-top: 2px solid @table-border-color;
         > td {
            /* Behave like a row */
           border: none;
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -187,8 +187,8 @@
                         <tbody ng-if="(builds | hashSize) == 0">
                           <tr><td colspan="3"><em>{{emptyMessage}}</em></td></tr>
                         </tbody>
-                        <tbody ng-repeat="build in orderedBuilds">
-                          <tr>
+                        <tbody ng-if="(builds | hashSize) > 0">
+                          <tr ng-repeat="build in orderedBuilds">
                             <td data-title="Build">
                               <!-- Build number and link -->
                               <span ng-if="build | annotation : 'buildNumber'">
--- a/app/views/browse/config-maps.html
+++ b/app/views/browse/config-maps.html
@@ -44,8 +44,8 @@
                   <tbody ng-if="(configMaps | hashSize) == 0">
                     <tr><td colspan="3"><em>No config maps to show</em></td></tr>
                   </tbody>
-                  <tbody ng-repeat="configMap in configMaps">
-                    <tr>
+                  <tbody ng-if="(configMaps | hashSize) > 0">
+                    <tr ng-repeat="configMap in configMaps">
                       <td data-title="Name">
                         <a href="{{configMap | navigateResourceURL}}">{{configMap.metadata.name}}</a>
                       </td>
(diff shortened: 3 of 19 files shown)


app/styles/_core.less
```diff
@@ -1209,3 +1209,14 @@ copy-to-clipboard .input-group.limit-width {
     width: 100%;
   }
 }
+
+// handle top-border in table rows inside of nested loops
+tbody.nested-loop {
+  > tr {
+    + tr {
+      &:nth-child(2) {
+        border-top-width: 0;
```
juanvallejo: @rhamilto I was not sure if this was the correct approach to fixing the double border on tables that need a hidden `<tr>` before the actual first visible `<tr>` element. wdyt?

rhamilto: I think @sg00dwin is probably a better qualified as I know he has recently fixed some stuff in _tables.less.

app/styles/_tables.less
```diff
@@ -143,6 +143,19 @@
...
+    // since we cannot select the "first element of parent with class ..."
+    // select every <tr> with the .nested-loop class, apply a 0px border-top-width,
+    // and then re-apply the 2px border-top-width to every tr.nested-loop child
+    // except the first one to avoid tables with 4px border-top-widths.
+    > tbody.nested-loop-container {
+      > tr.nested-loop {
+        border-top-width: 0;
+      }
+      > tr.nested-loop ~ tr.nested-loop {
+          border-top-width: 2px;
```
juanvallejo: @sg00dwin @jwforres @spadgett 
This works for all cases where there must be sibling elements (with the first one set to `display: none;`) in order to have nested loops (`traffic-table.html` and `deployments.html` in this PR). I first remove the border from all `<tr>` elements in the loop, and then re-apply it to every `<tr>` in that loop except the first one with `tr.nested-loop ~ tr.nested-loop`.

spadgett: I guess I'm not clear on why there is a top border on the first hidden row. If `display: none`, shouldn't the border be hidden as well?

juanvallejo: @spadgett 

The border comes from the first non-hidden `<tr>` (with className  `nested-loop`) (see attached image), but normally, this first non-hidden `<tr>` would not have a `border-top-width` due to the css rule `tbody > tr + tr {...` which would only apply the border to `<tr>` tags that immediately follow other `<tr>` tags. Normally the top `2px` border that is seen is the overall table's top-border. Because the hidden `<tr>` right above the one in the image still exists in the dom, it forces this first visible `<tr>` to have a `border-top-width`, which combined with the table's existing `2px` results in the top border with extra width that we are currently seeing in these tables.

To get around this, I select all non-hidden `<tr>` elements by adding the class `nested-loop` to them, remove their `border-top-width`, and then re-apply it to every one except the first `.nested-loop` tr. 

service-page-first-element-w-class

spadgett:
Does it work if we change the `ng-repeat-start` to just `ng-repeat` and remove the hidden `ng-repeat-end` table rows?

juanvallejo:
@spadgett 
> Does it work if we change the ng-repeat-start to just ng-repeat and remove the hidden ng-repeat-end table rows?

I have tested this on both [imagestream.html](https://github.com/openshift/origin-web-console/pull/1045/files#diff-5e6514233d7275a46f60dc9c94b50b98), and [deployments.html](https://github.com/openshift/origin-web-console/pull/1045/files#diff-f70bf6fb410fa0a40a8998349a7a1431) and it only seems to work in `imagestreams.html`. I have updated `imagestreams.html` in the latest commit https://github.com/openshift/origin-web-console/pull/1045/commits/58a95e48afade79413be1c290def3741172228cc.

juanvallejo:
@spadgett 
> @juanvallejo looks good, please squash your commits and I'll merge

Thanks! done

spadgett:
[merge]

PR #1046, opened 2016-12-15
Remove hover opacity change from the metrics-donut to be consistent with the pods-donut

sg00dwin:
@spadgett

### DIFF
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -280,7 +280,8 @@ code.command {
   cursor: pointer;
 }
 
-.pod-donut {
+.pod-donut,
+.metrics-donut {
   .c3-defocused {
     // Removing hover opacity change which caused the "c3-Running" state to look very similar to "c3-Not-Ready" state
     // https://github.com/openshift/origin/issues/9253
--- a/dist/styles/main.css
+++ b/dist/styles/main.css
@@ -3697,11 +3697,11 @@ code.command{display:inline-block;line-height:1.3;margin-right:2px}
 }
 .builds-block .builds .build:first-child{margin-top:0}
 .builds-block .builds .build+.build:after{background-color:#e0e0e0;top:-6px;content:"";display:block;height:1px;left:0;right:0;position:absolute}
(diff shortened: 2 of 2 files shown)


spadgett:
We have other donuts if you want to remove the opacity from them all. But I like the effect. I'm not sure we should remove it everywhere. We had a specific reason to on the overview.

To be honest, I'd almost rather revert the change to the overview donuts.

cc @jwforres @ajacobs21e

sg00dwin:
The opacity change from donut charts on the overview addressed usability issue openshift/origin#9253 I assumed we would be consistent is how these behave.

spadgett:
Right, I understand why we made the change. I feel like we made the general usability worse to fix an edge case, though. The opacity change is another helpful cue that the tooltip is only for that arc you're hovering over.

ajacobs21e:
I think it's difficult since we use so many close shades of blue. Is it an option to change some of the colors?
Is it possible to fade the non-highlighted portions more (aka make them very light) so that it doesn't conflict with our status colors?

spadgett:
> I think it's difficult since we use so many close shades of blue. Is it an option to change some of the colors?

This was deliberate in that the pod transitions from running, not ready (light blue) to running and ready (blue). We wanted to use blue consistently to indicate running. The states are similar. (If the pod takes too long to get ready, we change the color to orange as a warning.)

> Is it possible to fade the non-highlighted portions more (aka make them very light) so that it doesn't conflict with our status colors?

This might be possible. @sg00dwin Do we have control over the opacity?

sg00dwin:
@spadgett @ajacobs21e yes we can override the default opacity change for the defocused state, which is `.c3-defocused {opacity: .3} ` 

**Here are some example for an opacity of .2**
defocused2-pulling
defocused2-running
defocused2-terminating

**And here is .15**
at this point the light gray pretty much falls out
defocused1 5-pulling
defocused1 5-running
defocused1 5-terminating
defocused1 5-pending

spadgett:
Replaced by #1079

Issue #785, opened 2016-11-01
Different select styles on add to project page


spadgett:
We have a mix of normal HTML `select` elements and `ui-select` elements on the create from source page now that we've added secrets. They have different styles that clash. We should consider switching them all to `ui-select` for consistency.

![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/19904448/03c1efa6-a049-11e6-94a8-dadbac3ac0bb.png)

cc @jwforres @rhamilto

jhadvig:
Part 1 : https://github.com/openshift/origin-web-console/pull/1024

PR #1049, opened 2016-12-19
Replace select tag with ui-select - part 2

jhadvig:
This is the first part of work which is moving all remaining select tags to ui-select, which contains :
- select container for logs (monitoring page, pod browse page)
- selecting jenkinsfile type on BC editor
- pod browse page -> metrics tab
- selecting compute resource unit on `_compute-resource.html` and `osc-persistant-colume-claim` directive

@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/785

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -581,6 +581,45 @@ label.checkbox {
   }
 }
 
+.compute-resource {
+  .resource-size {
+    display: flex;
+    height: 27px;
+    .resource-amount {
+      width: 70%;
+      input {
+        width: 100%;
+      }
+    }
+    .resource-unit {
+      width: 30%;
+      .ui-select-container {
+        height: 27px;
+      }
+      .ui-select-choices,
+      .ui-select-toggle {
+        width: 100%;
+      }
+    }
+    @media (min-width: @screen-sm-min) {
+      .resource-amount {
+        width: 80%;
+      }
+      .resource-unit {
+        width: 20%;
+      }
+    }
+    @media (min-width: @screen-md-min) {
+      .resource-amount {
+        width: 85%;
+      }
+      .resource-unit {
+        width: 15%;
+      }
+    }
+  }
+}
+
 .next-steps {
   .tile {
     margin-bottom: 0;
--- a/app/styles/_log.less
+++ b/app/styles/_log.less
@@ -11,6 +11,19 @@
     .text-muted();
     margin: 0 2px;
   }
+  .ui-select-container {
+    min-width: 150px;
+    max-width: 250px;
+    margin: 0 5px;
+  }
+  .container-details {
+    display: inline-flex;
+    align-items: center;
+    flex-wrap: wrap;
+    .container-state {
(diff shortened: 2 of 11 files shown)


app/styles/_core.less
```diff
@@ -571,6 +571,32 @@ label.checkbox {
   }
 }
 
+.compute-resource {
+  .resource-size {
+    display: flex;
```
spadgett: Since we've switch to flexbox and changing these styles already, is it possible to make the input 100% wide? I think it would look much better in our forms since the other controls are 100%.

For instance, instead of:

screen shot 2016-12-19 at 9 06 30 am

We could have this:

screen shot 2016-12-19 at 9 06 07 am

jhadvig:
@spadgett updated.
Here are the screens:
screenshot

screenshot-1

jhadvig:
@spadgett not sure why the travis is failing. I've generated the `dist` and also run the `verify-dist.sh` successfully.

app/styles/_core.less
```diff
@@ -571,6 +571,46 @@ label.checkbox {
...
+      }
+    }
+  }
+}
+
+@media (max-width: @screen-sm-max) {
+  .compute-resource {
+    .inline-select {
+      // Spacing below input at mobile sizes.
+      margin-top: 5px;
```
spadgett: I don't think we need this style anymore since we no longer stack.

app/styles/_core.less
```diff
@@ -571,6 +571,46 @@ label.checkbox {
...
+    }
+    .resource-size {
+      .resource-amount {
+        width: 70%;
+      }
+      .resource-unit {
+        width: 30%;
```
spadgett: Better to make these widths the default and then change them for larger sizes ("mobile first" styles). Maybe

```css
.compute-resource {
  .resource-size {
    display: flex;
    height: 27px;
    .resource-amount {
      width: 70%;
      input {
        width: 100%;
      }
    }
    .resource-unit {
      width: 30%;
      .ui-select-container {
        height: 27px;
      }
      .ui-select-choices,
      .ui-select-toggle {
        width: 100%;
      }
    }
    @media (min-width: @screen-sm-min) {
      .resource-amount {
        width: 80%;
      }
      .resource-unit {
        width: 20%;
      }
    }
    @media (min-width: @screen-md-min) {
      .resource-amount {
        width: 85%;
      }
      .resource-unit {
        width: 15%;
      }
    }
  }
}
```

spadgett: Interesting. For me there is plenty of room. Do we have a different font for ui-select on Linux?

screen_shot_2016-12-19_at_11_43_18_am

spadgett:
[merge]

spadgett:
> Verifying that checked in built files under dist match the source...

@jhadvig can you try `make clean build`?

jhadvig:
@spadgett I did but nothing changed. Not sure whats the issue here :-/

spadgett:
[merge]

PR #1050, opened 2016-12-19
Make tabs work with one tap on iOS

### LINKED ISSUE #811: Have to press tabs twice to switch on iOS
I suspect the latest Patternfly added a hover style to the tabs, making it so I have to press twice.

/cc @rhamilto @jwforres

spadgett:
Remove the `a:hover:before` pseudo-class on tabs so that iOS doesn't treat the first tap as a hover.

Fixes #811

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -477,6 +477,14 @@ angular
         }
       });
     }, 1000);
+  })
+  .run(function(IS_IOS) {
+    if (IS_IOS) {
+      // Add a class for iOS devices. This lets us disable some hover effects
+      // since iOS will treat the first tap as a hover if it changes the DOM
+      // content (e.g. using :before pseudo-elements).
+      $('body').addClass('ios');
+    }
   });
 
 hawtioPluginLoader.addModule('openshiftConsole');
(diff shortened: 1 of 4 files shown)


app/scripts/app.js
```diff
@@ -477,6 +477,14 @@ angular
         }
       });
     }, 1000);
+  })
+  .run(function(IS_IOS) {
+    if (IS_IOS) {
+      // Add a class for iOS devices. This lets us disable some hover effects
+      // since iOS will treat the first tap as a hover if it changes the DOM
+      // content (e.g. using :before pseudo-elements).
+      $('body').addClass('ios');
```
jwforres: trying to decide if it should be more specific like browser-ios

@rhamilto opinions?

jwforres:
otherwise LGTM

app/scripts/app.js
rhamilto: I'm good with IS_IOS.

rhamilto: D'oh.  I'm good with ios, too (thanks, @spadgett for noticing my oversight).

jwforres:
k works for me [merge]

PR #1053, opened 2016-12-19
Show a slider to select service weights

### LINKED ISSUE #241: Should be able to enter % when doing load balance on multiple services
Currently in the UI, we're asking the user to enter a number from 1-256 to load balance across multiple services.  This is not intuitive because the rest of the UI presents the information in %.  Feedback came from @gshipley

spadgett:
When editing a route with two services, let users change the service weights as percentages using a slider. For 3 or more services, fall back to entering the weight values as integers.

openshift_web_console

Fixes #241

### DIFF
--- a/app/views/directives/osc-routing.html
+++ b/app/views/directives/osc-routing.html
@@ -109,7 +109,7 @@
     <div ng-if="services">
       <osc-routing-service model="route.to"
                            services="services"
-                           show-weight="route.alternateServices.length">
+                           show-weight="route.alternateServices.length > 1 || (controls.hideSlider && route.alternateServices.length)">
       </osc-routing-service>
     </div>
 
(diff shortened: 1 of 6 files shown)


app/views/directives/osc-routing.html
```diff
@@ -155,6 +155,47 @@
...
+            id="weight-slider"
+            type="range"
+            min="0"
+            max="100"
+            step="1"
+            list="ticks"
+            ng-model="controls.rangeSlider"
+            aria-describedby="weight-slider-help"
+            class="mar-top-md">
+        <!-- Use class="hidden" to make sure the options don't render for browsers that don't support datalist. -->
```
jwforres: how does it look on the ones that do support, it looked like it was just Safari and iOS that didnt

spadgett: I can probably take it out. It doesn't show up in Safari and iOS.

spadgett:
Removed `class="hidden"` from `datalist`

jwforres:
LGTM, you can merge if you are done making changes

spadgett:
[merge]

PR #1055, opened 2016-12-19
Fixing bug where word-break wasn't working in Fx and IE/Edge

rhamilto:
on the project list page

Before:

screen shot 2016-12-19 at 4 22 33 pm

After:

screen shot 2016-12-19 at 4 22 00 pm

@spadgett, PTAL


spadgett:
[merge]

Issue #1054, opened 2016-12-19
White block at bottom of create route form


spadgett:
Not sure if this is a problem on other pages. 

![screen_shot_2016-12-19_at_3_58_40_pm](https://cloud.githubusercontent.com/assets/1167259/21328926/24b32596-c604-11e6-826b-7f7fee7851a2.png)

cc @rhamilto

PR #1056, opened 2016-12-19
Move padding-bottom to the middle-content child element so that it doesn't…

sg00dwin:
create a white block at the bottom of gray background pages.

Fixes https://github.com/openshift/origin-web-console/issues/1054

screen shot 2016-12-19 at 4 53 35 pm

screen shot 2016-12-19 at 4 35 53 pm

screen shot 2016-12-19 at 4 28 20 pm

### DIFF
--- a/app/styles/_substructure.less
+++ b/app/styles/_substructure.less
@@ -79,17 +79,20 @@ html,body {
         }
         .middle-content {
           .flex(@columns: 1 1 auto);
-          // default 50px space beneath content
-          padding-bottom: @middle-content-bottom-margin;
           position: relative;
           width: 100%;
         }
       }
     }
   }
-  & .overview .middle .middle-section .middle-container .middle-content {
-      // reduce 20px to account for overview tiles bottom margin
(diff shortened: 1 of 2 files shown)


sg00dwin:
@spadgett @rhamilto Moved the bottom spacing to the container element which corrected this issue

app/styles/_substructure.less
```diff
@@ -79,17 +79,20 @@ html,body {
...
+            padding-bottom: @middle-content-bottom-margin;
+          }
         }
       }
     }
   }
-  & .overview .middle .middle-section .middle-container .middle-content {
-      // reduce 20px to account for overview tiles bottom margin
-      padding-bottom: (@middle-content-bottom-margin - 20px);
+  & .overview .middle .middle-section .middle-container .middle-content > .container-fluid,
```
spadgett: This is extremely specific. Does it need to be? It makes it hard to change the structure of the page or override styles when we string so many class selectors together.

sg00dwin:
@spadgett I made the rule to less specific.

spadgett:
[merge]

PR #1059, opened 2016-12-21
Improve Git URL error messages

spadgett:
Previously if you tabbed through the Git URL field without setting a
value, it would highlight the input orange without any text explaining.
Change it to NOT highlight the field until the user types something. If
the user types something and removes it, highlight the field red and
show a required field error. This is consistent with other forms like
"New Project" and the "Name" field just above Git repository on the "Add
to Project" page.

Also add a missing required field message to the build config editor if
you remove the Git URL.


benjaminapetersen:
LGTM, tested & works well!

spadgett:
[merge]

PR #1062, opened 2016-12-21
Follow logs automatically

### LINKED ISSUE #1061: Scrollbar visible when log doesn't fill screen
screen shot 2016-12-21 at 10 37 52 am

spadgett:
* Follow logs as soon as the tab is selected.
* Always show follow log links for live logs. This lets you follow or
  stop following without having to wait for the log to go past the
  bottom of the visible page. (The go to top and go to end links will
  still be hidden unless the log is taller than the visible page.)
* Fix issue with affix-bottom being set on the follow log link, causing
  the link to sometimes appear at the bottom of the page.
* Fix references to unused `$scope.autoScroll` variable. It has since
  been renamed `$scope.autoScrollActive`.
* Fix issue where scrollbars appear even if the log doesn't fill the page.

Fixes #1061

@stevekuznetsov fyi

### DIFF
--- a/app/scripts/directives/logViewer.js
+++ b/app/scripts/directives/logViewer.js
@@ -194,7 +194,8 @@ angular.module('openshiftConsole')
               var fill = $scope.fixedHeight ? $scope.fixedHeight : Math.floor($(window).height() - contentTop - pulserHeight);
               if (!$scope.chromeless && !$scope.fixedHeight) {
                 // Add some bottom margin if not chromeless.
-                fill = fill - 35;
+                // 40px, same as `@middle-content-bottom-margin`
+                fill = fill - 40;
               }
               if (animate) {
                 content.animate({ 'min-height': fill +'px' }, 'fast');
--- a/app/scripts/directives/logViewer.js
+++ b/app/scripts/directives/logViewer.js
@@ -60,7 +60,6 @@ angular.module('openshiftConsole')
         templateUrl: 'views/directives/logs/_log-viewer.html',
         scope: {
           followAffixTop: '=?',
-          followAffixBottom: '=?',
           object: '=',
           fullLogUrl: '=?',
           name: '=',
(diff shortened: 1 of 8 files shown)


stevekuznetsov:
Cool!

app/scripts/directives/logViewer.js
```diff
@@ -197,7 +194,8 @@ angular.module('openshiftConsole')
               var fill = $scope.fixedHeight ? $scope.fixedHeight : Math.floor($(window).height() - contentTop - pulserHeight);
               if (!$scope.chromeless && !$scope.fixedHeight) {
                 // Add some bottom margin if not chromeless.
```
benjaminapetersen: I cant think of a clever way to do it, but I wish we could share variable values between JS & LESS. Hmm.

app/styles/_variables.less
```diff
@@ -40,6 +40,8 @@
 @link-hover-color-bright: lighten(@link-color, 13%);
 @list-view-expanded-bg: #f5f5f5;
 @log-bg-color: #101214;
+// The log viewer needs to be udpated if this value changes.  It accounts for
```
benjaminapetersen: Glad you added this comment!

app/views/directives/logs/_log-viewer.html
```diff
@@ -68,14 +68,13 @@
 <div ng-show="state=='logs'">
   <div class="log-view" id="{{logViewerID}}" ng-class="{'log-fixed-height': fixedHeight}">
```
benjaminapetersen: Just noticed this, for dynamic id on line 69, are we supposed to use `ng-attr-id`?  Its not part of your PR (and I bet I overlooked this initially) so prob not a big deal.

app/scripts/directives/logViewer.js
spadgett: Yeah, I can't find a cleaner way. I was thinking of trying to get the padding/margin on the element in JS, but the element with the margin keeps changing.

spadgett:
[merge]

PR #1067, opened 2016-12-21
Fix builds not showing up in builds table

spadgett:
Fixes this problem:

screen shot 2016-12-21 at 3 54 12 pm

Also fixes some problems introduced when an additional duration column was added.

Relate to #1045

### DIFF
--- a/app/views/builds.html
+++ b/app/views/builds.html
@@ -41,29 +41,28 @@
                   </tr>
                 </thead>
                 <tbody ng-if="!(latestByConfig | hashSize)">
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
Nice, this is fixed here:
screen shot 2016-12-22 at 9 54 33 am
screen shot 2016-12-22 at 9 54 57 am

spadgett:
Yeah we added a column and forgot to change the empty message `colspan`.

spadgett:
thanks @benjaminapetersen 

[merge]

PR #1068, opened 2016-12-22
Let users paste certificate values

### LINKED ISSUE #964: Allow users to paste certificate values when creating or editing routes
Right now you must upload from a file. In some cases, it is easier to copy and paste a value.

@jwforres @smarterclayton

spadgett:
When creating or editing a route, let users paste in the certificate values if they need to. Previously you could only upload from a file.

Fixes #964

screen shot 2016-12-22 at 9 32 33 am

### DIFF
--- a/app/scripts/directives/oscFileInput.js
+++ b/app/scripts/directives/oscFileInput.js
@@ -8,8 +8,7 @@ angular.module('openshiftConsole')
         model: "=",
         required: "=",
         disabled: "=ngDisabled",
-        // Show the file contents below the file input.
-        showValues: "=",
+        showTextArea: '=',
         helpText: "@?",
         dropZoneId: "@?"
       },
@@ -24,7 +23,7 @@ angular.module('openshiftConsole')
         var dropMessageSelector = "#" + scope.dropMessageID,
             highlightDropZone = false,
             showDropZone = false,
-            inputFileField = element.find('input[type=file]')[0];
+            inputFileField = element.find('input[type=file]');
 
         setTimeout(addDropZoneListeners);
 
@@ -68,11 +67,11 @@ angular.module('openshiftConsole')
         scope.cleanInputValues = function() {
           scope.model = '';
           scope.fileName = '';
-          inputFileField.value = "";
+          inputFileField[0].value = "";
         };
 
-        element.change(function() {
-          addFile(inputFileField.files[0]);
+        inputFileField.change(function() {
+          addFile(inputFileField[0].files[0]);
         });
 
         // Add listeners for the dropZone element
--- a/app/scripts/directives/oscRouting.js
+++ b/app/scripts/directives/oscRouting.js
@@ -41,22 +41,31 @@ angular.module("openshiftConsole")
         hostReadOnly: "="
       },
       templateUrl: 'views/directives/osc-routing.html',
-      controller: function($scope) {
-        $scope.disableCertificateInputs = function() {
-          var termination = _.get($scope, 'route.tls.termination');
+      link: function(scope, element, attrs, formCtl) {
+        scope.form = formCtl;
+        scope.controls = {};
+        scope.options = {
+          secureRoute: false,
+          alternateServices: false
+        };
+
+        scope.disableWildcards = Constants.DISABLE_WILDCARD_ROUTES;
+        scope.disableCertificateInputs = function() {
+          var termination = _.get(scope, 'route.tls.termination');
           return !termination || termination === 'passthrough';
         };
-        $scope.insecureTrafficOptions = [
+
+        scope.insecureTrafficOptions = [
           {value: '', label: 'None'},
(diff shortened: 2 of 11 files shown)


benjaminapetersen:
Working through this, a few thoughts, but probably not to hold up your PR.
- I skimmed right past the "secure routes" checkbox a few times expecting to see a "Security" heading at the same level as "Labels" below.  Perhaps it would help draw the eye?  
- I'm curious if we should hide the boxes that are not relevant to a selection in the TLS Termination box.  For example, if "Edge" doesn't need "Destination CA Certificate", do we need to show it? 
- I was surprised to see 4 sets, expected one option with just 2.   Clearly I need to read up more on secure routes. :smile:

spadgett:
> I skimmed right past the "secure routes" checkbox a few times expecting to see a "Security" heading at the same level as "Labels" below. Perhaps it would help draw the eye?

Agree on this, although we'd have to decide how to make it look on the add to project page, which uses those heading for major sections. Maybe a smaller header?

openshift_web_console

> I'm curious if we should hide the boxes that are not relevant to a selection in the TLS Termination box. For example, if "Edge" doesn't need "Destination CA Certificate", do we need to show it?

Yeah, maybe not. Trying to decide if we should show it and keep the existing warning if a value will be removed.

openshift_web_console

Maybe a stronger warning there saying the value will be removed from the route.

spadgett:
@benjaminapetersen I'm glad you made these comments because it got me testing. There's some stuff broken with routes (although they're fine in our 1.4 release). Working on fixes.

benjaminapetersen:
Ah, that totally makes sense if it has a value already but the type is changed so the value is no longer needed.

spadgett:
@benjaminapetersen I have showing and hiding the fields working, but it feels strange. The different termination types all support different certs. It's a bit disconcerting to see things appear and disappear when changing types. Since the certs are the last thing on the page, it can cause the page to jump/scroll when removed. Thinking we should table that.

benjaminapetersen:
Yeah, that works for me. Figured it was a note/discussion, but prob not holding up the PR if it makes things quirky.

benjaminapetersen:
Oh wait, just saw your comment about broken stuff with routes.  Not a concern now, but will be?

spadgett:
> Oh wait, just saw your comment about broken stuff with routes. Not a concern now, but will be?

About to push an update. There's an error creating edge routes currently. Also includes some improvement to the form (section headers).

spadgett:
@benjaminapetersen updated, PTAL at the second commit

benjaminapetersen:
Checking it now

benjaminapetersen:
Couple observations:
- I definitely like the headings for "Alternate Services" and "Security".  
- Checking "split traffic" is a bit jumpy.  Couple things I struggled with here:
  - The slider works well, but once I click "edit weights as integers" I couldn't go back to the slider without refreshing the page.
  - When I switch to "edit weights as integers" it took me a while to figure out why I had a dropdown with only one service, even though the slider shows 2 services:
screen shot 2016-12-22 at 2 47 03 pm
    I think its because the first service shows up above and the "weight" input now appears next to it.  I had to think about that a bit as I expected everything I was interacting with to appear below the heading.  I wonder if we could leave the above selector to be exclusive to "service to route to" and just list them all again under "split traffic" with the weight inputs (somehow show the primary service linked to the above "service to route to").
   - Once I added additional services beyond 2, it was a little easier to understand.

spadgett:
Yeah I don't disagree on the a/b services usability. I think we should revisit that (although probably a follow on at this point). We changed it to use the slider very recently.

spadgett:
Related PR #1053

benjaminapetersen:
Was just typing the same thing--I don't think that these observations relate to "paste cert value" so can def pull that out.  I'll create an issue.

app/scripts/directives/oscFileInput.js
```diff
@@ -24,7 +23,7 @@ angular.module('openshiftConsole')
         var dropMessageSelector = "#" + scope.dropMessageID,
             highlightDropZone = false,
             showDropZone = false,
-            inputFileField = element.find('input[type=file]')[0];
+            inputFileField = element.find('input[type=file]');
```
benjaminapetersen: nit: we could do `inputFileField = element.find('input[type=file]')[0];` or `inputFileField = _.first(element.find())` here & not have to repeat the `[0]` a few places below.

spadgett: Unfortunately I can't because I need the jQuery object in one place :/

spadgett:
thanks @benjaminapetersen !

[merge]

app/scripts/directives/oscFileInput.js
benjaminapetersen: Ah, true, because of the `.change` on 73, no worries!

PR #1071, opened 2016-12-22
Remove unnecessary gutter-top style from routes table

spadgett:
`middle-header` already has a bottom margin, so the style is not needed. This makes the routes table consistent with others like services, deployments, and builds.


spadgett:
[merge]

PR #1073, opened 2017-01-03
Show advanced source options when used by "Try It"

spadgett:
If the "Try It" link fills in advanced source options like git ref or context dir, show those fields (without expanding all advanced options).

Fixes https://github.com/openshift/origin/issues/9297


jwforres:
[merge]

Issue #1064, opened 2016-12-21
Service page Traffic table missing margin-bottom


rhamilto:
![screen shot 2016-12-21 at 1 47 46 pm](https://cloud.githubusercontent.com/assets/895728/21401998/8238ab06-c784-11e6-91ce-41c2d08a7ad6.PNG)

spadgett:
@sg00dwin minor thing, but can you add a period at the end of that sentence as well

spadgett:
Fixed by #1080

PR #1080, opened 2017-01-04
Increase spacing beneath traffic-table and help text.

sg00dwin:
screen shot 2017-01-04 at 1 46 35 pm

### DIFF
--- a/app/styles/_substructure.less
+++ b/app/styles/_substructure.less
@@ -290,7 +290,8 @@ html,body {
       margin-bottom: 0;
     }
   }
-  pods-table {
+  pods-table,
+  traffic-table {
     display: block;
     .gutter-bottom();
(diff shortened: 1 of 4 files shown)


sg00dwin:
@spadgett 
Fixes https://github.com/openshift/origin-web-console/issues/1064

spadgett:
Nit, but I think it's a bit too much bottom margin for the traffic table since the text is describing resources in the table. I might use something like 5px.

"Show Annotations" is different since it's not related to the pods table.

 @rhamilto opinion?

rhamilto:
> @rhamilto opinion?

I'm ok with it as is because it's still reads as part of the "Traffic" section.

spadgett:
[merge]

PR #1081, opened 2017-01-05
Use `metadata.name` as secondary sort for projects

### LINKED ISSUE #915: orderByDisplayName filter isn't stable
screen shot 2016-11-22 at 11 04 39 am

When display names are the same, the filter should consistently sort by the name.  It randomly flakes.

spadgett:
Update the projects page and `orderByDisplayName` filter to use
`metadata.name` as a secondary sort when projects have the same display
name.

Fixes #915


spadgett:
@benjaminapetersen thanks

[merge]

PR #1089, opened 2017-01-09
Bug 1411258, Bug 1411266 - Fix editing existing commands

spadgett:
The first edit made when editing an existing command is ignored. The
`argsChanged` value is not properly reset on the first `input.args` watch
callback, so changes are not saved unless there is a second edit.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1411258
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1411266


jwforres:
[merge]

PR #1090, opened 2017-01-09
Add/Edit env vars from secrets and config maps

rhamilto:
TODOs:

- [x] Correct display of valueFrom values on edit/dc (the pickers don't populate with the values)
- [x] Make second add link function
- [x] Add link to config map or secret on readonly valueFrom if user has permissions to view them
- [x] Move altTextForValueFrom filter in to kve code so the use of the filter isn't optional and overlooked
- [x] Add to additional contexts:
  - [x] create from image
  - [x] deploy-image
  - [x] deployment config custom type
  - [x] edit lifecycle hook
- [x] clean cruft (e.g., valueFrom.selected) from entry object
- [x] BUG:  after save of env vars, env var values become read only
- [x] BUG:  a user may not have permission to view secrets or config maps
  - [x] but a value may have been set by a user who does have permissions --> need readonly display for these values in the editable form
  - [x] so (s)he should not see the link to add using config maps or secrets
- [x] BUG:  unbroken strings cause faux-input-groups to grow
- [ ] [NICE TO HAVE] What to do about the default case for valueFrom, and additional options like this:
```
- name: OTHER
   valueFrom:
   resourceFieldRef:
   resource: requests.cpu 
   divisor: '0'
```
- [ ] [NICE TO HAVE] Add links to create a config map or secret (in a modal)


rhamilto:
@benjaminapetersen, @jwforres, I need an assist getting KVE running.  I'm not receiving any errors, but the editor isn't rendering.  Not sure how to troubleshoot.

screen shot 2017-01-09 at 9 31 33 am

spadgett:
@rhamilto did you add the JS files to index.html?

rhamilto:
Thanks, @spadgett.  I had overlooked the addition.

app/scripts/directives/keyValueEditor.js
```diff
@@ -286,6 +302,36 @@
                 }
               });
 
+              var findReferenceValue = function(entry) {
```
benjaminapetersen: Think its possible to break this one down a bit?  There are a lot of `return`s at various levels.  Might be helpful to clarify (via function names) what is going on?

app/scripts/directives/keyValueEditor.js
```diff
@@ -0,0 +1,422 @@
...
+            switch(refType) {
+              case 'configMapKeyRef':
+                var configMapObj = {
+                  kind: "ConfigMap",
+                  metadata: {
+                    name: from.name
+                  }
+                };
+                console.log('configMapObj', configMapObj);
+                entry.valueAlt = 'Set to the key ' + from.key + ' in config map <a ng-if="\'configMaps\' | canI : \'get\'" ng-href="{{configMapObj | navigateResourceURL}}">' + from.name + '</a>';
```
rhamilto: @benjaminapetersen or @jwforres, this doesn't create a valid href, and I'm not sure why.  Help?

benjaminapetersen: Looking into using `$compile` to get the filters to render correctly.

rhamilto:
screen shot 2017-02-10 at 9 42 47 am

@jwforres, @spadgett  Thoughts?  We're currently checking to see if the user has access to config maps or secrets, but not if the config map or secret exists.

app/views/directives/key-value-editor.html
```diff
@@ -0,0 +1,239 @@
...
+              data-placement="top"
+              data-original-title="{{entry.valueIconTooltip || valueIconTooltip}}"
+              title="{{entry.valueIconTooltip || valueIconTooltip}}"></span>
+            <div
+              class="faux-form-control readonly">
+              <span ng-switch="entry.refType">
+                <span ng-switch-when="configMapKeyRef">
+                  Set to the key {{entry.valueFrom.configMapKeyRef.key}} in config map
+                  <span
+                    ng-if="entry.isReadonlyValue">
```
jwforres: why is the appearance of the name as a link dependent on whether the value is readonly?  shouldn't the appearance of the link depend on whether the user has rights to view secrets / view config maps using a canI check?

app/views/directives/key-value-editor.html
```diff
@@ -0,0 +1,239 @@
...
+                <span ng-switch-default>
+                  {{entry.valueAlt}}
+                </span>
+              </span>
+            </div>
+          </div>
+          <div ng-if="(!isReadonlyAny) && (!entry.isReadonlyValue)">
+            <div class="ui-select">
+              <ui-select ng-model="entry.selectedValueFrom" ng-required="true" on-select="valueFromObjectSelected(entry, $select.selected)">
+                <ui-select-match placeholder="Select config map or secret">
```
jwforres: we made the link configurable to support other options down the road, but then hardcoded this one into the editor.  We made the list of objects generic to support other types, would this be better off as just "Select a resource" or something.

app/scripts/controllers/create/createFromImage.js
```diff
@@ -151,14 +162,38 @@ angular.module("openshiftConsole")
...
 
-          DataService.list("secrets", context, function(secrets) {
-            var secretsByType = SecretsService.groupSecretsByType(secrets);
-            var secretNamesByType =_.mapValues(secretsByType, function(secrets) {return _.map(secrets, 'metadata.name')});
-            // Add empty option to the image/source secrets
-            $scope.secretsByType = _.each(secretNamesByType, function(secretsArray) {
-              secretsArray.unshift("");
-            });
-          });
+          // TODO: make valueFromObjects properly
```
rhamilto: ^ @benjaminapetersen, I'm thinking this comment can simply be deleted?

app/scripts/directives/lifecycleHook.js
```diff
@@ -10,6 +10,8 @@ angular.module("openshiftConsole")
         hookParams: "=model",
         availableVolumes: "=",
         availableContainers: "=",
+        availableSecrets: "=",
+        availableConfigmaps: "=",
```
rhamilto: Good catch.  Fixed.

app/views/directives/key-value-editor.html
```diff
@@ -0,0 +1,239 @@
...
+      <!-- the value block -->
+      <div
+        class="form-group key-value-editor-input"
+        ng-class="forms.keyValueEditor[uniqueForValue(unique, $index)].$invalid ? 'has-error' : ''">
+
+        <label for="uniqueForValue(unique, $index)" class="sr-only">{{valuePlaceholder}}</label>
+
+        <div ng-if="(!entry.valueFrom)">
+          <input
+            ng-if="(!entry.valueAlt)"
```
rhamilto: Good catch.  Fixed.

app/scripts/services/images.js
```diff
@@ -2,6 +2,17 @@
...
 angular.module("openshiftConsole")
   .factory("ImagesService", function($filter, ApplicationGenerator, DataService) {
+
+    // maps an env object: { key: 'val', key2: 'val2'}
+    // to an array: [{},{},{}]
+    var makeEnvArray = function(input) {
+      return _.isArray(input) ?
+              input :
+              _.map(input, function(value, key){
+                return {name: key, value: value};
```
rhamilto: ^ @benjaminapetersen

app/views/directives/key-value-editor.html
rhamilto: We are [using canI to check if the user has rights to view config maps or secrets](https://github.com/rhamilto/origin-web-console/blob/3940968a6e3e5ab3b881f396788b85236c5f746c/app/scripts/directives/keyValueEditor.js#L197-L198), we're just using a single scope parameter (entry.isReadonlyValue) to track multiple conditions as this was easier than having multiple params.  The thought being it was better to keep this logic in the controller and have one less key to clean off entry.  Yea/nay?

1.  If the user can get config maps or secrets and the kve instance is NOT read only, we show the ui-selects for config maps and secrets.

screen shot 2017-02-15 at 9 03 35 am

2.  If the user can get config maps or secrets and the kve instance is read only, we show the faux input with the link.

screen shot 2017-02-15 at 9 03 46 am

3.  If the user can NOT get config maps or secrets and the kve instance is NOT read only, we should the faux input without the link.

screen shot 2017-02-15 at 9 11 42 am

4.  If the user can NOT get config maps or secrets and the kve instance is read only, we should the faux input without the link.

screen shot 2017-02-15 at 9 11 36 am

benjaminapetersen: Should we doc this somewhere?  Comments @ the top of the directive for future reference?

app/scripts/controllers/create/createFromImage.js
benjaminapetersen: agree... i don't think that has meaning anymore.

app/scripts/services/images.js
benjaminapetersen: This is just a switch for backwards compatibility.
- if it gets a hash, this is the "old way" that needs to transform into `[{name:'',value:''}, etc]`
-  If it gets an array, return the array (input arg).  We've already  

Off the top of my head I'm not sure why it received a hash of `{'<name>': '<value>'}` to begin with.

app/scripts/controllers/deployment.js
```diff
@@ -112,6 +124,7 @@ angular.module('openshiftConsole')
             $scope.loaded = true;
             $scope.deployment = deployment;
             updateHPAWarnings();
+            EnvironmentService.toggleReadonlyForContainerEnvsValueFrom(_.get($scope.deployment, 'spec.template.spec.containers'), envSecretAndConfigMapAreReadonly);
```
benjaminapetersen: Other than the gorp, looking @ the toggle again per our discussion to see if we can clean up.

benjaminapetersen: We considered moving the `DataService.list("secrets")` and `DataService.list("configmaps")` into the directive & eliminating the controller noise... might be a good way to fix this.

jwforres:
No that is risky, you'll end up re-listing in some cases if the directive
gets re-created for any reason (like an ng-if changing state)

jwforres:
Yeah I definitely disagree with this approach. You are lumping together
multiple things onto an attribute that no longer means what the variable is
called.  readonly is readonly, readonly doesn't mean don't let me have a
link to something.

Why is the canI check not right here in the view, you have everything you
need there already, you know exactly which type of resource is being
referenced.  The whole point of us pulling in the KVE was to stop doing
hacky things to work around limitations of making stuff generic.

benjaminapetersen:
- [x] Everything is being cleaned, either via `compact()` or directly.

benjaminapetersen:
I believe we did it to reduce template logic, we began in the view file but ended up branching/copy paste more than we liked.  Doing the check up front simplified the branching.

benjaminapetersen:
@jwforres `isReadonly` still just controls wether or not you get the select pickers.  In the readonly state, we use `canI` to render either plain text, or text with a link, based on if you have perms to work with `secrets`/`configmaps` ([outlined in this comment](https://github.com/openshift/origin-web-console/pull/1090#discussion_r101284780)).

app/views/directives/key-value-editor.html
```diff
@@ -0,0 +1,238 @@
...
+    <div
+      class="key-value-editor-entry form-group"
+      ng-if="(!cannotAdd)">
+      <a
+        href=""
+        class="add-row-link"
+        role="button"
+        aria-label="Add row"
+        ng-click="onAddRow()">{{ addRowLink }}</a>
+      <span
```
rhamilto: Moar simpler:  a single span wrapping both the pipe and the link with ng-if="valueFromSelectorOptions.length

app/views/directives/key-value-editor.html
```diff
@@ -0,0 +1,238 @@
...
+                  <a
+                    ng-if="!(entry.isReadonlyValue)"
+                    ng-href="{{entry.apiObj | navigateResourceURL}}">
+                    {{entry.valueFrom.configMapKeyRef.name}}
+                  </a>
+                </span>
+                <span ng-switch-when="secretKeyRef">
+                  Set to the key {{entry.valueFrom.secretKeyRef.key}} in secret
+                  <span
+                    ng-if="entry.isReadonlyValue">
```
rhamilto: canI is much clearer to anyone reading the template.  Let's use it instead.

benjaminapetersen:
@jwforres gonna jot down the flow here quick as well, with deployments as an example.  It will explain the odd `envSecretAndConfigMapAreReadonly` variable.  I'm open to improvements.
- the `DataService.get('deployments')` is called initially & sets `envSecretAndConfigMapAreReadonly` to `true` since we don't have `secrets`/`configmaps` yet.  Nothing should be editable.
- then the `DataService.watchObject('deployments')` is called, along with `DataService.list('configmaps')` and `DataService.list('secrets')`.  We don't know which will return first, which is why we have the quirky logic.
- If the `DataService.watchObject('deployments')` happens to fire, `envSecretAndConfigMapAreReadonly` does not flip, there is no new info for us to toggle state (we already had this).  This watch, if fired, needs to regenerate the state as it is, not knowing if it has `secrets`/`configmaps` or not.
- But, if either `DataService.list("configmaps")` or `DataService.list("secrets")` return first, we do flip the bit, and we can now show the pickers.  
- Since the watch may update many times, we need it to just trust the state of `envSecretAndConfigMapAreReadonly` wherever it has landed based on the other 3 calls.

Hopefully that makes sense... and moves along some discussion about simplifying if we can make it happen!

app/scripts/controllers/create/createFromImage.js
rhamilto: Comment removed.

app/views/directives/key-value-editor.html
rhamilto: > Moar simpler: a single span wrapping both the pipe and the link with ng-if="valueFromSelectorOptions.length"

Done.

app/views/directives/key-value-editor.html
rhamilto: > canI is much clearer to anyone reading the template. Let's use it instead.

Done

app/views/directives/key-value-editor.html
```diff
@@ -0,0 +1,238 @@
...
+
+        <div ng-if="entry.valueFrom">
+          <div ng-if="isReadonlyAny || entry.isReadonlyValue" class="faux-input-group">
+            <span
+              class="faux-form-control-addon {{entry.valueIcon}}"
+              aria-hidden="true"
+              data-toggle="tooltip"
+              data-placement="top"
+              data-original-title="{{entry.valueIconTooltip || valueIconTooltip}}"
+              title="{{entry.valueIconTooltip || valueIconTooltip}}"></span>
```
jwforres: while we are fixing things... can you remove the extra title attribute here, it is causing https://bugzilla.redhat.com/show_bug.cgi?id=1395079

jwforres:
[merge]

jwforres:
[merge]

jwforres:
take three [merge]

jwforres:
[merge]

jwforres:
[merge]

Issue #1083, opened 2017-01-05
Missing margin for "Show Annotations" link


spadgett:
The table is using the `:last-child` style even though there is content below it on the config map page. This could be a problem on other pages.

![screen shot 2017-01-05 at 9 31 41 am](https://cloud.githubusercontent.com/assets/1167259/21684009/e0d15fd6-d329-11e6-8501-37f22ffc93d2.png)

cc @rhamilto

spadgett:
![screen_shot_2017-01-05_at_9_33_41_am](https://cloud.githubusercontent.com/assets/1167259/21684053/1909ba74-d32a-11e6-8733-523608202570.png)

PR #1093, opened 2017-01-09
Updates to control margin of annotations block

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/1083
screen shot 2017-01-09 at 2 30 33 pm
screen shot 2017-01-09 at 2 42 06 pm
screen shot 2017-01-09 at 2 43 46 pm
screen shot 2017-01-09 at 2 43 36 pm

### DIFF
--- a/app/styles/_substructure.less
+++ b/app/styles/_substructure.less
@@ -285,10 +285,7 @@ html,body {
 // middle-content area in order to achieve consistent bottom spacing across all pages
 .middle-content {
   annotations {
-    &:last-child,
-    p {
-      margin-bottom: 0;
-    }
+    display: block;
   }
   pods-table,
   traffic-table {
--- a/app/views/browse/config-map.html
+++ b/app/views/browse/config-map.html
@@ -70,7 +70,7 @@
                   </tbody>
                 </table>
               </div>
-              <annotations annotations="configMap.metadata.annotations"></annotations>
+              <annotations annotations="configMap.metadata.annotations" class=" mar-top-xl"></annotations>
             </div><!-- /col-* -->
           </div>
         </div>
--- a/app/views/directives/annotations.html
+++ b/app/views/directives/annotations.html
@@ -1,10 +1,10 @@
-<p>
+<p ng-class="{'mar-bottom-md': expandAnnotations, 'mar-none':!expandAnnotations}">
   <a href="" ng-click="toggleAnnotations()" ng-if="!expandAnnotations">Show Annotations</a>
   <a href="" ng-click="toggleAnnotations()" ng-if="expandAnnotations">Hide Annotations</a>
 </p>
 <div ng-if="expandAnnotations">
   <div ng-if="annotations" class="table-responsive">
-    <table class="table table-bordered table-bordered-columns key-value-table table-top-margin">
+    <table class="table table-bordered table-bordered-columns key-value-table">
       <tbody>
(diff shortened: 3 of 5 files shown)


sg00dwin:
@spadgett PTAL
I'm now setting the appropriate margin on the annotations content.

spadgett:
@sg00dwin I'd argue the problem is that the `table:last-child` rules are being applied to tables when they shouldn't be. We should only have 0 margin for a table when it's the last thing on the page. If we have to special case margins anyway, then I'd almost rather remove the `:last-child` rules.

cc @rhamilto

sg00dwin:
@spadgett the bottom margin of the table needs to be 0 so that extra space isn't added to the the global bottom spacing set on the `container-fluid`. Are you suggesting, instead of removing it with [this rule](https://github.com/sg00dwin/origin-web-console/blob/4f87fadd75e94b9e4e621cf4ef3a8c56abd7b3fe/app/styles/_substructure.less#L295) do you want `class="mar-none"` add to these tables.

spadgett:
@sg00dwin I'm suggesting we try to fix the rule so it only applies when the table is the last thing on the page, if possible. If not, we should discuss alternatives. I don't think we should keep a rule that applies to any table that is the last child of any element. It will cause more problems than it helps.

sg00dwin:
@spadgett To my knowledge their is no way, using just css, to target only the last table on any page, since their nested markup structures vary from page to page. @rhamilto let me know if I'm wrong.

On almost all pages that have table(s) within `.middle-content` the table is the last item on the page, which is the reason I used `.middle-content .table:last-child` to remove it's bottom margin and then selectively add margin on the few cases where bottom margin is needed. specifically the `pods-table` & `traffic-table` 

This seemed more preferable than going through each page and manually adding `.mar-bottom-none` to the last table. But then there could be cases where we don't know what the last table will be.

Which way would you prefer?

spadgett:
My opinion: I'd rather remove the `table:last-child` rule, remove the bottom margin from `.middle-content`, and use the normal bottom margin of the last element on the page. If the last element doesn't have a bottom margin, we'd just add one.

I think a rule that sometimes applies to tables and sometimes doesn't is going to lead to other unexpected bugs like #1083 as we change the markup. Wrapping a table in a `<div class="table-responsive">` or an `<div ng-if="...">` element seems really common and will unexpectedly change the margin. Stuff like that is really hard to catch on code reviews.

Open to suggestions, though. @rhamilto thoughts?

rhamilto:
> I think a rule that sometimes applies to tables and sometimes doesn't is going to lead to other unexpected bugs like #1083 as we change the markup.

Agreed since our templates seem to vary so much from page to page.  When I initially proposed this rule, I assumed we could make it specific enough to avoid these issues by using a [child selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_selectors), but that didn't work in practice because of the markup inconsistency.  :(

sg00dwin:
Ok, so I will put together a pr remove the [global bottom margin](https://github.com/openshift/origin-web-console/blob/master/app/styles/_substructure.less#L89) and [these rules](https://github.com/openshift/origin-web-console/blob/master/app/styles/_substructure.less#L284-L317) that zeroed out the bottom margin on various elements positioned at the end of pages. 

Then go through each page and ensure there is a bottom margin on all pages and add one if needed. Noting that the margin could vary based on the type of ui element that is positioned last.

PR #1096, opened 2017-01-10
Switching to have the last ui element within .middle-content set the …

sg00dwin:
…margin-bottom. And add margin to elements that appear last but have none.

fixes https://github.com/openshift/origin-web-console/issues/1083

replaces https://github.com/openshift/origin-web-console/pull/1093

### DIFF
--- a/app/styles/_log.less
+++ b/app/styles/_log.less
@@ -65,6 +65,10 @@
   font-family: @font-family-monospace;
   padding: 0;
   position: relative;
+  .mar-bottom-xl();
+  .chromeless & {
+    margin-bottom: 0;
+  }
   pre, pre code {
     background-color: transparent;
     border: 0;
--- a/app/views/browse/deployment-config.html
+++ b/app/views/browse/deployment-config.html
@@ -338,7 +338,7 @@
                   </uib-tab>
                   <uib-tab heading="Environment" active="selectedTab.environment" ng-if="deploymentConfig">
                     <uib-tab-heading>Environment</uib-tab-heading>
-                    <ng-form name="forms.dcEnvVars">
+                    <ng-form name="forms.dcEnvVars" class="mar-bottom-xl block">
                       <div ng-repeat="container in updatedDeploymentConfig.spec.template.spec.containers">
                         <h3>Container {{container.name}} Environment Variables</h3>
                         <key-value-editor
--- a/app/views/directives/annotations.html
+++ b/app/views/directives/annotations.html
@@ -1,10 +1,10 @@
-<p>
+<p ng-class="{'mar-bottom-xl': !expandAnnotations}">
   <a href="" ng-click="toggleAnnotations()" ng-if="!expandAnnotations">Show Annotations</a>
   <a href="" ng-click="toggleAnnotations()" ng-if="expandAnnotations">Hide Annotations</a>
 </p>
 <div ng-if="expandAnnotations">
   <div ng-if="annotations" class="table-responsive">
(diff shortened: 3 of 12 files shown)


app/views/browse/deployment-config.html
```diff
@@ -338,7 +338,7 @@ <h3 class="hidden visible-lg visible-xl">Details</h3>
                   </uib-tab>
                   <uib-tab heading="Environment" active="selectedTab.environment" ng-if="deploymentConfig">
                     <uib-tab-heading>Environment</uib-tab-heading>
-                    <ng-form name="forms.dcEnvVars">
+                    <ng-form name="forms.dcEnvVars" class="ng-margin-bottom">
```
spadgett: Is `ng-margin-bottom` a typo?

app/views/directives/annotations.html
```diff
@@ -20,7 +20,7 @@
       </tbody>
     </table>
   </div>
-  <p ng-if="!annotations">
+  <p ng-if="!annotations" class="mar-bottom-xl">
```
spadgett: Is the normal margin for `<p>` tags not enough?

app/views/directives/annotations.html
```diff
@@ -1,10 +1,10 @@
-<p>
+<p ng-class="{'mar-bottom-xl': !expandAnnotations}">
```
spadgett: Same question here: Is the default `p` margin not OK?

sg00dwin: `p` have 10.5px bottom margin, but in this case, since almost all annotations appear last, I give these 20px to match the bottom margin of what most other elements have `.table, dl, forms`.

spadgett: Can you post a screenshot showing the difference?

sg00dwin: screen shot 2017-01-11 at 11 13 07 am
screen shot 2017-01-11 at 11 13 33 am

spadgett: I don't mind the default margin, and it would save us some special case classes, but I'm okay with whatever you and @rhamilto think

app/styles/_log.less
```diff
@@ -65,6 +65,10 @@
   font-family: @font-family-monospace;
   padding: 0;
   position: relative;
+  .mar-bottom-xl();
```
spadgett: Have you checked the bottom margin on the monitoring page?

sg00dwin: yes the `list-view-pf` for each section has a bottom margin by default

spadgett: Sorry, asking if there is too much margin below on the logs on that page.

sg00dwin: It looks fine. The `log-view` bottom margin and `metrics mar-top-xl` don't combine to make too much space. 

screen shot 2017-01-11 at 5 35 56 pm

sg00dwin:
@spadgett anything else needed for this?

spadgett:
> the bottom margin needs to beneath the save button, but<key-value editor> doesn't wrap it so I placed the margin on the first element that did ng-form. Which in this case was a typo should have been mar-bottom-xl block not ng-margin-bottom.

I was thinking we should add a `button-group` class with margins. Then we have consistent margins for the buttons on all of our forms. But that touches a lot of other places, so probably not for this PR.

spadgett:
[merge]

spadgett:

sg00dwin:
ok, hopefully this one is correct

spadgett:
[merge]

Issue #1091, opened 2017-01-09
Donut scaling buttons can be clipped on deployment page


spadgett:
At certain screen widths, the scaling buttons for a deployment can be clipped. This was at ~785px wide.

![screen_shot_2017-01-09_at_9_24_59_am](https://cloud.githubusercontent.com/assets/1167259/21769677/edbc24a2-d64d-11e6-8ae7-39274fffc730.png)

PR #1105, opened 2017-01-12
Removal of legacy css rule causing clipping of pod donut arrows.

sg00dwin:
There is a min-width,  that is not needed, on `.deployment-donut` inherited from some [legacy code](url) that was originally associated with `.overview-pods`. Removing it tests fine on Chrome, FF, IE, Safari

Fixes https://github.com/openshift/origin-web-console/issues/1091

200donutissue

200donutissueie

### DIFF
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -309,7 +309,6 @@ code.command {
 
 .deployment-donut {
   justify-content: center;
-  min-width: 200px;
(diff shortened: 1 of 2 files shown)


spadgett:
Needs dist

sg00dwin:
Now with dist :)

spadgett:
[merge]

PR #1107, opened 2017-01-12
Fix bug displaying read-only build config env vars

### LINKED ISSUE #1106: Environment variables not visible for build configs by viewers
If I log in as a user with role `view`, I can't see any of the env vars for a build config.

spadgett:
The read-only view was incorrectly inside a `canI : 'update'` block.

Fixes #1106


jwforres:
[merge]

PR #1108, opened 2017-01-12
Support volume mount subpaths and read-only flag

spadgett:
Show subpath and read-only values for volume mounts. Allow users to set these fields on the add storage page.

### DIFF
--- a/app/views/attach-pvc.html
+++ b/app/views/attach-pvc.html
@@ -113,6 +113,32 @@
                       </div>
 
                       <div class="form-group">
+                        <label for="sub-path">Subpath</label>
+                        <input
+                            id="sub-path"
+                            class="form-control"
+                            type="text"
+                            name="subPath"
+                            ng-model="attach.subPath"
+                            placeholder="example: application/resources"
+                            ng-pattern="RELATIVE_PATH_PATTERN"
+                            autocorrect="off"
+                            autocapitalize="off"
+                            spellcheck="false"
+                            aria-describedby="sub-path-help">
+                        <div id="sub-path-help" class="help-block">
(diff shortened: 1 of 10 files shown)


app/views/attach-pvc.html
```diff
@@ -154,6 +180,18 @@ <h2 class="text-center">No persistent volume claims.</h2>
...
                       </div>
 
+                      <div class="form-group">
+                        <div class="checkbox">
+                          <label>
+                            <input type="checkbox" ng-model="attach.readOnly" aria-describedby="read-only-help">
+                            Read only
+                          </label>
+                          <div id="read-only-help" class="help-block">
+                            Mount the volume read-only.
```
jwforres: Mount the volume as read-only.

app/views/attach-pvc.html
```diff
@@ -113,6 +113,32 @@ <h2 class="text-center">No persistent volume claims.</h2>
...
+                            ng-model="attach.subPath"
+                            placeholder="example: application/resources"
+                            ng-pattern="RELATIVE_PATH_PATTERN"
+                            autocorrect="off"
+                            autocapitalize="off"
+                            spellcheck="false"
+                            aria-describedby="sub-path-help">
+                        <div id="sub-path-help" class="help-block">
+                          Optional path within the volume from which the container's volume should
+                          be mounted. Defaults to the volume's root.
```
jwforres: Maybe:
Optional path within the volume from which it will be mounted into the container. Defaults to the volume's root.

jwforres:
also this needs a rebase

spadgett:
@jwforres thanks, updated

jwforres:
[merge]

PR #1109, opened 2017-01-12
Display custom metrics for pods if they are available.

mwringe:
The Hawkular OpenShift Agent is capable of gathering a few custom metrics from pods which are deployed.

This PR will allow these pods to be displayed under the pod's metric page

### DIFF
--- a/app/scripts/directives/podMetrics.js
+++ b/app/scripts/directives/podMetrics.js
@@ -95,6 +95,41 @@ angular.module('openshiftConsole')
           });
         }
 
+        if (!window.OPENSHIFT_CONSTANTS.DISABLE_CUSTOM_METRICS) {
+          // Load any custom metrics onto the page
+          MetricsService.getCustomMetrics(scope.pod).then(
+            function(response) {
+              angular.forEach(response, function(metric) {
+
+                // set the label to the description if specified
+                var label = metric.description || metric.name;
+              
+                // get the unit value if specified
+                var unit =  metric.unit || "";
+
+                scope.metrics.push({
+                  label: label,
+                  units: unit,
+                  chartPrefix: "custom-" + _.uniqueId('custom-metric-'),
+                  chartType: "spline",
+
+                  datasets: [
+                    {
+                      id: "custom/" + metric.name,
+                      label: label,
+                      type: metric.type,
+                      data: []
+                    },
+                  ]
+                });
+
+              });
+              // update the page with the new charts.
+              update();
+            }
+          );
+        }
+
         // Set to true when any data has been loaded (or failed to load).
         scope.loaded = false;
         scope.noData = true;
@@ -260,6 +295,7 @@ angular.module('openshiftConsole')
           var lastPoint;
           var config = {
             metric: dataset.id,
+            type: dataset.type,
             bucketDuration: getBucketDuration()
           };
 
--- a/app/scripts/services/metrics.js
+++ b/app/scripts/services/metrics.js
@@ -3,6 +3,7 @@
 angular.module("openshiftConsole")
   .factory("MetricsService", function($filter, $http, $q, $rootScope, APIDiscovery) {
     var POD_GAUGE_TEMPLATE = "/gauges/{containerName}%2F{podUID}%2F{metric}/data";
+    var POD_COUNTER_TEMPLATE = "/counters/{containerName}%2F{podUID}%2F{metric}/data";
 
     var metricsURL;
     function getMetricsURL() {
@@ -50,7 +51,13 @@ angular.module("openshiftConsole")
 
     function getRequestURL(config) {
       return getMetricsURL().then(function(metricsURL) {
-        var template = metricsURL + POD_GAUGE_TEMPLATE;
+        var template;
+        // if no type is specified, it is assumed the metric is a gauge
+        if (config.type === "counter") {
+          template = metricsURL + POD_COUNTER_TEMPLATE;
+        } else {
+          template = metricsURL + POD_GAUGE_TEMPLATE;
+        }
         return URI.expand(template, {
           podUID: config.pod.metadata.uid,
           containerName: config.containerName,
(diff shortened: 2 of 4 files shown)


app/scripts/directives/podMetrics.js
```diff
@@ -95,6 +95,38 @@ angular.module('openshiftConsole')
           });
         }
 
+        // Load any custom metrics onto the page
+        MetricsService.getCustomMetrics(scope.pod).then(
+          function(response) {
+            angular.forEach(response, function(metric) {
+
+              // set the label to the description if specified
+              var label = (metric.description !== undefined) ? metric.description : metric.name;
```
spadgett:
```js
var label = metric.description || metric.name;
```

app/scripts/directives/podMetrics.js
```diff
@@ -95,6 +95,38 @@ angular.module('openshiftConsole')
...
+              
+              // get the unit value if specified
+              var unit = (metric.unit !== undefined) ? metric.unit : "unknown";
```
spadgett:
```js
var unit = metric.unit || "unknown";
```

But maybe the empty string is better if no unit set?

```js
var unit = metric.unit || "";
```

app/scripts/services/metrics.js
```diff
@@ -169,6 +169,46 @@ angular.module("openshiftConsole")
...
+          var metrics = [];
+          angular.forEach(response.data, function(value) {
+            var metric = {
+              name: value.tags.metric_name,
+              unit: value.tags.units,
+              description: value.tags.description,
+            };
+            metrics.push(metric);
+          });
+           return metrics;
```
spadgett: indentation

app/scripts/services/metrics.js
```diff
@@ -169,6 +169,46 @@ angular.module("openshiftConsole")
...
+        }).then(function(response) {
+          var metrics = [];
+          angular.forEach(response.data, function(value) {
+            var metric = {
+              name: value.tags.metric_name,
+              unit: value.tags.units,
+              description: value.tags.description,
+            };
+            metrics.push(metric);
+          });
```
spadgett: Suggest using `_.map`. (Untested, but I think this will work.)

```js
return _.map(response.data, function(value) {
  return {
    name: value.tags.metric_name,
    unit: value.tags.units,
    description: value.tags.description
  };
});
```

app/scripts/services/metrics.js
```diff
@@ -169,6 +169,46 @@ angular.module("openshiftConsole")
...
+            'Hawkular-Tenant': namespace
+          },
+          params: params
+        }).then(function(response) {
+          var metrics = [];
+          angular.forEach(response.data, function(value) {
+            var metric = {
+              name: value.tags.metric_name,
+              unit: value.tags.units,
+              description: value.tags.description,
```
spadgett: Extra trailing comma

spadgett:
cc @jwforres @derekwaynecarr @DirectXMan12

mwringe:
@spadgett Ok, I made the requested changes to the code.

Custom metrics is something we want for 1.5, we are hoping to have something in the console to display these metrics. If we can not get it the console at this point in time, we will look to create a custom console extension or rely on something like grafana to display them.

spadgett:
@mwringe can you add a flag to disable custom metrics to constants.js? something like

https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js#L66

mwringe:
@spadgett do we just want a flag to be added and the rest kept the same? Or a flag to be added and the drop down to still be used instead?

spadgett:
my opinion: flag added and the rest kept the same.

@jwforres let me know if you disagree

jwforres:
yeah based on @spadgett 's assessment about the risk of introducing the
dropdown at this point in the release I agree

mwringe:
Update to include a flag to enable/disable custom metrics.

I also fixed an issue I found where we were not properly fetching counter based metrics and fixed a potential issue with encoding the metric name for the chart prefix.

app/scripts/directives/podMetrics.js
```diff
@@ -95,6 +95,45 @@ angular.module('openshiftConsole')
...
+
+                // set the label to the description if specified
+                var label = metric.description || metric.name;
+              
+                // get the unit value if specified
+                var unit =  metric.unit || "";
+
+                // note: the prefix needs to follow selector rules
+                // converting to base64 and cleaning up the special /,+,= characters
+                var prefix = btoa(metric.name).replace(/\+/g,"_62_").replace(/\//g,"_63_").replace(/\=/g,"");
```
spadgett: Suggest this instead for the prefix:

```js
var prefix = _.uniqueId('custom-metric-');
```

app/scripts/services/metrics.js
```diff
@@ -50,7 +51,12 @@ angular.module("openshiftConsole")
 
     function getRequestURL(config) {
       return getMetricsURL().then(function(metricsURL) {
-        var template = metricsURL + POD_GAUGE_TEMPLATE;
+        var template;
+        if (config.type !== undefined && config.type === "counter") {
```
spadgett: Maybe just

```js
if (config.type === 'counter') { ... }
```

although probably worth a comment saying gauge is assumed when type is not set

app/scripts/services/metrics.js
```diff
@@ -169,6 +175,44 @@ angular.module("openshiftConsole")
...
+      return getMetricsURL().then(function(metricsURL) {
+
+        if (!metricsURL) {
+          return null;
+        }
+
+        var url = metricsURL + "/metrics";
+
+        var params = {
+          tags: "custom_metric:true,pod_name:" + podName
```
spadgett: Missed this on first review, but should we use pod UID instead in case a pod with the same name is created twice? (For pods that don't have generated names.)

app/scripts/directives/podMetrics.js
mwringe: awesome, that will work

app/scripts/services/metrics.js
mwringe: ok

app/scripts/services/metrics.js
mwringe: oh, good catch. Thanks. I will fix that.

mwringe:
Ok, updated with the requested changes.

spadgett:
[merge]

PR #1111, opened 2017-01-13
Don't show quota warnings when max is 0

### LINKED ISSUE #1110: Ignore quota warning when the max=0?
If we our quotas include things like `services.loadbalancers=0` or `services.nodeports=0` it'd be helpful to have a way to ignore these from the quota warning in the overview page

spadgett:
Don't show any quota warnings when the max is 0 since they'll always be displayed and aren't a problem.

Fixes #1110

### DIFF
--- a/app/scripts/services/quota.js
+++ b/app/scripts/services/quota.js
@@ -227,47 +227,51 @@ angular.module("openshiftConsole")
     };
 
     var getLatestQuotaAlerts = function(resources, context) {
-      var result = $q.defer();
-      var quotaAlerts = [];
-      var quotas, clusterQuotas;
-      var remaining = 2;
-      function _checkDone() {
-        if (remaining === 0) {
-          quotaAlerts = getQuotaAlerts(resources, quotas, clusterQuotas);
-          result.resolve({quotaAlerts: quotaAlerts});
-        }
-      }
+      var quotas, clusterQuotas, promises = [];
 
       // double check using the latest quotas
-      DataService.list("resourcequotas", context, function(quotaData) {
+      promises.push(DataService.list("resourcequotas", context).then(function(quotaData) {
         quotas = quotaData.by("metadata.name");
         Logger.log("quotas", quotas);
-        remaining--;
-        _checkDone();
-      });
+      }));
 
-      DataService.list("appliedclusterresourcequotas", context, function(clusterQuotaData) {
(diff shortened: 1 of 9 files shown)


app/scripts/services/quota.js
```diff
@@ -227,36 +227,25 @@ angular.module("openshiftConsole")
...
-      return result.promise;
+      return $q.all(promises).then(function() {
+        var quotaAlerts = getQuotaAlerts(resources, quotas, clusterQuotas);
+        return {
+          quotaAlerts: quotaAlerts
+        };
+      });
     };
 
     var isAnyQuotaExceeded = function(quotas, clusterQuotas) {
```
jwforres: i feel like it needs a comment added to explicitly say that it will not trigger on a quota of zero

app/views/quota.html
```diff
@@ -84,11 +84,11 @@ <h3 class="text-center">Memory <small>Limit</small></h3>
                         <tr ng-repeat="(resourceType, specMax) in quota.spec.quota.hard"
                             ng-if="resourceType !== 'resourcequotas'"
                             ng-class="{
-                              warning: (quota.status.total.used[resourceType] | usageValue) >= (quota.status.total.hard[resourceType] | usageValue)
+                              warning: isAtLimit(quota, resourceType)
```
jwforres: i wonder if there is another visual way to treat rows where the max is set to zero, gray/disabled?  it feels like something special to me because you are being completely denied the use of this resource

spadgett: Here's what it might look like. Opinion?

openshift_web_console

We need to humanize some additional quota resources.

spadgett:
@jwforres updated

openshift_web_console

app/views/quota.html
```diff
@@ -23,8 +23,8 @@
                 </div>
 
                 <div ng-repeat="quota in clusterQuotas | orderBy: 'metadata.name'" class="gutter-bottom">
-                  <h2 ng-if="(clusterQuotas | hashSize) > 1">{{quota.metadata.name}}</h2>
                   <div ng-if="$first" class="help-block">Limits resource usage across a set of projects.</div>
+                  <h2 ng-if="(clusterQuotas | hashSize) > 1">{{quota.metadata.name}}</h2>
```
jwforres: was this moved intentionally?

spadgett: No, will fix.

spadgett:
@jwforres updated

jwforres:
[merge]

PR #1113, opened 2017-01-13
Refresh membership page when removing 'admin' role from current user

### LINKED ISSUE #1103: Page should redirect after removing admin role from self
If you remove the admin role from yourself in the membership page the page should redirect/refresh otherwise you continue to click around on other membership management functions and wonder why nothing is happening

benjaminapetersen:
Fixes #1103 

- Adds a forceRefresh flag to `AuthorizationService.getProjectRules` to allow busting the rules cache
- Ensures the edit mode is tested after every update
- Falls back to manually updating a roleBindings if the list on roleBindings fails (which it will if the user removes `admin` from self)

@spadgett


spadgett:
[merge]

Issue #1118, opened 2017-01-16
Editing standalone replication controller env vars via KVE results in error


rhamilto:
![screen shot 2017-01-16 at 10 25 48 am](https://cloud.githubusercontent.com/assets/895728/21988748/5e544e00-dbd6-11e6-8cc1-85b49f41dac6.PNG)

To reproduce: 

1.  Use https://github.com/openshift/origin-web-console/blob/master/test/testdata/overview-edge-cases.yaml to create a standalone replication controller "frontend".  
1.  Reorder the env vars using KVE.

@benjaminapetersen, looks like another KVE bug with secrets and configmaps.

benjaminapetersen:
Looking at this now.

jwforres:
fixed by #1120

PR #1120, opened 2017-01-16
Fix console error while editing envs for standalone rc

benjaminapetersen:
Fixes issue #1118 

@rhamilto @jwforres @spadgett for review!  I think all it needs is a guard in case the `$scope.deploymentConfigDeploymentsInProgress` does not exist.

### DIFF
--- a/app/scripts/controllers/replicaSet.js
+++ b/app/scripts/controllers/replicaSet.js
@@ -436,6 +436,7 @@ angular.module('openshiftConsole')
             deploymentConfigName = annotation(replicaSet, 'deploymentConfig');
             rsName = replicaSet.metadata.name;
           }
+          $scope.deploymentConfigDeploymentsInProgress = $scope.deploymentConfigDeploymentsInProgress || {};
           if (!action) {
             // Loading of the page that will create deploymentConfigDeploymentsInProgress structure, which will associate running deployment to his deploymentConfig.
             $scope.deploymentConfigDeploymentsInProgress = DeploymentsService.associateRunningDeploymentToDeploymentConfig($scope.deploymentsByDeploymentConfig);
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/replicaSet.js
```diff
@@ -445,7 +446,9 @@ angular.module('openshiftConsole')
             $scope.deploymentConfigDeploymentsInProgress[deploymentConfigName][rsName] = replicaSet;
           } else if (action === 'MODIFIED') {
             // After the deployment ends remove him from the deploymentConfigDeploymentsInProgress structure.
-            delete $scope.deploymentConfigDeploymentsInProgress[deploymentConfigName][rsName];
+            if($scope.deploymentConfigDeploymentsInProgress && $scope.deploymentConfigDeploymentsInProgress[deploymentConfigName]) {
```
jwforres: pretty sure `$scope.deploymentConfigDeploymentsInProgress` will always be defined, its going to get defined during the initial data returned by the watch.  were you seeing a case where it wasn't? because that would break line 445 as well.

My guess is the problem is just `$scope.deploymentConfigDeploymentsInProgress[deploymentConfigName]` not being defined

benjaminapetersen: Yeah I was seeing the whole thing `undefined`.  Shall I update `$scope.deploymentConfigDeploymentsInProgress` to create an empty object outside the if statements?

jwforres:
yeah probably, its either that or also patch the other place that could
break

benjaminapetersen:
Updated @jwforres

app/scripts/controllers/replicaSet.js
jwforres: cant you remove the first part of this if stmt now?

benjaminapetersen: yup will do

jwforres:
[merge]

PR #1124, opened 2017-01-16
Don't disable overview start pipeline button when pipeline running

spadgett:
We made this change to the build config page when `runPolicy` was added, but not here.


jwforres:
[merge]

PR #1125, opened 2017-01-17
Bug 1413842 - Use dot in ng-model path for compute resource units select

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1413842

See https://github.com/angular-ui/ui-select/wiki/FAQs#ng-model-not-working-with-a-simple-variable-on-scope

@jwforres This is not a problem in 1.4. I swept the code and didn't see this problem anywhere else we use ui-select.


jwforres:
[merge]

PR #1129, opened 2017-01-17
Let users pause rollouts when editing deployment configs

spadgett:
Add a checkbox to pause rollouts when making changes to a deployment config (or deployment). This lets you make edits without immediately deploying.

openshift_web_console

### DIFF
--- a/app/views/edit/deployment-config.html
+++ b/app/views/edit/deployment-config.html
@@ -356,6 +356,11 @@
                         </div>
                       </div>
 
+                      <pause-rollouts-checkbox
+                          deployment="updatedDeploymentConfig"
+                          always-visible="true">
+                      </pause-rollouts-checkbox>
+
                       <div class="buttons gutter-top-bottom">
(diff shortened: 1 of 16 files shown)


app/views/edit/deployment-config.html
```diff
@@ -356,6 +356,11 @@ <h3 class="with-divider">Environment Variables</h3>
                         </div>
                       </div>
 
+                      <pause-rollouts-checkbox
+                          deployment="updatedDeploymentConfig"
+                          always-visible="true">
```
jwforres: was this because you can change the triggers within this page?  why not show/hide this depending on whether the dc trigger is enabled?

spadgett: I thought you might want to pause/resume rollouts in this editor for other reasons (e.g. to keep image change triggers from firing)

jwforres: ok makes sense

jwforres:
this needs to be rebased, otherwise LGTM

spadgett:
Rebased, no other changes

[merge]

PR #1130, opened 2017-01-17
Making placement of actions menu consistent across pages

### LINKED ISSUE #1112: Actions menu misplaced on some pages at mobile
When the object name is long, the actions menu wraps to the next line. It might be difficult to notice. Better to keep it to the right of the name (and maybe truncate the name if necessary?).

openshift_web_console

rhamilto:
And adding word-wrap() to the h1 so that long, unbroken strings don’t
cause the layout to break.

Fixes #1112

screen shot 2017-01-17 at 4 07 09 pm
screen shot 2017-01-17 at 4 07 51 pm

@jwforres or @spadgett, PTAL

### DIFF
--- a/app/styles/_typography.less
+++ b/app/styles/_typography.less
@@ -35,13 +35,18 @@
   }
 }
 
-h1 small.meta,
-.build-config-summary .meta,
-.deployment-config-summary .meta {
-  font-size: 12px;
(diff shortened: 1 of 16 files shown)


spadgett:
[merge]

Issue #1128, opened 2017-01-17
.table-mobile td without value results in "collapsed" row


rhamilto:
<img width="471" alt="screen shot 2017-01-17 at 3 08 58 pm" src="https://cloud.githubusercontent.com/assets/895728/22037927/286448da-dcc7-11e6-93ab-6827e01aa56b.PNG">

PR #1131, opened 2017-01-17
Prevent mobile table td collapsing issue

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/1128

screen shot 2017-01-17 at 4 44 41 pm

### DIFF
--- a/app/styles/_tables.less
+++ b/app/styles/_tables.less
@@ -8,8 +8,8 @@
     > thead >tr > th, > tbody > tr > td {
       border-left: 0;
       border-right: 0;
-      padding-bottom: 8px;
-      padding-top: 8px;
+      padding-bottom: @table-cell-padding-bottom;
+      padding-top: @table-cell-padding-top;
       vertical-align: middle;
     }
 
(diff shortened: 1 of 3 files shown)


spadgett:
In other places, we used a non-breaking space to make sure the table cell had height. For instance,

https://github.com/spadgett/origin-web-console/blob/695c74dc0fac848444c79c555fb689c4d8431973/app/views/browse/routes.html#L82

If we have a different solution, we should probably change those to be consistent.

app/styles/_tables.less
```diff
@@ -123,6 +123,7 @@
            /* Behave like a row */
           border: none;
           border-bottom: 1px solid lighten(@table-border-color, 5%);
+          min-height: 38px; // prevent empty td collapsing
```
spadgett: This seems like a magic number. Is there any existing variable that makes sense for min-height?

spadgett:
cc @rhamilto

sg00dwin:
Updated PR, switched to using non-breaking space to maintain td height

spadgett:
I'm good with the CSS fix. Just saying we should remove the `nbsp` in other places if it's no longer needed to clean up our markup.

sg00dwin:
Updated pr to use min-height calculated. 

Note: related issue pertaining to switching out other `&nbsp;` used to give height, will be corrected in other pr.

spadgett:
[merge]

PR #1133, opened 2017-01-18
Create storage form improvements

spadgett:
* Correctly show error messages for invalid capacity values
* Make GiB the default unit
* Remove PiB unit since it's unlikely to be needed
* Add decimal units in addition to current binary units (to be consistent with set limits page)
* Add units help modal
* Group units in the ui-select (for both create storage and set limits)

@erinboyd @zherman0 fyi


spadgett:
@jwforres I plan to add the limit range check in a follow-on.

jwforres:
awesome thanks for fixing these [merge]

spadgett:
Limit range validation added in #1134.

PR #1134, opened 2017-01-18
Bug 1414229: Display and validate limit ranges on create storage page

spadgett:
Show the min and max storage capacity and validate the entered value. If min and max are the same, show the value as a read-only field.

https://bugzilla.redhat.com/show_bug.cgi?id=1414229

@erinboyd @zherman0 fyi


jwforres:
[merge]

PR #1135, opened 2017-01-18
Escaping <> chars when included in display name within ui-select drop…

rhamilto:
…down

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1414195

Note:  this breaks search highlighting when the search query includes < or > since < and &amp;lt; and > and &amp;gt; are not the same.  This seems like the lesser of two evils since the inclusion of < or > in ui-select dropdowns is [a limitation of ui-select itself](https://github.com/angular-ui/ui-select/issues/846#issuecomment-226226811).

Before:

screen shot 2017-01-18 at 2 31 14 pm

After:

screen shot 2017-01-18 at 2 28 32 pm

Search highlighting ok as < or > not included in search query:
screen shot 2017-01-18 at 2 28 43 pm

Search highlighting broken as < included in search query:
screen shot 2017-01-18 at 2 28 51 pm

@jwforres or @spadgett, PTAL.

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -120,16 +120,29 @@ angular.module('openshiftConsole')
       });
       return nameCount;
     }
-    return function (resource, projects){
+    function escapeLessThanGreaterThan(str) {
+      return str.
+        replace(/</g, '&lt;').
+        replace(/>/g, '&gt;');
+    }
+    return function (resource, projects, escape){
       if (!resource) {
         return '';
       }
       var displayName = displayNameFilter(resource);
       var name = resource.metadata.name;
       if (displayName !== name && countNames(projects)[displayName] > 1 ){
-        return displayName + ' (' + name + ')';
+        if (!!escape) {
+          return escapeLessThanGreaterThan(displayName) + ' (' + name + ')';
+        } else {
+          return displayName + ' (' + name + ')';
+        }
       }
-      return displayName;
+      if (!!escape) {
+        return escapeLessThanGreaterThan(displayName);
+      } else {
+        return displayName;
+      }
     };
   })
   .filter('searchProjects', function(annotationNameFilter) {
--- a/app/views/create-from-url.html
+++ b/app/views/create-from-url.html
@@ -28,7 +28,7 @@
                           {{$select.selected | uniqueDisplayName : projects}}
                         </ui-select-match>
                         <ui-select-choices repeat="project in projects | searchProjects : $select.search">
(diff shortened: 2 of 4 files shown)


spadgett:
We should look at all the places we're using ui-select values that might have special characters.

app/scripts/filters/resources.js
```diff
@@ -120,16 +120,29 @@ angular.module('openshiftConsole')
       });
       return nameCount;
     }
-    return function (resource, projects){
+    function escapeLessThanGreaterThan(str) {
+      return str.
+        replace(/</g, '&lt;').
+        replace(/>/g, '&gt;');
+    }
```
spadgett: You can use `_.escape`, although I suggest adding it as a separate filter to util.js we can call anywhere.

```js
  .filter('escape', function() {
    return _.escape;
  });
```

then

```html
<div ng-bind-html="project | uniqueDisplayName : projects | escape | highlight : $select.search"></div>
```

spadgett: We might write our own highlight filter at some point to correctly handle the entities.

rhamilto: I explicitly opted for this custom function so I could scope it to only < and > (`_.escape` covers &, <, >, ", ', and \`).   Given &, ", ', and \` don't  cause the bug, I thought it was better to leave them out of the escaping so search highlighting isn't broken for them, too.  Yea/nay?

spadgett: Are we sure there isn't a problem with other content? Need to be careful when setting something as HTML to avoid XSS attacks (although Angular helps sanitize unsafe markup I believe). I'd rather be conservative and escape `&` too.

spadgett: I just checked and there's a problem with `&`. Set the project display name to an HTML entity to see. But Angular sanitizes things like `<script>` tags.

rhamilto: > Are we sure there isn't a problem with other content?

No, but wouldn't this be a much larger issue for anyone using ui-select?  Is `ng-bind-html` not protecting us?

spadgett: > No, but wouldn't this be a much larger issue for anyone using ui-select?

Yes.

> Is ng-bind-html not protecting us?

From XSS attacks, yes, but there are still display issues with HTML entities.

rhamilto: Are you proposing what would amount to a patch to ui-select?

spadgett: I think the easiest for now is to use `_.escape` to fix the display issue.

spadgett:
Replaced by #1145, which handles the special characters in searches.

PR #1136, opened 2017-01-18
Use image listing from registry-image-widgets

stefwalter:
This shares code with the Registry Console and Openshift Web Console for the ImageStream image listing.


stefwalter:
Openshift Web Console View:

screenshot from 2017-01-18 20-55-26

stefwalter:
Existing Registry Console Views:

screenshot from 2017-01-18 20-54-31

screenshot from 2017-01-18 20-54-36

stefwalter:
CC @andreasn @jwforres

stefwalter:
CC @dperpeet

spadgett:
Very cool

jwforres:
I see this is pulling in gettext now @stefwalter is this going to cause the image stream bits to be translated when the rest of the console is not? or does something have to be enabled for gettext to use a different language other than the default

stefwalter:
> I see this is pulling in gettext now @stefwalter is this going to cause the image stream bits to be translated when the rest of the console is not? or does something have to be enabled for gettext to use a different language other than the default.

It won't. The because project does not initiallize angular-gettext ... nor does the project load any translations.

jwforres:
OK sounds good, I will pull this down today and try it out

jwforres:
When no tags have been pushed it says "No image streams are present"  seems like that should say "No tags are present"

image-stream-1


When no tags have been pushed the Image count is empty instead of saying '0'. Also noticed it is inconsistent on this screen whether the labels have a ':' after them.  The rest of the console does have ':' after any labels.

image-stream-2



This version of the image widgets has some kind of issue with the image layers widget, when i revert back to master and use the image widgets version that points to this same image shows up just fine in the layers widget:

image-stream-3

jwforres:
When I hover over the dropdown arrow it messes up the row highlighting, part of the row is gray and part of the row is blue.

image-stream-4

jwforres:
The "pushed from" in the originates from column is not on the same horizontal as the tag and the SHA, which creates a strange white gap, seems like it should move up.

image-stream-5

jwforres:
The text "To push to an image to this image stream:"  isnt right, should be "To push an image to this image stream:"

jwforres:
What is expanding a tag supposed to do?  I thought maybe it was supposed to show the history for that tag, but when I expand it then it isn't showing me anything different, even when I know a tag has had multiple images pushed to it.  It just seems to add a line between the sha and the timestamp.

jwforres:
The sha causes problems at various responsive sizes, pushing content off screen. It needs to have word-break on it

image-stream-6

jwforres:
I'm actually seeing similar issues on the image widget page with long identifiers pushing off the page at mobile

jwforres:
When no images are pushed yet, the table doesn't extend all the way to the right of the screen, as soon as there as an image it does

image-stream-7

andreasn:
There is a couple of things I'm not currently satisfied about how this view works, also in the registry.
 * Tags appearance looks out of place. Maybe this would do well from using the badge look. http://www.patternfly.org/pattern-library/widgets/#badges
 * The rows are very tall.
 * The list view looks and works a bit different from both Cockpit and Patternfly (mostly because it's older code from before we refined the appearance of things) http://www.patternfly.org/pattern-library/content-views/list-view/#/expanding-rows
 * The SHA string is reeeeally long and just breaks the rows apart. This will only get worse with SHA512. I understand the idea about not hiding part of it (opening the question of "will I be able to copy this out?"), but maybe we can ellipsis it a bit and then show the full value on hover.

I'll put together a suggestion in a mockup for how to solve these.

stefwalter:
Thanks for the review @andreasn. May I suggest changes as a follow up pull request, and the first one gets the two projects mostly in sync on the view? Are you okay with that @jwforres 

@jwforres and @andreasn: The inline expansion stuff wasn't supposed to show up in the Openshift Web Console. That's a bug.

@andreasn In the registry the entire sha256 is necessary as the docker command line doesn't support abbreviated SHA identifiers for images. But perhaps that's a place where the Openshift Web Console can diverge and the implementation can take a CSS option to truncate this ... just thinking outloud.

andreasn:
I'm OK with doing it as a followup if @jwforres is.

stefwalter:
> When no tags have been pushed it says "No image streams are present" seems like that should say "No tags are present"

Fixed.

> When no tags have been pushed the Image count is empty instead of saying '0'.

Fixed.

> Also noticed it is inconsistent on this screen whether the labels have a ':' after them. The rest of the console does have ':' after any labels.

The Registry Console doesn't use colons, and uses different styling for metadata/data. So to make it consistent I've just added these colons via CSS in the Openshift Web Console use of the widgets.

> This version of the image widgets has some kind of issue with the image layers widget, when i revert back to master and use the image widgets version that points to this same image shows up just fine in the layers widget:

Fixed. This was a regression from when we marked is translatable. Now works.

> When I hover over the dropdown arrow it messes up the row highlighting, part of the row is gray and part of the row is blue.

This doesn't happen for me. Maybe I tagged the bower registry-image-widgets build incorrectly or something. Should work now. The dropdown arrow is not part of the changes we intended to share with the Openshift Web Console. The Registry Console has a listing pattern that allows expansion of data inline, but that doesn't need to be carried over into the Openshift Web Console.

> The "pushed from" in the originates from column is not on the same horizontal as the tag and the SHA, which creates a strange white gap, seems like it should move up.

I'll leave this up to @andreasn's follow up pull request.

> The text "To push to an image to this image stream:" isnt right, should be "To push an image to this image stream:"

Fixed.

> When no images are pushed yet, the table doesn't extend all the way to the right of the screen, as soon as there as an image it does

Fixed.

> The sha causes problems at various responsive sizes, pushing content off screen. It needs to have word-break on it

Added word-break.

stefwalter:
Everything fixed except for the row redesign, which should be done as a follow up to getting this feature "in" before the freeze.

pweil-:
@stefwalter please rebase

@jwforres @spadgett bump for another review since a lot of things have been fixed :+1:

stefwalter:
Rebased.

stefwalter:
However origin-web-console no longer authenticates my version of openshift, and I'm running out of time to sort that all out.

```
[root@f1 ~]# openshift version
openshift v1.4.1+3f9807a
kubernetes v1.4.0+776c994
etcd 3.1.0-rc.0
```

jwforres:
it is probably because of the change to remove localhost:9000 from the
default set of redirectURIs allowed in the master for the console oauth
client.  As a cluster admin you need to run

oc patch oauthclient/openshift-web-console -p '{"redirectURIs":["
https://localhost:9000/"]}'

We all use oc cluster up for development, which runs that step for us.

stefwalter:
Thanks @jwforres. That did the trick. Smoke testing done.

andreasn:
I've opened an issue in Cockpit for the changes I

petervo:
@jwforres opened https://github.com/openshift/registry-image-widgets/pull/7 with the redesign changes. Once that gets in i'll update this PR.

jwforres:
@petervo thanks! merged the other PR and released v0.0.8

petervo:
@jwforres, Thanks, rebased and updated this one.

jwforres:
thanks @petervo 

I'm seeing two bugs currently

1) When you hover over a row the visual style is a little weird where the blue hover color appears on three sides of the tag but not on the left.  This one is lower severity and I'd be ok fixing it in a follow-up as long as we get an issue open to track it.
hover-state

2) Bad overflow can occur at mobile res in a couple places.  The table can overflow, and the docker pull commands can overflow.
overflow-mobile

app/views/browse/imagestream.html
```diff
@@ -44,105 +44,14 @@ <h1 class="contains-actions">
...
-                    </td>
-                    <td data-title="Pull Spec">
-                      <div ng-if="item.dockerImageReference" ng-attr-title="{{item.dockerImageReference}}" class="td-long-string">
-                        {{item.dockerImageReference}}
-                      </div>
-                    </td>
-                  </tr>
-                </tbody>
-              </table>
+              <registry-imagestream-body imagestream="imageStream">
```
jwforres: sorry I didn't notice it before, but the switch to use registry-imagestream-body changes the behavior of how annotations are presented to the user, everywhere else in the UI we do not show these to the user unless they click a link to reveal them.  This is because annotations are typically intended for machine use and not end user facing.

petervo:
@jwforres, for the docker commands  do you want ellipsis or line wrapping with small sizes. Line wraping looks weird but is better for copy/paste. Is that ok?

app/views/browse/imagestream.html
```diff
@@ -44,105 +44,14 @@ <h1 class="contains-actions">
...
           <div class="row" ng-if="imageStream">
             <div class="col-md-12">
-              <div class="resource-details">
-                <dl class="dl-horizontal left">
-                  <dt ng-if-start="imageStream.spec.dockerImageRepository">Follows docker repo:</dt>
-                  <dd ng-if-end>{{imageStream.spec.dockerImageRepository}}</dd>
-                  <dt>Docker pull spec:</dt>
-                  <dd>{{imageStream.status.dockerImageRepository}}</dd>
-                </dl>
-                <annotations annotations="imageStream.metadata.annotations"></annotations>
```
jwforres: for reference, this is the annotations directive we use today that handles the show/hide behavior plus how we format annotations when visible

jwforres:
agree wrapping is probably better.  i know we do multi-line code style in
other places and managed to make it look alright, one example is in
_pod-template.html L94, i suspect the "command" class was probably adding
some special multi-line stuff

petervo:
@jwforres, opened https://github.com/openshift/registry-image-widgets/pull/8 and built this with that so you can check this out first before merging the other one.

Once that is in i can do a rebuild

jwforres:
looks like the table still has an issue at iphone 5 res which is 320px width

the quick fix is in imagestream.html to wrap `<registry-imagestream-body>` in a `<div class="table-responsive" style="padding: 0 5px;">`

this will prevent the layout like the nav from getting messed up when the table overflows the page, it wraps the table in a scrollable content area with a border when at mobile resolutions

if you are fine just making that change in this PR then we can merge the PR to image-registry-widgets and cut that release, let me know

petervo:
@jwforres pushed an update that fixes it for me.

jwforres:
@petervo I have your latest changes but I can still reproduce it with a tag name that is long

long-tag-name

petervo:
@jwforres updated again.

jwforres:
thanks LGTM now, think we can merge the other PR

jwforres:
I'll LGTM the other PR and you can merge/release it if you are ready

petervo:
@jwforres, i don't have access to marge or deploy to npm. I'll see if can I track down someone who does.

jwforres:
@petervo ah k i can merge and create the tag for bower, i dont think i have the npm access but bower is all we need to the console

jwforres:
@petervo ok 0.0.9 should be available for bower now

jwforres:
[merge]

PR #1143, opened 2017-01-19
Fix dangling "none" on statefulset template

benjaminapetersen:
Just noticed this, after shifting around the location of the volume claim templates.

Without a volume claim template there is a dangling "none":
screen shot 2017-01-19 at 10 17 05 am

Fix will eliminate:
screen shot 2017-01-19 at 10 17 46 am

What it looks like with a volume claim template, for reference:

screen shot 2017-01-19 at 10 20 58 am

@jwforres

### DIFF
--- a/app/views/browse/stateful-set.html
+++ b/app/views/browse/stateful-set.html
@@ -106,9 +106,6 @@
                         detailed="true">
                       </pod-template>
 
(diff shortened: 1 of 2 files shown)


jwforres:
[merge]

PR #1088, opened 2017-01-06
[WIP] PetSets / StatefulSets

benjaminapetersen:
Initial run at PetSets/StatefulSets

Currently supports:
- Edit YAML
- Delete
- Readonly env vars (cannot edit env vars on a PetSet)
- Annotations
- Labels

Screenshots:

screen shot 2017-01-06 at 1 46 15 pm
screen shot 2017-01-06 at 1 46 24 pm

I have a running list of about 15 more tasks, but wanted to get this up here sooner than later & see if we can merge in PetSets in stages.  

@jwforres @spadgett

PR #1144, opened 2017-01-19
[WIP] stateful sets updated volume claim template selector

benjaminapetersen:
Initial swap, will require a minor CSS change and probably some template updates elsewhere, pinging @sg00dwin to make sure changes don't cause other bugs.

Initial discussion @ [this comment](https://github.com/openshift/origin-web-console/pull/1088#discussion_r96746640)

### DIFF
--- a/app/views/_volume-claim-templates.html
+++ b/app/views/_volume-claim-templates.html
@@ -36,9 +36,12 @@
       </div>
       <div flex class="word-break">
         <span class="pod-template-key">Selector:</span>
+        <!--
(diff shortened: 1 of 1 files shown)


benjaminapetersen:
I think we probably just need to swap to using a class name rather than `:first-child` selector:
```css
.icon-row div:first-child {
    text-align: center;
    width: 30px;
}
```
instead:
```css
.icon-row .icon-wrap { 
    text-align: center;
    width: 30px;
}
```
to eliminate the recursion.

PR #1146, opened 2017-01-19
Don't show full editor for one-line post-commit scripts

spadgett:
Doesn't change the build config editor, only the browse page.


jwforres:
[merge]

Issue #1037, opened 2016-12-14
Empty events table missing right table header border at some widths


spadgett:
![screen_shot_2016-12-14_at_12_13_58_pm](https://cloud.githubusercontent.com/assets/1167259/21192533/eb504f74-c1f6-11e6-833f-d14f57782ae9.png)

spadgett:
@rhamilto fyi

rhamilto:
I'm pretty sure this is a known Chrome bug.  https://bugs.chromium.org/p/chromium/issues/detail?id=2902  :(

PR #1151, opened 2017-01-19
Adjust colspan to correct events table border issue

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/1037

screen shot 2017-01-19 at 3 43 15 pm

### DIFF
--- a/app/views/directives/events.html
+++ b/app/views/directives/events.html
@@ -48,14 +48,14 @@
     </thead>
     <tbody ng-if="(filteredEvents | hashSize) === 0">
       <tr>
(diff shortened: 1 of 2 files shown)


rhamilto:
@sg00dwin, did you check the Events table?  I think your fix may break it.

rhamilto:
@sg00dwin, yeah.  It's broken.  

screen shot 2017-01-19 at 4 21 40 pm

I think you need to be dynamically setting the colspan based on whether or not kind and name are present.  I think I've got a working fix.  Lemme PR.

PR #1160, opened 2017-01-20
Avoid race condition on pod metrics page load

spadgett:
Fix a race between the first call to `update()` and the request for
custom metrics. The race caused the same data to sometimes load twice,
resulting in loops around the duplicated points in the spline chart.

cc @mwringe


jwforres:
[merge]

PR #1163, opened 2017-01-20
Add StatefulSets to monitoring

benjaminapetersen:
Progress so far:

screen shot 2017-01-20 at 1 37 28 pm

screen shot 2017-01-20 at 1 38 09 pm

I think this is doable today, is just a lot of copy/paste & update.  No logs (since we don't have the endpoint) but metrics works. 

@jwforres @spadgett

### DIFF
--- a/app/scripts/controllers/monitoring.js
+++ b/app/scripts/controllers/monitoring.js
@@ -46,6 +46,9 @@ angular.module('openshiftConsole')
       },
       {
         kind: "Builds"
+      },
+      {
+        kind: "StatefulSets"
       }
     ];
     $scope.kindSelector = {
@@ -55,26 +58,30 @@ angular.module('openshiftConsole')
     $scope.logOptions = {
       pods: {},
       replicationControllers: {},
-      builds: {}
+      builds: {},
+      statefulSets: {}
     };
 
     $scope.logCanRun = {
       pods: {},
       replicationControllers: {},
-      builds: {}
+      builds: {},
+      statefulSets: {}
     };
 
     $scope.logEmpty = {
       pods: {},
       replicationControllers: {},
-      builds: {}
+      builds: {},
+      statefulSets: {}
     };
 
     $scope.expanded = {
       pods: {},
       replicationControllers: {},
       replicaSets: {},
-      builds: {}
+      builds: {},
+      statefulSets: {}
     };
 
     var isNil = $filter('isNil');
@@ -86,7 +93,6 @@ angular.module('openshiftConsole')
 
     var ageFilteredBuilds, ageFilteredReplicationControllers, ageFilteredReplicaSets, ageFilteredPods;
 
-    // Check if the metrics service is available so we know when to show the tab.
     MetricsService.isAvailable().then(function(available) {
       $scope.metricsAvailable = available;
     });
@@ -105,7 +111,17 @@ angular.module('openshiftConsole')
       $scope.filteredReplicationControllers = KeywordService.filterForKeywords(ageFilteredReplicationControllers, filterFields, filterKeywords);
       $scope.filteredReplicaSets = KeywordService.filterForKeywords(ageFilteredReplicaSets, filterFields, filterKeywords);
       $scope.filteredBuilds = KeywordService.filterForKeywords(ageFilteredBuilds, filterFields, filterKeywords);
+      $scope.filteredStatefulSets = KeywordService.filterForKeywords(_.values($scope.statefulSets), filterFields, filterKeywords);
     };
+
+    // TODO: logs for stateful sets will come later
+    // logging endpoint for stateful sets coming in the future.
+    // var setStatefulSetLogVars = function(set) {
+    //   $scope.logOptions.statefulSets[set.metadata.name] = {
+    //     container: set.spec.template.spec.containers[0].name
+    //   };
+    //   $scope.logCanRun.statefulSets[set.metadata.name] = !(_.includes(['New', 'Pending', 'Unknown'], set.status.phase));
+    // };
 
     var setPodLogVars = function(pod) {
       $scope.logOptions.pods[pod.metadata.name] = {
@@ -126,6 +142,11 @@ angular.module('openshiftConsole')
     var setBuildLogVars = function(build) {
       $scope.logOptions.builds[build.metadata.name] = {};
       $scope.logCanRun.builds[build.metadata.name] = !(_.includes(['New', 'Pending', 'Error'], build.status.phase));
+    };
+
+
+    var filterStatefulSets = function() {
+      $scope.filteredStatefulSets = KeywordService.filterForKeywords(_.values($scope.statefulSets), filterFields, filterKeywords);
     };
 
     var filterPods = function() {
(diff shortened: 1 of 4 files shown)


jwforres:
so the interesting problem here is that when metrics isnt enabled there
would be nothing to show when the row is expanded :-\

spadgett:
Here's what we show for replica sets:

https://github.com/openshift/origin-web-console/blob/master/app/views/monitoring.html#L248-L252

benjaminapetersen:
TODO yet:
- [ ] fix the above issue to show something when metrics is not available
- [ ] fix the dropdown picker for "stateful sets"
- [ ] fix the replicas count, since it will always be 0 until 1.6

benjaminapetersen:
No logs message: 

screen shot 2017-01-20 at 3 05 53 pm

same fn works to load up the pods for a stateful sets as is used for the replica sets.

benjaminapetersen:
squashed

app/scripts/controllers/monitoring.js
```diff
@@ -105,6 +112,18 @@ angular.module('openshiftConsole')
       $scope.filteredReplicationControllers = KeywordService.filterForKeywords(ageFilteredReplicationControllers, filterFields, filterKeywords);
       $scope.filteredReplicaSets = KeywordService.filterForKeywords(ageFilteredReplicaSets, filterFields, filterKeywords);
       $scope.filteredBuilds = KeywordService.filterForKeywords(ageFilteredBuilds, filterFields, filterKeywords);
+      // TODO: make this work
+      $scope.filteredStatefulSets = KeywordService.filterForKeywords(ageFilteredStatefulSets, filterFields, filterKeywords);
+    };
+
+    // TODO: make setStatefulSetLogVars work
+    // logging endpoint for stateful sets coming in the future.
+    var setStatefulSetLogVars = function(set) {
```
jwforres: can we just comment out all of this, no point in running code we arent using

app/scripts/controllers/monitoring.js
```diff
@@ -128,6 +147,18 @@ angular.module('openshiftConsole')
       $scope.logCanRun.builds[build.metadata.name] = !(_.includes(['New', 'Pending', 'Error'], build.status.phase));
     };
 
+    var filterStatefulSets = function() {
+      ageFilteredStatefulSets = _.filter($scope.statefulSets, function(set) {
+        if (!$scope.filters.hideOlderResources) {
+          return true;
+        }
+        // TODO: statefulSets don't have status.phase. As far as I understand, they don't age the
+        // way other things age, so perhaps this is irrelevant?
```
jwforres: yep i think you are right, I don't think we would hide any statefulsets ever, so you don't need to change anything when the filter is checked/unchecked

app/views/monitoring.html
```diff
@@ -260,6 +262,108 @@
...
+
+                      <div class="list-view-pf-main-info">
+                        <div class="list-view-pf-body">
+                          <div class="list-view-pf-description">
+                            <div class="list-group-item-heading">
+                              <a ng-href="{{set | navigateResourceURL}}"><span ng-bind-html="set.metadata.name | highlightKeywords : filterKeywords"></span></a>
+                              <small>created <span am-time-ago="set.metadata.creationTimestamp"></span></small>
+                            </div>
+                            <div class="list-group-item-text">
+                              <!-- TODO: is this the right status? -->
```
jwforres: probably just needs to be the replica counts like we have in the stateful sets table, we dont have a status for stateful sets

app/views/monitoring.html
```diff
@@ -260,6 +262,108 @@
...
+                              <status-icon status="set | deploymentStatus" disable-animation></status-icon>
+                              {{set | deploymentStatus | sentenceCase}}
+                              <span ng-if="set.status.replicas !== set.spec.replicas">{{set.status.replicas}}/</span>{{set.spec.replicas}} replica<span ng-if="set.spec.replicas != 1">s</span>
+                            </div>
+                          </div>
+                          <div class="list-view-pf-additional-info">
+                            <div class="list-view-pf-additional-info-item">
+                              <span class="pficon fa-fw pficon-image"></span>
+                              <!-- TODO: not sure what to do with this, gotta figure it out -->
```
jwforres: looks like this is right, if you copied it from the dep stuff should be fine

app/views/monitoring.html
```diff
@@ -260,6 +262,108 @@
...
+                              <image-names pod-template="set.spec.template" pods="podsByOwnerUID[set.metadata.uid]"></image-names>
+                            </div>
+                          </div>
+                        </div>
+                      </div>
+                    </div>
+                    <div ng-repeat-end ng-if="expanded.statefulSets[set.metadata.name]" class="list-group-expanded-section" ng-class="{'expanded': expanded.statefulSets[set.metadata.name]}">
+
+                      <!-- TODO: add logs once we have logs subresource for statefulSets
```
jwforres: honestly i would probably just remove this completely, we dont have logs for replicasets either and it may be awhile before we do, I dont think we left any commented out code for those

app/views/monitoring.html
```diff
@@ -260,6 +262,108 @@
...
+                            </ui-select-choices>
+                          </ui-select>
+                        </span>
+
+                      </log-viewer>
+                      -->
+
+                      <!-- Until the metrics directive pulls the metrics from the time range of the life of the pod, hide the metrics for old stuff -->
+                      <div class="mar-top-lg" ng-if="metricsAvailable">
+                        Logs are not available for stateful sets.
```
jwforres: this doesnt seem right, this message only shows up if metricsAvailable is true, you'll still end up with any empty section when metrics arent available

app/views/monitoring.html
benjaminapetersen: works for me, removed.

app/views/monitoring.html
benjaminapetersen: ha, sorry, that was an early comment I should have removed :)

app/views/monitoring.html
benjaminapetersen: Ok. I got this for the other pages, will see if I can work in in here quick as well.

app/scripts/controllers/monitoring.js
benjaminapetersen: good, will wipe out most of this fn then.

app/scripts/controllers/monitoring.js
benjaminapetersen: can do.

app/views/monitoring.html
benjaminapetersen: Um, yeah... that doesn't make sense. Fixed.

benjaminapetersen:
Fixed the replicas count as well by counting pods

screen shot 2017-01-20 at 4 38 59 pm

benjaminapetersen:
rebased

jwforres:
[merge]

spadgett:
```
oc --server https://localhost:8443 patch oauthclient/openshift-web-console -p {"redirectURIs":["https://localhost:9000/"]}
>> Error: The connection to the server localhost:8443 was refused - did you specify the right host or port?
```

[merge]

jwforres:
[merge]

PR #1165, opened 2017-01-20
Fix stateful sets replicas count

benjaminapetersen:
@jwforres this will take care of the Stateful Set list & individual page. 
I'll fix it for monitoring in a separate follow-on.

### DIFF
--- a/app/scripts/controllers/statefulSets.js
+++ b/app/scripts/controllers/statefulSets.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module('openshiftConsole')
-  .controller('StatefulSetsController', function($scope, $routeParams, AlertMessageService, DataService, ProjectsService, LabelFilter) {
+  .controller('StatefulSetsController', function($scope, $routeParams, AlertMessageService, DataService, ProjectsService, LabelFilter, LabelsService) {
     $scope.projectName = $routeParams.project;
     $scope.alerts = $scope.alerts || {};
     $scope.labelSuggestions = {};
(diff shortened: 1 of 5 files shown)


app/scripts/controllers/statefulSets.js
```diff
@@ -34,6 +34,14 @@ angular.module('openshiftConsole')
           updateFilterWarning();
         }));
 
+        // TODO: 1.6 eliminate this block, we dont actually need pods on this page,
+        // we are just using to fix the fact that the replicas count in inaccurate
+        watches.push(DataService.watch('pods', context, function(podData) {
+          $scope.pods = podData.by('metadata.name');
+          var allOwners = _.toArray($scope.statefulSets);
```
jwforres: i actually dont think you need to convert it to an array, i think it will accept $scope.statefulSets directly.  The only reason we did it on overview is because we have to combine a bunch of different sets of owners

app/views/browse/stateful-set.html
```diff
@@ -88,13 +88,16 @@
...
+                          <!-- TODO: swap back in 1.6
                           <replicas
                             status="statefulSet.status.replicas"
                             spec="statefulSet.spec.replicas"
                             disable-scaling="!isScalable()"
                             scale-fn="scale(replicas)"
                             deployment="statefulSet">
                           </replicas>
+                          -->
+                          {{podsForStatefulSet | hashSize}} current / {{statefulSet.spec.replicas}} desired
```
jwforres: when current === desired we should just say 'X replicas'

app/scripts/controllers/statefulSets.js
benjaminapetersen: ok cool, ill nix this

app/views/browse/stateful-set.html
benjaminapetersen: will do, i have this on the other page, easy swap

benjaminapetersen: Will now be `3/5 replicas` or `5 replicas`, depending on `===`

benjaminapetersen:
squashed

jwforres:
[merge]

PR #1166, opened 2017-01-20
Fixing issue where long, unbroken strings in resource name cell URLs breaks layout

### LINKED ISSUE #1161: Long, unbroken names in route table pushes rest of content off-screen
screen shot 2017-01-20 at 11 44 41 am

@jwforres FYI

rhamilto:
Fixes #1161

screen shot 2017-01-24 at 10 17 45 am
screen shot 2017-01-24 at 10 17 59 am
screen shot 2017-01-24 at 10 18 08 am
screen shot 2017-01-24 at 10 18 20 am
screen shot 2017-01-24 at 10 18 44 am
screen shot 2017-01-24 at 10 19 02 am
screen shot 2017-01-24 at 10 21 10 am
screen shot 2017-01-24 at 10 21 54 am
screen shot 2017-01-24 at 10 22 24 am
screen shot 2017-01-24 at 10 22 31 am
screen shot 2017-01-24 at 10 25 14 am

Plus bonus bug fix for Pods table where colspan was incorrect

screen shot 2017-01-23 at 4 24 22 pm

@jwforres or @spadgett, PTAL.  cc: @sg00dwin

### DIFF
--- a/app/styles/_mixins.less
+++ b/app/styles/_mixins.less
@@ -56,7 +56,7 @@
      min-width: 0;          // firefox (1)
 }
 // DESCRIPTION: Breaks long non-breaking strings under the following circumstances:
-// - if within a table not set to table-layout: fixed
+// - if within a table not set to table-layout: fixed (but have a very good reason for doing so!  otherwise, see .table-layout-fixed)
 // - fieldset
 // - flex container
 // EXAMPLE:
--- a/app/styles/_tables.less
+++ b/app/styles/_tables.less
@@ -19,6 +19,14 @@
       }
     }
   }
+  // use if you have long, unbroken strings in cells you need to wrap
+  // you will probably want to use in conjunction with colgroup to size cols since this will make all cols equal size
+  &.table-layout-fixed {
+    table-layout: fixed;
+    td {
+      .word-break();
+    }
+  }
 }
 
 .table th .pficon-help {
@@ -169,16 +177,12 @@ td[role="presentation"].arrow:after {
 
 // table for events
 .events-table {
-  table-layout: fixed;
   th#time {width:90px;}
   th#kind-name {width:190px;}
   th#kind {width:160px;}
   th#severity {width:10px;padding:0;}
   th#reason {width:150px;}
   th#message   {width:100%;} // ensures it gets remaining width
-  td {
-    .word-break();
-  }
   .pficon {
     vertical-align: middle;
   }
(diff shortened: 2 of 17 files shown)


spadgett:
Should we apply the break-all styles to just the columns that need it? Some columns like status and relative times shouldn't need break-all.

rhamilto:
> Should we apply the break-all styles to just the columns that need it? Some columns like status and relative times shouldn't need break-all.

@sg00dwin, I was following [your lead](https://github.com/openshift/origin-web-console/blob/master/app/views/secrets.html#L31) on this fix.  What do you think?

sg00dwin:
There is a `.word-break-normal` class that can be set on individual `<td>`, so we can isolate in those cases. 

As for whether we should apply word-break-all by default... I assume it would depend on whether we think there's a greater risk of breaking content when it's not intended or not breaking long strings where needed. Either way we approach it, will require individual examination of a tables column content.

spadgett:
I feel like we have the default backwards: we should only apply the style to the few columns that need it, not the whole table. I'd rather add word-break-all to the td elements that need it and remove the class that adds it to the whole table.

rhamilto:
I switched the default to opt-in to word-break.  PTAL @spadgett.

rhamilto:
@spadgett, changed so that all td, th elements in .table-layout-fixed get word-break by default.

spadgett:
I'm seeing at least one place (deployments page) where we have overlapping text, likely because we have a min-width on the column.

openshift_web_console

spadgett:
Suggest we give less width to "Last Build" and more to "Created."

We might want duration value and unit to wrap together. (Doesn't have to be in this PR, but noticed while testing.)

@jwforres We might have too many columns in some of our tables :/

openshift_web_console

spadgett:
We should consider updating build-config.html, deployment-config.html, deployment.html, and other-resources.html, too.

rhamilto:
> I'm seeing at least one place (deployments page) where we have overlapping text, likely because we have a min-width on the column.

I think it should be better now.

> Suggest we give less width to "Last Build" and more to "Created."

Done.  And now we hide "Source" at small viewport sizes.  And fixed the th wrapping issue by removing word-break for th elements.

> We should consider updating build-config.html, deployment-config.html, deployment.html, and other-resources.html, too.

Done:

* deployment.html 
screen shot 2017-01-24 at 2 44 46 pm
* other-resources.html
screen shot 2017-01-24 at 3 59 53 pm

Not done as I don't think it's necessary (am I overlooking something on these?):

* build-config.html
screen shot 2017-01-24 at 3 57 48 pm
* deployment-config.html
screen shot 2017-01-24 at 3 57 26 pm

Let's give this another whirl, @spadgett?

spadgett:
[merge]

PR #1167, opened 2017-01-21
Add fullscreen terminal support

spadgett:
@jwforres FYI I have fullscreen support working for terminals. Need to test more across browsers.

@csrwng CC

TODO:

- [x] Add expand and collapse icons for entering and exiting fullscreen (`fa-expand` / `fa-compress`)
- [ ] Investigate sending ESC to the terminal since it's trapped by the browser to exit fullscreen

### DIFF
--- a/app/scripts/controllers/pod.js
+++ b/app/scripts/controllers/pod.js
@@ -15,6 +15,7 @@ angular.module('openshiftConsole')
                                          Logger,
                                          DataService,
                                          EnvironmentService,
+                                         FullscreenService,
                                          ImageStreamResolver,
                                          MetricsService,
                                          PodsService,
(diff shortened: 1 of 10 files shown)


app/scripts/controllers/pod.js
```diff
@@ -316,6 +317,17 @@ angular.module('openshiftConsole')
           }
         };
 
+        var focusTerminal = function() {
+          $('.terminal').focus();
```
jwforres: do we actually have multiple terminals when there are multiple containers (with only one visible)? or do we just maintain the connections?

or does jquery only find the visible ones by default, cant remember

spadgett: Good catch. It looks like we keep the terminal once it's been used, so it's possible to have multiple. I'll change it to only focus the visible terminal.

spadgett:
Updated

spadgett:
Rebased

jwforres:
[merge]

PR #1169, opened 2017-01-23
Bug 1415602: Fix next steps for deploy image

spadgett:
Don't try to calculate the previous URL for breadcrumbs. Instead, use
history back, which simplifies the next steps controller code and makes
it work from any page (including deploy image). This also fixes a bug
where the display name for the image stream was not preserved.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1415602


jwforres:
[merge]

PR #1170, opened 2017-01-23
Correcting word-break-all comment

rhamilto:
A silly PR, but I noticed this and figure it out to be fixed

@spadgett, PTAL


spadgett:
[merge]

PR #1171, opened 2017-01-23
Better conflict handling for env var edits

### LINKED ISSUE #621: ENV variable manager and crashing pods
This left a bad experience for a user who's pod kept crashing due to an incorrectly set ENV variable. 

When he went to edit the ENV variable in the console, because the pod kept restarting it was triggering a restart. By the time he had edited the ENV variable, the pod would've crashed and tried again, the web console would then force him to refresh the ENV variables and start again.

### LINKED ISSUE #676: Got both success and warning alerts when saving env variables
Sometimes when you save env variables, you see both a success alert saying the resource was updated and a warning saying the resource was updated in the background. Neither of these alerts were visible before I clicked "Save." This was editing a deployment.

openshift_web_console

/cc @jwforres @benjaminapetersen

spadgett:
Fixes #621
Fixes #676

@jwforres @rhamilto FYI

Resources updated:
- [x] Deployment Config
- [x] Deployment
- [x] Replica Set
- [x] Replication Controller
- [x] Stateful Set

### DIFF
--- a/app/scripts/controllers/deployment.js
+++ b/app/scripts/controllers/deployment.js
@@ -14,14 +14,14 @@ angular.module('openshiftConsole')
                         AlertMessageService,
                         DataService,
                         DeploymentsService,
+                        EnvironmentService,
                         HPAService,
                         ImageStreamResolver,
                         ModalsService,
                         Navigate,
                         Logger,
                         ProjectsService,
-                        StorageService,
-                        keyValueEditorUtils) {
+                        StorageService) {
     var imageStreamImageRefByDockerReference = {}; // lets us determine if a particular container's docker image reference belongs to an imageStream
 
     $scope.projectName = $routeParams.project;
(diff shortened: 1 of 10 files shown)


app/scripts/controllers/deploymentConfig.js
```diff
@@ -56,18 +56,43 @@ angular.module('openshiftConsole')
...
+      // deployment config object.
+      if (EnvironmentService.isEnvironmentEqual(current, previous)) {
+        $scope.updatedDeploymentConfig = EnvironmentService.mergeEdits($scope.updatedDeploymentConfig, current);
+        return;
+      }
+
+      previousEnvConflict = true;
+      $scope.alerts["background_update"] = {
+        type: "warning",
+        message: "This deployment configuration has been updated in the background. Saving your changes may create a conflict or cause loss of data.",
```
jwforres: now that we know the env vars are what caused the conflict, i'm thinking we should update the message to be more specific

app/scripts/services/environment.js
```diff
@@ -0,0 +1,77 @@
...
+      // Note: This modifies object. It should only be called on a copy.
+      compactEnvironment: function(object) {
+        var containers = getContainers(object);
+        _.each(containers, function(container) {
+          container.env = keyValueEditorUtils.compactEntries(container.env);
+        });
+      },
+
+      // Copy and normalize the environment for editing using the key value editor.
+      copy: function(object) {
```
jwforres: hmm, i wouldn't expect a function called copy to also be doing normalization

app/scripts/services/environment.js
```diff
@@ -0,0 +1,77 @@
...
+      isEnvironmentEqual: function(left, right) {
+        var leftContainers = getContainers(left);
+        var rightContainers = getContainers(right);
+        if (leftContainers.length !== rightContainers.length) {
+          return false;
+        }
+
+        var i, leftEnv, rightEnv;
+        for (i = 0; i < leftContainers.length; i++) {
+          leftEnv = leftContainers[i].env || [];
```
jwforres: i feel like we should be validating that the container names are actually the same, what if i took out a container, added a brand new container, the length of containers array is the same but they are no longer equal. Especially since merge edits below assumes containers in the same positions are in fact the same.

spadgett:
@jwforres Updated. I'd like to do some more testing, but the code should be ready for another review.

spadgett:
Still need to sort out #676, which I suspect is a race between the PUT callback and watch callback.

app/scripts/services/environment.js
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        var i, leftEnv, rightEnv;
+        for (i = 0; i < leftContainers.length; i++) {
+          // If a container name has changed, consider it a conflict.
+          if (leftContainers[i].name !== rightContainers[i].name) {
+            return false;
+          }
+
+          // Check if any of the variable names are values are different.
```
jwforres: typo in the comment 'variable names or values'

jwforres:
just one nit, otherwise LGTM

spadgett:
Thanks @jwforres 

Updated with a fix for the PUT/watch race (#676).

app/scripts/controllers/deployment.js
```diff
@@ -131,28 +156,23 @@ angular.module('openshiftConsole')
...
+              updateHPAWarnings();
 
-              if ($scope.forms.deploymentEnvVars.$pristine) {
-                copyDeploymentAndEnsureEnv(deployment);
+              updateEnvironment(deployment, previous);
+
+              // Wait for a pending save to complete to avoid a race between the PUT and the watch callbacks.
+              if (saveEnvPromise) {
+                saveEnvPromise.finally(function() {
+                  updateEnvironment(deployment, previous);
```
spadgett: @jwforres It looks like calling `updateEnvironment` from `saveEnvPromise.finally` is enough.

* If the PUT succeeds, the success callback in `saveEnvVars` will set the form to pristine before `updateEnvironment` is called. When the form is pristine, `updateEnvironment` will simply copy the new values.
* If the PUT fails, the normal conflict checking is performed by `updateEnvironment`.

jwforres:
Update LGTM

spadgett:
> Update LGTM

Thanks @jwforres. Tested various scenarios with all of the types (deployment configs, deployments, replication controllers, and replica sets). Also checked the read-only view of the stateful set environment.

[merge]

spadgett:
```
Running "add-redirect-uri" task
oc --server https://localhost:8443 patch oauthclient/openshift-web-console -p {"redirectURIs":["https://localhost:9000/"]}
>> Error: The connection to the server localhost:8443 was refused - did you specify the right host or port?
```

[merge]

PR #1172, opened 2017-01-23
Update membership role filter to use new annotation

benjaminapetersen:
@jwforres @spadgett

### DIFF
--- a/app/scripts/services/membership/membership.js
+++ b/app/scripts/services/membership/membership.js
@@ -114,18 +114,10 @@ angular
       return _.sortBy(roles, 'metadata.name');
     };
 
-    // TODO: follow-on PR, there will be an annotation for this
     var filterRoles = function(roles) {
       return _.filter(roles, function(item) {
-        // image-puller & image-pusher ok, other system: prob no
-        return (_.isEqual(item.metadata.name, 'system:image-puller')   ||
-                _.isEqual(item.metadata.name, 'system:image-pusher')   ||
-                _.isEqual(item.metadata.name, 'system:image-builder')  ||
-                _.isEqual(item.metadata.name, 'system:deployer')     ) ||
-                ! _.startsWith(item.metadata.name, 'cluster-') &&
-                ! _.startsWith(item.metadata.name, 'system:') &&
-                ! _.startsWith(item.metadata.name, 'registry-') &&
-                ! _.startsWith(item.metadata.name, 'self-');
+        // system-only must be explicitly <string> 'true' to hide the annotation
+        return annotation(item, 'systemOnly') !== 'true';
       });
     };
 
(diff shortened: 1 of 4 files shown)


benjaminapetersen:
@enj fyi

app/scripts/services/membership/membership.js
```diff
@@ -114,18 +114,9 @@ angular
...
-        // image-puller & image-pusher ok, other system: prob no
-        return (_.isEqual(item.metadata.name, 'system:image-puller')   ||
-                _.isEqual(item.metadata.name, 'system:image-pusher')   ||
-                _.isEqual(item.metadata.name, 'system:image-builder')  ||
-                _.isEqual(item.metadata.name, 'system:deployer')     ) ||
-                ! _.startsWith(item.metadata.name, 'cluster-') &&
-                ! _.startsWith(item.metadata.name, 'system:') &&
-                ! _.startsWith(item.metadata.name, 'registry-') &&
-                ! _.startsWith(item.metadata.name, 'self-');
+        return !annotation(item, 'systemOnly');
```
jwforres: you need to actually check the content of the annotation and compare it to the string 'true' in theory someone could set the annotation and set it to false

jwforres:
@benjaminapetersen have you run the latest origin images and made sure this works as expected in combination with that?

benjaminapetersen:
@jwforres yup, built it this morn

app/scripts/services/membership/membership.js
benjaminapetersen: Right, & verified with @enj that is consistent with some other annotations.  So only `'true'` will be honored. Updated.

jwforres:
@benjaminapetersen the travis failure is because your spec tests need to be updated

benjaminapetersen:
fixed
[TEST]

test/spec/services/membership/membershipSpec.js
```diff
@@ -87,27 +87,14 @@ describe('MembershipService', function() {
...
-        {metadata: {name: 'system:deployer'}},
-        // roles the filter will explicitly reject
-        {metadata: {name: 'cluster-magical-scary-role'}},
-        {metadata: {name: 'system:special-scary-role'}},
-        {metadata: {name: 'registry-ninja'}},
-        {metadata: {name: 'self-destruction'}},
-        // ignored
-        {metadata: {name: 'admin'}},
-        {metadata: {name: 'edit'}}
+        {metadata: {name: 'system-only-role', annotations: {'authorization.openshift.io/system-only': 'true'}}},
```
jwforres: Can you add one with the annotation set to false

benjaminapetersen: Yeah ill add a few more examples of things that wont actually pass the filter, just to be clear.

jwforres:
[merge]

PR #1179, opened 2017-01-25
Make empty msg for stateful sets consistant with other ones

jhadvig:
@spadgett PTAL

### DIFF
--- a/app/views/browse/stateful-sets.html
+++ b/app/views/browse/stateful-sets.html
@@ -43,7 +43,7 @@
                 </thead>
                 <tbody ng-if="(statefulSets | hashSize) == 0">
                   <tr>
-                    <td colspan="3"><em>No stateful sets</em></td>
+                    <td colspan="3"><em>No stateful sets to show</em></td>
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/statefulSets.js
```diff
@@ -24,6 +24,7 @@ angular.module('openshiftConsole')
           group: 'apps',
           version: 'v1beta1'
         }, context, function(statefulSets) {
+          $scope.emptyMessage = "No stateful sets to show";
```
spadgett: I'd rather inline the message than use a scope var since it doesn't change. (We do this in other places, but I'd like to get away from that.)

jhadvig:
@spadgett updated ... makes sense to have the msg inline :+1: . Also created a [issue](https://github.com/openshift/origin-web-console/issues/1182) to move all the empty msgs inline and assigned to myself if you dont mind, just to stay in shape :smile:

spadgett:
Sounds good, thanks

[merge]

PR #1183, opened 2017-01-26
Fix error changing route termination to passthrough

spadgett:
Fix a problem when changing an existing route to passthrough from
another termination type. The web console was incorrectly sending
certificates (and path if set) in the PUT request even though those
inputs were disabled in the form and invalid for passthrough routes.


jwforres:
[merge]

PR #1184, opened 2017-01-26
Let users specify insecureEdgeTerminationPolicy for other termination types

spadgett:
Previously we only let users specify the insecureEdgeTerminationPolicy for edge-terminated routes. Allow it for any termination type. The value `Allow` should be disabled for passthrough routes, however.

cc @smarterclayton

### DIFF
--- a/app/views/directives/osc-routing.html
+++ b/app/views/directives/osc-routing.html
@@ -247,15 +247,28 @@
       <!-- Insecure Edge Termination Policy -->
       <div class="form-group">
         <label for="insecureTraffic">Insecure Traffic</label>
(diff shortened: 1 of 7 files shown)


app/views/directives/osc-routing.html
```diff
@@ -247,15 +247,28 @@
...
         </ui-select>
         <div>
           <span id="route-insecure-policy-help" class="help-block">
-            Policy for traffic on insecure schemes like HTTP for edge termination.
+            Policy for traffic on insecure schemes like HTTP.
+          </span>
+        </div>
+        <div ng-if="routeForm.insecureTraffic.$error.passthrough" class="has-warning">
+          <span class="help-block">
+            Passthrough routes can't have insecure traffic policy <var>Allow</var>.
```
jwforres: Passthrough routes can't use the insecure traffic policy Allow

spadgett:
@jwforres thanks, updated

jwforres:
[merge]

Issue #1175, opened 2017-01-24
Long, unbroken strings on overview overflow containers


rhamilto:
![screen shot 2017-01-24 at 10 10 41 am](https://cloud.githubusercontent.com/assets/895728/22253001/0d91c60c-e21e-11e6-87f7-a079ee69970b.PNG)

This example shows DC and SS.  There may be others.

spadgett:
We should make sure that metrics tooltips aren't clipped when we fix this since they can extend outside the tile.

PR #1188, opened 2017-01-27
Addition of truncate class to tile headers and when no deployments have started

sg00dwin:
Include ng-class to selectively apply margin-right for tiles that have rc shield & number
Tested in FF, Chrome, Safari, IE

Fixes https://github.com/openshift/origin-web-console/issues/1175

screen shot 2017-01-27 at 3 07 00 pm

### DIFF
--- a/app/views/overview/_dc.html
+++ b/app/views/overview/_dc.html
@@ -1,8 +1,8 @@
 <div class="overview-tile" ng-class="{ 'deployment-in-progress': inProgressDeployment }">
   <ng-include src="'views/overview/_service-header.html'"></ng-include>
   <div class="overview-tile-header">
-    <div class="rc-header">
-      <div>
+    <div class="rc-header" ng-class="{ 'rc-header-shield' : activeReplicationController}">
+      <div class="truncate">
         Deployment Config
         <a ng-href="{{deploymentConfig | navigateResourceURL}}">{{deploymentConfig.metadata.name}}</a>
         <small class="overview-timestamp" ng-if="activeReplicationController && !inProgressDeployment">
@@ -43,8 +43,7 @@
         A new deployment will start automatically when
         <span ng-if="imageChangeTriggers.length === 1">
           an image is available for
-          <a ng-href="{{urlForImageChangeTrigger(imageChangeTriggers[0], deploymentConfig)}}">
-            {{imageChangeTriggers[0].imageChangeParams.from | imageObjectRef : deploymentConfig.metadata.namespace}}</a>.
+          <a ng-href="{{urlForImageChangeTrigger(imageChangeTriggers[0], deploymentConfig)}}">{{imageChangeTriggers[0].imageChangeParams.from | imageObjectRef : deploymentConfig.metadata.namespace}}</a>.
         </span>
         <span ng-if="imageChangeParams.length > 1">
           one of the images for this deployment config changes.
@@ -63,11 +62,8 @@
           </div>
         </div>
         <div ng-if="!pipelinesForDC[deploymentConfig.metadata.name].length">
-          <p>
-            No deployments have started for
-            <a ng-href="{{deploymentConfig | navigateResourceURL}}">{{deploymentConfig.metadata.name}}</a>.
-          </p>
-
+            <p>No deployments have started for
+            <a ng-href="{{deploymentConfig | navigateResourceURL}}">{{deploymentConfig.metadata.name}}</a>.</p>
           <button ng-if="'deploymentconfigs' | canI : 'update'" class="btn btn-primary" ng-click="startDeployment(deploymentConfig)">
             Start Deployment
           </button>
--- a/app/views/overview/_deployment.html
+++ b/app/views/overview/_deployment.html
@@ -1,8 +1,8 @@
 <div class="overview-tile" ng-class="{ 'deployment-in-progress': inProgressDeployment }">
   <ng-include src="'views/overview/_service-header.html'"></ng-include>
   <div class="overview-tile-header">
-    <div class="rc-header">
-      <div>
+    <div class="rc-header" ng-class="{ 'rc-header-shield' : latestReplicaSet && latestRevision && !inProgressDeployment}">
+      <div class="truncate">
         Deployment
         <a ng-href="{{deployment | navigateResourceURL}}">{{deploymentName}}</a>
         <small class="overview-timestamp" ng-if="latestReplicaSet">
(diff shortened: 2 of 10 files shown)


sg00dwin:
@rhamilto @spadgett

app/views/overview/_dc.html
```diff
@@ -1,8 +1,8 @@
 <div class="overview-tile" ng-class="{ 'deployment-in-progress': inProgressDeployment }">
   <ng-include src="'views/overview/_service-header.html'"></ng-include>
   <div class="overview-tile-header">
-    <div class="rc-header">
-      <div>
+    <div class="rc-header" ng-class="{ 'rc-header-shield' : activeReplicationController}">
```
spadgett: You'll need to add the `rc-header-shield` class to the shield in _deployment.html as well.

rhamilto:
Service names also have the same problem.  Want to fix that here?  Or another issue?

sg00dwin:
@rhamilto @spadgett 
Added `rc-header-shield` to _deployment.html and service names are now truncated also.

Tested in FF, Chrome, Safari, IE

screen shot 2017-01-30 at 2 18 01 pm

app/views/overview/_deployment.html
```diff
@@ -1,8 +1,8 @@
 <div class="overview-tile" ng-class="{ 'deployment-in-progress': inProgressDeployment }">
   <ng-include src="'views/overview/_service-header.html'"></ng-include>
   <div class="overview-tile-header">
-    <div class="rc-header">
-      <div>
+    <div class="rc-header" ng-class="{ 'rc-header-shield' : activeReplicationController}">
```
spadgett:
```html
<div class="rc-header" ng-class="{ 'rc-header-shield': latestReplicaSet && latestRevision && !inProgressDeployment }">
```

spadgett: @sg00dwin You can test with

https://github.com/openshift/origin-web-console/blob/master/test/testdata/k8s-deployments.yaml

spadgett:
@sg00dwin I see you pushed some updates. Ready for another review?

sg00dwin:
@spadgett Yes please review.

I enabled the truncation of service names. To get that to work, required some changes to `overview-service` and `.service-name` and removing the flex from tiles for `.no-deployments-message` and  then applied the `overview-tile` rules, so that they wouldn't over expand. Also, the truncation along with  `line-height: 1.1` inherited from `.h3()` caused clipping of text descenders. So I increased line-height and offset the additional height by reducing the top and bottom margins so it maintained it's current height.

Additionally
Since the links in `.empty-tile` and `.empty-dc` where in the middle of a sentence I couldn't truncate so I applied `.word-break-all` so that they would wrap in every browser. Since this can cause weird breaking mid word, I reduced the left and right padding so that it would allow more room for long strings and reduce the chance of it occurring.


screen shot 2017-01-31 at 4 55 06 pm
screen shot 2017-01-31 at 4 52 45 pm

app/views/overview/_service.html
```diff
@@ -1,11 +1,10 @@
 <div ng-if="!tileCount" class="no-deployments-block">
-  <div column class="no-deployments-message">
+  <div column class="no-deployments-message overview-tile">
     <ng-include src="'views/overview/_service-header.html'"></ng-include>
     <div class="empty-tile">
       <h2>No deployments or pods.</h2>
       <p>
-        Service
-        <a ng-href="{{service | navigateResourceURL}}">{{service.metadata.name}}</a>
+        <div class="word-break-all">Service <a ng-href="{{service | navigateResourceURL}}">{{service.metadata.name}}</a></div>
```
spadgett: Isn't this going to add a line break in the middle of a sentence?

app/views/overview/_service-group.html
```diff
@@ -71,12 +71,10 @@ <h3 class="route-title truncate">
             <div class="no-child-services-message">
               <div class="empty-tile">
                 <h2>No grouped services.</h2>
-                <p>
-                  No services are grouped with <a ng-href="{{service | navigateResourceURL}}">{{service.metadata.name}}</a>.
+                  No services are grouped with <div class="word-break-all"><a ng-href="{{service | navigateResourceURL}}">{{service.metadata.name}}</a>.</div>
```
spadgett: Isn't this going to add a line break in the middle of a sentence?

app/views/overview/_dc.html
```diff
@@ -43,8 +43,8 @@
         A new deployment will start automatically when
         <span ng-if="imageChangeTriggers.length === 1">
           an image is available for
-          <a ng-href="{{urlForImageChangeTrigger(imageChangeTriggers[0], deploymentConfig)}}">
-            {{imageChangeTriggers[0].imageChangeParams.from | imageObjectRef : deploymentConfig.metadata.namespace}}</a>.
+          <div class="truncate mar-bottom-xl"><a ng-href="{{urlForImageChangeTrigger(imageChangeTriggers[0], deploymentConfig)}}">
```
spadgett: Isn't this going to add a line break in the middle of a sentence?

app/views/overview/_service.html
sg00dwin: yep, these will be fixed.

sg00dwin:
@spadgett PTAL

sg00dwin:
screen shot 2017-02-02 at 3 36 29 pm

spadgett:
This change breaks sentences mid-word in Firefox. Can we use `word-break` instead of `word-break-all`?

openshift_web_console

sg00dwin:
With just `<p class="word-break">...` There would be scenarios where the long string wouldn't break and cause the tile width to expand, causing it to bump down below it's sibling tile.

screen shot 2017-02-06 at 10 43 22 am

sg00dwin:
I'm looking further to see if we can incorporate [hyphens](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens) with our `word-break` rule

sg00dwin:
@spadgett  I have `.break-word` working  in those empty tiles, ftw!
Cross browser screenshots...

screen shot 2017-02-06 at 3 52 21 pm
screen shot 2017-02-06 at 3 51 43 pm
screen shot 2017-02-06 at 3 46 39 pm
screen shot 2017-02-06 at 3 44 57 pm

rhamilto:
#1205 is a related bug that can also be fixed in this PR.

spadgett:
@sg00dwin Approved, but needs rebase

sg00dwin:
@spadgett this is now rebased

sg00dwin:
@spadgett removed vendor.js

spadgett:
[merge]

PR #1193, opened 2017-01-30
Show builders and templates that match no subcategory

### LINKED ISSUE #1189: category system isn't properly showing custom java, has usability issues
Referencing the source:
https://github.com/openshift/origin-web-console/blob/c9474fc80ed7b5174e97ff168531384dccb049ba/app/scripts/constants.js#L283-L319

And the docs:
https://docs.openshift.org/latest/install_config/web_console_customization.html#configuring-catalog-categories

And the following `ImageStream` gist:
https://gist.github.com/thoraxe/23449ba2aa50fd32c4b21ff4a0017a24

* This Java IS does not appear natively in the Java category
* The docs describe how to add featured and additional language categories, but not how to add subcategories
* Java subcategories are hard-coded as opposed to being introspected by some kind of labeling or tagging system

Ultimately this adds up to a bit of a usability issue. As an example, customers may be using alternative Java middleware stacks (MuleSoft, Spring, etc) and there does not appear to be a way for them to add a category for their own solutions and then have those show underneath.

I can find my Java IS by searching, but it will not appear any way in the current OCP 3.4.0.40 console in the Add to Project dialog.

spadgett:
Fix a problem where image streams and templates that matched a category, but did not match any subcategories, were hidden on the add to project page.

Fixes #1189 

Unfiltered:

screen shot 2017-01-30 at 12 04 26 pm

Filtered:

screen shot 2017-01-30 at 12 04 09 pm

### DIFF
--- a/app/scripts/controllers/create/browseCategory.js
+++ b/app/scripts/controllers/create/browseCategory.js
@@ -15,7 +15,6 @@ angular.module('openshiftConsole')
                                                     $routeParams,
                                                     $uibModal,
                                                     AlertMessageService,
-                                                    CatalogService,
                                                     Constants,
                                                     DataService,
                                                     LabelFilter,
--- a/app/scripts/directives/catalog.js
+++ b/app/scripts/directives/catalog.js
@@ -48,6 +48,7 @@ angular.module('openshiftConsole')
             var categoryKeywords = _.reject(keywords, matchesCategoryLabel);
             $scope.filteredBuildersByCategory[id] = CatalogService.filterImageStreams(builders, categoryKeywords);
           });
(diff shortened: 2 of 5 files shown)


app/scripts/controllers/create/browseCategory.js
```diff
@@ -15,7 +15,6 @@ angular.module('openshiftConsole')
                                                     $routeParams,
                                                     $uibModal,
                                                     AlertMessageService,
-                                                    CatalogService,
```
spadgett: CatalogService was not used by the controller

jwforres:
[merge]

PR #1200, opened 2017-01-31
Add note for storage class used on PVC grid

zherman0:
We have a request to show which provisioner was used to create a claim.  I have added a muted note next to the PVC name when a storage class was assigned.

### DIFF
--- a/app/views/storage.html
+++ b/app/views/storage.html
@@ -55,7 +55,9 @@
                 </tbody>
                 <tbody ng-if="(pvcs | hashSize) > 0">
                   <tr ng-repeat="pvc in pvcs | orderObjectsByDate : true">
-                    <td data-title="Name"><a ng-href="{{pvc | navigateResourceURL}}">{{pvc.metadata.name}}</a></td>
+                    <td data-title="Name"><a ng-href="{{pvc | navigateResourceURL}}">{{pvc.metadata.name}}</a>
+                      <span ng-if="pvc | storageClass" class="text-muted"> using storage class {{pvc | storageClass}}</span>
+                    </td>
                     <td data-title="Status">
                       <status-icon status="pvc.status.phase" disable-animation></status-icon>
                       {{pvc.status.phase}}
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -98,6 +98,11 @@ angular.module('openshiftConsole')
              annotationFilter(resource, 'kubernetes.io/description');
     };
   })
+  .filter('storageClass', function(annotationFilter) {
+     return function(pvc) {
+       return annotationFilter(pvc, 'volume.beta.kubernetes.io/storage-class');
+     };
+  })
   .filter('displayName', function(annotationFilter) {
     // annotationOnly - if true, don't fall back to using metadata.name when
     //                  there's no displayName annotation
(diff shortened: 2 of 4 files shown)


zherman0:
storage_view_sc2

spadgett:
I think I'd prefer it as another table column. Easier to see at a glance.

spadgett:
cc @jwforres @erinboyd

erinboyd:
@spadgett @zherman0  I think a column might be confusing as not all PVCs will be associated with a Storage Class. And you really don't have a way to auto hide it as this view shows all PVCs.
Thoughts?

spadgett:
@erinboyd I don't feel that strongly. Will defer to you and @jwforres. We do have tables with columns that aren't always populated like routes.

openshift_web_console

jwforres:
I think the question boils down to how often we expect people to be using storage classes vs not.

app/views/storage.html
```diff
@@ -55,7 +55,9 @@
                 </tbody>
                 <tbody ng-if="(pvcs | hashSize) > 0">
                   <tr ng-repeat="pvc in pvcs | orderObjectsByDate : true">
-                    <td data-title="Name"><a ng-href="{{pvc | navigateResourceURL}}">{{pvc.metadata.name}}</a></td>
+                    <td data-title="Name"><a ng-href="{{pvc | navigateResourceURL}}">{{pvc.metadata.name}}</a>
+                      <span ng-if="pvc | annotation : 'volume.beta.kubernetes.io/storage-class'" class="text-muted"> using storage class {{pvc | annotation : 'volume.beta.kubernetes.io/storage-class'}}</span>
```
spadgett: If we expect the `beta` to eventually drop from the annotation name or storage class to ever be added as another API field, it's probably worth adding a `storageClass` filter. Then we can change it in one place (and fall back to looking for this annotation if needed). Something like we do for the description annotation.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/filters/resources.js#L94-L100

```js
  .filter('storageClass', function(annotationFilter) {
    return function(pvc) {
      return annotationFilter(pvc, 'volume.beta.kubernetes.io/storage-class');
    };
  })
```

Then here in the view

```html
 <span ng-if="pvc | storageClass" class="text-muted"> using storage class {{pvc | storageClass}}</span>
```

This also makes the markup more concise.

zherman0: I agree, this will make things much nicer and be easier to update in the futures.  Changes added.

zherman0:
@spadgett @erinboyd @jwforres  I think whether to add a column or not will come down to how much people use the storage classes.  Ideally, we would like them to be quickly adopted and every new claim is associated with a storage class.  But time will tell.  Let's leave it as is and we can very easily tweak it in the future.

erinboyd:
@jwforres @spadgett @zherman0 Ideally, yes, we would hope that the majority of customers will leverage a storage class..so I guess in that case a column seems appropriate

app/views/storage.html
erinboyd: I like that!

spadgett:
> Ideally, yes, we would hope that the majority of customers will leverage a storage class..so I guess in that case a column seems appropriate

I'm OK with the PR as-is if we want to merge, unless @erinboyd you'd rather see this change

erinboyd:
@spadgett LGTM

spadgett:
thanks @zherman0 

[merge]

erinboyd:
LGTM

PR #1201, opened 2017-01-31
Let users reset "Don't Show Me Again" alerts

spadgett:
Add a link to reset hidden alerts on the about page. Lets users reset alerts they might need to see again like those from #1194

Since this resets notifications for all projects and we don't have a settings page, I've added it to the about page. Open to other suggestions.

openshift_web_console

cc @ajacobs21e @lizsurette @ncameronbritt

### DIFF
--- a/app/scripts/controllers/about.js
+++ b/app/scripts/controllers/about.js
@@ -7,13 +7,26 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('AboutController', function ($scope, AuthService, Constants) {
+  .controller('AboutController',
+              function($scope,
+                       AlertMessageService,
+                       AuthService,
+                       Constants) {
     AuthService.withUser();
-    
+
+    $scope.alerts = {};
     $scope.version = {
       master: {
         openshift: Constants.VERSION.openshift,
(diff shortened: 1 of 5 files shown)


jwforres:
I would expect this to be more along the lines of a user preferences page.  I hesitate to shove this onto the About page just because we have nowhere else for it to live currently.

cc @ajacobs21e @lizsurette

spadgett:
Fair enough. Do you think it's worth adding a setting page with one option, though?

I thought about the user dropdown as well, although it makes showing an alert message difficult without toast notifications.

spadgett:
Or maybe there are a few simple settings we can add for existing things to justify a new page?

lizsurette:
@spadgett One idea is to add a "Settings" option with just a fly out with one option of "Reset Alerts". It could be a temporary solution until you have enough settings to justify a new page.

lizsurette:
@spadgett Trying to think of other things that might be nice to have on a settings page...

-Date/time format
-Time before login times out for security

I think I'm just reaching and creating unnecessary requirements 👎

ajacobs21e:
@spadgett @jwforres does a setting like this make more sense on the project list page?

PR #1202, opened 2017-01-31
Don't link to image stream tags that haven't synced yet

### LINKED ISSUE #1199: Remove tag link from image stream page when unable to resolve
We shouldn't show the tag as a link when we know it will 404.

cc @stefwalter 

openshift_web_console

spadgett:
Don't link to image stream tags that haven't synced or had an error syncing. The page will give a not found error.

Fixes #1199


jwforres:
[merge]

Issue #1182, opened 2017-01-26
Inline empty resource message


jhadvig:
Inline empty resource message instead of using a scope variables since they don't change afterwards.
Eg: https://github.com/openshift/origin-web-console/blob/master/app/views/browse/config-maps.html#L48

PR #1203, opened 2017-02-01
Remove emptyMessage var if its not changing

jhadvig:
Found out that there are place in the code where we use `emptyMessage` var as part of the scope, where we just set it to "No $resource to show" and thats it. So I've got rid of that and use the scoped `loaded` variable to check if the resources are loaded since we do it on most parts of the code.
Other option would be to get rid even of the `loaded` var and check only the resource variable, which wont be set by default in the controller... eg:
```
<div ng-if="!services">Loading...</div>
<div class="row" ng-if="services">
     <tbody ng-if="(services | hashSize) == 0">
         <tr><td colspan="4"><em>No image streams to show</em></td></tr>
     </tbody>
     <tbody ng-if="(imageStreams | hashSize) > 0">
```

Also removed `emptyMessage` var on places we didn't even used that variable in views.

Will add `dist` after review

@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/1182

### DIFF
--- a/app/scripts/directives/resources.js
+++ b/app/scripts/directives/resources.js
@@ -182,8 +182,6 @@ angular.module('openshiftConsole')
         services: '=',
         portsByRoute: '=',
         showNodePorts: '=?',
-        // Optional empty message to display when there are no pods.
-        emptyMessage: '=?',
         // Alternative header text to display in the 'Name' column.
         customNameHeader: '=?',
       },
--- a/app/scripts/controllers/create/createFromImage.js
+++ b/app/scripts/controllers/create/createFromImage.js
@@ -86,7 +86,6 @@ angular.module("openshiftConsole")
         }
         function initAndValidate(scope){
 
-          scope.emptyMessage = "Loading...";
           scope.name = $routeParams.name;
           scope.imageName = imageName;
           scope.imageTag = $routeParams.imageTag;
--- a/app/scripts/controllers/create/nextSteps.js
+++ b/app/scripts/controllers/create/nextSteps.js
@@ -13,7 +13,6 @@ angular.module("openshiftConsole")
     var displayNameFilter = $filter('displayName');
     var watches = [];
 
-    $scope.emptyMessage = "Loading...";
     $scope.alerts = [];
     $scope.loginBaseUrl = DataService.openshiftAPIBaseUrl();
     $scope.buildConfigs = {};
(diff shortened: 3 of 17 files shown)


spadgett:
@jhadvig I think the places we should make this change are

* When the message doesn't switch from "Loading..." to "No resources to show" (`$scope.emtpyMessage` is set once and never changed)
* When we want different styles for empty states like #497

Hesitant to change a lot working code if it's not for one of the two above reasons.

jhadvig:
@spadgett ok, will update the PR so the changes only tackle mentioned reasons.
THanks !

spadgett:
So we do have the other #497 for empty states. I think for now we should probably only look at places when `$scope.emptyMessage` is set once and never changed.

jhadvig:
@spadgett PTAL

app/scripts/directives/resources.js
```diff
@@ -176,8 +176,6 @@ angular.module('openshiftConsole')
         pods: '=',
         // Optional active pods map to display whether or not pods have endpoints
         activePods: '=?',
-        // Optional empty message to display when there are no pods.
-        emptyMessage: '=?',
```
spadgett: This breaks the loading message on the browse -> pods page.

app/scripts/directives/resources.js
```diff
@@ -194,8 +192,6 @@ angular.module('openshiftConsole')
         services: '=',
         portsByRoute: '=',
         showNodePorts: '=?',
-        // Optional empty message to display when there are no pods.
-        emptyMessage: '=?',
```
spadgett: Need to update `traffic-table.html`

https://github.com/spadgett/origin-web-console/blob/859a4cb2689c7bd8b5ef1a1c17ce8efcf267c758/app/views/directives/traffic-table.html#L15

spadgett:
@jhadvig Thanks for removing some completely unused variables

jhadvig:
@spadgett updated .. PTAL :)

spadgett:
Thanks @jhadvig, will merge when the stream opens for 1.6

jwforres:
@jhadvig needs to be rebased

jwforres:
@jhadvig actually this PR shouldn't have been touching vendor.js at all, I'm not seeing a reason for it to, make sure you clean/install deps/build after you rebase

spadgett:
> actually this PR shouldn't have been touching vendor.js at all

It was from when we were pulling in the bootstrap-switch dependency update. Should disappear on the clean/install

jhadvig:
> It was from when we were pulling in the bootstrap-switch dependency update. Should disappear on the clean/install

yep

@jwforres @spadgett I've updated the PR, should be fine now

spadgett:
[merge]

PR #1213, opened 2017-02-01
Add namespace picker to service account tab on membership page

### LINKED ISSUE #850: Adding service account a little confusing
The project dropdown is in the roles column, so it looks like I'm adding role `oc`. Also the other names have a prefix, so it looks like you should type "oc/my-name" in the input, which is an invalid name. I think we at least need labels for the inputs since they don't correspond to the column headings, or maybe move them. Placeholder text in the name input wouldn't hurt.

openshift_web_console

cc @benjaminapetersen @jwforres @ajacobs21e

### LINKED ISSUE #1191: Can not add a role to "default" serviceaccount through the console
When you have a project and you want to add a role to the default serviceaccount, this one does not show in the console, so it's not possible to add an additional role, like "view".
image

But the service account is there in the system, as "default" service account is always there even if you delete it, it will get recreated by openshift.
image

benjaminapetersen:
Fix #1191
Fix #850 

@jwforres @spadgett

screen shot 2017-02-01 at 2 14 05 pm

mobile:

screen shot 2017-02-01 at 2 13 53 pm

### DIFF
--- a/app/scripts/controllers/membership.js
+++ b/app/scripts/controllers/membership.js
@@ -70,6 +70,24 @@ angular
         $scope.newBinding.newRole = null;
       };
 
+      var refreshServiceAccountsList = function(ctx) {
+        DataService
+          .list('serviceaccounts', ctx)
+          .then(function(resp) {
+            var serviceAccounts = _.keys(resp.by('metadata.name')).sort();
+            angular.extend($scope, {
+              serviceAccounts: serviceAccounts,
+              refreshServiceAccounts: function(search) {
+                if(search && !_.includes($scope.serviceAccounts, search)) {
+                  $scope.serviceAccounts = [search].concat(serviceAccounts);
+                } else {
+                  $scope.serviceAccounts = serviceAccounts;
+                }
+              }
+            });
+          });
+      };
+
       var refreshRoleBindingList = function(toUpdateOnError) {
         DataService
           .list('rolebindings', requestContext, null , {
@@ -159,6 +177,7 @@ angular
         },
         selectTab: function(selected) {
           $scope.newBinding.kind = selected;
+          $scope.newBinding.name = '';
         }
       });
 
@@ -219,15 +238,18 @@ angular
           $scope.user = resp;
         });
 
-
       DataService
         .list('projects', {})
         .then(function(resp) {
-          var projects = _.map(resp.by('metadata.name'), function(project) {
-            return project.metadata.name;
-          });
+          var projects = _.keys(resp.by('metadata.name')).sort();
           angular.extend($scope, {
             projects: projects,
+            selectProject: function(projectName) {
+              $scope.newBinding.name = '';
+              refreshServiceAccountsList({
+                namespace: projectName
+              });
+            },
             refreshProjects: function(search) {
               if(search && !_.includes($scope.projects, search)) {
                 $scope.projects = [search].concat(projects);
@@ -243,6 +265,7 @@ angular
         .then(_.spread(function(project, context) {
           requestContext = context;
           refreshRoleBindingList();
+          refreshServiceAccountsList(requestContext);
           angular.extend($scope, {
             project: project,
             subjectKinds: subjectKinds,
@@ -364,5 +387,6 @@ angular
                 }
               });
             });
+
         }));
     });
--- a/app/views/membership.html
+++ b/app/views/membership.html
@@ -167,13 +167,16 @@
                   <div
                     ng-if="mode.edit"
                     class="item-row form-new-role" row mobile="column">
-                    <div class="col-name" column>
+                    <div class="col-name pad-bottom-none" row mobile="column" tablet="column">
                       <label
                         ng-attr-for="newBindingName"
                         class="sr-only">
                         Name
                       </label>
+
+                      <!-- name entry field for everything but service accounts -->
                       <input
+                        ng-if="newBinding.kind !== 'ServiceAccount'"
                         type="text"
                         class="form-control input-name"
                         placeholder="Name"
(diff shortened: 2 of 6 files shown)


benjaminapetersen:
Is it necessary to watch service accounts rather than list? I don't know how common it is to manipulate them.

spadgett:
Suggest making the selects

```
[my-project       v] / [service-account v]
```

mirroring how we display them above. We do something similar with image streams. Should be able to take up two columns since there will be no assigned roles, right?

app/scripts/controllers/membership.js
```diff
@@ -364,5 +363,25 @@ angular
...
               });
             });
+
+
+          DataService
+            .list('serviceaccounts', requestContext)
+            .then(function(resp) {
+              var serviceAccounts = _.map(resp.by('metadata.name'), function(sa) {
+                return sa.metadata.name;
+              });
```
spadgett: Maybe

```js
var serviceAccountNames = _.keys(resp.by('metadata.name'));
```

spadgett: Should we remove service accounts from the list that already have roles above?

app/scripts/controllers/membership.js
```diff
@@ -364,5 +363,25 @@ angular
                 }
               });
             });
+
+
+          DataService
+            .list('serviceaccounts', requestContext)
```
spadgett: Don't we need to be getting service accounts from the selected namespace? Won't this only get the service accounts from the current namespace?

app/views/membership.html
```diff
@@ -173,14 +173,42 @@
...
                         class="form-control input-name"
                         placeholder="Name"
                         ng-model="newBinding.name">
+
+                      <!-- name entry/picker for service accounts -->
                       <div
                         ng-if="newBinding.kind === 'ServiceAccount'"
-                        class="service-account-namespace hidden-sm hidden-md hidden-lg"
+                        class="service-account-name pad-bottom-sm"
+                        aria-hidden="true">
```
spadgett: Why is this aria-hidden?

app/views/membership.html
```diff
@@ -207,27 +235,7 @@
...
-                            </ui-select-match>
-                            <ui-select-choices
-                              repeat="projectName in projects | filter: $select.search"
-                              refresh="refreshProjects($select.search)"
-                              refresh-delay="200">
-                              <div>{{projectName}}</div>
-                            </ui-select-choices>
-                          </ui-select>
-                        </div>
+                        &nbsp;
```
spadgett: Why the non-breaking space? If we need whitespace, we should use CSS

app/scripts/controllers/membership.js
benjaminapetersen: Yeah thats helpful.  I have the ability to enter an unlisted, but should def request & update whenever the namespace changes.

app/scripts/controllers/membership.js
benjaminapetersen: Doesn't matter to me, the update will put the 2nd role on the SA either way.  Prefer to eliminate it from the dropdown?

app/views/membership.html
benjaminapetersen: Good catch, I need to remove that tag now as I previously had 2 `<ui-select>` for namespace, but now only have 1.

app/views/membership.html
benjaminapetersen: Yeah, to keep the flex from collapsing completely, habit, I think its irrelevant, I'll remove & the extra axis stuff.

benjaminapetersen:
@spadget on this:
```
[my-project       v] / [service-account v]
```
& your comments on #850, I like the 2 columns side-by-side, but it will put the 2nd select back under the 'Roles' heading... any concerns?

app/scripts/controllers/membership.js
spadgett: > Prefer to eliminate it from the dropdown?

I think so to avoid confusion. Probably want to sort as well.

```js
var serviceAccountNames = _.keys(resp.by('metadata.name')).sort();
```

benjaminapetersen:
WIP, updated to refresh the service accounts list to match the selected project:
screen shot 2017-02-02 at 5 12 47 pm

jwforres:
this is looking way better to me

benjaminapetersen:
Tablet / mobile:
screen shot 2017-02-03 at 11 15 09 am
screen shot 2017-02-03 at 11 15 30 am

There is a spot between tablet & larger screens where multiple roles added to a user/SA can cause some layout wobble.  Is unrelated to this change, so I'll create an issue & get to work on tweaking that.

@spadgett ready for another pass

app/scripts/controllers/membership.js
```diff
@@ -70,6 +70,24 @@ angular
         $scope.newBinding.newRole = null;
       };
 
+      var refreshServiceAccountsList = function(ctx) {
+        DataService
+          .list('serviceaccounts', ctx)
+          .then(function(resp) {
+            var serviceAccounts = _.keys(resp.by('metadata.name'));
```
spadgett: Maybe

```js
_.keys(resp.by('metadata.name')).sort();
```

so the list is sorted?

app/scripts/controllers/membership.js
```diff
@@ -219,15 +237,17 @@ angular
...
         });
 
-
       DataService
         .list('projects', {})
         .then(function(resp) {
-          var projects = _.map(resp.by('metadata.name'), function(project) {
-            return project.metadata.name;
-          });
+          var projects = _.keys(resp.by('metadata.name'));
```
spadgett:
```js
var projects = _.keys(resp.by('metadata.name')).sort();
```

?

app/views/membership.html
```diff
@@ -203,35 +208,38 @@
...
+                          class="select-role">
+                          <ui-select-match
+                            placeholder="Service account">
+                            <span ng-bind="newBinding.name"></span>
+                          </ui-select-match>
+                          <ui-select-choices
+                            repeat="saName in serviceAccounts | filter: $select.search"
+                            refresh="refreshServiceAccounts($select.search)"
+                            refresh-delay="200">
+                            <div>{{saName}}</div>
```
spadgett:
```html
<div ng-bind-html="saName | highlight : $select.search"></div>
```

app/views/membership.html
```diff
@@ -167,30 +167,35 @@
...
                       <input
+                        ng-if="newBinding.kind !== 'ServiceAccount'"
                         type="text"
                         class="form-control input-name"
                         placeholder="Name"
                         ng-model="newBinding.name"
                         autocorrect="off"
                         autocapitalize="off"
                         spellcheck="false">
+	  				  <!-- namespace entry/picker for service accounts -->
```
spadgett: indentation

app/views/membership.html
spadgett: Probably should add this to the project select above, too.

app/views/membership.html
```diff
@@ -203,35 +208,38 @@
...
+                        class="service-account-name">
+                        <ui-select
+                          ng-model="newBinding.name"
+                          theme="bootstrap"
+                          search-enabled="true"
+                          title="Select a service account"
+                          class="select-role">
+                          <ui-select-match
+                            placeholder="Service account">
+                            <span ng-bind="newBinding.name"></span>
```
spadgett: Any reason to prefer `ng-bind` instead of just using the curly braces for expressions?

spadgett:
Nit: In the very last screenshot, the project/sa selects should probably fill the width like the role select.

app/scripts/controllers/membership.js
benjaminapetersen: yeah sounds good.

app/scripts/controllers/membership.js
benjaminapetersen: good catch, will update!

app/views/membership.html
benjaminapetersen: nice, ill add.

app/views/membership.html
benjaminapetersen: I think I have that because it was part of `ui-select` docs when i first dropped them in, I'll swap to the `{{}}`.

app/views/membership.html
benjaminapetersen: got it.

benjaminapetersen:
Agree, this looks better:
screen shot 2017-02-03 at 2 01 29 pm

benjaminapetersen:
updated & rebased.

spadgett:
I noticed when I select a service account then immediately switch back to the users tab, it pre-fills the service account name in the user input. Same if I enter a user and switch to service accounts.

benjaminapetersen:
Hmm. Also a bit of a flicker going back and forth between the SA tab & others as the input appears/disappears.

benjaminapetersen:
Updated to clear the name when tabs are changed

spadgett:
Thanks @benjaminapetersen 

[merge]

PR #1216, opened 2017-02-02
Set larger label-filter default input click targets

sg00dwin:
Set reasonable widths on the default input

screen shot 2017-02-02 at 5 23 08 pm
screen shot 2017-02-02 at 5 18 40 pm

screen shot 2017-02-02 at 5 17 34 pm

### DIFF
--- a/app/styles/_data-toolbar.less
+++ b/app/styles/_data-toolbar.less
@@ -1,6 +1,13 @@
 /* Data toolbar
    - Includes customizations for label filter input and active filter
 ---------------------------------------------------------------------------- */
+// @screen-md-min - (sidebar-left + left + right margin); 992 - (145 + 30 + 30) = 787
+@data-toolbar-filter-max-width:   (@screen-md-min - (@sidebar-left-width-md + @middle-content-container-padding-lg + @middle-content-container-padding-lg));
+@label-filter-input-width-lg: 730px; // max width available for input at > 992
+@label-filter-input-width-md: 540px; // max width available for input at > 768
+@label-filter-input-width-sm: 400px; // max width available for input at > 480
+@label-filter-input-width-xs: 240px; // 320 - (40 + 40); container - ((left + right margin) + button)
+
 
 .data-toolbar {
   padding: 5px 0;
@@ -26,12 +33,19 @@
       .flex(@columns: 1 1 0%);
     }
     @media (min-width: @screen-md-min) {
-      max-width: 600px;
+      max-width: @data-toolbar-filter-max-width;
     }
     .filter .navbar-filter-widget .label-filter .selectize-control .selectize-input {
       &, & input {
         font-size: @font-size-base;
       }
+    }
+    .label-filter-key .selectize-input.not-full {
+      // Increase default input click target
+      width: @label-filter-input-width-xs;
+      @media (min-width: @screen-xs-min) { width: @label-filter-input-width-sm;}
+      @media (min-width: @screen-sm-min) { width: @label-filter-input-width-md;}
+      @media (min-width: @screen-md-min) { width: @label-filter-input-width-lg;}
     }
     .form-group {
       margin-bottom: 0;
--- a/dist/styles/main.css
+++ b/dist/styles/main.css
@@ -4137,9 +4137,16 @@ copy-to-clipboard .input-group.limit-width{max-width:300px}
 .data-toolbar{padding:5px 0}
 .data-toolbar.other-resources-toolbar .data-toolbar-dropdown{min-width:210px}
 .data-toolbar .checkbox{margin-bottom:0;margin-top:10px}
+.data-toolbar .data-toolbar-filter .label-filter-key .selectize-input.not-full{width:240px}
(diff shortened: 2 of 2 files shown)


jwforres:
what happens after you have filled out the first box? it goes back to only the width it needs?

sg00dwin:
Yes. It only targets the initial default state for the label key `.label-filter-key .selectize-input.not-full`

app/styles/_data-toolbar.less
```diff
@@ -33,6 +33,12 @@
         font-size: @font-size-base;
       }
     }
+    .label-filter-key .selectize-input.not-full {
+      // Increase default input click target
+      width: 220px;
+      @media (min-width: @screen-xs) and (max-width: @screen-sm-max) { width: (@screen-xs - 100)} // 380px
```
spadgett: I don't think you need the `max-width` here since you have a rule just below that sets the width for `screen-md` and larger

app/styles/_data-toolbar.less
```diff
@@ -33,6 +33,12 @@
...
+      @media (min-width: @screen-md) { width: (@screen-xs + 20)} // 500px
```
spadgett: I'm not sure how the width relates to `screen-xs` here. You set `max-width: 600px;` of the entire control above, so I'd think you would make width relative to that.

sg00dwin:
@spadgett Updated pr
At the standard media queries I set the input width to fill the data-toolbar-filter. 
Since the `data-toolbar-filter max-width:600px` was initially an arbitrarily created width. I decided to make `data-toolbar-filter` width match that of its container at the 992 media query break point.
832 - 60 = 772; container - (left + right margin)


screen shot 2017-02-05 at 9 33 46 pm
screen shot 2017-02-05 at 9 26 20 pm
screen shot 2017-02-06 at 9 59 12 am
screen shot 2017-02-05 at 8 20 45 pm
screen shot 2017-02-05 at 8 20 23 pm

sg00dwin:
[test]

app/styles/_data-toolbar.less
```diff
@@ -1,6 +1,12 @@
 /* Data toolbar
    - Includes customizations for label filter input and active filter
 ---------------------------------------------------------------------------- */
+@data-toolbar-filter-max-width:   772px; // 832 - 60; container - (left + right margin) at 992px viewport width
```
spadgett: Thanks for making these variables.

Unclear where the 832 comes from. Would be better to calculate from variables like `@grid-gutter-width` and `@screen-md` etc where possible.

Same comment for some of the magic numbers below.

app/styles/_data-toolbar.less
```diff
@@ -26,13 +32,20 @@
...
       &, & input {
         font-size: @font-size-base;
       }
     }
+    .label-filter-key .selectize-input.not-full {
+      // Increase default input click target
+      width: @label-filter-default-input-xs;
+      @media (min-width: @screen-xs) { width: @label-filter-default-input-sm;}
+      @media (min-width: @screen-sm) { width: @label-filter-default-input-md;}
+      @media (min-width: @screen-md) { width: @label-filter-default-input-lg;}
```
spadgett: Shouldn't these media queries use `@screen-xs-min`, `@screen-sm-min` and `@screen-md-min` ?

sg00dwin:
@spadgett Yep, updated the media query min variable. And for the `data-toolbar-filter` set it max width using the following.
@screen-md-min - (sidebar-left + left + right margin); 992 - (145 + 30 + 30) = 787

sg00dwin:
@spadgett did a clean install and updated pr

app/styles/_data-toolbar.less
```diff
@@ -1,6 +1,13 @@
 /* Data toolbar
    - Includes customizations for label filter input and active filter
 ---------------------------------------------------------------------------- */
+// @screen-md-min - (sidebar-left + left + right margin); 992 - (145 + 30 + 30) = 787
+@data-toolbar-filter-max-width:   (@screen-md-min - (@sidebar-left-width-md + @middle-content-container-padding-lg + @middle-content-container-padding-lg));
+@label-filter-default-input-lg: (@label-filter-default-input-md + 190); // 730px
+@label-filter-default-input-md: (@label-filter-default-input-sm + 140); // 540px
+@label-filter-default-input-sm: (@label-filter-default-input-xs + 160); // 400px
```
spadgett: @sg00dwin I don't think adding arbitrary amounts to the previous width variables here adds much value. I'd rather just set pixel sizes for each.

```less
@label-filter-default-input-xs: 240px;
@label-filter-default-input-sm: 400px;
@label-filter-default-input-md: 540px;
@label-filter-default-input-lg: 730px;
```

Or even inline the values if they're only used once.

spadgett: If you keep the variables names, maybe use `label-filter-input-width-*`? `-width` should be in there somewhere.

sg00dwin:
@spadgett updated pr with rebase and clean/install

sg00dwin:
@spadgett PTAL vendor.js diff is removed

spadgett:
[merge]

Issue #1221, opened 2017-02-06
Console say dc change saved but on review no change was saved


magick93:
OpenShift Master: v1.4.1+3f9807a
Kubernetes Master: v1.4.0+776c994

In an attempt to workaround https://github.com/fabric8io/jenkins-docker/issues/110 we try to edit Jenkins Deployment Config in the OpenShift console. 
We navigate to `Deployments  >  jenkins,` then click `Actions > Edit YAML,` change:

```
      serviceAccountName: jenkins
      serviceAccount: jenkins
      securityContext: {}
```

to (as described on https://kubernetes.io/docs/user-guide/security-context/)

```
      serviceAccountName: jenkins
      serviceAccount: jenkins
      securityContext:
        privileged: true
```

then save, and see
![selection_667](https://cloud.githubusercontent.com/assets/2173213/22647068/b80a22ae-ec6f-11e6-85a3-6fccf016cb3c.png)

Yet when we again click on Actions > Edit YAML, the change we added is gone.

Additional
When I try this in vim I get the same result. The change is not saved. But no indication of error.
```
$ oc edit dc jenkins
deploymentconfig "jenkins" skipped
```

spadgett:
It looks like you're changing the pod security context, which doesn't have a `privileged` field. The container security context does. Currently, OpenShift ignores unrecognized fields when editing YAML, so there is no error when you save (and why you see the skipped message when using the CLI).

Can you try changing this one instead and see if it works?

![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/22648364/530d67ce-ec44-11e6-90a4-145233224d08.png)

spadgett:
@jwforres We might be able to warn about this case. If the resource version in the PUT response body is the same as before, no changes were applied. @liggitt can you confirm?

magick93:
Im trying to change the security context for the containers, as described on https://kubernetes.io/docs/user-guide/security-context/

```
    spec:
      containers:
        - name: fabric8-container
          image: 'fabric8/fabric8-console:2.2.199'
          ports:
            - name: http
              containerPort: 9090
              protocol: TCP
          env:
            - name: OAUTH_AUTHORIZE_URI
              valueFrom:
                configMapKeyRef:
                  name: fabric8
                  key: oauth-authorize-uri
            - name: GITHUB_OAUTH_CLIENT_SECRET
              valueFrom:
                configMapKeyRef:
                  name: fabric8
                  key: github-oauth-client-secret
[...]

PR #1222, opened 2017-02-06
Warn when editing YAML did not change the object

spadgett:
Show a warning and stay on the edit YAML page when no edits were applied. This can happen when trying to add a property that is not recognized or when misspelling a field name.

See #1221

### DIFF
--- a/app/scripts/controllers/edit/yaml.js
+++ b/app/scripts/controllers/edit/yaml.js
@@ -26,6 +26,7 @@ angular.module('openshiftConsole')
     }
 
     var humanizeKind = $filter('humanizeKind');
+    $scope.alerts = {};
     $scope.name = $routeParams.name;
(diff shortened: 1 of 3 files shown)


jwforres:
what happens if someone modifies something else in etcd between the time Edit YAML is opened and when you saved?  resourceVersion isn't per object

spadgett:
Just tested: it still works.

We're checking that the resource version hasn't changed at all, which seems to be the case always if I add an unrecognized property.

spadgett:
@jwforres travis is complaining about dist, and I'm not sure why

jwforres:
cleared the caches and told it to rebuild, if that doesn't work maybe try cleaning your deps?  its vendor.js its complaining about, maybe some dependency has changed.

spadgett:
Updated to fix dist

jwforres:
[merge]

PR #1223, opened 2017-02-06
Don't show failed pods in pod donut

spadgett:
Prevent evicted and other failed pods from showing up in the pod donut.

cc @derekwaynecarr


spadgett:
[merge]

PR #1226, opened 2017-02-08
Remove old routes watcher in the Services controller

jhadvig:
... since we are not using it in the view
@benjaminapetersen @spadgett PTAL


spadgett:
[merge]

PR #1229, opened 2017-02-08
Fix membership addRoleToUser() fn, check namespace w/subject

benjaminapetersen:
Fix bugzilla #1420247

Will now allow you to add roles to service accounts with the same name that exist in different projects, oversight from last update.

@jwforres @spadgett


jwforres:
[merge]

Issue #1234, opened 2017-02-09
Extra whitespace around access mode radio buttons on Create Storage page


jwforres:
There is extra whitespace both above and below the radio buttons for the Access Mode when compared with the whitespace around the textboxes for the other inputs.

![access-mode-whitespace](https://cloud.githubusercontent.com/assets/1467629/22797414/21a2cf0a-eecc-11e6-8598-abd3673fec09.png)

jwforres:
cc @zherman0

PR #1239, opened 2017-02-09
Removing an inappropriate class used to wrap inline radio form controls.

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/1234

Removed the `.radio` class from two places that use `radio-inline` controls which in not needed and adds excessive top and bottom margin.

screen shot 2017-02-09 at 3 52 31 pm
screen shot 2017-02-09 at 3 44 53 pm

### DIFF
--- a/app/styles/_forms.less
+++ b/app/styles/_forms.less
@@ -86,3 +86,12 @@
   // the sections just above.
   margin-top: @line-height-computed;
 }
+
+// Switch default bootstrap margin from left to the right side to enable left alignment of inputs at mobile
+.radio-inline,
+.radio-inline + .radio-inline,
+.checkbox-inline,
+.checkbox-inline + .checkbox-inline {
+  margin-left: 0;
+  margin-right: 10px;
+}
--- a/app/views/directives/edit-lifecycle-hook.html
+++ b/app/views/directives/edit-lifecycle-hook.html
@@ -10,24 +10,24 @@
 
       <div class="form-group">
         <label for="actionType" class="required">Lifecycle Action</label><br/>
-        <div class="radio">
-          <label class="radio-inline">
-            <input type="radio"
-                    name="{{type}}-action-newpod"
-                    ng-model="action.type"
-                    value="execNewPod"
-                    aria-describedby="action-help">
-            Run a specific command in a new pod
-          </label>
-          <label class="radio-inline">
-            <input type="radio"
-                  name="{{type}}-action-images"
-                  ng-model="action.type"
-                  value="tagImages"
-                  aria-describedby="action-help">
-              Tag image if the deployment succeeds
-          </label>
-        </div>
+        <label class="radio-inline">
+          <input
+            type="radio"
+            name="{{type}}-action-newpod"
+            ng-model="action.type"
+            value="execNewPod"
+            aria-describedby="action-help">
+          Run a specific command in a new pod
+        </label>
+        <label class="radio-inline">
+          <input
+            type="radio"
+            name="{{type}}-action-images"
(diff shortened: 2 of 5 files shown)


app/views/directives/edit-lifecycle-hook.html
```diff
@@ -10,7 +10,7 @@
 
       <div class="form-group">
         <label for="actionType" class="required">Lifecycle Action</label><br/>
-        <div class="radio">
+        <div>
```
spadgett: Can you remove the unnecessary div?

spadgett:
+1 to removing `class="radio"`. It looks like that's not mean to be used with `radio-inline`

sg00dwin:
I use the div as a parent wrap for the radio-inline inputs, so that aren't displayed inline with the main `.form-group` label "Access Mode" and "Lifecycle Action".  Also to address the left alignment issue https://github.com/openshift/origin-web-console/issues/1240 at mobile we need to remove the bootstrap default left margin on siblings, so < 768 I switched it to the right side, so that they still space apart while inline, but align left when stacked.

screen shot 2017-02-10 at 10 40 21 am

screen shot 2017-02-10 at 10 41 18 am

app/styles/_forms.less
```diff
@@ -86,3 +86,16 @@
...
+// Wrap for inline radio and/or checkbox groups that are displayed inline.
+.radio-checkbox-group-inline {
+  display: block;
+  @media (max-width: @screen-xs-max) {
+    // switch default bootstrap margin to right side to enable left alignment of inputs at mobile
+    .radio-inline,
+    .checkbox-inline {
+      margin-left: 0;
+      margin-right: 10px;
+    }
```
spadgett: Any reason not to do this always (instead of max-width xs)? It seems like it could still be a problem with long labels at any width.

Would this work without needing the parent div `radio-checkbox-group-inline`?

```css
.radio-inline,
.checkbox-inline {
  margin-left: 0;
  margin-right: 10px;
}
```

spadgett: We only use it on two pages, so should be easy to test the change.

```
origin-web-console (master) ➜  ag radio-inline
app/views/directives/edit-lifecycle-hook.html
14:          <label class="radio-inline">
22:          <label class="radio-inline">

app/views/directives/osc-persistent-volume-claim.html
95:        <label class="radio-inline">
100:        <label class="radio-inline">
105:        <label class="radio-inline">
origin-web-console (master) ➜  ag checkbox-inline
origin-web-console (master) ➜
```

sg00dwin:
Ok, updated so that it always applies and it works fine as well. 
Needs to include the adjacent rule since that's what bootstrap uses. 
https://github.com/openshift/origin-web-console/blob/master/dist/styles/main.css#L708

spadgett:
Thanks @sg00dwin , can you fix the indentation since you took out the parent div?

sg00dwin:
Indentation corrected

app/views/directives/edit-lifecycle-hook.html
```diff
@@ -10,24 +10,22 @@
...
-                  aria-describedby="action-help">
-              Tag image if the deployment succeeds
-          </label>
-        </div>
+        <label class="radio-inline">
+          <input type="radio"
+          name="{{type}}-action-newpod"
+          ng-model="action.type"
+          value="execNewPod"
+          aria-describedby="action-help">
```
spadgett: Sorry, I'm picking nits, but can you indent the attributes as before? I find it harder to read this way.

jwforres:
[merge]

PR #1245, opened 2017-02-10
Add Scopes service to provide scope defaults to controllers

benjaminapetersen:
A possible setup towards reducing some of the copy/paste? Example w/a couple files.

@jwforres @spadgett

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -214,6 +214,7 @@
         <script src="scripts/services/cliHelp.js"></script>
         <script src="scripts/services/html.js"></script>
         <script src="scripts/services/environment.js"></script>
+        <script src="scripts/services/scopes.js"></script>
         <script src="scripts/controllers/projects.js"></script>
         <script src="scripts/controllers/pods.js"></script>
         <script src="scripts/controllers/pod.js"></script>
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -18,14 +18,12 @@ angular.module('openshiftConsole')
                                                  LabelFilter,
                                                  ModalsService,
                                                  ProjectsService,
-                                                 keyValueEditorUtils) {
+                                                 keyValueEditorUtils,
+                                                 Scopes) {
+    angular.extend($scope, Scopes.getControllerDefaults());
     $scope.projectName = $routeParams.project;
     $scope.buildConfigName = $routeParams.buildconfig;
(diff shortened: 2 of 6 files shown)


spadgett:
Agree the duplicate code is a problem, but I'm hesitant to pollute scopes with variables that might not be used. In fact, I'm not sure any of these variables is used in every controller.

benjaminapetersen:
I'm sure they are not all used in all controllers.  That said, I think scope variables, if named:

- should mean the same thing on every page & not drift
- should therefore be able to be safely ignored
- should, by default, be in one place so its easy to reference existing options
- should help us create naming patterns

IE, its good to reduce the cognitive load of each person having to know all the details and/or go fishing to find things.

spadgett:
> each person having to know all the details and/or go fishing to find things.

If the service could set meaningful values, I'd agree. But you have to actually set the values yourself, so you have to know the details. It **increases** the need to fish because now the you're setting values to scope variables that are initialized in a service somewhere else entirely. For instance,

```js
angular.extend($scope, Scopes.getControllerDefaults());
$scope.breadcrumbs.push({ ... });
```

Now I have to know that `breadcrumbs` was initialized to an array by this service. Is that really less cognitive load than this?

```js
$scope.breadcrumbs = [{ ... }];
```

I'd only prefer the former if there was some initial breadcrumbs that were used in all controllers that `Scopes` service could initialize. But there aren't. Even in that case, I'd rather have something like `BreadcrumbsService.initializeBreadcrumbs()` to set the common values.

It just feels weird to have services initialize scope variables. Services shouldn't know anything about scope variables being used by controllers and views.

Something like this might be useful:

```js
$scope.alerts = AlertsMessage.initializeAlerts();
```

To replace this copypasta controller logic:

```js
    $scope.alerts = $scope.alerts || {};
    AlertMessageService.getAlerts().forEach(function(alert) {
      $scope.alerts[alert.name] = alert.data;
    });
```

This is the kind of stuff that is really problematic and we should target. I don't see a lot of value in initializing scope variables that may or may not be used to `false` or `[]`.

benjaminapetersen:
Agree, though the Scopes service isn't actually doing anything to scope, its returning an object that then gets merged into scope (don't want to pass scopes into services).  

I also prefer Individual services for things, but I'm not sure we can meaningfully do that for everything.  Alerts & Breadcrumbs are great candidates for reducing noise.

PR #1248, opened 2017-02-10
Fix sample pipeline button

### LINKED ISSUE #1247: TypeError on pipelines page
Just noticed this while navigating around a new empty project:

screen shot 2017-02-10 at 1 04 19 pm

I think its prob a refactor bug, missed rename.
@jwforres @spadgett

spadgett:
We were calling a removed Navigate service API to get the template URL, which caused a runtime error.

Fixes #1247


jwforres:
[merge]

PR #1250, opened 2017-02-10
Upgrade to angular 1.5 and patternfly 3.18.1

jwforres:


spadgett:
Seeing a diff in `dist.java`

jwforres:
cc @dtaylor113 @jeff-phillips-18 @spadgett started this branch that contains dave's changes plus my fixes for orderBy.  I got all the other repos updated on the angular dependency so there are no bower conflicts now.

jwforres:
Looks like travis is happy now

dtaylor113:
LGTM!  Thanks @jwforres :+1:

jwforres:
@spadgett think this is ready for review now, particularly the second commit

bower.json
```diff
@@ -2,34 +2,35 @@
...
+    "angular": "1.5.11",
     "json3": "3.3.2",
     "es5-shim": "3.1.1",
-    "angular-resource": "1.3.20",
-    "angular-cookies": "1.3.20",
-    "angular-sanitize": "1.3.20",
-    "angular-animate": "1.3.20",
-    "angular-touch": "1.3.20",
-    "angular-route": "1.3.20",
+    "es5-dom-shim": "*",
```
spadgett: Concerns me a bit we have to use `"*"` here :/

app/scripts/controllers/imagestream.js
```diff
@@ -48,7 +48,7 @@ angular.module('openshiftConsole')
                 };
               }
               $scope.imageStream = imageStream;
-              $scope.tagsByName = ImageStreamsService.tagsByName($scope.imageStream);
+              $scope.tags = _.sortBy(ImageStreamsService.tagsByName($scope.imageStream), 'name');
```
spadgett: 👍  sorting in the watch callback seems better

bower.json
jwforres: We effectively already were since that is what hawtio-core was asking for before, there is not other option as the es5-dom-shim repo doesn't actually have releases.

We can probably drop this dependency altogether now, I don't think any of the browsers we support now even need this.  But I would rather do that independent of this set of changes so we aren't trying to figure out what caused the issue if something ends up breaking.

spadgett: Yeah, I'm fine with it, although something we might look at. Maybe we can just exclude (not for this PR)

app/scripts/controllers/quota.js
```diff
@@ -102,13 +102,13 @@ angular.module('openshiftConsole')
         $scope.project = project;
 
         DataService.list("resourcequotas", context, function(quotas) {
-          $scope.quotas = quotas.by("metadata.name");
+          $scope.quotas = _.sortBy(quotas.by("metadata.name"), 'name');
```
jwforres: on the data service question I tend to agree, but since we are in the middle of moving the services out to the other repo, I'd rather do that separately from this PR

spadgett: Works for me

jwforres:
PR updated to address review feedback plus the the stuff around image stream tags we just discussed

jwforres:
[merge]

PR #1251, opened 2017-02-13
[WIP] prettier autoformatter

benjaminapetersen:
Related to #682 eslint & style guide.
- Run prettier (atom package) against builds.js to compare output to see how far this falls from what we'd like. 

Opening just to see how y'all feel about this.

@spadgett @jwforres @rhamilto @sg00dwin

### DIFF
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -27,6 +27,7 @@
     "angular": false,
     "by": false,
     "before": false,
+    "beforeAll": false,
     "beforeEach": false,
     "browser": false,
     "describe": false,
--- a/test/integration/features/user_adds_imagestream_to_project.spec.js
+++ b/test/integration/features/user_adds_imagestream_to_project.spec.js
@@ -1,0 +1,43 @@
+'use strict';
+
+var h = require('../helpers');
+var projectHelpers = require('../helpers/project');
+var OverviewPage = require('../page-objects/overview').OverviewPage;
+var CreateProjectPage = require('../page-objects/createProject').CreateProjectPage;
+var ImageStreamsPage = require('../page-objects/imageStreams').ImageStreamsPage;
+var centosImageStream = require('../fixtures/image-streams-centos7.json');
+
+describe('User adds an image stream to a project', function() {
+
+  beforeEach(function() {
+    h.commonSetup();
+    h.login();
+    projectHelpers.deleteAllProjects();
+  });
+
+  afterEach(function() {
+    h.commonTeardown();
+  });
+
+  describe('after creating a new project', function() {
+    describe('using the "Import YAML/JSON" tab', function() {
+      it('should process and create the images in the image stream', function() {
+        var project = projectHelpers.projectDetails();
+        var createProjectPage = new CreateProjectPage(project);
+        createProjectPage.visit();
+        createProjectPage.createProject();
+        var overviewPage = new OverviewPage(project);
+        overviewPage.visit();
+        var catalogPage = overviewPage.clickAddToProject();   // implicit redirect to catalog page
+        catalogPage
+          .processImageStream(JSON.stringify(centosImageStream))
+          .then(function() {
+            // verify we have the nodejs image stream loaded
+            var imageStreamsPage = new ImageStreamsPage(project);
+            imageStreamsPage.visit();
+            expect(element(by.cssContainingText('td', 'nodejs')).isPresent()).toBe(true); // TODO: use fixture
+          });
+      });
+    });
+  });
+});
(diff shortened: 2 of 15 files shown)


app/scripts/controllers/build.js
```diff
@@ -111,101 +131,139 @@ angular.module('openshiftConsole')
...
         };
 
         $scope.cloneBuild = function() {
-          var name = _.get($scope, 'build.metadata.name');
+          var name = _.get($scope, "build.metadata.name");
           if (name && $scope.canBuild) {
-            BuildsService
-              .cloneBuild(name, context)
-              .then(function resolve(build) {
+            BuildsService.cloneBuild(name, context).then(
```
rhamilto: Hmm.  I can't say I'm a fan of the formatting change here.

jawnsy: Over the years, I've found that the precise style doesn't matter to me so much as consistency between files. I've worked in codebases that used 3-space indents, which seem like some sort of bizarre compromise between 2- and 4-space (and maybe it was), but I got used to it eventually. We could bikeshed all day about the best way to format code, but ultimately what matters is consistency, and that's something tools like `go fmt` do well.

I'd like to suggest that perhaps we (as developers and humans) underestimate our ability to adapt :smile:

benjaminapetersen: Agree! I already opened an [issue](https://github.com/jlongster/prettier/issues/682) re the `.then` formatting.

spadgett: > I'd like to suggest that perhaps we (as developers and humans) underestimate our ability to adapt

I find that I adapt the best when I'm multitasking on several things at once

app/scripts/controllers/build.js
```diff
@@ -111,101 +131,139 @@ angular.module('openshiftConsole')
...
           if (!logLink) {
             return [];
           }
 
-          return [{
-            href: logLink,
-            label: "View Log"
-          }];
+          return [
+            {
```
jwforres: this change is a waste of space but its not terrible.

app/scripts/controllers/build.js
```diff
@@ -94,15 +110,19 @@ angular.module('openshiftConsole')
...
       };
     };
 
     var buildConfigResolved = function(buildConfig, action) {
       if (action === "DELETED") {
         $scope.alerts["deleted"] = {
           type: "warning",
-          message: "Build configuration " + $scope.buildConfigName + " has been deleted."
+          message: (
+            "Build configuration " +
```
jwforres: while this makes it easier to see the parts that are being appended, i feel like it actually makes it harder to read the sentence its creating

benjaminapetersen:
Yeah, there is definitely some little things that are quirky, but I tend to think the gains are good and I'd vote running it the first time "out of the box" on everything to normalize & get a quick win.  We could tinker with individual config options as a future second pass?

app/scripts/controllers/build.js
benjaminapetersen: Yeah, thats a little unnecessary.

app/scripts/controllers/build.js
benjaminapetersen: I think I probably prefer this....

jawnsy: Maybe the solution is to use format strings? Are those a thing in Angular?

benjaminapetersen: We don't use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in the console yet to avoid a transpiler, but we do use them in the tests since they run in node.

benjaminapetersen:
Updated & ran against all the scripts related to `builds` to give us a few more files to look at. In addition, here is the options via the atom plugin:

screen shot 2017-02-15 at 9 55 07 am

And what is available (same stuff) for creating a grunt task:

screen shot 2017-02-15 at 9 58 05 am

So it is fairly limited, but I think that may be a good thing (for now).  We could easily use `prettier` & update the code base in one swoop.  If we wanted finer grained control, I could then run the `eslint` rule detector against the `/src` (follow-up PR) which would generate us a `yaml` or `json` file that reflects what is set by prettier & we could decide if we want to tweak from there (eslint is more configurable).

That said, I like the quick win of:
- running `prettier` against our `src` & committing this PR
- we all install the atom plugin & tick the `format on save` option (and/or add a grunt task) and let it maintain from here on out.

benjaminapetersen:
This change from `buildConfig.js` is a big waste of space... I'm inclined to think its too much:
```javascript
            {
              http: {
                params: {
                  labelSelector: (
                    $filter('labelName')('buildConfig') +
                      '=' +
                      _.trunc($scope.buildConfigName, {
                        length: 63,
                        omission: ''
                      })
                  )  // uh...
                } // really?
              } // hmmm...
            } // omg.
          ) // really?
        ); // yikes.
```

PR #1255, opened 2017-02-14
Removing orphaned Settings templates and controller

### LINKED ISSUE #1253: app/views/settings.html is orphaned
We should probably remove the file since it's no longer in use?

rhamilto:
Fixes #1253


jwforres:
[merge]

PR #1257, opened 2017-02-14
Removing orphaned environment directive

### LINKED ISSUE #1252: Environment directive no longer in use and should be deprecated
Since adding the key-value-editor, we are no longer using the environment directive.  Presumably, the directive should either be commented out a la [pods and podContent](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/resources.js#L34-L79) or removed.

See

https://github.com/openshift/origin-web-console/blob/master/app/views/directives/environment.html

and 

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/resources.js#L139-L150

rhamilto:
Fixes #1252


jwforres:
[merge]

PR #1276, opened 2017-02-20
Reduce tpl logic in deployment view

benjaminapetersen:
Any objection to this kind of reduction?

@spadgett @jwforres

### DIFF
--- a/app/scripts/controllers/deployment.js
+++ b/app/scripts/controllers/deployment.js
@@ -38,6 +38,17 @@ angular.module('openshiftConsole')
         title: $routeParams.deployment
       }
     ];
+    var canI = $filter('canI');
+    var resourceGroups = {
+      k8sDeployment: { group: 'extensions', resource: 'deployments' },
+      hpa: {resource: 'horizontalpodautoscalers', group: 'extensions'}
+    };
+    $scope.canI = {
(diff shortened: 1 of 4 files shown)


spadgett:
It's definitely an improvement. If we have this many checks, it might be better to evaluate `canI` once in the controller, though. What do you think?

spadgett:
I'm thinking a check like this, only assigning the result to a scope var

https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/addConfigVolume.js#L101

benjaminapetersen:
I don't mind doing that, tho in kve we just decided yesterday that `canI` is better in the template @rhamilto @jwforres.  Any objections to this check?

So long as we do less object literal in the tpl, I'm good either way on the filter.

jwforres:
i don't mind if its clear what the scope variable is for.  in KVE what was in the view had no obvious correlation back to the actual check that was happening.

benjaminapetersen:
Cool, I will update.

benjaminapetersen:
Updated.  Thoughts? I can condense the `ctrl` block down a bit if we don't want the `$filter('canI')` & `resourceGroup` bit expanded.

benjaminapetersen:
rebased

PR #1280, opened 2017-02-21
Fix missing "Create Source Secret" link

### LINKED ISSUE #1279: "Create New Secret" link sometimes missing
If I load the create from source page without visiting and other pages first, the create new secret link is not present on the page.

spadgett:
Make sure the project has loaded before the `canI` check is evaluated when creating from source.

Fixes #1279


jwforres:
[merge]

PR #1281, opened 2017-02-21
[enterprise-3.5] Fix missing "Create Source Secret" link

spadgett:
Make sure the project has loaded before the `canI` check is evaluated when creating from source.


jwforres:
[merge]

PR #1287, opened 2017-02-22
[enterprise-3.5] Bug 1425728 - Fix cancel from add config files page

spadgett:
If you click create secret, then cancel out both forms, you go back to the create secret page. Use `$window.history.back()` instead of a `then` parameter from the create secret page to prevent this.

https://bugzilla.redhat.com/show_bug.cgi?id=1425728


jwforres:
[merge]

PR #1288, opened 2017-02-22
Resolve bug where list-view-item top border disappeared

### LINKED ISSUE #1283: Monitoring page missing borders between list view rows
Since the Patternfly update #1250, the monitoring page no longer has borders between rows.

openshift_web_console

rhamilto:
Caused by backward incompatible changes in PatternFly v3.12+.

Resolves #1283

Note:  we’ll likely replace the instance using this LESS/CSS on the
Monitoring page with the new list-pf, but until then, this fixes the
bug.

### DIFF
--- a/app/styles/_list-view.less
+++ b/app/styles/_list-view.less
@@ -44,25 +44,6 @@
   }
 }
 
-.list-group-item-expandable {
-  cursor: pointer;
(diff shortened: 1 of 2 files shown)


spadgett:
[merge]

PR #1263, opened 2017-02-17
Aligning kebab styles with PatternFly

rhamilto:
Fixes #1262 

Before:

screen shot 2017-02-17 at 3 31 03 pm
screen shot 2017-02-17 at 3 31 11 pm

After:

screen shot 2017-02-17 at 3 29 33 pm
screen shot 2017-02-17 at 3 29 42 pm

PR #1293, opened 2017-02-22
Making rule less specific so it applies even if another node (ng)

rhamilto:
is inserted in the markup.  Enhancement/fix for #1263 

Was a problem on browse/rc pages (possibly others).


spadgett:
[merge]

PR #1307, opened 2017-02-27
Bug 1427084 - Fix problem showing project usage for cluster quota

spadgett:
Since we switched from a `clusterQuotas` map to a `clusterQuotas` array
for sorting, we incorrectly keyed the `namespaceUsageByClusterQuota` map
by array index instead of namespace. This prevented the project usage
for a cluster quota from displaying.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1427084

@benjaminapetersen ptal
@jwforres fyi


spadgett:
Thanks @benjaminapetersen 

[merge]

PR #1308, opened 2017-02-27
Replace common services with imports from origin-web-common

jeff-phillips-18:
This removes the common services from Web Console and imports them from origin-web-common.
Functionality remains the same just a relocation to make the services available outside of web console.


app/scripts/services/ws.js
```diff
@@ -1,91 +0,0 @@
...
-        return false;
-      }
-    };
-
-    return $ws;
-  };
-})
-
-/* A WebSocket factory for kubernetesContainerTerminal */
-.factory("ContainerWebSocket", function(API_CFG, $ws) {
```
spadgett: We should remove this from origin-web-common and keep it here since it's specific to container terminal.

app/scripts/services/containerWebSocket.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+/* A WebSocket factory for kubernetesContainerTerminal */
+angular.module('openshiftCommon')
```
jwforres: think this module name needs to get switched back to the main console one

jeff-phillips-18:
Thanks @jwforres , updated.

spadgett:
@jeff-phillips-18 Should we remove the `ContainerWebSocket` initialization from origin-web-common and cut a 0.0.3?

https://github.com/openshift/origin-web-common/blob/82ec00ff08179d87621d8dd003fdbc5130779567/src/services/ws.js#L82-L91

jwforres:
Pulled the branch down, it looks like you need to add
`<script src="scripts/services/containerWebSocket.js"></script>`
to index.html

jwforres:
If they aren't already we probably need to get the spec tests running in the origin-web-common repo.  Travis is probably sufficient for now.

Otherwise this all LGTM now, everything seems to be working ok.

jeff-phillips-18:
I was able to move the spec tests that existed for the (now) common services running in the origin-web-common repo.

jwforres:
Yeah i think my question was whether we actually have Travis running on that repo, but we can take that discussion to another thread i think.

@jeff-phillips-18 can you squash your commits?

jeff-phillips-18:
Squished em.

app/scripts/app.js
```diff
@@ -514,120 +501,3 @@ angular
     }
   });
 
-hawtioPluginLoader.addModule('openshiftConsole');
```
jwforres: Missed this earlier, I'm actually surprised you were able to remove this.  How is the openshiftConsole module getting registered with hawtioPluginLoader now?

jeff-phillips-18:
Hmmm.... Odd that it still loaded. Well, it's back now.

jeff-phillips-18:
@jwforres @spadgett Any further issues?

jwforres:
nope thanks @jeff-phillips-18 !

[merge]

jwforres:
looks like Travis and Jenkins are both failing on verify-dist, @jeff-phillips-18 can you re-run grunt build

jwforres:
looks like Travis is having a partial outage right now so I can't see in the logs why it just failed, will try jenkins [test]

jwforres:
[merge]

PR #1312, opened 2017-03-01
Bug 1427360 - Correctly handle 0 values in MetricsService

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1427360

### DIFF
--- a/app/scripts/services/metrics.js
+++ b/app/scripts/services/metrics.js
@@ -25,9 +25,10 @@ angular.module("openshiftConsole")
 
       _.each(data, function(point) {
         // Set point.value to the average or null if no average.
-        if (!point.value || point.value === "NaN") {
-          var avg = point.avg;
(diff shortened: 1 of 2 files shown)


app/scripts/services/metrics.js
```diff
@@ -25,9 +25,10 @@ angular.module("openshiftConsole")
 
       _.each(data, function(point) {
         // Set point.value to the average or null if no average.
-        if (!point.value || point.value === "NaN") {
```
spadgett: @jwforres `point.value` is never set by Hawkular, so I removed this check.

spadgett: http://www.hawkular.org/docs/rest/rest-metrics.html#NumericBucketPoint

jwforres:
[merge]

Issue #1309, opened 2017-02-28
Not enough margin below log container select


spadgett:
![openshift_web_console](https://cloud.githubusercontent.com/assets/1167259/23407987/5b465e44-fd94-11e6-9717-d3b20ff6aadf.png)

PR #1314, opened 2017-03-01
Include conditional style to set spacing for ui-select

sg00dwin:
This is also used on the monitoring page to set spacing beneath the select if it exists. 

Fixes https://github.com/openshift/origin-web-console/issues/1309

screen shot 2017-03-01 at 11 21 29 am


spadgett:
[merge]

Issue #1237, opened 2017-02-09
Config map table at mobile may not be obvious that the full values isn't visible


jwforres:
If you have a config map with a lot of key/value pairs in it, or even just on with a lot of content in a value, then at mobile the horizontal scrollbar that appears may end up off the bottom of the screen.  Could we do something like a fade effect on the right side of the content area to indicate that there is more content that is pushing to the right and that its a scrollable content area?

![config-map-scrolling](https://cloud.githubusercontent.com/assets/1467629/22797731/6ecc9e18-eecd-11e6-8aad-3ea88a1ea7ac.png)

jwforres:
@ncameronbritt @rhamilto do we have anything in patternfly today where there is stylistic indicator that content is scrolling behind an edge?

rhamilto:
Not that I'm aware of.

PR #1319, opened 2017-03-03
Introduction of a horizontal scroll mixin that displays left and right dropshadow as visual cue.

sg00dwin:
Current usage on `.table-responsive` and multi line secrets with `pre`
Can be added to other elements as needed.

Fixes https://github.com/openshift/origin-web-console/issues/1237

screen shot 2017-03-03 at 10 43 15 am

screen shot 2017-03-03 at 5 19 33 pm

screen shot 2017-03-03 at 1 13 43 pm

screen shot 2017-03-03 at 10 44 27 am

### DIFF
--- a/app/styles/_scroll-shadows-horizontal.less
+++ b/app/styles/_scroll-shadows-horizontal.less
@@ -1,0 +1,36 @@
+// Horizontal scroll drop shadows
+// Adds drop shadows to left and right edges that display when scrollable content is hidden
+// Test cases http://codepen.io/sg00dwin/pen/GWWRLR/
+@scroll-shadows-horizontal-bg:         rgba(255,255,255,1);
+@bg-color:                             @scroll-shadows-horizontal-bg;
+
+.scroll-shadows-horizontal(@bg-color: @scroll-shadows-horizontal-bg) {
+  background:
+    radial-gradient(ellipse at left, rgba(0, 0, 0, 0.25) 0%, rgba(0, 0, 0, 0) 75%) 0 center, // left shadow
+    radial-gradient(ellipse at right, rgba(0, 0, 0, 0.25) 0%, rgba(0, 0, 0, 0) 75%) 100% center, // right shadow
+    @bg-color;
+  background-size:
+    10px 100%,
+    10px 100%;
+  background-attachment:
+    scroll,
+    scroll;
+  background-repeat: no-repeat;
+	// make horizontally scrollable
+	overflow-x: auto;
+  min-height: .01%;
+
+	// White gradient to hide the scroll shadows when at the farthermost scroll positions
+	// change these gradients from white to your background colour if it differs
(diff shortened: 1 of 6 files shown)


spadgett:
@sg00dwin 

Love the effect.

I still see the shadow when scrolled all the way over on some tables, though.

openshift_web_console

Travis is complaining about a dist diff. You might need to clean / install deps and build again.

benjaminapetersen:
This def looks helpful!

app/styles/_tables.less
```diff
@@ -215,24 +215,31 @@
   }
 }
 
+// for testing
```
rhamilto: Seems like 219 should be removed based on this comment.

serenamarie125:
@sg00dwin curious as to why we are doing this?  Wondering if it addresses a specific customer issue or usability issue.  If so, would love to hear about it, so that we can consider whether it's something we should contribute back to PatternFly

sg00dwin:
@rhamilto I thought it was ready, but in my last set of browser testing, IE 11 wasn't positioning the white scroll cover gradient correctly and couldn't find a solution at that moment.

app/styles/_scroll-shadows-horizontal.less
```diff
@@ -0,0 +1,36 @@
+// Horizontal scroll drop shadows
+// Adds drop shadows to left and right edges that display when scrollable content is hidden
+// Test cases http://codepen.io/sg00dwin/pen/GWWRLR/
+@scroll-shadows-horizontal-bg:         rgba(255,255,255,1);
+@bg-color:                             @scroll-shadows-horizontal-bg;
+
+.scroll-shadows-horizontal(@bg-color: @scroll-shadows-horizontal-bg) {
```
rhamilto: Nit:  alpha and consistent formatting of background-* amongst rules.

rhamilto: screen shot 2017-04-26 at 4 26 09 pm

* The aforementioned blue background.
* The outer border is missing at desktop resolutions.
* The shadows aren't being fully masked (there is a ever-so-slight ghosting of the shadows).

sg00dwin:
@serenamarie125 Jessica opened https://github.com/openshift/origin-web-console/issues/1237 noting that on the annotations tables that use the bootstrap table-responsive rules which make the table have a horizontal scrollbar. But on some of those tables the content is taller than the screen so the scrollbar isn't visible and it might not be apparent that the content extends off the screen.[1] So we tried the idea of adding left and right scroll shadows.


[1] screen shot 2017-04-26 at 4 27 19 pm

app/styles/_scroll-shadows-horizontal.less
rhamilto: The application of .scroll-shadows-horizontal() is missing from .copy-to-clipboard.

PR #1334, opened 2017-03-09
Update tech preview constant flags

jwforres:


app/scripts/constants.js
```diff
@@ -85,7 +85,7 @@ window.OPENSHIFT_CONSTANTS = {
   ],
 
   ENABLE_TECH_PREVIEW_FEATURE: {
-    pipelines: true
+    catalog_landing_page: false
```
jwforres: @sspeiche any preference on what we call this thing, "catalog" seemed to generic since we already have a "catalog" in the console.   Keep in mind this would show up in the docs for how to turn it on for tech previewing.

sspeiche: maybe expanding to service_catalog_landing_page OR just have some extra vague like new_landing_page_experience

jwforres:
[merge]

Issue #1272, opened 2017-02-20
Route warnings don't warn about missing alternate backends


spadgett:
We warn when `spec.to` doesn't exist, but we don't warn when one of `spec.alternateBackends` is missing.

spadgett:
Also need to check target port on alternate backends.

PR #1335, opened 2017-03-09
Redesigned Overview

### LINKED ISSUE #236: Overview is missing Container status info
When I create an application from the web UI (add to project) I'll be redirected to the overview page. There, there is no indication of what's happening in the background, so I don't see if the image is being pulled down or the container is being created, or anything else. 
As a user/developer I want to see in the overview the states the pod is going through. 

cc/ @marekjelen

### LINKED ISSUE #619: Option to allow the user to toggle individual metrics on overview
We've had to disable the overview metrics because it became just too resource intensive (both client and server sided), this is very obvious when the user has a lot of deployments in one project.

It'd be nice if we could give the user the option to click where the overview metrics used to be to show the metrics for that specific deployment/service. Metrics end up being quite hidden when they have to be disabled on the overview.

### LINKED ISSUE #658: RFE: be able to get to route details from overview page
Currently the overview page allows me to get to details of the objects it presents (deployment, service), except for route:
untitled

Clicking route will take me to its destination, which is perfectly valid, but sometimes I'd like to be able to get to the route details quickly. 
Currently, the only option is to go through Applications->Routes->my route. How about having small (details) next to the route link which would do just that?

### LINKED ISSUE #729: Routes are not visible on dependent services
If a dependent (grouped) service has a route, you can't see it from the overview.

spadgett:
https://trello.com/c/QxbqJsoK

Updated overview using the Patternfly list view with rows that expand to reveal more detail.

screen shot 2017-03-09 at 5 02 30 pm

screen shot 2017-03-09 at 6 52 11 pm

Contains code from @rhamilto. Based on designs from @ajacobs21e @lizsurette and @ncameronbritt 

Fixes #236
Fixes #619
Fixes #658 
Fixes #729

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1389458
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1401134

**TODO**

- [x] Generalize some directives so they're not specific to the overview.
- [x] Add in-view logic back to overview metrics
- [x] Move `getBreakpoint` to a service
- [x] Use namespaced resize event listener
- [x] Make `view-by` local storage key specific to overview
- [x] Don't show scaled down replication controllers from a deployment config that was deleted
- [x] Add some missing `canI` checks for kebab actions
- [x] Hide kebab menu when none of the actions are visible
- [x] Fix bug where filtering doesn't trigger in-view updates for metrics
- [x] Remove `renderOptions`
- [x] Colorize and linkify mini log
- [x] Use `row.state.breakpoint` rather than `hidden-*` classes to show and hide the metrics summary
- [x] Reorder DataService calls
- [x] Order RCs by deployment number instead of date
- [x] Test failed rollouts of replica sets
- [x] Fix pipeline sorting
- [x] Consistently use `DEFAULT_POLL_INTERVAL`
- [x] Fix `visibleReplicaSets` typo
- [x] get/setBuildConfigsForDeploymentConfig -> get/setBuildConfigsForObject 
- [x] Don't pass dummy pod template to `LabelSelector.matches`
- [x] Don't sort `recentPipelinesByBuildConfig` in the view
- [x] Use `AlertMessageService.permanentlyHideAlert` in `ResourceAlertsService`
- [x] Fall back to comparing names if service weights are equal when sorting
- [x] Fix initial tab selection for mobile when there are no services
- [x] Improve "deployment no longer latest" error message
- [x] Handle NaN deployment versions when sorting
- [x] Show < 0.01 cores for very small CPU usage

**Follow-on Work**

- [ ] Support types other than deployment configs for pipeline annotation `pipeline.alpha.openshift.io/uses`
- [ ] Show headless services
- [ ] Remove previous overview code
- [x] Show basic route info in collapsed state
- [ ] Make cancel deployment dialog generic and reuse it on the deployment config page

### DIFF
--- a/app/scripts/controllers/newOverview.js
+++ b/app/scripts/controllers/newOverview.js
@@ -1,0 +1,1240 @@
+'use strict';
+
+// TODO: Rename file and controller when old overview is removed.
+angular.module('openshiftConsole').controller('NewOverviewController', [
+  '$scope',
+  '$filter',
+  '$routeParams',
+  'AlertMessageService',
+  'AppsService',
+  'BuildsService',
+  'Constants',
+  'DataService',
+  'DeploymentsService',
+  'HPAService',
+  'HTMLService',
+  'ImageStreamResolver',
+  'KeywordService',
+  'LabelFilter',
+  'LabelsService',
+  'Logger',
+  'MetricsService',
+  'Navigate',
+  'ProjectsService',
+  'ResourceAlertsService',
+  'RoutesService',
+  OverviewController
+]);
+
+function OverviewController($scope,
+                            $filter,
+                            $routeParams,
+                            AlertMessageService,
+                            AppsService,
+                            BuildsService,
+                            Constants,
+                            DataService,
+                            DeploymentsService,
+                            HPAService,
+                            HTMLService,
+                            ImageStreamResolver,
+                            KeywordService,
+                            LabelFilter,
+                            LabelsService,
+                            Logger,
+                            MetricsService,
+                            Navigate,
+                            ProjectsService,
+                            ResourceAlertsService,
+                            RoutesService) {
+  var overview = this;
+  var limitWatches = $filter('isIE')() || $filter('isEdge')();
+  var DEFAULT_POLL_INTERVAL = 60 * 1000; // milliseconds
+
+  $scope.projectName = $routeParams.project;
+
+  // Filters used by this controller.
+  var annotation = $filter('annotation');
+  var getBuildConfigName = $filter('buildConfigForBuild');
+  var deploymentIsInProgress = $filter('deploymentIsInProgress');
+  var getErrorDetails = $filter('getErrorDetails');
+  var imageObjectRef = $filter('imageObjectRef');
+  var isJenkinsPipelineStrategy = $filter('isJenkinsPipelineStrategy');
+  var isNewerResource = $filter('isNewerResource');
+  var label = $filter('label');
+  var getPodTemplate = $filter('podTemplate');
+
+  var imageStreams;
+  var labelSuggestions = {};
+  var pipelineLabelSuggestions = {};
+
+  // The most recent replication controller by deployment config name. This
+  // might not be the active deployment if failed or cancelled.
+  var mostRecentByDeploymentConfig = {};
+
+  // `overview.state` tracks common state that is shared with overview-list-row.
+  // This avoids having to pass the same values as attributes again and again
+  // for different types, but lets us update these maps in one place as needed
+  // from watch callbacks in the overview controller.
+  //
+  // NOTE: Do not change or remove properties without updating overview-list-row.
+  var state = overview.state = {
+    alerts: {},
+    builds: {},
+    clusterQuotas: {},
+    imageStreamImageRefByDockerReference: {},
+    imagesByDockerReference: {},
+    limitRanges: {},
+    limitWatches: limitWatches,
+    notificationsByObjectUID: {},
+    pipelinesByDeploymentConfig: {},
+    podsByOwnerUID: {},
+    quotas: {},
+    recentPipelinesByDeploymentConfig: {},
+    routesByService: {},
+    servicesByObjectUID: {},
+    // Set to true below when metrics are available.
+    showMetrics: false
+  };
+
+  AlertMessageService.getAlerts().forEach(function(alert) {
+    state.alerts[alert.name] = alert.data;
+  });
+  AlertMessageService.clearAlerts();
+
+  // Track the breakpoint ourselves so we can remove elements from the page,
+  // rather than hiding them using CSS. This avoids rendering charts more than
+  // once for the responsive layout, which switches to tabs at smaller screen
+  // widths.
+  overview.state.breakpoint = HTMLService.getBreakpoint();
+  var onResize = _.throttle(function() {
+    var breakpoint = HTMLService.getBreakpoint();
+    if (overview.state.breakpoint !== breakpoint) {
+      $scope.$evalAsync(function() {
+        overview.state.breakpoint = breakpoint;
+      });
+    }
+  }, 50);
+
+  $(window).on('resize.overview', onResize);
+
+  overview.showGetStarted = false;
+  overview.showLoading = true;
+
+  overview.filterByOptions = [{
+    id: 'name',
+    label: 'Name'
+  }, {
+    id: 'label',
+    label: 'Label'
+  }];
+  overview.filterBy = 'name';
+
+  overview.viewByOptions = [{
+    id: 'app',
+    label: 'Application'
+  }, {
+    id: 'resource',
+    label: 'Resource Type'
+  }, {
+    id: 'pipeline',
+    label: 'Pipeline'
+  }];
+
+  var getName = function(apiObject) {
+    return _.get(apiObject, 'metadata.name');
+  };
+
+  var getUID = function(apiObject) {
+    return _.get(apiObject, 'metadata.uid');
+  };
+
+  // The size of all visible top-level items.
+  var size = function() {
+    return _.size(overview.deploymentConfigs) +
+           _.size(overview.vanillaReplicationControllers) +
+           _.size(overview.deployments) +
+           _.size(overview.vanillaReplicaSets) +
+           _.size(overview.statefulSets) +
+           _.size(overview.monopods);
+  };
+
+  // The size of all visible top-level items after filtering.
+  var filteredSize = function() {
+    return _.size(overview.filteredDeploymentConfigs) +
+           _.size(overview.filteredReplicationControllers) +
+           _.size(overview.filteredDeployments) +
+           _.size(overview.filteredReplicaSets) +
+           _.size(overview.filteredStatefulSets) +
+           _.size(overview.filteredMonopods);
+  };
+
+  // Show the "Get Started" message if the project is empty.
+  var updateShowGetStarted = function() {
+    overview.size = size();
+    overview.filteredSize = filteredSize();
+
+    // Check if there is any data visible in the overview.
+    var projectEmpty = overview.size === 0;
+
+    // Check if we've loaded the top-level items we show on the overview.
+    var loaded = overview.deploymentConfigs &&
+                 overview.replicationControllers &&
+                 overview.deployments &&
+                 overview.replicaSets &&
+                 overview.statefulSets &&
+                 overview.pods;
+
+    state.expandAll = loaded && overview.size === 1;
+
+    overview.showGetStarted = loaded && projectEmpty;
+    overview.showLoading = !loaded && projectEmpty;
+
+    overview.everythingFiltered = !projectEmpty && !overview.filteredSize;
+  };
+
+  // Group a collection of resources by app label. Returns a map where the key
+  // is the app label value and the value is an array of objects, sorted by
+  // `metadata.name`.
+  var groupByApp = function(collection) {
+    return AppsService.groupByApp(collection, 'metadata.name');
+  };
+
+  var getBestRoute = function(routes) {
+    var bestRoute = null;
+    _.each(routes, function(candidate) {
+      if (!bestRoute) {
+        bestRoute = candidate;
+        return;
+      }
+
+      // Is candidate better than the current display route?
+      bestRoute = RoutesService.getPreferredDisplayRoute(bestRoute, candidate);
+    });
+
+    return bestRoute;
+  };
+
+  // Debounce so we're not reevaluating this too often.
+  var updateRoutesByApp = _.debounce(function() {
+    $scope.$evalAsync(function() {
+      overview.bestRouteByApp = {};
+
+      if (!overview.routes) {
+        return;
+      }
+
+      // Any of the following can have services that have routes.
+      var toCheck = [
+        overview.filteredDeploymentConfigsByApp,
+        overview.filteredReplicationControllersByApp,
+        overview.filteredDeploymentsByApp,
+        overview.filteredReplicaSetsByApp,
+        overview.filteredStatefulSetsByApp,
+        overview.filteredMonopodsByApp
+      ];
+
+      // Find the best route for each app.
+      _.each(overview.apps, function(app) {
+        // Create a map of routes, keyed by route name to avoid adding them twice.
+        var routesForApp = {};
+        _.each(toCheck, function(byApp) {
+          var apiObjects = _.get(byApp, app, []);
+          _.each(apiObjects, function(apiObject) {
+            var uid = getUID(apiObject);
+            var services = _.get(state, ['servicesByObjectUID', uid], []);
+            _.each(services, function(service) {
+              // Only need to get the first route, since they're already sorted by score.
+              var routes = _.get(state, ['routesByService', service.metadata.name], []);
+              _.assign(routesForApp, _.indexBy(routes, 'metadata.name'));
+            });
+          });
+        });
+
+        overview.bestRouteByApp[app] = getBestRoute(routesForApp);
+      });
+    });
+  }, 300, { maxWait: 1500 });
+
+  // Group each resource kind by app and update the list of app label values.
+  var updateApps = function() {
+    overview.filteredDeploymentConfigsByApp = groupByApp(overview.filteredDeploymentConfigs);
+    overview.filteredReplicationControllersByApp = groupByApp(overview.filteredReplicationControllers);
+    overview.filteredDeploymentsByApp = groupByApp(overview.filteredDeployments);
+    overview.filteredReplicaSetsByApp = groupByApp(overview.filteredReplicaSets);
+    overview.filteredStatefulSetsByApp = groupByApp(overview.filteredStatefulSets);
+    overview.filteredMonopodsByApp = groupByApp(overview.filteredMonopods);
+    overview.apps = _.union(_.keys(overview.filteredDeploymentConfigsByApp),
+                            _.keys(overview.filteredReplicationControllersByApp),
+                            _.keys(overview.filteredDeploymentsByApp),
+                            _.keys(overview.filteredReplicaSetsByApp),
+                            _.keys(overview.filteredStatefulSetsByApp),
+                            _.keys(overview.filteredMonopodsByApp));
+
+    AppsService.sortAppNames(overview.apps);
+    updateRoutesByApp();
+  };
+
+  var updatePipelineOtherResources = function() {
+    // Find deployment configs not associated with a pipeline.
+    var otherDeploymentConfigs = _.filter(overview.deploymentConfigs, function(deploymentConfig) {
+      var name = getName(deploymentConfig);
+      return _.isEmpty(state.pipelinesByDeploymentConfig[name]);
+    });
+    overview.deploymentConfigsNoPipeline = _.sortBy(otherDeploymentConfigs, 'metadata.name');
+    overview.pipelineViewHasOtherResources =
+      !_.isEmpty(overview.deploymentConfigsNoPipeline) ||
+      !_.isEmpty(overview.vanillaReplicationControllers) ||
+      !_.isEmpty(overview.deployments) ||
+      !_.isEmpty(overview.vanillaReplicaSets) ||
+      !_.isEmpty(overview.statefulSets) ||
+      !_.isEmpty(overview.monopods);
+  };
+
+  var updateFilterDisabledState = function() {
+    overview.disableFilter = overview.viewBy === 'pipeline' && _.isEmpty(overview.pipelineBuildConfigs);
+  };
+
+  var filterByLabel = function(items) {
+    return LabelFilter.getLabelSelector().select(items);
+  };
+
+  // Updated on viewBy changes to include the app label when appropriate.
+  var filterFields = ['metadata.name'];
+  var filterByName = function(items) {
+    return KeywordService.filterForKeywords(items, filterFields, state.filterKeywords);
+  };
+
+  var filterItems = function(items) {
+    switch (overview.filterBy) {
+    case 'label':
+      return filterByLabel(items);
+    case 'name':
+      return filterByName(items);
+    }
+
+    return items;
+  };
+
+  var isFilterActive = function() {
+    switch (overview.filterBy) {
+    case 'label':
+      return !LabelFilter.getLabelSelector().isEmpty();
+    case 'name':
+      return !_.isEmpty(state.filterKeywords);
+    }
+  };
+
+  var updateFilter = function() {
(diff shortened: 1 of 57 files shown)


spadgett:
cc @sspeiche @jorgemoralespou

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+    var vanillaReplicationControllers = [];
+    overview.replicationControllersByDeploymentConfig = {};
+    overview.currentByDeploymentConfig = {};
+    mostRecentByDeploymentConfig = {};
+
+    // Add the replication controllers to a temporary map until we have them all and can sort.
+    var rcByDC = {};
+    _.each(overview.replicationControllers, function(replicationController) {
+      var dcName = getDeploymentConfigName(replicationController) || '';
+      if (!dcName || !overview.deploymentConfigs[dcName]) {
```
jwforres: does this mean if i delete a DC and it doesn't cascade to delete the RCs that I'm going to see a bunch of old RCs showing up that aren't scaled up?

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+
+    // Add any failed / canceled deployment notifications.
+    var mostRecent = getMostRecentReplicationController(deploymentConfig);
+    notifications = ResourceAlertsService.getDeploymentStatusAlerts(deploymentConfig, mostRecent);
+
+    // Roll up notifications like pod warnings for any visible replication controller.
+    var visibleReplicationControllers = getVisibleReplicationControllers(deploymentConfig);
+    _.each(visibleReplicationControllers, function(replicationController) {
+      var rcNotifications = getNotifications(replicationController);
+      _.assign(notifications, rcNotifications);
```
spadgett: The alert ID for RC notifications contains the object UID. (And for pods, we have the grouping logic.)

app/scripts/controllers/newOverview.js
spadgett: Thanks, fixed.

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+    overview.renderOptions.showLoading = !loaded && projectEmpty;
+
+    overview.everythingFiltered = !projectEmpty && !overview.filteredSize;
+  };
+
+  // Group a collection of resources by app label. Returns a map where the key
+  // is the app label value and the value is an array of object, sorted by
+  // `metadata.name`.
+  var groupByApp = function(collection) {
+    return AppsService.groupByApp(collection, 'metadata.name');
```
spadgett: > Or was the thinking that if we kept it as group by app but changed how the grouping worked that it would just be under this same service.

Yeah, that was the thought, although I'm OK moving it to LabelsService as well.

spadgett:
> Not done with newOverview.js yet, will continue later. This is what I've got for now.

Thanks @jwforres. I added a checklist for feedback and follow-on work to the description, and I have local changes for most of your comments to this point. Let me know if you'd like me to push.

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+  overview.filterByOptions = [{
+    id: 'name',
+    label: 'Name'
+  }, {
+    id: 'label',
+    label: 'Label'
+  }];
+  overview.filterBy = 'name';
+
+  overview.viewByOptions = [{
```
spadgett: It's funny, I was debating removing `renderOptions` since some directives that don't have isolated scope look for it. For instance,

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/nav.js#L150-L154

The settings here apply only to the overview, but `renderOptions` implies that they might be used outside of this controller. We have some copied code in other controllers as well that sets `renderOptions` when not needed.

As an aside, we might remove that check from `project-filter`. I don't think it's needed anymore since it was removed from the nav a while back?

spadgett:
@jwforres Noticed I'm missing some `canI` checks for the actions kebab. Working on adding these.

app/scripts/controllers/newOverview.js
jwforres: Good point, we may not even need renderOptions any more

app/scripts/controllers/newOverview.js
jwforres: Ok thats fine, I think I'm ok with it either way.

spadgett:
@jwforres Should we reorder the builds and pipelines menu items in the nav to match the order in the overview?

screen_shot_2017-03-10_at_11_54_18_am

jwforres:
I would hesitate to switch nav order on people if we can avoid it. Muscle memory is a thing :)

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+      });
+  };
+
+  var watches = [];
+  ProjectsService.get($routeParams.project).then(_.spread(function(project, context) {
+    // Project must be set on `$scope` for the projects dropdown.
+    $scope.project = project;
+    state.context = context;
+
+    var updateReferencedImageStreams = function() {
```
jwforres: should we pull this up out of the watches since all the other functions are defined above?

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+    }));
+
+    watches.push(DataService.watch({
+      group: "extensions",
+      resource: "horizontalpodautoscalers"
+    }, context, function(hpaData) {
+      overview.horizontalPodAutoscalers = hpaData.by("metadata.name");
+      groupHPAs();
+      Logger.log("autoscalers (subscribe)", overview.horizontalPodAutoscalers);
+    }, {poll: limitWatches, pollInterval: 60 * 1000}));
```
jwforres: any reason you didn't switch these next few to use DEFAULT_POLL_INTERVAL

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+      updateQuotaWarnings();
+    }, {poll: true, pollInterval: 60 * 1000}));
+
+    // List limit ranges in this project to determine if there is a default
+    // CPU request for autoscaling.
+    DataService.list("limitranges", context, function(response) {
+      state.limitRanges = response.by("metadata.name");
+    });
+
+    watches.push(DataService.watch("imagestreams", context, function(imageStreamData) {
```
jwforres: thinking we should have all the watches first, then the limited watches, then the always polled, then the lists, and then the gets.  I think most of them were in that order besides this one.

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+  //   key: deployment config name
+  //   value: array of builds, sorted in descending order by creation date
+  var updateRecentBuildsForDeploymentConfig = function(deploymentConfig) {
+    var builds = [];
+    var buildConfigs = getBuildConfigsForDeploymentConfig(deploymentConfig);
+    _.each(buildConfigs, function(buildConfig) {
+      var recentForConfig = _.get(state, ['recentBuildsByBuildConfig', buildConfig.metadata.name], []);
+      builds = builds.concat(recentForConfig);
+    });
+    builds = BuildsService.sortByBuildNumber(builds, true);
```
spadgett: This is recent by config, so there should be no duplicates on build number unless there's a bug or someone has changed the annotation.

`sortByBuildNumber` falls back to name when there's no number.

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+    _.each(overview.deploymentConfigs, updateRecentBuildsForDeploymentConfig);
+  };
+
+  var groupBuilds = function() {
+    if(!state.builds || !overview.buildConfigs) {
+      return;
+    }
+
+    // Reset these maps.
+    overview.recentPipelinesByBuildConfig = {};
```
spadgett: I'm hesitant to put things in `state` that aren't needed by list row. Then it's not obvious when you change something here if it impacts the list row component.

app/scripts/controllers/newOverview.js
spadgett: It needs access to `context` returned from `ProjectsService.get` :/

app/scripts/controllers/newOverview.js
spadgett: No, that was an oversight.

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+      pipelineBuildConfigs.push(buildConfig);
+
+      // TODO: Handle other types.
+      // Preserve the order they appear in the annotation.
+      var dcNames = BuildsService.usesDeploymentConfigs(buildConfig);
+      var bcName = getName(buildConfig);
+      _.set(overview, ['deploymentConfigsByPipeline', bcName], dcNames);
+      _.each(dcNames, function(dcName) {
+        state.pipelinesForDeploymentConfig[dcName] = state.pipelinesForDeploymentConfig[dcName] || [];
+        state.pipelinesForDeploymentConfig[dcName].push(buildConfig);
```
spadgett: > should this be sorted

It's not displayed directly, although `recentPipelinesByDeploymentConfig` is.

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+  //   value: array of pipeline builds
+  var groupPipelineByDC = function(build) {
+    var bcName = getBuildConfigName(build);
+    var buildConfig = overview.buildConfigs[bcName];
+    if (!buildConfig) {
+      return;
+    }
+
+    overview.recentPipelinesByBuildConfig[bcName] = overview.recentPipelinesByBuildConfig[bcName] || [];
+    overview.recentPipelinesByBuildConfig[bcName].push(build);
```
spadgett: Yes, it's sorted in `groupBuilds`

app/scripts/controllers/newOverview.js
jwforres: unless im misreading this its getting recent for config across multiple build configs and then concat'ing those into one array, and then you are sorting that concat'ed array

app/scripts/controllers/newOverview.js
jwforres: Ok thats fine

app/scripts/controllers/newOverview.js
jwforres: Ok probably something we should refactor later then, it doesn't really need the context, i think it really just needs to know the project

app/scripts/controllers/newOverview.js
spadgett: You're right. I was looking at a different part of the code because my numbers locally no longer line up.

But this is only used for build counts, so shouldn't need to be sorted at all.

jwforres: OK

spadgett:
@jwforres I've updated the build sorting with a `BuildsService.sortBuilds` method we can reuse on the build config page. Right now it has the same problem where the builds are in the wrong order if `creationTimestamp` is the same for both.

app/scripts/controllers/newOverview.js
```diff
@@ -0,0 +1,1200 @@
...
+    if (!apiObjects || !overview.services) {
+      return;
+    }
+
+    _.each(apiObjects, function(apiObject) {
+      var services = [];
+      var uid = getUID(apiObject);
+      var podTemplate = getPodTemplate(apiObject) || { metadata: { labels: {} } };
+      _.each(selectorsByService, function(selector, serviceName) {
+        if (selector.matches(podTemplate)) {
```
spadgett: It looks like `LabelSelector.matches` handles null and undefined values. I'll remove the dummy object.

https://github.com/kubernetes-ui/label-selector/blob/54cc2dba5a2229073088f6dee589981ace485e6d/labelSelector.js#L98-L101

spadgett:
@jwforres I believe I've addressed all comments to this point

app/scripts/directives/routeServiceBarChart.js
```diff
@@ -0,0 +1,62 @@
...
+  var compareWeights = function(left, right) {
+    if (left.name === routeServices.highlightService) {
+      return -1;
+    }
+
+    if (right.name === routeServices.highlightService) {
+      return 1;
+    }
+
+    return right.weight - left.weight;
```
jwforres: if weights are equal fallback to name

app/scripts/services/deployments.js
```diff
@@ -401,6 +401,45 @@ angular.module("openshiftConsole")
...
+        if (!leftVersion) {
+          return descending ? 1 : -1;
+        }
+
+        if (!rightVersion) {
+          return descending ? -1 : 1;
+        }
+
+        leftVersion = parseInt(leftVersion, 10);
+        rightVersion = parseInt(rightVersion, 10);
```
jwforres: shouldn't we handle bad version annotations that parse as NaN?  would expect these to be treated like non-existing versions

app/scripts/services/resourceAlerts.js
```diff
@@ -0,0 +1,170 @@
...
+
+    var isAlertHidden = function(alertID, namespace) {
+      var key = alertHiddenKey(alertID, namespace);
+      return localStorage.getItem(key) === 'true';
+    };
+
+    var hideAlert = function(alertID, namespace) {
+      var key = alertHiddenKey(alertID, namespace);
+      localStorage.setItem(key, 'true');
+    };
```
jwforres: isnt this duplicating these that we already had put into AlertMessageService, which is actually referenced in the quota warnings below.

app/scripts/services/deployments.js
spadgett: Yeah, they should parse as NaN. The second parameter is radix.

app/scripts/services/resourceAlerts.js
spadgett: Yeah, this code was moved from `serviceGroupNotifications`, which probably was written before the helpers were added to `AlertMessageService`. I'll update it to use the `AlertMessageService` methods.

app/scripts/services/deployments.js
jwforres: Right but if they parse as NaN don't we want to fall back to sorting by name?

spadgett: > Right but if they parse as NaN don't we want to fall back to sorting by name?

Fixed

app/scripts/directives/overview/listRow.js
```diff
@@ -0,0 +1,330 @@
...
+      });
+  };
+
+  row.startDeployment = function() {
+    DeploymentsService.startLatestDeployment(row.apiObject, {
+      namespace: row.apiObject.metadata.namespace
+    }, { alerts: row.state.alerts });
+  };
+
+  // TODO: Pulled from dc.js, but we should probably make the dialog generic and reuse for the deployment config page.
```
spadgett: > agreed 

Added a follow on item to use the same dialog for the deployment config page.

app/scripts/directives/miniLog.js
```diff
@@ -0,0 +1,105 @@
...
+    var linkifiedHTML = HTMLService.linkify(html, '_blank', true);
+
+    lineNum++;
+    buffer.push({
+      markup: linkifiedHTML,
+      id: lineNum
+    });
+
+    if (buffer.length > numLines) {
+      buffer = _.takeRight(buffer, numLines);
```
spadgett: Yeah, not sure what's best. At least in this case, the buffer is max 8 lines, so we aren't ever keeping too much data in memory. And we're only calling `_.takeRight` on a small array.

app/scripts/directives/overview/listRow.js
```diff
@@ -0,0 +1,330 @@
...
+          };
+        }
+      }
+    });
+
+    modalInstance.result.then(function() {
+      if (replicationController.metadata.uid !== row.current.metadata.uid) {
+        row.state.alerts["cancel-deployment"] = {
+          type: "error",
+          message: "Deployment " + rcName + " no longer latest."
```
spadgett: Fixed.

screen shot 2017-03-14 at 7 32 57 pm

spadgett:
Addressed latest comments.

app/views/overview/_list-row-empty-state.html
```diff
@@ -0,0 +1,38 @@
...
+</div>
+<div ng-if="!row.imageChangeTriggers.length">
+  <p>
+    No deployments for {{row.apiObject.kind | humanizeKind}}
+    <a ng-href="{{row.apiObject | navigateResourceURL}}">{{row.apiObject.metadata.name}}</a>.
+  </p>
+  <div ng-if="row.apiObject.kind === 'DeploymentConfig'">
+    <div ng-if="pipeline = row.pipelines[0]">
+      <p>
+        This deployment config is part of pipeline
```
jwforres: part of the pipeline

app/views/overview/_list-row-expanded.html
```diff
@@ -0,0 +1,186 @@
...
+                class="overview-metrics">
+              </deployment-metrics>
+              <h4 class="h5">Usage <small>Last 15 Minutes</small></h4>
+            </div>
+            <div ng-if="row.apiObject.kind !== 'Pod'">
+              <deployment-metrics
+                pods="row.getPods(row.current)"
+                containers="row.current.spec.template.spec.containers"
+                profile="compact"
+                alerts="row.sate.alerts"
```
jwforres: typo
row.state.alerts

app/views/overview/_list-row-expanded.html
```diff
@@ -0,0 +1,186 @@
...
+                  class="overview-metrics">
+                </deployment-metrics>
+                <h4 class="h5">Usage <small>Last 15 Minutes</small></h4>
+              </div>
+              <div ng-if="row.apiObject.kind !== 'Pod'">
+                <deployment-metrics
+                  pods="row.getPods(row.current)"
+                  containers="row.current.spec.template.spec.containers"
+                  profile="compact"
+                  alerts="row.sate.alerts"
```
jwforres: same typo for state

app/views/overview/_networking.html
```diff
@@ -0,0 +1,59 @@
...
+  <div ng-repeat="service in networking.services" class="row">
+    <div class="col-sm-5">
+      <div class="component-label">
+        Service
+        <span class="sublabel">Internal Traffic</span>
+      </div>
+      <h3>
+        <a ng-href="{{service | navigateResourceURL}}">{{service.metadata.name}}</a>
+      </h3>
+      <!-- <span class="text-muted mar-right-xs"> -->
```
jwforres: can we remove this

app/views/overview/_networking.html
```diff
@@ -0,0 +1,59 @@
...
+      </div>
+      <div ng-if="networking.routesByService[service.metadata.name] | hashSize">
+        <div ng-repeat="route in networking.routesByService[service.metadata.name] | limitTo : 2 track by (route | uid)" class="overview-routes">
+          <h3>
+            <span ng-if="route | isWebRoute">
+              <a ng-href="{{route | routeWebURL}}" target="_blank">{{route | routeLabel}}</a>
+              <i class="fa fa-external-link small" aria-hidden="true"></i>
+            </span>
+            <span ng-if="!(route | isWebRoute)">{{route | routeLabel}}</span>
+            <route-warnings route="route" service="service"></route-warnings>
```
jwforres: wondering if route warnings should be taking into account alternate backends, i know we didnt already just thinking as something for later

spadgett:
Added the route beside the app name:

openshift_web_console

app/styles/_overview.less
```diff
@@ -13,6 +15,376 @@
...
+    stroke-width: 1px;
+  }
+  .component-label {
+    font-size: @component-label;
+    font-weight: 500;
+    padding: 0 0 4px 0;
+    .text-muted();
+    text-transform: uppercase;
+    .sublabel {
+      font-size: (@component-label - 1);
```
spadgett:
```
app/styles/_variables.less
44:@component-label: 11px;
```

> does the variable subtraction do the right thing on px sizes?

Yes, that works.

app/views/overview/_builds.html
```diff
@@ -0,0 +1,46 @@
...
+  <div ng-repeat="buildConfig in overviewBuilds.buildConfigs track by (buildConfig | uid)" class="row">
+    <div class="col-sm-5">
+      <h3 class="mar-top-xs">
+        <a ng-href="{{buildConfig | navigateResourceURL}}">{{buildConfig.metadata.name}}</a>
+      </h3>
+    </div>
+    <div class="col-sm-7">
+      <div ng-if="!(overviewBuilds.recentBuildsByBuildConfig[buildConfig.metadata.name] | hashSize)">
+        No builds.
+        <!-- <a href="" ng-if="('buildconfigs/instantiate' | canI : 'create')" ng-click="startBuild(buildConfig)" class="small mar-left-sm">Start Build</a> -->
```
spadgett: No, I'll take it out.

app/views/directives/route-service-bar-chart.html
```diff
@@ -0,0 +1,15 @@
+<div class="route-service-bar-chart">
+  <h5>Traffic Split</h5>
+  <div ng-repeat="backend in routeServices.backends">
```
spadgett: No UID available, we just have name. We could do a `track by backend.name`, but I think it's possible for the same name to be there twice (even if it doesn't make sense), which would cause a runtime error.

app/views/overview/_list-row-actions.html
```diff
@@ -0,0 +1,64 @@
...
+        <div uib-dropdown>
+          <a href=""
+          uib-dropdown-toggle
+          class="actions-dropdown-kebab"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
+          <ul class="dropdown-menu dropdown-menu-right" uib-dropdown-menu role="menu">
+            <li role="menuitem" ng-if="row.rgv | canI : 'update'">
+              <a ng-href="{{row.apiObject | editYamlURL}}">Edit YAML</a>
+            </li>
+            <li ng-if="(pod = row.firstPod(row.current)) && ('pods/log' | canI : 'get')" role="menuitem">
+              <a ng-href="{{pod | navigateResourceURL}}?tab=logs">View Logs</a>
```
spadgett: > This really makes me want to have a logs tab on deployments / statefulsets that just adds a Pod switcher in addition to the container switcher. Follow-on though

+ 1

app/views/overview/_networking.html
spadgett: Yeah I had previously opened an issue, #1272

jwforres:
[merge]

spadgett:
\o/

PR #1336, opened 2017-03-10
Include origin-web-catalogs library.

jeff-phillips-18:
This PR pulls in the origin-web-catalogs dependency. None of the catalog components are used yet but this gives the ability to start doing so.


jeff-phillips-18:
@jwforres @spadgett

jeff-phillips-18:
Please let me know when you think we are at a point to consider this and we can rev origin-web-catalog and I will update this PR to use the new version.

jwforres:
@jeff-phillips-18 lets go ahead and do it, I think we will probably just need to rev catalog and update console several times over the next couple of weeks.

jeff-phillips-18:
This PR now includes v0.0.3 of origin-web-catalog

jeff-phillips-18:
Modified code to prevent overwriting OPENSHIFT_CONSTANTS settings that may already be set.

jwforres:
[merge]

Issue #1324, opened 2017-03-07
Pod status can overlap container ready in pods tables


spadgett:
This was the pods table on the replication controllers page.

<img width="679" alt="screen shot 2017-03-07 at 5 58 22 pm" src="https://cloud.githubusercontent.com/assets/1167259/23682042/c43688a4-035f-11e7-882e-e8a4b1749388.png">

PR #1338, opened 2017-03-10
Addition of grid width class to the status column so that message doesn't overlap adjacent column

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/1324

screen shot 2017-03-10 at 10 47 28 am


spadgett:
[merge]

Issue #799, opened 2016-11-02
DataService.list to update


benjaminapetersen:
The instances of `DataService.list` that we need to update now that it supports promises.

- [x] create/createFromImage.js
- [x] attachPVC.js
- [x] create/browseCategory.js
- [x] create.js
- [x] createRoute.js
- [x] deployment.js
- [x] edit/buildConfig.js
- [x] deploymentConfig.js
- [x] edit/autoscaler.js
- [x] edit/deploymentConfig.js
- [x] edit/route.js
- [x] membership.js
- [x] otherResources.js
- [x] quota.js
- [x] secrets.js
- [x] setLimits.js
- [x] replicaSet.js

spadgett:
Let's skip overview.js and topology.js, which will both be removed soon

benjaminapetersen:
Sounds good.

MarkDeMaria:
@benjaminapetersen I'm not seeing settings.js. Is that deprecated code?

spadgett:
It's been removed.

benjaminapetersen:
@MarkDeMaria if the file is not there, you can cut the todo.  Also, if the actual `.controller('WhateverName')` isn't listed in the `app.js` file under routes, its stale & we can eliminate it rather than update.

spadgett:
Added newOverview.js

spadgett:
Verified that the remaining items are done.

/close

Issue #220, opened 2016-07-25
Remove previous overview code


spadgett:
Tech debt issue to track removing the old overview code.

/cc @jwforres @rhamilto

spadgett:
Need to be careful with styles in _components.less that might be used outside of the overview.

rhamilto:
Orphaned styles include:

* .builds-block

spadgett:
We should wait for #539 before doing this.

PR #1340, opened 2017-03-13
Promises - Topology.js

MarkDeMaria:
Part of #799 for Topology


spadgett:
Hey @MarkDeMaria this is dead code for the old overview we're planning to remove soon (tracked by #220). Closing since I don't think it's worth updating code that won't be used.

Issue #1322, opened 2017-03-07
Affixed "Follow" link misaligned on Firefox for Mac


spadgett:
Extra gap between the edge of the log and the follow link. It should be flush with the black background's edge. We appear to be accounting for the scrollbar width, but on Firefox for Mac, there's no scrollbar.

<img width="288" alt="screen_shot_2017-03-06_at_8_13_20_pm" src="https://cloud.githubusercontent.com/assets/1167259/23637427/9b6eacfa-02a9-11e7-8326-2394bf56d339.png">

sg00dwin:
@spadgett this is another issue that i'm unable to reproduce. Is this something you consistently see? 

Following log before scrollbar needed
<img width="392" alt="screen shot 2017-03-10 at 12 19 00 pm" src="https://cloud.githubusercontent.com/assets/1874151/23805519/2ceffef6-058c-11e7-8660-8fe3a5900916.png">


Following log with scrollbar
<img width="395" alt="screen shot 2017-03-10 at 12 19 04 pm" src="https://cloud.githubusercontent.com/assets/1874151/23805478/08b2942c-058c-11e7-9bd4-eb6d38707db6.png">

spadgett:
My version of Firefox hides the scrollbar when not active. It might depend on your mac system settings.

sg00dwin:
ok, I think i've found the instance of when it happens. looking further...

PR #1341, opened 2017-03-13
Position "Follow" log btn based on whether chromeless log-view has a scrollbar

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/1322

**When no scrollbar is present**
screen shot 2017-03-13 at 2 53 49 pm
screen shot 2017-03-13 at 2 53 44 pm

**With scrollbar**

screen shot 2017-03-13 at 2 46 21 pm

screen shot 2017-03-13 at 2 47 14 pm

### DIFF
--- a/app/styles/_log.less
+++ b/app/styles/_log.less
@@ -203,18 +203,25 @@
   bottom: 5px;
   left: 10px;
 }
+
 .chromeless {
   background-color:  @log-bg-color;
   color: #d1d1d1;
-  .log-pending-ellipsis {
-    background-color: @log-bg-color;
-    padding: @grid-gutter-width 0 0 (@grid-gutter-width - 10px);
+  .log-view {
+    .log-pending-ellipsis {
+      background-color: @log-bg-color;
+      padding: @grid-gutter-width 0 0 (@grid-gutter-width - 10px);
+    }
+    .log-scroll-top.affix {
+      right: 0px; // align to edge of log-view when showScrollLinks is false
+      &.log-has-scroll {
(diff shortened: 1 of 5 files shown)


spadgett:
Won't these changes only apply to chromeless logs? I'm seeing the bug without expanding the log.

spadgett:
@rhamilto fyi

spadgett:
@sg00dwin I think we should try to get this fix in, but leave #1322 open since it describes a different problem. @rhamilto has also been looking at scrollbar issues.

app/styles/_log.less
```diff
@@ -203,18 +203,25 @@
...
+    .log-scroll-top.affix {
+      right: 0px; // align to edge of log-view when showScrollLinks is false
+      &.log-has-scroll {
+        right: @scrollbar-width; // align to edge of scrollbar when showScrollLinks is true
+      }
+    }
   }
-  .log-scroll-top.affix.target-logger-node {
-    right: @scrollbar-width;  // override
+  ::-webkit-scrollbar-track {
```
rhamilto: Would love to see this moved to _scrollbars.less since it's specific to scrollbar styling.

rhamilto:
Thinking about this last night and came to the conclusion we should reconsider the styling of the Follow/Stop Following/Go to End/Go to Top buttons so the positioning isn't so obvious when things are misaligned (e.g., get rid of the border and change the background color, etc.).  

For reference, Travis uses a button like this:

screen shot 2017-05-05 at 8 53 54 am

sg00dwin:
agreed

rhamilto:
Replaced by #1528

PR #1343, opened 2017-03-14
Make route link icon clickable

matskiv:
Service group header on 'Overview' screen contains a link to service route. Next to a link there is small icon that indicates that this link will open in new window. But clicking on this icon is handled as click on header and not on service route link (collapses the service group instead of opening the link).

This PR is supposed to improve this behavior - open service route link on icon click. 

screenshot from 2017-03-14 14-17-16

### DIFF
--- a/app/views/overview/_service-group.html
+++ b/app/views/overview/_service-group.html
@@ -12,15 +12,13 @@
       </span>
     </h2>
     <h3 class="route-title truncate">
-      <span ng-if="appName && (displayRoute | isWebRoute)">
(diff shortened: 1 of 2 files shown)


matskiv:
Hello @spadgett, can I ask you for a review ? :)

spadgett:
Hi @matskiv. It's a reasonable change, but we're planning to replace the overview soon with #1335. I suggest we fix this in the new overview.

matskiv:
@spadgett Thank you for the response. I will close this PR and later send new one for new overview screen.

PR #1344, opened 2017-03-14
Fix Custom Metrics Graphs

jmazzitelli:
Custom metrics can be defined with IDs which are different than their names. We need to use the ID and not name for the dataset ID. Without this fix, some metrics graphs in the console will be blank (particularly those from Prometheus endpoints that have labeled time series data).

### DIFF
--- a/app/scripts/directives/podMetrics.js
+++ b/app/scripts/directives/podMetrics.js
@@ -358,6 +358,11 @@ angular.module('openshiftConsole')
               // get the unit value if specified
               var unit =  metric.unit || "";
 
+              // A typical metric ID is of the form "pod/<pod-id>/custom/<some-endpoint-info>"
+              // such as "pod/be381d4b-87fc-11e5-b2a3-525400b33d1d/custom/JVM-Heap-Memory-Used".
+              // We only want part of the metric ID - "custom/" and everything after it.
+              var datasetId = "custom/" + metric.id.replace(/.*\/custom\//, '');
+
               scope.metrics.push({
                 label: label,
                 units: unit,
@@ -366,7 +371,7 @@ angular.module('openshiftConsole')
 
                 datasets: [
                   {
-                    id: "custom/" + metric.name,
+                    id: datasetId,
                     label: label,
                     type: metric.type,
                     data: []
--- a/app/scripts/services/metrics.js
+++ b/app/scripts/services/metrics.js
@@ -204,6 +204,7 @@ angular.module("openshiftConsole")
         }).then(function(response) {
           return _.map(response.data, function(value) {
             return {
+              id: value.id,
               name: value.tags.metric_name,
               unit: value.tags.units,
               description: value.tags.description,
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -1876,6 +1876,7 @@ params:f
 }).then(function(a) {
 return _.map(a.data, function(a) {
 return {
+id:a.id,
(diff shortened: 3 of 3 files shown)


app/scripts/directives/podMetrics.js
```diff
@@ -358,6 +358,9 @@ angular.module('openshiftConsole')
               // get the unit value if specified
               var unit =  metric.unit || "";
 
+              // we only want part of the metric ID - "custom/" and everything after it
+              var datasetId = "custom/" + metric.id.replace(/.*\/custom\//, '');
```
spadgett: @jmazzitelli Do you have an example of the metrics ID? It would be good to give an example in the comment as well.

cc @mwringe

jmazzitelli: Committed additional information to the comments to explain what a typical metric ID in full looks like.

app/scripts/directives/podMetrics.js
```diff
@@ -358,6 +358,11 @@ angular.module('openshiftConsole')
               // get the unit value if specified
               var unit =  metric.unit || "";
 
+              // A typical metric ID is of the form "pod/<pod-id>/custom/<some-endpoint-info>"
+              // such as "pod/be381d4b-87fc-11e5-b2a3-525400b33d1d/custom/JVM-Heap-Memory-Used".
```
spadgett: OK, seeing the example, I suggest we expose `MetricsService.getMetricInfo()` and use that instead.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/metrics.js#L105-L113

@mwringe Do you know if the format we assume in that function always holds?

spadgett: The good thing about using `getMetricInfo` is it's the same key MetricsService uses when it creates the map, so they should always match.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/metrics.js#L126-L131

mwringe: > @mwringe Do you know if the format we assume in that function always holds?

It should hold. We control what the metric id looks like coming from the agent and from heapster.

Technically metric id should be considered more opaque, but we understand that some clients are expecting a certain format and constructing the metric id directly instead of using tag queries.

jmazzitelli:
> Technically metric id should be considered more opaque

This is why my regex merely strips everything prior to /custom/ - since I did not want to assume anything about the metric ID up until that part of the metric ID string, I just stripped everything up to that point in the metric ID string. This will allow us to change the metric ID prefix in the future (if needed for whatever reason), and still not break this code. So long as our custom metric ID ends with "/custom/<some ID info>" this code will work. I would thus not want to assume anything about this metric ID string other than that.

spadgett:
It's possible MetricService will be broken in that case anyway. We wrote that assuming specifically "memory/usage", "cpu/usage", etc. If we think it might be a problem, we need to fix it there.

jmazzitelli:
OK, so the suggestion then is that we should be using "descriptor" field of the returned object from getMetricId (assuming that function variable is exposed). Something like this (line 364 in app/scripts/directives/podMetrics.js of the PR):

```
var datasetId = getMetricInfo(metric.id).descriptor;
```

spadgett:
OK, digging deeper I see `getMetricInfo` is only used for `MetricsService.getPodMetrics`, which is not called by this directive. So I'm OK with the fix as-is.

spadgett:
[merge]

PR #1349, opened 2017-03-15
Removing pficon variables (overrides)

rhamilto:
PatternFly now includes these vars in its source less, so defining them
in our less can (and will with PatternFly v3.23.0) result in a bug
where icons disappear (PatternFly v3.23.0 changes the unicode values of
a number of icons, thus causing the bug).

@spadgett, PTAL

### DIFF
--- a/app/styles/_variables.less
+++ b/app/styles/_variables.less
@@ -4,12 +4,6 @@
 @icon-font-path: "@{font-path}@{font-path-trailing-slash}"; // Bootstrap glyphicons
 @image-path: "../images";
 
-// Icons
-@pficon-var-error-circle-o: '\e61d';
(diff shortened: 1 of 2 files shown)


spadgett:
@rhamilto @sg00dwin do we know why we were overriding these icons before?

rhamilto:
> @rhamilto @sg00dwin do we know why we were overriding these icons before?  

I'm assuming it was because PatternFly didn't initially have the variables (I added them to PatternFly at @jwforres' request).

spadgett:
[merge]

jwforres:
Yep what @rhamilto said, we needed the vars before they made it into Patternfly

PR #1355, opened 2017-03-17
Update Autoscaler controller to use new DataService.list

MarkDeMaria:
Make Autoscaler Promise Compliant #799


spadgett:
[merge]

PR #1356, opened 2017-03-17
Update DeploymentConfig Controller to use new DataService.list

MarkDeMaria:
Make Deployment Config Promise Compliant #799


spadgett:
[merge]

PR #1357, opened 2017-03-17
Update edit/Route.js Controller to use new DataService.list

MarkDeMaria:
Make edit/Route controller promise compliant #799


spadgett:
[merge]

PR #1358, opened 2017-03-17
Update OtherResources.js Controller to use new DataService.list

MarkDeMaria:
Make Other Resources controller promise compliant #799


spadgett:
[merge]

PR #1359, opened 2017-03-17
Update Quota Controller to use new DataService.list

MarkDeMaria:
Make quota controller promise compliant #799


spadgett:
[merge]

PR #1366, opened 2017-03-20
Correctly hide overview metrics when unavailable

spadgett:


jwforres:
[merge]

spadgett:
@rhamilto fyi, I think this is the problem you saw when you had metrics disabled

PR #1367, opened 2017-03-20
Hide service weight slider when all weights are 0

### LINKED ISSUE #1339: Route editor: don't show slider when all service weights are 0
When editing routes, I should not default to the slider view when both service weights are set to 0.

spadgett:
When editing a route with alternate backends where all service weights
are 0, don't use the slider since it can't represent all 0 values.

Fixes #1339


jwforres:
[merge]

PR #1368, opened 2017-03-20
Resolving bugs with system-status in header

rhamilto:
Before:

screen shot 2017-03-20 at 9 14 02 am
screen shot 2017-03-20 at 10 19 59 am
screen shot 2017-03-20 at 10 20 21 am
screen shot 2017-03-20 at 10 20 37 am

After:

screen shot 2017-03-20 at 4 28 41 pm
screen shot 2017-03-20 at 4 28 59 pm
screen shot 2017-03-20 at 4 29 21 pm
screen shot 2017-03-20 at 4 29 30 pm

FYI @luciddreamz

### DIFF
--- a/app/styles/_navbar-alt.less
+++ b/app/styles/_navbar-alt.less
@@ -63,10 +63,10 @@
     }
   }
   .nav.navbar-iconic {
-    margin-right: 20px;
+    margin-right: 10px;
     .dropdown-menu {
       margin-top: -10px;
-      right: -5px;
+      right: 0;
       &:after {
         border: 6px solid transparent;
         border-bottom-color: #fff;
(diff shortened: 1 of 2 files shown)


rhamilto:
@spadgett, don't merge yet, please.  I found a bug.

rhamilto:
[test]

rhamilto:
Ok.  Should be ready for review now, @spadgett.

extensions/examples/online-extensions.js
```diff
@@ -41,8 +41,8 @@ angular
     'extensionRegistry',
     function(extensionRegistry) {
 
-      var system_status_elem = $('<a href="http://status.openshift.com/" target="_blank" class="nav-item-iconic system-status project-action-btn">');
-      var system_status_elem_mobile = $('<div row flex class="navbar-flex-btn system-status-mobile">');
+      var system_status_elem = $('<a href="http://status.openshift.com/" target="_blank" class="nav-item-iconic system-status project-action-btn" style="display: none;">');
+      var system_status_elem_mobile = $('<div row flex class="navbar-flex-btn system-status-mobile" style="display: none;">');
```
spadgett: I'm unclear why we set `display: none` and then reset it below. It looks like the DOM elements are only appended when there's an issue, so it should be unnecessary?

rhamilto: Agreed.  I'm unclear, too, but that is what is [in production for online](https://github.com/openshift/online/blob/master/ui/assets/extensions/online-extensions.js#L102-L118), so I figured it was better to match that since that's the real world use case?  It looks like this example has been more recently updated than what's in online, so maybe that's a bad assumption?  @benjaminapetersen, it looks like [you removed the inline display: none styles and javascript style undo](https://github.com/rhamilto/origin-web-console/blame/7f8960af97f07793c657f9f64b1b3d6fc1c5e988/extensions/examples/online-extensions.js#L55-L59).  Thoughts?

rhamilto:
@spadgett, I removed the extension example changes and opened a separate PR to fix the downstream.  PTAL

spadgett:
[merge]

extensions/examples/online-extensions.js
benjaminapetersen: Oi, good Q. That was a good long time ago.  I vaguely remember eliminating it as it was extra logic not really doing much, but w/o pulling it and running it again prob cant tell you.

PR #1369, opened 2017-03-20
Improve pods table performance

### LINKED ISSUE #1350: 10 $digest() loop iteration reached if scaling Replica Sets to large enough number
Intentionally trying to break it, I scaled up to a point where Angular freaked out:

screen shot 2017-03-15 at 4 40 11 pm

I hit the replica editor and entered "50" to case the error.

@spadgett

spadgett:
* Sort only on pod updates (don't sort in the view)
* Debounce the sort
* Track pods by UID in ng-repeat

Fixes #1350

cc @benjaminapetersen


jwforres:
[merge]

PR #1370, opened 2017-03-20
Use BuildsService.sortBuilds for sorting

spadgett:
Reuse the BuildsService.sortBuild method added in #1335. The BuildsService method falls back to build number when the creation dates for two builds are the same.


jwforres:
[merge]

Issue #1365, opened 2017-03-19
Label filter extends offscreen on other resources page


spadgett:
<img width="668" alt="screen shot 2017-03-19 at 4 22 37 pm" src="https://cloud.githubusercontent.com/assets/1167259/24084411/53415b28-0cc0-11e7-9fbd-ba5c1021e293.png">

Related to #1216

spadgett:
@sg00dwin I also had to put in a workaround for the overview:

https://github.com/openshift/origin-web-console/blob/master/app/styles/_overview.less#L122-L126

PR #1372, opened 2017-03-21
Updates to to label filter so that it maintains a large click target …

sg00dwin:
…at empty state, by allowing it to expand, but doesn't use hard widths that causes breakage when combined with select options on overview and other resources.
Tested in Chrome, Firefox, Safari, IE

Fixes https://github.com/openshift/origin-web-console/issues/1365

empty
full


spadgett:
[merge]

PR #1377, opened 2017-03-23
Reducing widths to fix wrapping issues with very long usernames

rhamilto:
I found a bug with very long usernames, so I've adjusted the widths of the username and project picker.  I should have noticed this yesterday, but, alas, is fixed now.


spadgett:
[merge]

PR #1378, opened 2017-03-23
Update to origin-web-common 0.0.7, remove what was made common

jeff-phillips-18:


app/scripts/filters/resources.js
```diff
@@ -238,12 +97,6 @@ angular.module('openshiftConsole')
       return icon;
     };
   })
-  .filter('imageStreamTagIconClass', function(imageStreamTagAnnotationFilter) {
-    return function(resource, /* optional */ tagName) {
-      var icon = imageStreamTagAnnotationFilter(resource, "iconClass", tagName);
-      return (icon) ? icon : "fa fa-cube";
-    };
-  })
   .filter('imageName', function() {
```
dtaylor113: .filter('imageStreamTagTags' is in common and needs to be removed here in console.

dtaylor113:
Hi, I also see /services/base64util.js existing in both repos.

jeff-phillips-18:
Thanks @dtaylor113. Removed those redundancies

spadgett:
It looks like the `mark` style moved to common is not pulled in

jeff-phillips-18:
@spadgett I removed it from _core.less:

https://github.com/openshift/origin-web-console/pull/1378/files#diff-e5367f9b3c5f479f2f08f2b050a22608L166

spadgett:
Right, but the style from common is not picked up

jeff-phillips-18:
It should be picked up from origin-web-common.css:

https://github.com/openshift/origin-web-console/pull/1378/files#diff-b65cb2cce313c83cf5b8352e3da4f343R461

spadgett:
OK I see it, but it's losing to the style in Bootstrap type.less

screen shot 2017-03-23 at 4 27 26 pm

jeff-phillips-18:
So, I guess I leave it in _core.less ? Either that or distribute and build the common repo's less files into the console's css. Thoughts?

spadgett:
I suggest leaving the mark style here for now and opening a follow on issue in the common repo. We might want to pull in the less source files, although this is the only style I see right now that would be a problem.

cc @jwforres

jwforres:
Yeah as a follow-on lets use the less files instead of the css coming from
common.  Not urgent to do right now.

jeff-phillips-18:
Put mark styling back. Added issues:
 https://github.com/openshift/origin-web-common/issues/18
and
#1380

spadgett:
[merge]

spadgett:
> Error: Wait timed out after 2172ms

[merge]

spadgett:
@jeff-phillips-18 Looks like the change might have broken create from URL.

https://docs.openshift.org/latest/dev_guide/create_from_url.html
https://ci.openshift.redhat.com/jenkins/job/test_pull_requests_origin_web_console/1189/s3/

jeff-phillips-18:
@spadgett The create project directive takes a submitButtonLabel parameter. For this particular case it is passed as 'Next'. However, the directive in console did not use the parameter and always used 'Create' for the submit button. The test was written to check for 'Create' so it worked until I fixed this issue in the common directive.

I can either:
a) fix the test (button label will now be 'Next' rather that 'Create')
b) change the use of the directive to use 'Create' for the button label
c) change the directive back to always use 'Create' and remove the option to set the label

Thoughts?

spadgett:
OK thanks for tracking that down @jeff-phillips-18 

"Next" looks a little strange to me on that page with the button positioning ([screenshot][screenshot]). I'm tempted to say leave it as "Create."

@jwforres @rhamilto opinion?

[screenshot]: https://ci.openshift.redhat.com/jenkins/job/test_pull_requests_origin_web_console/1189/s3/download/screenshots/authenticated%20e2e-user%20create%20from%20URL%20using%20a%20template%20supplied%20as%20a%20query%20string%20param%20should%20load%20the%20template%20in%20to%20a%20newly%20created%20project.png

jwforres:
I think saying Next there is misleading, Next doesn't imply to me that it
is going to create the project when I hit that button.

spadgett:
@jeff-phillips-18 My vote is for (c). Agree with @jwforres -- I'm not sure we ever want the directive to say something other than create.

jeff-phillips-18:
I will remove the setting from this case and I will add an issue to remove the setting from the common directive? The default is "Create' so tests will pass and Create will be used as desired.

spadgett:
Sounds good, we'd want to remove the `submit-button-label` attribute anyway if we're removing it from the directive.

spadgett:
LGTM. Typo in commit message, but please squash anyway. Thanks!

spadgett:
[merge]

PR #1382, opened 2017-03-25
Add to project options using the ordering overlay component

jwforres:
Prototype of what it looks like embedding our existing deploy image form into the overlay ordering panel.

This is a bit clunky to interact with until we can actually split these forms into wizards.

cc @serenamarie125 @spadgett

### DIFF
--- a/app/scripts/directives/create/nextSteps.js
+++ b/app/scripts/directives/create/nextSteps.js
@@ -1,9 +1,10 @@
 'use strict';
 
-(function() {  
+(function() {
   angular.module('openshiftConsole').component('nextSteps', {
     controller: [
       'ProcessedTemplateService',
+      'Navigate',
       NextSteps
     ],
     bindings: {
@@ -11,12 +12,13 @@
       projectName: '<',
       loginBaseUrl: '<',
       fromSampleRepo: '<',
-      createdBuildConfig: '<'
+      createdBuildConfig: '<',
+      onContinue: '<'
     },
     templateUrl: 'views/directives/next-steps.html'
   });
 
-  function NextSteps(ProcessedTemplateService) {
+  function NextSteps(ProcessedTemplateService, Navigate) {
     var ctrl = this;
     ctrl.showParamsTable = false;
 
@@ -65,8 +67,11 @@
     ctrl.erroredTasks = erroredTasks;
     ctrl.pendingTasks = pendingTasks;
 
-    ctrl.toggleParamsTable = function() {
-      ctrl.showParamsTable = true;
+    ctrl.goToOverview = function() {
+      if (_.isFunction(ctrl.onContinue)) {
+        ctrl.onContinue();
+      }
+      Navigate.toProjectOverview(ctrl.projectName);
     };
   }
 })();
--- a/app/scripts/directives/fromFileDialog.js
+++ b/app/scripts/directives/fromFileDialog.js
@@ -1,0 +1,95 @@
+'use strict';
+
+(function() {
+  angular.module('openshiftConsole').component('fromFileDialog', {
+    controller: [
+      '$scope',
+      '$timeout',
+      'DataService',
+      FromFileDialog
+    ],
+    controllerAs: '$ctrl',
+    bindings: {
+      visible: '<',
+      project: '<', //handle create project optionally
+      context: '<',
+      onDialogClosed: '&'
+    },
+    templateUrl: 'views/directives/from-file-dialog.html'
+  });
+
+  function FromFileDialog($scope, $timeout, DataService) {
+    var ctrl = this;
+
+    ctrl.$onInit = function() {
+      ctrl.alerts = {};
+      ctrl.loginBaseUrl = DataService.openshiftAPIBaseUrl();
+    };
+
+    function getIconClass() {
+      var icon = _.get(ctrl, 'template.metadata.annotations.iconClass', 'fa fa-cubes');
+      return (icon.indexOf('icon-') !== -1) ? 'font-icon ' + icon : icon;
+    }
+
+    ctrl.importFile = function() {
+      $scope.$broadcast('importFileFromYAMLOrJSON');
+    };
+
+    ctrl.instantiateTemplate = function() {
+      $scope.$broadcast('instantiateTemplate');
+    };
+
+    $scope.$on('fileImportedFromYAMLOrJSON', function(event, message) {
+      ctrl.selectedProject = message.project;
+      ctrl.template = message.template;
+      ctrl.iconClass = getIconClass();
+      // Need to let the current digest loop finish so the template config step becomes visible or the wizard will throw an error
+      // from the change to currentStep
+      $timeout(function() {
+        ctrl.currentStep = ctrl.template ? "Template Configuration" : "Results";
+      },0);
+    });
+
+    $scope.$on('templateInstantiated', function(event, message) {
+      ctrl.selectedProject = message.project;
+      ctrl.currentStep = "Results";
+    });
+
+    ctrl.close = function() {
+      ctrl.template = null;
+      var cb = ctrl.onDialogClosed();
+      if (_.isFunction(cb)) {
+        cb();
+      }
+      ctrl.wizardDone = false;
+      return true;
+    };
+
+    $scope.$on("wizard:stepChanged", function (e, parameters) {
+      if (parameters.step.stepId === 'results') {
+        ctrl.nextButtonTitle = "Close";
+        ctrl.wizardDone = true;
+      } else {
+        ctrl.nextButtonTitle = "Create";
+      }
+    });
+
+    ctrl.currentStep = "JSON / YAML";
+
+    ctrl.nextCallback = function (step) {
+      if (step.stepId === 'file') {
+        ctrl.importFile();
+        return false;  // don't actually navigate yet
+      }
+      else if (step.stepId === 'template') {
+        ctrl.instantiateTemplate();
+        return false;
+      }
+      else if (step.stepId === 'results') {
+        ctrl.close();
+        return false;
+      }
+      return true;
+    };
+  }
+})();
--- a/app/views/directives/deploy-image-dialog.html
+++ b/app/views/directives/deploy-image-dialog.html
@@ -1,0 +1,52 @@
+<overlay-panel show-panel="$ctrl.visible" show-close="true" handle-close="$ctrl.close">
+  <div pf-wizard
+       on-cancel="$ctrl.close()"
+       on-finish="$ctrl.close()"
+       hide-header="true"
+       hide-back-button="true"
+       hide-sidebar="true"
+       next-title="$ctrl.nextButtonTitle"
+       next-callback="$ctrl.nextCallback"
+       current-step="$ctrl.currentStep"
+       step-class="order-service-wizard-step"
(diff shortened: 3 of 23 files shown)


smarterclayton:
Crikey - that scroll is going to make me stab someone.  The moment I hit "want more details" I'm going to want to use the rest of the real estate.   How do we turn "in-context" into "usable with rich options" for end users?

We should vet for each of these flows the expected depth of interaction (successive choices made).  For a lot of templates / builders 1-3 choices is def reasonable.  But the moment I want to go deep, I'm going to get angry if I can't use my screen real estate.

Seems like we need a way to turn the restricted in-context into unrestricted nav in a compelling way.  Is that on the design radar?

jwforres:
Updated to use the full height version of the overlay, stuff still to do:

- [ ] overlay-panel-body overflow causing unnecessary horizontal scroll with forms that use bootstrap row
- [ ] what to do with the Next Steps page, navigating there loses the context which defeats the purpose of doing this in the overlay
- [ ] same question if i do import yaml and it needs to go to the template processing page
- [ ] if import only creates / updates a single resource its using AlertMessageService to show the success message, this only gets used by the overview when it loads, so the message is lost
- [ ] import sometimes not closing on success, its trying to navigate to the overview which is a no-op if you are already on the overview
- [ ] if I use these overlays on other pages, should they navigate to the overview on success? depends on what im doing maybe as to what i might expect?

jwforres:
We are triggering a number of things about template processing from the isDialog flag, the interesting thing is in the context of the Import YAML dialog we have significantly more height in order to account for the ACE editor on the previous step.  So with our default templates you end up in a situation like this one:

process template from import dialog

For reference this is the previous step:

import-yaml-dialog




I can:
1) Reduce the height of the ACE editor when its in a dialog... but I kind of like the height its at now.  Much shorter would make it hard to see the context of the object you are editing if you want to make changes.  There is whitespace beneath the editor we could eliminate, currently it only is helpful if there are syntax errors in your JSON/YAML because the validation error alert is pretty tall.  That will help, but not eliminate the problem.

2) Not do the simple vs advanced modes in this context.  Which would be a little strange that it behaves differently than template processing from the catalog.

3) Leave as is in the screenshot.  Which looks a bit silly.

jwforres:
realized its missing the description for the template and other info that we show in the left column in the catalog, that should help fill out that page.

ncameronbritt:
@spadgett and I were talking about removing the "manage your app" and "command line tools" stuff from the results step of the wizard. 

Given the space constraints of the wizard, I think it makes sense to focus more on the things directly related to what you have just created. The command line tools stuff isn't specifically about what you've just created, but is more general to the project and I worry that having it in the results step could distract from other more pertinent information.

Sam's thought, which I agree with, is that the landing page is probably the place to advertise the command line tools, in that right panel. Within the project, the command line info is in that info menu, which makes sense to me.

@openshift/team-ux-review

app/scripts/directives/fromFileDialog.js
```diff
@@ -0,0 +1,93 @@
...
+function FromFileDialog($scope, $timeout, DataService) {
+  var ctrl = this;
+
+  ctrl.$onInit = function() {
+    ctrl.alerts = {};
+    ctrl.loginBaseUrl = DataService.openshiftAPIBaseUrl();
+  };
+
+  function getIconClass() {
+    var icon = _.get(ctrl, 'template.metadata.annotations.iconClass', 'fa fa-cubes');
```
spadgett: Wondering if we should use the `iconClass` filter, which handles template icons. But it defaults to `fa-clone` for templates. This makes me wonder if we shouldn't use `fa-clone` as our default in the catalog.

I'm OK with this as-is, but we might open an issue to decide what to do since we can show different default icons for the new experience vs. old.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/filters/resources.js#L53

app/styles/_project-menu.less
```diff
@@ -2,6 +2,11 @@
 // Project list menu
 // -----------------------------------------------
 
+// TEMPORARY HACK
```
spadgett: Still needed?

app/scripts/directives/create/nextSteps.js
```diff
@@ -3,19 +3,21 @@
...
     NextSteps
   ],
   bindings: {
     project: '<',
     projectName: '<',
     loginBaseUrl: '<',
     fromSampleRepo: '<',
-    createdBuildConfig: '<'
+    createdBuildConfig: '<',
+    onContinue: '<'
```
spadgett: We're handling the callback differently here than `onDailogClosed` below. Would be good to be consistent.

app/views/directives/deploy-image-dialog.html
```diff
@@ -0,0 +1,53 @@
+<overlay-panel show-panel="$ctrl.visible" show-close="true" handle-close="$ctrl.close">
+  <!-- TODO what to do with alerts? -->
```
spadgett: TODO still needed?

app/views/directives/from-file-dialog.html
```diff
@@ -0,0 +1,95 @@
+<overlay-panel show-panel="$ctrl.visible" show-close="true" handle-close="$ctrl.close" ng-if="$ctrl.project">
+  <!-- TODO what to do with alerts? -->
```
spadgett: TODO still needed?

app/views/directives/header/project-header.html
```diff
@@ -25,10 +25,19 @@
...
+    <div row class="dropdown" ng-if="project.metadata.name | canIAddToProject" uib-dropdown>
+      <a row class="nav-item-iconic dropdown-toggle" href="" ng-disabled="project.status.phase != 'Active'" title="Add to Project" uib-dropdown-toggle>
+        <i class="fa fa-plus visible-xs-inline-block"></i><span class="hidden-xs add-to-project">Add to Project</span><span class="hidden-xs caret" aria-hidden="true"></span>
       </a>
+      <ul class="uib-dropdown-menu dropdown-menu dropdown-menu-right">
+        <li ng-if-start="!catalogLandingPageEnabled"><a ng-href="project/{{projectName}}/create?tab=fromCatalog">Browse Catalog</a></li>
+        <li><a ng-href="project/{{projectName}}/create?tab=deployImage">Deploy Image</a></li>
+        <li ng-if-end><a ng-href="project/{{projectName}}/create?tab=fromFile">Import YAML / JSON</a></li>
+
+        <li ng-if-start="catalogLandingPageEnabled"><a href="/">Browse Catalog</a></li>
```
spadgett: Hm, I thought it should be `href="./"` but it must be working?

app/views/directives/header/project-header.html
```diff
@@ -25,10 +25,19 @@
...
     </div>
     <!-- create buttons show at mobile -->
-    <div row class="navbar-flex-btn project-action" ng-if="project.metadata.name | canIAddToProject">
-      <a row class="project-action-btn" href="project/{{project.metadata.name}}/create" ng-disabled="project.status.phase != 'Active'" title="Add to project">
-        <i class="fa fa-plus visible-xs-inline-block"></i><span class="hidden-xs">Add to project</span>
+    <div row class="dropdown" ng-if="project.metadata.name | canIAddToProject" uib-dropdown>
+      <a row class="nav-item-iconic dropdown-toggle" href="" ng-disabled="project.status.phase != 'Active'" title="Add to Project" uib-dropdown-toggle>
+        <i class="fa fa-plus visible-xs-inline-block"></i><span class="hidden-xs add-to-project">Add to Project</span><span class="hidden-xs caret" aria-hidden="true"></span>
       </a>
+      <ul class="uib-dropdown-menu dropdown-menu dropdown-menu-right">
```
spadgett: Probably should use  `<ul role="menu">` and each `<li role="menuitem">`

app/views/directives/header/project-header.html
```diff
@@ -38,3 +47,6 @@
   </div> <!-- /navbar-project-menu -->
   <navbar-utility class="hidden-xs"></navbar-utility>
 </nav>
+<!-- clean up ordering.* -->
```
spadgett: Not clear on what this means. Is it a TODO?

app/scripts/directives/create/nextSteps.js
```diff
@@ -68,4 +70,11 @@ function NextSteps(ProcessedTemplateService) {
     ctrl.showParamsTable = true;
   };
 
+  ctrl.goToOverview = function() {
+    if (_.isFunction(ctrl.onContinue)) {
+      ctrl.onContinue();
+    }
+    Navigate.toProjectOverview(ctrl.projectName);
```
spadgett: Do we need the `onContinue` handler since we're always navigating away here?

app/scripts/directives/fromFileDialog.js
jwforres: Yeah this is consistent with the catalog and this dialog, for now, doesn't show up in the old experience. I'll open an issue.

app/views/directives/header/project-header.html
jwforres: hmm.. weird yeah its working fine as is

app/views/directives/header/project-header.html
jwforres: at one point it was a todo but i think i changed my mind, will remove comment

jwforres:
updated for the review feedback but i'm seeing a CSS issue at mobile now with my header, probably something related to all the flex changes and merge conflicts i had during rebase

jwforres:
fixed the bad visual issues at mobile, there is one thing I want to have @sg00dwin look at with me tmrw afternoon

app/scripts/directives/create/nextSteps.js
```diff
@@ -1,22 +1,24 @@
...
       NextSteps
     ],
     bindings: {
       project: '<',
       projectName: '<',
       loginBaseUrl: '<',
       fromSampleRepo: '<',
-      createdBuildConfig: '<'
+      createdBuildConfig: '<',
+      onContinue: '<'
```
spadgett: Handle the callback like `onDialogClosed` below?

app/views/directives/deploy-image-dialog.html
```diff
@@ -0,0 +1,52 @@
...
+         substeps="false"
+         ok-to-nav-away="true"
+         allow-click-nav="false">
+      <div class="wizard-pf-main-inner-shadow-covers">
+        <div class="order-service-config-single-column">
+          <div class="wizard-pf-main-form-contents">
+            <next-steps project="$ctrl.selectedProject"
+                        project-name="$ctrl.selectedProject.metadata.name"
+                        login-base-url="$ctrl.loginBaseUrl"
+                        on-continue="$ctrl.close">
```
spadgett: I guess I'm still unclear why this is necessary since on-continue always navigates to the overview anyway?

jwforres: If you are already on the overview then navigating to the overview does nothing. Angular treats it like a no-op, so the overlay never closes.

jwforres:
@spadgett it looked like you hadn't yet implemented toast notifications for deployImage, i'd rather do that separate from this PR I think

otherwise I think this PR is ready

spadgett:
[merge]

jwforres:
updated tests [test]

jwforres:
integration tests passed [merge]

PR #1387, opened 2017-03-29
[WIP] Add serviceInstanceRow, extract common func from OverviewListRow

benjaminapetersen:
Add serviceInstanceRow component, extract common functionality & view fragments from OverviewListRow:
- rowMethods service to share methods & dedupe
- chevron view partial

Def WIP, but opening to get opinions from @spadgett and @jwforres on the method I'm using to share bits of the OverviewListRow & avoid copy/paste of some of these functions.

### DIFF
--- a/app/scripts/controllers/newOverview.js
+++ b/app/scripts/controllers/newOverview.js
@@ -93,6 +93,7 @@ function OverviewController($scope,
     recentPipelinesByDeploymentConfig: {},
     routesByService: {},
     servicesByObjectUID: {},
+    serviceInstances: {},
     // Set to true below when metrics are available.
     showMetrics: false
   };
@@ -156,7 +157,8 @@ function OverviewController($scope,
            _.size(overview.deployments) +
            _.size(overview.vanillaReplicaSets) +
            _.size(overview.statefulSets) +
-           _.size(overview.monopods);
+           _.size(overview.monopods) +
+           _.size(overview.state.serviceInstances);
   };
 
   // The size of all visible top-level items after filtering.
@@ -166,7 +168,8 @@ function OverviewController($scope,
            _.size(overview.filteredDeployments) +
            _.size(overview.filteredReplicaSets) +
            _.size(overview.filteredStatefulSets) +
-           _.size(overview.filteredMonopods);
+           _.size(overview.filteredMonopods) +
+           _.size(overview.state.serviceInstances);
   };
 
   // Show the "Get Started" message if the project is empty.
@@ -183,7 +186,8 @@ function OverviewController($scope,
                  overview.deployments &&
                  overview.replicaSets &&
                  overview.statefulSets &&
-                 overview.pods;
+                 overview.pods &&
+                 overview.state.serviceInstances;
 
     state.expandAll = loaded && overview.size === 1;
 
@@ -1061,6 +1065,33 @@ function OverviewController($scope,
       });
   };
 
+  var refreshSecrets = _.debounce(function(context) {
+    DataService.list("secrets", context, null, { errorNotification: false }).then(function(secretData) {
+      state.secrets = secretData.by("metadata.name");
+    });
+  }, 300);
+
+  // TODO: code duplicated from directives/bindService.js
+  // extract & share 
+  var sortServiceInstances = function() {
+    if(!state.serviceInstances && !state.serviceClasses) {
+      return;
+    }
+    state.orderedServiceInstances = _.toArray(state.serviceInstances).sort(function(left, right) {
+      var leftName = _.get(state.serviceClasses, [left.spec.serviceClassName, 'osbMetadata', 'displayName']) || left.spec.serviceClassName;
+      var rightName = _.get(state.serviceClasses, [left.spec.serviceClassName, 'osbMetadata', 'displayName']) || right.spec.serviceClassName;
+
+      // Fall back to sorting by `metadata.name` if the display names are the
+      // same so that the sort is stable.
+      if (leftName === rightName) {
+        leftName = _.get(left, 'metadata.name', '');
+        rightName = _.get(right, 'metadata.name', '');
+      }
+
+      return leftName.localeCompare(rightName);
+    });
+  };
+
   var watches = [];
   ProjectsService.get($routeParams.project).then(_.spread(function(project, context) {
     // Project must be set on `$scope` for the projects dropdown.
@@ -1221,6 +1252,8 @@ function OverviewController($scope,
         resource: 'instances'
       }, context, function(serviceInstances) {
         state.serviceInstances = serviceInstances.by('metadata.name');
+        sortServiceInstances();
+        updateFilter();
       }, {poll: limitWatches, pollInterval: DEFAULT_POLL_INTERVAL}));
     }
 
@@ -1228,8 +1261,10 @@ function OverviewController($scope,
       watches.push(DataService.watch({
         group: 'servicecatalog.k8s.io',
         resource: 'bindings'
-      }, context, function(serviceBindings) {
-        state.serviceBindings = serviceBindings.by('metadata.name');
+      }, context, function(bindings) {
+        state.bindings = bindings.by('metadata.name');
+        refreshSecrets(context);
+        updateFilter();
       }, {poll: limitWatches, pollInterval: DEFAULT_POLL_INTERVAL}));
     }
 
@@ -1247,6 +1282,8 @@ function OverviewController($scope,
         resource: 'serviceclasses'
       }, context, function(serviceClasses) {
         state.serviceClasses = serviceClasses.by('metadata.name');
+        sortServiceInstances();
+        updateFilter();
       });
     }
 
--- a/app/scripts/directives/overview/listRow.js
+++ b/app/scripts/directives/overview/listRow.js
@@ -8,6 +8,7 @@ angular.module('openshiftConsole').component('overviewListRow', {
     'BuildsService',
     'DeploymentsService',
     'Navigate',
+    'ListRowUtils',
     OverviewListRow
   ],
   controllerAs: 'row',
@@ -27,8 +28,11 @@ function OverviewListRow($filter,
                          APIService,
                          BuildsService,
                          DeploymentsService,
-                         Navigate) {
+                         Navigate,
+                         rowMethods) {
   var row = this;
+
+  _.extend(row, rowMethods.ui);
 
   var canI = $filter('canI');
   var deploymentIsInProgress = $filter('deploymentIsInProgress');
@@ -70,14 +74,6 @@ function OverviewListRow($filter,
     }
   };
 
-  var getNotifications = function(object) {
-    var uid = _.get(object, 'metadata.uid');
-    if (!uid) {
-      return null;
-    }
-    return _.get(row, ['state', 'notificationsByObjectUID', uid]);
-  };
-
   // Return the same empty array each time. Otherwise, digest loop errors occur.
   var NO_HPA = [];
   var getHPA = function(object) {
@@ -95,54 +91,9 @@ function OverviewListRow($filter,
     return _.get(row.state.hpaByResource, [kind, name], NO_HPA);
   };
 
-  var expandedKey = function(apiObject) {
-    var uid = _.get(apiObject, 'metadata.uid');
-    if (!uid) {
-      return null;
-    }
-
-    return 'overview/expand/' + uid;
-  };
-
-  row.toggleExpand = function(e, always) {
-    // Don't toggle if clicking on a link inside the row unless `always` is set.
-    if (!always && $(e.target).closest("a").length > 0) {
-      return;
-    }
-
-    var key = expandedKey(row.apiObject);
-    if (!key) {
-      return;
-    }
-
-    row.expanded = !row.expanded;
-    sessionStorage.setItem(key, row.expanded ? 'true' : 'false');
-  };
-
-  var setInitialExpandedState = function() {
-    var key = expandedKey(row.apiObject);
-    if (!key) {
-      row.expanded = false;
-      return;
-    }
-
-    var item = sessionStorage.getItem(key);
-    if (!item && row.state.expandAll) {
-      row.expanded = true;
-      return;
-    }
-
-    row.expanded = item === 'true';
-  };
-
-  row.$onInit = function() {
-    _.set(row, 'selectedTab.networking', true);
-    setInitialExpandedState();
-  };
-
   row.$doCheck = function() {
     // Update notifications.
-    row.notifications = getNotifications(row.apiObject);
+    row.notifications = rowMethods.getNotifications(row.apiObject, row);
 
     // Update HPA.
     row.hpa = getHPA(row.apiObject);
--- a/app/scripts/directives/overview/serviceInstanceRow.js
+++ b/app/scripts/directives/overview/serviceInstanceRow.js
@@ -1,0 +1,104 @@
+'use strict';
+
+angular.module('openshiftConsole').component('serviceInstanceRow', {
+  controller: [
+    '$filter',
+    'DataService',
+    'ListRowUtils',
+    '$uibModal',
+    ServiceInstanceRow
+  ],
+  controllerAs: 'row',
+  bindings: {
+    apiObject: '<',
+    state: '<'
+  },
+  templateUrl: 'views/overview/_service-instance-row.html'
+});
+
+function ServiceInstanceRow($filter, DataService, rowMethods, $uibModal) {
+  var row = this;
+  _.extend(row, rowMethods.ui);
+
+  var getErrorDetails = $filter('getErrorDetails');
+
+  var getDisplayName = function() {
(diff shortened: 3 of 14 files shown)


benjaminapetersen:
For a visual:

screen shot 2017-03-29 at 4 32 35 pm

@ncameronbritt keep you in the loop!

benjaminapetersen:
The alternative approach would be to do something like:

```html
<overview-list-row>
  <node-with-pods></node-with-pods> <!-- or whatever generic -->
</overview-list-row>

<overview-list-row>
  <service-instance-nodes></service-instance-nodes>
</overview-list-row>
```
Might reduce the use of includes, but would require more passing of attribs (unless the generic list row shell didnt care about data, just about open,close,collapse).

app/scripts/services/rowMethods.js
```diff
@@ -0,0 +1,51 @@
...
+        // Don't toggle if clicking on a link inside the row unless `always` is set.
+        if (!always && $(e.target).closest("a").length > 0) {
+          return;
+        }
+
+        var key = expandedKey(this.apiObject);
+        if (!key) {
+          return;
+        }
+        console.log('we togglin');
```
spadgett: Can we toast this message?

app/scripts/services/rowMethods.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .factory('rowMethods', function() {
```
spadgett: Since almost all of this is controlling remembering expand/collapse state, I wonder a generic service we could use for that outside the overview.

app/scripts/services/rowMethods.js
benjaminapetersen: hopefully a toast and a dancing raptor to sing a song 😄

app/scripts/services/rowMethods.js
benjaminapetersen: sounds good to me. a generic storage state service, or just for expand/collapse? i assume we may eventually want to save other state.

spadgett: Maybe just expand/collapse? We have some services already saving state in local/session storage like AlertsMessageService

app/scripts/services/rowMethods.js
jwforres: stealing one of @fabianofranz 's favorites https://www.youtube.com/watch?v=8oP6xwzV7OQ

app/scripts/services/rowMethods.js
```diff
@@ -0,0 +1,51 @@
...
+
+    var setInitialExpandedState = function(row) {
+      var key = expandedKey(row.apiObject);
+      if (!key) {
+        row.expanded = false;
+        return;
+      }
+
+      var item = sessionStorage.getItem(key);
+      if (!item && row.state.expandAll) {
```
benjaminapetersen: is `expandAll` used anymore? I'm not sure I see it referenced anywhere else.

spadgett:
```
app/scripts/controllers/newOverview.js
188:    state.expandAll = loaded && overview.size === 1;
```

benjaminapetersen:
Still trouble creating bindings, but pushed up some more. :frowning:

benjaminapetersen:
Have bindings now, just working out some cleanup w/the way our rows use flex. @rhamilto did I hear you may have done some cleanup work on new overview? Want to chime in before I hack away?

- [ ] most of the links are using `navigateResourceURL` atm, which is probably not idea, esp since I took `Binding` out of the blacklist.  
- promoted instances to top level items in the overview. They will list instead of the "Getting Started" message, even if they are the only items.

screen shot 2017-04-05 at 4 32 35 pm
screen shot 2017-04-05 at 4 33 23 pm

rhamilto:
> @rhamilto did I hear you may have done some cleanup work on new overview? 

Not on my plate given the work remaining on catalog, but @sg00dwin may be able to help.

benjaminapetersen:
@rhamilto no worries, was more curious if you had prev done anything & had a direction going for this.  I have something in place that works.

benjaminapetersen:
Updating a few items from @ncameronbritt, will push shortly.

app/scripts/services/rowMethods.js
benjaminapetersen: The only odd thing is `setInitialExpandedState` takes a "row" and sets `row.expanded` and `row.state.expandAll`.  Hmm.

app/scripts/controllers/newOverview.js
```diff
@@ -1221,15 +1225,22 @@ function OverviewController($scope,
         resource: 'instances'
       }, context, function(serviceInstances) {
         state.serviceInstances = serviceInstances.by('metadata.name');
+        updateFilter();
       }, {poll: limitWatches, pollInterval: DEFAULT_POLL_INTERVAL}));
     }
 
     if(Constants.ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page) {
+
+      DataService.list("secrets", context, null, { errorNotification: false }).then(function(secretData) {
```
jwforres: need to think about how we are loading these, any time you create a new binding that is going to create new secrets.  This probably either needs to be a limited watch, or we need to refresh the list any time we see an update to bindings (debounced so it doesn't spam the server with requests if there are a lot binding updates at once).  I'm leaning towards the latter.  @spadgett thoughts?

app/scripts/directives/overview/listRow.js
```diff
@@ -27,9 +28,12 @@ function OverviewListRow($filter,
                          APIService,
                          BuildsService,
                          DeploymentsService,
-                         Navigate) {
+                         Navigate,
+                         rowMethods) {
```
jwforres: what about calling this something like ListRowUtils instead of rowMethods ?

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -0,0 +1,101 @@
...
+
+function ServiceInstanceRow($filter, DataService, rowMethods, $uibModal) {
+  var row = this;
+  _.extend(row, rowMethods.ui);
+
+  // returns the best user-friendly name available for the instance
+  var getDisplayName = function() {
+    var serviceClassName = row.apiObject.spec.serviceClassName;
+    var instanceName = row.apiObject.metadata.name;
+    var serviceClassDisplayName = _.get(row, ['state','serviceClasses', serviceClassName, 'OSBMetadata', 'displayName']);
```
jwforres: osbMetadata

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -0,0 +1,101 @@
...
+        modalConfig: function() {
+          return {
+            alerts: {
+              deprovision: {
+                type: 'error',
+                message: 'Service \'' + row.apiObject.spec.serviceClassName + '\' will be deleted and no longer available.'
+              }
+            },
+            detailsMarkup: 'Deprovision Service?',
+            okButtonText: 'Deprovision Service',
```
jwforres: thats a lot of text for an ok button, probably just Deprovision

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -0,0 +1,101 @@
...
+    })
+    .result.then(function() {
+      DataService.delete({
+        group: 'servicecatalog.k8s.io',
+        resource: 'instances'
+      },
+      row.apiObject.metadata.name,
+      { namespace: row.apiObject.metadata.namespace })
+      .then(function(resp) {
+        console.log('instances delete success/deprovision', resp);
```
jwforres: clean up these two console.logs and handle the two cases correctly.  success should just be to close the dialog and add an inline alert to the page for now for consistency with our other actions.  failure should be showing a message in the dialog

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -0,0 +1,101 @@
...
+      }, function(err) {
+        console.log('instances delete err', err);
+      });
+    });
+  };
+
+
+  // TODO: MVP may/may not have these actions.
```
jwforres: go ahead and remove these comments

app/views/new-overview.html
```diff
@@ -355,6 +355,31 @@ <h2 ng-if="!app">
...
+                    states: pending, failed to provision, deleting/deprovisioning...
+                    todo: try to pretty name:
+                          serviceClassDisplayName then
+                          serviceClassName then
+                          serviceInstanceName
+                          try to be pretty, if 2x+ of the same, will neeed the non-pretty
+                          name below... check specs and stuff.
+                  -->
+                  <service-instance-row
+                    ng-repeat="serviceInstance in overview.state.serviceInstances"
```
jwforres: we have started stable sorting most of our lists, i believe overview is sorted alphabetically @spadgett ?  You'll want to leave state.serviceInstances as is (my bind dialog is expecting it) and then have an orderedServiceInstances

You can rip off the code I used for the bind dialog to sort service instances based on their pretty name https://github.com/openshift/origin-web-console/pull/1395/files#diff-9b6377c25f28da45ed50c1d71ef2863aR60

you'll want to wait to sort (and show) them till both service classes and service instances are loaded, otherwise stuff will jump around

app/views/overview/_service-instance-row.html
```diff
@@ -0,0 +1,133 @@
...
+  <div class="list-pf-container" ng-click="row.toggleExpand($event)">
+
+    <div class="list-pf-chevron">
+      <div ng-include src=" 'views/overview/_list-row-chevron.html' " class="list-pf-content"></div>
+    </div>
+
+    <div class="list-pf-content">
+      <div class="list-pf-name">
+        <h3>
+          <a ng-href="{{row.apiObject | navigateResourceURL}}">
```
jwforres: i'm thinking about not making this a link right now, I know it will send you to Other Resources but thats a really crappy experience, and i dont think we even want people doing Edit YAML on service instances right now, and we already have Deprovision from the overview

app/views/overview/_service-instance-row.html
```diff
@@ -0,0 +1,133 @@
...
+            <span ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></span>
+          </a>
+          <div class="component-label" style="padding-top: 4px;">
```
jwforres: get rid of this inline style, if 5px will look ok then use the pad-top-sm, if not move this to the CSS

app/views/overview/_service-instance-row.html
```diff
@@ -0,0 +1,133 @@
...
+        </h3>
+        <div class="status-icons">
+          <notification-icon ng-if="!row.expanded" alerts="row.notifications"></notification-icon>
+        </div>
+      </div>
+    </div>
+
+    <div
+      class="list-pf-content hidden-xs hidden-sm">
+      <div style="width: 300px">
```
jwforres: we shouldn't be setting a fixed width here, you should be able to use bootstrap row / col classes if needed, that was the point of using the new list-pf

app/views/overview/_service-instance-row.html
```diff
@@ -0,0 +1,133 @@
...
+        <div>
+          <strong ng-if="!row.instanceBindings.length">No Bindings</strong>
```
jwforres: i know @ncameronbritt was thinking about changing this to a Create Binding link, but i guess that is dependent on me finishing the other direction of binding so this is fine for this PR.

app/views/overview/_service-instance-row.html
```diff
@@ -0,0 +1,133 @@
...
+          <strong ng-if="row.instanceBindings.length">Bindings</strong>
+        </div>
+        <a g-if="row.instanceBindings.length >= 1" href="{{row.getSecretForBinding(row.instanceBindings[0]) | navigateResourceURL}}">{{row.instanceBindings[0].metadata.name}}</a><span ng-if="row.instanceBindings.length > 1">,&nbsp;</span>
```
jwforres: same point about other resources, we may want to just not make this a link right now

app/views/overview/_service-instance-row.html
```diff
@@ -0,0 +1,133 @@
...
+        </div>
+        <span ng-if="!row.instanceBindings.length">There are no bindings.</span>
+        <div
+          ng-if="row.instanceBindings.length"
+          class="row"
+          ng-repeat="binding in row.instanceBindings">
+          <div class="col-sm-4">
+
+            <a href="{{row.getSecretForBinding(binding) | navigateResourceURL}}">
+              {{binding.metadata.name}}
```
jwforres: it seems weird for the link for the binding to go to the secret, maybe a second link for "View Secret" that navigates to the secret?

app/views/overview/_service-instance-row.html
benjaminapetersen: Yeah, we don't have fully baked patterns for flex in these lists yet, but I'll move this to a class or nest a bootstrap class.

app/scripts/directives/overview/serviceInstanceRow.js
benjaminapetersen: I think this is in the [design docs](https://docs.google.com/document/d/1w2Tor1hgJiNfCClr3q7hBmW6B5zJAqs9saZHn5OJoLY/edit#heading=h.pem26cddew12), but I agree, one word is prob sufficient.

app/views/overview/_service-instance-row.html
benjaminapetersen: Yeah that seems misleading/cause confusion.  So list the binding, no link, second 'view secret' link seems much better. I'll update.

app/scripts/controllers/newOverview.js
benjaminapetersen: Hmm... so `DEFAULT_POLL_INTERVAL` is 60 seconds.  I like being consistent with our polling if we don't watch, but that seems a long time to wait if things change.  Do we want to do the logic of #2 or have a 2nd slightly more frequent poll interval?

app/scripts/directives/overview/listRow.js
benjaminapetersen: updated

app/scripts/directives/overview/serviceInstanceRow.js
benjaminapetersen: got it

app/scripts/controllers/newOverview.js
benjaminapetersen: - [ ]TODO:  refresh secrets whenever bindings change

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -0,0 +1,101 @@
...
+    row.displayName = getDisplayName();
+    row.instanceBindings = getBindings();
+  };
+
+  row.getSecretForBinding = function(binding) {
+    return binding && row.state.secrets[binding.spec.secretName];
+  }
+
+
+  row.deprovision = function() {
```
benjaminapetersen: @jwforres @ncameronbritt Q... so deprovision is pretty destructive.  Do we want to do the same thing that we do for deleting a project?  IE, type in the name of the instance before we let you go forward?  Far as I understand, there is no going back, so if you deprovision something like a database you might 😢 😭 .

jwforres:
the name is going to be pretty painful to type in, the service instance
name is pretty gorpy.  don't worry about it for this PR, but we can think
about it for a follow-on

app/scripts/directives/overview/serviceInstanceRow.js
benjaminapetersen: done

app/views/overview/_service-instance-row.html
benjaminapetersen: done

app/views/overview/_service-instance-row.html
benjaminapetersen: done

app/views/overview/_service-instance-row.html
benjaminapetersen: Follow-on?

jwforres: yep

app/scripts/directives/overview/serviceInstanceRow.js
benjaminapetersen: Do we keep the dialog open & show an error elsewhere?  I'm searching but haven't found that.  Seems like it might be a little strange to replace the warning message with a new message:

screen shot 2017-04-07 at 3 27 18 pm

jwforres: Not sure, check what delete link does?

app/views/new-overview.html
benjaminapetersen: done

app/views/overview/_service-instance-row.html
benjaminapetersen: done

app/views/overview/_service-instance-row.html
benjaminapetersen: done

jwforres:
[merge]

PR #1389, opened 2017-03-31
Add catalog to web console

spadgett:
screen shot 2017-04-07 at 5 37 27 pm

@jwforres @jeff-phillips-18 FYI, not ready to merge, but it's working


spadgett:
Requires https://github.com/openshift/origin-web-catalog/pull/73

app/scripts/services/navigate.js
```diff
@@ -70,6 +70,10 @@ angular.module("openshiftConsole")
         return "project/" + encodeURIComponent(projectName) + "/overview";
       },
 
+      toProjectList: function(){
+        $location.path('projects');
```
jwforres: this should always work whether catalog is turned on or not, maybe we should add a redirect from projects to / when the landing page isn't enabled

spadgett: ok

app/scripts/app.js
```diff
@@ -51,11 +52,23 @@ angular
...
     }
 
+    var landingPageRoute;
+    var projectsPageRoute = {
+      templateUrl: 'views/projects.html',
+      controller: 'ProjectsController'
+    };
+    if (_.get(window, 'OPENSHIFT_CONSTANTS.ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page')) {
+      landingPageRoute = {
+        templateUrl: 'views/landing-page.html',
```
jeff-phillips-18: Might want to use home.html rather than landing-page.html just for the look of the URL to users.

spadgett: The url should just be `/console`. Although maybe I shouldn't name the controller the same as the component

jeff-phillips-18: sorry, read it wrong.

jwforres: users should never see this URL, our templates are embedded into $templateCache at build time

jwforres:
[merge]

PR #1392, opened 2017-04-03
Update masthead background color and border/inset colors.

sg00dwin:
New masthead hex color #1c2127
Slight adjustments to the border and inset colors so they maintain the current level of contrast between themselves and the background color.

**New header**
new-closed
new-open

**Old header**
old-closed
old-open


sg00dwin:
@rhamilto @jwforres @spadgett

rhamilto:
Nice job, @sg00dwin!  LGTM.  

We'll want to double check this against #1382, however.   I don't think there will be a problem based on a cursory look at both, but I may have missed something...

jwforres:
I'm without question going to have to rebase #1382 so don't hold this out on that, if there is an issue we can tweak it there.

jwforres:
[merge]

PR #1401, opened 2017-04-06
Assemble script changes for split console based on httpd

jwforres:


jwforres:
@spadgett i think we should merge this version that works with the tagged 2.4 image

jwforres:
[merge]

PR #1410, opened 2017-04-10
Fix the binding list update failure in service instance rows on the overview

benjaminapetersen:
- `$onChanges` does not see properties of object `row.state` change
- moving `getBindings` to `$doCheck` solves the problem
- this does call the function on every digest loop, but it does not trigger infinite digests
   - it is possible that storing a temp reference to the old bindings & testing against the new before assigning to `row. instanceBindings` could be slightly more performant as it would ensure no DOM updates, but that may be overkill in this situation.

### DIFF
--- a/app/scripts/controllers/newOverview.js
+++ b/app/scripts/controllers/newOverview.js
@@ -94,6 +94,7 @@ function OverviewController($scope,
     routesByService: {},
     servicesByObjectUID: {},
     serviceInstances: {},
+    bindingsByInstanceRef: {},
     // Set to true below when metrics are available.
     showMetrics: false
   };
@@ -1072,7 +1073,7 @@ function OverviewController($scope,
   }, 300);
 
   // TODO: code duplicated from directives/bindService.js
-  // extract & share 
+  // extract & share
   var sortServiceInstances = function() {
     if(!state.serviceInstances && !state.serviceClasses) {
       return;
@@ -1252,6 +1253,10 @@ function OverviewController($scope,
         resource: 'instances'
       }, context, function(serviceInstances) {
         state.serviceInstances = serviceInstances.by('metadata.name');
+        _.each(state.serviceInstances, function(instance) {
+          var notifications = ResourceAlertsService.getServiceInstanceAlerts(instance);
+          setNotifications(instance, notifications);  
+        });
         sortServiceInstances();
         updateFilter();
       }, {poll: limitWatches, pollInterval: DEFAULT_POLL_INTERVAL}));
@@ -1263,6 +1268,7 @@ function OverviewController($scope,
         resource: 'bindings'
       }, context, function(bindings) {
         state.bindings = bindings.by('metadata.name');
+        overview.bindingsByInstanceRef = _.groupBy(state.bindings, 'spec.instanceRef.name');
         refreshSecrets(context);
         updateFilter();
       }, {poll: limitWatches, pollInterval: DEFAULT_POLL_INTERVAL}));
--- a/app/scripts/directives/overview/serviceInstanceRow.js
+++ b/app/scripts/directives/overview/serviceInstanceRow.js
@@ -11,7 +11,8 @@ angular.module('openshiftConsole').component('serviceInstanceRow', {
   controllerAs: 'row',
   bindings: {
     apiObject: '<',
-    state: '<'
+    state: '<',
+    bindings: '<'
   },
   templateUrl: 'views/overview/_service-instance-row.html'
 });
@@ -21,7 +22,7 @@ function ServiceInstanceRow($filter, DataService, rowMethods, $uibModal) {
   _.extend(row, rowMethods.ui);
 
   var getErrorDetails = $filter('getErrorDetails');
-
+  
   var getDisplayName = function() {
     var serviceClassName = row.apiObject.spec.serviceClassName;
     var instanceName = row.apiObject.metadata.name;
@@ -34,17 +35,10 @@ function ServiceInstanceRow($filter, DataService, rowMethods, $uibModal) {
     return _.get(row, ['state','serviceClasses', serviceClassName, 'description']);
   };
 
-  var getBindings = function() {
-    return _.filter(row.state.bindings, function(binding) {
-      return binding.spec.instanceRef.name === row.apiObject.metadata.name;
-    });
-  };
-
   row.$onChanges = function() {
     row.notifications = rowMethods.getNotifications(row.apiObject, row.state);
     row.displayName = getDisplayName();
     row.description = getDescription();
-    row.instanceBindings = getBindings();
   };
 
   row.getSecretForBinding = function(binding) {
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -1084,6 +1084,8 @@ angular.module('openshiftConsole')
       return lastFinishTime;
     };
   })
+  // gets the status condition that matches provided type
+  // statusCondition(object, 'Ready')
   .filter('statusCondition', function() {
     return function(apiObject, type) {
       if (!apiObject) {
@@ -1092,6 +1094,14 @@ angular.module('openshiftConsole')
 
       return _.find(_.get(apiObject, 'status.conditions'), {type: type});
     };
+  })
+  .filter('isServiceInstanceReady', function(statusConditionFilter) {
+    return function(apiObject) {
+      return _.get(statusConditionFilter(apiObject, 'Ready'), 'status') === 'True';
+    };
+  })
+  .filter('isBindingReady', function(isServiceInstanceReadyFilter) {
+    return isServiceInstanceReadyFilter;
   })
   .filter('routeIngressCondition', function() {
     return function(ingress, type) {
--- a/app/scripts/services/resourceAlerts.js
+++ b/app/scripts/services/resourceAlerts.js
@@ -147,10 +147,39 @@ angular.module("openshiftConsole")
       return alerts;
     };
 
+    var makeConditionAlert = function(alerts, uid, condition, type) {
+      alerts[uid+'-'+condition.reason] = {
+        type: type,
+        message: condition.message
+      };
+    };
+
+    var getServiceInstanceAlerts = function(instance) {
+      var alerts = {};
+      if(!instance) {
+        return alerts;
+      }
+      var uid = instance.metadata.uid;
+      var namespaceError = _.find(instance.status.conditions, {reason: 'ErrorFindingNamespaceForInstance'});
+      var provisionFail = _.find(instance.status.conditions, {reason: 'ProvisionFailed'});
+      var deprovisionFail = _.find(instance.status.conditions, {reason: 'DeprovisioningFailed'});
+
+      if(namespaceError) {
+        makeConditionAlert(alerts, uid, namespaceError, 'warning');
+      }
+      if(provisionFail) {
+        makeConditionAlert(alerts, uid, provisionFail, 'error');
+      }
+      if(deprovisionFail) {
+        makeConditionAlert(alerts, uid, deprovisionFail, 'error');
+      }
+      return alerts;
+    };
+
     return {
       getPodAlerts: getPodAlerts,
       setGenericQuotaWarning: setGenericQuotaWarning,
-      getDeploymentStatusAlerts: getDeploymentStatusAlerts
+      getDeploymentStatusAlerts: getDeploymentStatusAlerts,
+      getServiceInstanceAlerts: getServiceInstanceAlerts
     };
   });
-
--- a/app/views/new-overview.html
+++ b/app/views/new-overview.html
@@ -365,6 +365,7 @@
                   <service-instance-row
                     ng-repeat="serviceInstance in overview.state.orderedServiceInstances"
                     api-object="serviceInstance"
+                    bindings="overview.bindingsByInstanceRef[serviceInstance.metadata.name]"
                     state="overview.state"></service-instance-row>
                 </div>
               </div>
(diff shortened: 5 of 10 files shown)


benjaminapetersen:
Eh, updated & added the check anyway.  Better be safe.

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -44,7 +44,15 @@ function ServiceInstanceRow($filter, DataService, rowMethods, $uibModal) {
     row.notifications = rowMethods.getNotifications(row.apiObject, row.state);
     row.displayName = getDisplayName();
     row.description = getDescription();
-    row.instanceBindings = getBindings();
+  };
+
+  var prevBindings = [];
+  row.$doCheck = function() {
+    var bindings = getBindings();
+    if(prevBindings.length !== bindings.length) {
```
jwforres: a length check isn't sufficient, you'll lose updates to the status of a binding

spadgett: Instead of having to filter `row.state.bindings` every digest loop, can we keep a map of bindings by instance ref name updated in the overview watch callback? Then `$doCheck` might not be necessary.

benjaminapetersen: @jwforres this just hits adding the additional binding row, I'm working on looking at the statuses now.

@spadgett i can do that, assuming that we would want something like this then:
```html 
<service-instance-row
  api-object="serviceInstance"
  bindings="bindingsByRef[serviceInstance.metadata.name]"></service-instance-row>
```

spadgett: @benjaminapetersen that works for me and should perform better

benjaminapetersen: && `overview.state.bindingsByInstanceRef`, where the other maps seem to be.

benjaminapetersen:
@jwforres @spadgett updated, this includes a couple filter additions that I'm using for other updates coming to the service instances.

benjaminapetersen:
Fixes the following:
- [x] bindings will now update on overview after "bind" action
- [x] secret link will not appear until the binding action completes, will show "Pending"

app/scripts/controllers/newOverview.js
```diff
@@ -1263,6 +1264,13 @@ function OverviewController($scope,
         resource: 'bindings'
       }, context, function(bindings) {
         state.bindings = bindings.by('metadata.name');
+        state.bindingsByInstanceRef = _.reduce(state.bindings, function(result, next) {
```
spadgett: Can this just be

```js
state.bindingsByInstanceRef = _.indexBy(state.bindings, 'spec.instanceRef.name');
```

app/scripts/filters/resources.js
```diff
@@ -1093,6 +1105,13 @@ angular.module('openshiftConsole')
       return _.find(_.get(apiObject, 'status.conditions'), {type: type});
     };
   })
+  // returns true/false based on the presence of a Ready status conditions
+  // statusConditionReady(obj)
+  .filter('statusConditionReady', function(statusConditionFilter) {
```
spadgett: Nit: I might give this a name to make it clear it's specific to service catalog items. Maybe `isInstanceOrBindingReady` since it doesn't apply to other types (a bit long I know)

app/views/new-overview.html
```diff
@@ -365,6 +365,7 @@ <h2 ng-if="overview.state.serviceInstances">
                   <service-instance-row
                     ng-repeat="serviceInstance in overview.state.orderedServiceInstances"
                     api-object="serviceInstance"
+                    bindings="overview.state.bindingsByInstanceRef[serviceInstance.metadata.name]"
```
spadgett: Generally the items in `state` are meant to be accessed directly by the row. So maybe `bindingsByInstanceRef` should just be on `overview` directly.

But I'm OK either way.

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -44,13 +46,15 @@ function ServiceInstanceRow($filter, DataService, rowMethods, $uibModal) {
...
     row.description = getDescription();
-    row.instanceBindings = getBindings();
   };
 
   row.getSecretForBinding = function(binding) {
     return binding && _.get(row, ['state', 'secrets', binding.spec.secretName]);
   };
 
+  row.showSecretLink = function(binding) {
+    return statusConditionReady(binding);
```
spadgett: I think it's more clear if you just call the filter directly in the view. I was confused why `!showSecretLink()` means show a pending icon, whereas

```html
          <span ng-if="!(binding | statusConditionReady)">
             <status-icon status="'Pending'"></status-icon> Pending
          </span>
```

is pretty clear

app/views/overview/_service-instance-row.html
```diff
@@ -93,19 +93,24 @@
...
           <!-- TODO: follow-on PR
           <a ng-href="{{binding | navigateResourceURL}}">View configuration details</a>
           -->
-          <a href="{{row.getSecretForBinding(binding) | navigateResourceURL}}">
+          <span ng-if="!row.showSecretLink(binding)">
+            <status-icon status="'Pending'"></status-icon> Pending
+          </span>
+          <a
+            ng-if="row.showSecretLink(binding)"
+            href="{{row.getSecretForBinding(binding) | navigateResourceURL}}">
```
spadgett: Probably worth a `canI` check if you can view secrets

app/views/new-overview.html
benjaminapetersen: Ok cool, I can move that.

app/scripts/directives/overview/serviceInstanceRow.js
benjaminapetersen: Agree, I'll update.  I originally had the filter logic in that function, factored it out to a filter but should have also made this change.

app/scripts/filters/resources.js
benjaminapetersen: Long, but works, I'll update.

spadgett: You could also have two names for the filter. Call this `isServiceInstanceReady` and then have an `isBindingReady` filter that returns `isServiceInstanceReadyFilter`.

These might be useful in origin-web-common, but don't worry about that for this PR.

benjaminapetersen:
@spadget i think i got all of the items, but doing a little more testing.  Having trouble with my instances getting stuck.

app/scripts/filters/resources.js
```diff
@@ -1093,6 +1104,16 @@ angular.module('openshiftConsole')
       return _.find(_.get(apiObject, 'status.conditions'), {type: type});
     };
   })
+  // returns true/false based on the presence of a Ready status conditions
+  // statusConditionReady(obj)
```
spadgett: Looks like the comment has the old filter name.

app/scripts/filters/resources.js
```diff
@@ -1093,6 +1104,16 @@ angular.module('openshiftConsole')
...
+  .filter('isServiceInstanceReady', function(statusConditionFilter) {
+    return function(apiObject) {
+      return _.get(statusConditionFilter(apiObject, 'Ready'), 'status') === 'True';
+    };
+  })
+  .filter('isBindingReady', function(isServiceInstanceReadyFilter) {
```
spadgett: 👍 I think these names read a lot better, thanks

app/views/overview/_service-instance-row.html
```diff
@@ -93,19 +93,24 @@
...
           <!-- TODO: follow-on PR
           <a ng-href="{{binding | navigateResourceURL}}">View configuration details</a>
           -->
-          <a href="{{row.getSecretForBinding(binding) | navigateResourceURL}}">
+          <span ng-if="!(binding | isBindingReady)">
+            <status-icon status="'Pending'"></status-icon> Pending
+          </span>
+          <a
+            ng-if="binding | isBindingReady"
+            href="{{row.getSecretForBinding(binding) | navigateResourceURL}}">
```
spadgett: Do you mind adding the `canI` check? Users with `view` access typically can't get secrets.

https://github.com/spadgett/origin-web-console/blob/a172029b6d375f52775f7b3d5b064f9a94c31fe4/app/views/_volumes.html#L22-L27

app/scripts/filters/resources.js
```diff
@@ -1084,7 +1084,18 @@ angular.module('openshiftConsole')
       return lastFinishTime;
     };
   })
+  // gets the first status condition.
   .filter('statusCondition', function() {
```
spadgett: Hm, you can't change the behavior of this filter since it's used elsewhere (bind dialog)

benjaminapetersen: Ill do a search again, but I thought I updated the other refrences. Let me check.

app/views/overview/_service-instance-row.html
benjaminapetersen: Yup, already did. let me push that up.  Sorry, didn't update after a bit more tinkering last night.

benjaminapetersen:
Pushed again this morning, but give me a few, rebuilding my env as things blew up.  I want to test a bit more before another review.

app/scripts/filters/resources.js
jwforres: No I disagree with this behavior change.  You may at any point have multiple status conditions, you need to actually ask for the one you want.

app/views/overview/_service-instance-row.html
```diff
@@ -14,13 +14,18 @@
           </div>
         </h3>
         <div class="status-icons">
+          <!--
+            TODO: this should show an info status icon if the apiObject is not yet ready.
+          -->
           <notification-icon ng-if="!row.expanded" alerts="row.notifications"></notification-icon>
+          <span ng-if="!row.isReady">
+            <status-icon status="row.status"></status-icon> {{row.statusReason | sentenceCase}}
```
jwforres: it is unlikely status-icon is going to work with the condition reasons we are getting back, and at any time they might add more and break us / cause us to show the ?  I would prefer we check for known failure conditions.

app/scripts/controllers/newOverview.js
```diff
@@ -1263,6 +1263,7 @@ function OverviewController($scope,
         resource: 'bindings'
       }, context, function(bindings) {
         state.bindings = bindings.by('metadata.name');
+        overview.bindingsByInstanceRef = _.indexBy(state.bindings, 'spec.instanceRef.name');
```
spadgett: Should be `_.groupBy` not `_.indexBy` (my fault)

jwforres:
FYI I pulled down your PR and validated switching to _.groupBy works

app/views/overview/_service-instance-row.html
benjaminapetersen: And this should probably be handled in the `<notification-icon>` above like other alerts, correct?

benjaminapetersen: I'm opting for #2, the less verbose. Objections @jwforres @spadgett ?
screen shot 2017-04-12 at 2 25 38 pm
screen shot 2017-04-12 at 2 29 34 pm

app/scripts/controllers/newOverview.js
benjaminapetersen: Heh. Was messing with that before I read this.  Getting only 1 item every time was confusing 😄

app/scripts/filters/resources.js
benjaminapetersen: Ok, working on this.

benjaminapetersen: `statusConditionIs` filter is the original code, so I aliased `statusCondition` back to it & updated my calls.

jwforres:
Not sure why we would keep both in that case. Just keep statusCondition

jwforres:
If we are going to show the error from the API lets show the full error
message, we don't typically truncate it

benjaminapetersen:
Ok cool, got those.

benjaminapetersen:
& rebased.

benjaminapetersen:
one more tweak coming

app/scripts/controllers/newOverview.js
```diff
@@ -1252,6 +1253,20 @@ function OverviewController($scope,
         resource: 'instances'
       }, context, function(serviceInstances) {
         state.serviceInstances = serviceInstances.by('metadata.name');
+        _.each(state.serviceInstances, function(instance) {
+          var failConditions = _.filter(instance.status.conditions, {status: 'False'});
```
jwforres: a condition with status false is not necessarily a failure condition

benjaminapetersen:
ok. think this is ready for one more pass?

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -21,6 +22,8 @@ function ServiceInstanceRow($filter, DataService, rowMethods, $uibModal) {
   _.extend(row, rowMethods.ui);
 
   var getErrorDetails = $filter('getErrorDetails');
+  var statusCondition = $filter('statusCondition');
```
jwforres: is anything actually referencing these?

app/scripts/services/resourceAlerts.js
```diff
@@ -147,10 +147,46 @@ angular.module("openshiftConsole")
       return alerts;
     };
```
benjaminapetersen: @jwforres @spadgett closer to the mark adding this?

benjaminapetersen: updating to `_.find` since `_.filter` doesn't make sense here.

jwforres:
[merge]

benjaminapetersen:
😂

jwforres:
[merge]

jwforres:
Looks like a flake [merge]

PR #1414, opened 2017-04-11
Create a binding from a svc to an app

jwforres:


app/views/directives/bind-service/select-application.html
```diff
@@ -0,0 +1,35 @@
+<h3 class="mar-top-none">Select an application to bind to service <strong>{{ctrl.serviceClasses[ctrl.target.spec.serviceClassName].osbMetadata.displayName || ctrl.target.spec.serviceClassName}}</strong></h3>
```
spadgett: > {{ctrl.serviceClasses[ctrl.target.spec.serviceClassName].osbMetadata.displayName || ctrl.target.spec.serviceClassName}}

Probably worth adding a `serviceClassDisplayName` filter at some point. Doesn't have to be in this PR. It might make sense in common.

app/views/overview/_service-instance-row.html
```diff
@@ -124,3 +127,9 @@
   </div>
 
 </div>
+<overlay-panel single-column="true" show-panel="row.overlay.panelVisible" show-close="true" handle-close="row.closeOverlayPanel">
+  <div ng-if="row.overlay.panelName === 'bindService'">
```
spadgett: I'm guessing we're anticipating more panels?

app/views/directives/bind-service/select-application.html
jwforres: Yeah I held off for right now because I think it does need to go in common

app/views/overview/_service-instance-row.html
jwforres: i wont be shocked if we have more in the future, so just set this up the same way i was doing in the Add to Project dropdown PR

jwforres:
[merge]

PR #1427, opened 2017-04-17
Hide "Start Build" actions for binary builds

spadgett:
Uploading a source directory for binary builds is not supported in the web console, so hide the actions that will result in errors.

### DIFF
--- a/app/scripts/directives/overview/listRow.js
+++ b/app/scripts/directives/overview/listRow.js
@@ -37,6 +37,7 @@ function OverviewListRow($filter,
   var canI = $filter('canI');
   var deploymentIsInProgress = $filter('deploymentIsInProgress');
   var getErrorDetails = $filter('getErrorDetails');
+  var isBinaryBuild = $filter('isBinaryBuild');
   var isJenkinsPipelineStrategy = $filter('isJenkinsPipelineStrategy');
 
   var updateTriggers = function(apiObject) {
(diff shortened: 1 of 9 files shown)


app/scripts/directives/overview/listRow.js
```diff
@@ -158,9 +159,14 @@ function OverviewListRow($filter,
         return true;
       }
       // Start pipeline or build is displayed.
-      if ((_.size(row.buildConfigs) === 1 || _.size(row.pipelines) === 1) &&
-           canI('buildconfigs/instantiate')) {
-        return true;
+      if (canI('buildconfigs/instantiate', 'create')) {
```
jwforres: we are duplicating a lot of logic from showStartBuildAction and showStartPipelineAction, is there a reason not to just make calls to each of those?

app/scripts/filters/resources.js
```diff
@@ -763,6 +763,12 @@ angular.module('openshiftConsole')
       }
     };
   })
+  .filter('isBinaryBuild', function() {
+    // Accepts a build or build config.
+    return function(build) {
+      return _.get(build, 'spec.source.type') === 'Binary';
```
jwforres: wasn't spec.source.type deprecated because of the support for multiple sources?

spadgett:
@jwforres updated

jwforres:
[merge]

spadgett:
[merge]

PR #1429, opened 2017-04-17
Always show latest deployment on overview

spadgett:
Fixes #1428

@jwforres k8s deployments don't have this problem, only deployment configs

PR #1433, opened 2017-04-18
Hide failed and cancelled deployments

spadgett:
Fixes a regression from #1429

@jwforres This works with the caveat that it will show the empty state message when the first deployment fails. I'm thinking the simplest might be to adjust the empty state message based on the deployment config conditions.

```yaml
  conditions:
    - type: Available
      status: 'False'
      lastUpdateTime: '2017-04-18T12:22:13Z'
      lastTransitionTime: '2017-04-18T12:22:13Z'
      message: Deployment config does not have minimum availability.
    - type: Progressing
      status: 'False'
      lastUpdateTime: '2017-04-18T12:32:55Z'
      lastTransitionTime: '2017-04-18T12:32:55Z'
      reason: ProgressDeadlineExceeded
      message: replication controller "failing-1" has failed progressing
```

The previous overview had some logic in the view here:

https://github.com/openshift/origin-web-console/blob/master/app/views/overview/_dc.html#L85

Screenshots:

screen shot 2017-04-18 at 8 37 35 am
screen shot 2017-04-18 at 8 37 40 am

For reference, here is how it looks with the previous overview:

screen shot 2017-04-18 at 8 45 55 am

### DIFF
--- a/app/scripts/controllers/newOverview.js
+++ b/app/scripts/controllers/newOverview.js
@@ -653,12 +653,6 @@ function OverviewController($scope,
       return true;
     }
 
-    // If the deployment is latest, always show it.
-    var latestVersion = _.get(overview, ['deploymentConfigs', dcName, 'status', 'latestVersion']);
-    if (latestVersion) {
-      return latestVersion === Number(annotation(replicationController, 'deploymentVersion'));
-    }
-
     return deploymentIsInProgress(replicationController);
   };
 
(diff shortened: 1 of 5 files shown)


spadgett:
@jwforres Should be working now. I also adjusted the empty state message for image change triggers. PTAL

screen shot 2017-04-18 at 10 12 24 am

jwforres:
[merge]

PR #1444, opened 2017-04-19
Support filtering provisioned services on overview

spadgett:
screen shot 2017-04-19 at 8 28 05 am

### DIFF
--- a/app/scripts/controllers/newOverview.js
+++ b/app/scripts/controllers/newOverview.js
@@ -170,7 +170,7 @@ function OverviewController($scope,
            _.size(overview.filteredReplicaSets) +
            _.size(overview.filteredStatefulSets) +
            _.size(overview.filteredMonopods) +
-           _.size(overview.state.serviceInstances);
+           _.size(overview.filteredServiceInstances);
   };
 
   // Show the "Get Started" message if the project is empty.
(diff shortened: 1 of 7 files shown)


app/scripts/controllers/newOverview.js
```diff
@@ -301,11 +301,15 @@ function OverviewController($scope,
   };
 
   var filterByLabel = function(items) {
-    return LabelFilter.getLabelSelector().select(items);
+    var labelSelector = LabelFilter.getLabelSelector();
+    if (labelSelector.isEmpty()) {
+      return items;
+    }
```
spadgett: @jwforres The label `LabelSelector.select()` rejects objects with no labels even when there are no active filters. Is this a label selector bug?

jwforres:
Not a bug, it's an option when you instantiate the LabelSelector. There are
times when we want empty to select all (when being used as a filter) and
times where we want empty to select nothing. Conceptually this would be
like a Service with no selector set wouldn't select any Pods.

spadgett:
Requires https://github.com/kubernetes-ui/label-selector/pull/31

spadgett:
@jwforres updated

jwforres:
[merge]

PR #1451, opened 2017-04-19
Improving copy-to-clipboard display

### LINKED ISSUE #1235: Not enough space between keys and values on Secrets at mobile
When viewing a secret at mobile the key and the value are a little close together. The margin could be increased some.

secrets-mobile-spacing

### LINKED ISSUE #1236: Copy to clipboard button is off screen on multiline content at mobile
This definitely appears on Secrets when viewing values like certs.  This is only really an issue for Android since we disable copy to clipboard buttons on iOS.  While you can scroll over to get to the button, its not as discoverable as it could be.  We could consider absolutely positioning the button so its always visible.

copy-off-screen

rhamilto:
Fixes #1235
Fixes #1236

Fixes issue where dl.secret-data didn’t display properly in IE

Fixes issue where .secret-data readonly inputs weren't getting the right background color and font color in Firefox, IE/Edge

Fixes issue where .btn inside .input-group-btn was 1px too tall in
Firefox, Safari, IE/Edge

screen shot 2017-04-19 at 3 23 58 pm

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -1300,11 +1300,6 @@ pre.clipped {
   }
 }
 
-// don't allow copy-to-clipboard .input-groups get bigger than 600px so Github webhook secrets aren't revealed within the input-group
-copy-to-clipboard .input-group.limit-width {
-  max-width: 300px;
-}
-
 .tech-preview-header {
   justify-content: space-between;
 }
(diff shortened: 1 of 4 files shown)


spadgett:
[merge]

spadgett:
[merge]

rhamilto:
[merge]

Issue #1454, opened 2017-04-20
Tooltip clipped behind list row header at mobile


spadgett:
![screen shot 2017-04-20 at 8 38 42 am](https://cloud.githubusercontent.com/assets/1167259/25231168/d51445f0-25a4-11e7-80f6-9b741c69ff6b.png)

spadgett:
It also gets clipped behind the alert.

![screen shot 2017-04-20 at 8 44 45 am](https://cloud.githubusercontent.com/assets/1167259/25231370/a9a6a100-25a5-11e7-82e3-0cc2679d2e43.png)

spadgett:
I understand that the tooltips don't currently appear for the donut at mobile, but I'd like to check that they aren't clipped for metrics before closing. (I don't currently have metrics enabled to try.)

In the previous overview, we had a conditional class that only set the overflow styles when a deployment was in progress:

https://github.com/openshift/origin-web-console/blob/master/app/views/overview/_dc.html#L1

rhamilto:
Metrics tooltips are getting clipped, too.  

![screen shot 2017-04-20 at 9 29 32 am](https://cloud.githubusercontent.com/assets/895728/25233174/ee7e8e04-25ab-11e7-8687-1251734beec1.PNG)

@sg00dwin, @spadgett suggests we add a class to control overflow so that it's only applied when a deployment is in progress.

PR #1464, opened 2017-04-24
Apply transient class during deployment animation to hide offset ui elements

sg00dwin:
Sets `display:hidden` on `.list-pf-item` by default and on `.list-pf-expansion` when animation enters vertically.

Fixes https://github.com/openshift/origin-web-console/issues/1454

screen shot 2017-04-24 at 10 11 34 am

tooltip above alert
screen shot 2017-04-24 at 10 11 10 am

### DIFF
--- a/app/styles/_overview.less
+++ b/app/styles/_overview.less
@@ -260,9 +260,6 @@
     }
     .list-pf-container {
       display: block; // so that .word-break() will work in IE
-      @media (max-width: @screen-xs-max) {
-        overflow: hidden; // so animation offsets are hidden when sliding in vertically
-      }
     }
     .pod-donut {
       .text-center();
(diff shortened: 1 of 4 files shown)


app/styles/_overview.less
```diff
@@ -272,8 +269,16 @@
...
-        overflow: hidden; // so animation offsets are hidden when sliding in horizontally
+    }
+    &.deployment-in-progress {
+      // So animation offsets are hidden when sliding in horizontally
+      // Only use this style when a deployment is in progress so that chart
+      // tooltips aren't clipped.
+      overflow:hidden;
+      .list-pf-expansion {
+        @media (max-width: @screen-xs-max) {
+          overflow: hidden; // Hidden when sliding in vertically
```
spadgett: Can we get away with only an `overflow: hidden` on `.list-pf-expansion`? It's not clear to me why it's needed on `.list-pf-item`

sg00dwin: Well, we could since both will hide elements outside the right border, but during the deployment the `.previous-donut` tooltip overlays outside of `.list-pf-expansion` and gets clipped.  So to alleviate it , I set hidden on the `.list-pf-item` moving the overflow:hidden above the title bar. And when >480px on `.list-pf-expansion`.

spadgett:
[merge]

PR #1456, opened 2017-04-20
Changing position of copy to clipboard tooltips

rhamilto:
Fixes 1453

screen shot 2017-04-20 at 3 05 08 pm

screen shot 2017-04-20 at 3 05 16 pm

**Bonus:**  changes tooltip and popover [container](http://getbootstrap.com/javascript/#tooltips-options) default to [.middle](https://github.com/openshift/origin-web-console/blob/master/app/index.html#L55) so tooltips and popovers scroll with the content.

PR #1461, opened 2017-04-21
Increase default Jasmine timeout interval

spadgett:
Some tests do a lot of setup in `beforeEach`, including deleting and
recreating projects, which can cause tests to timeout with the
current value.

Also remove quotes from some test descriptions that break download links
for failure screenshots in Jenkins.

PR #1467, opened 2017-04-24
Restoring popup container default to body to fix popover clipping bug

### LINKED ISSUE #1463: Popover masked by navbar at non-mobile resolutions
screen shot 2017-04-24 at 10 00 48 am

It looks like I created this bug with #1456.  I will fix.

rhamilto:
Fixes #1463

screen shot 2017-04-24 at 11 44 49 am

In restoring the popup container default to body, [popovers and tooltips once again scroll away
from their content](https://github.com/openshift/origin-web-console/pull/1456), but this is a less egregious bug than clipping of
popovers or tooltips; however, we can add a data-containter=“.middle”
to any popover or tooltip instances we don’t want to scroll away from
their content.  I’ve done so for copy-to-clipboard but not any other
existing instances.

### DIFF
--- a/app/scripts/directives/popups.js
+++ b/app/scripts/directives/popups.js
@@ -24,7 +24,7 @@ angular.module('openshiftConsole')
       },
       link: function($scope, element, attrs) {
         var popupConfig = {
-          container: attrs.container || ".middle",
+          container: attrs.container || "body",
           placement: attrs.placement || "auto"
         };
         if (attrs) {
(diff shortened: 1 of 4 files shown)


spadgett:
[merge]

spadgett:
@jwforres FYI, looks like @rhamilto hit the flake I was seeing Friday.

```
1) authenticated e2e-user create from URL using an image stream and image stream tag supplied as query string params should display details about the the image
  - timeout: timed out after 30000 msec waiting for spec to complete
```

I did open #1461 to increase the timeout. As far as I can tell, the test just takes more than 30s given all the set up it's doing.

[merge]

spadgett:
Opened #1468 for the flake in the last failure

[merge]

PR #1471, opened 2017-04-25
Add toast notification display

jeff-phillips-18:


app/views/landing-page.html
```diff
@@ -1,4 +1,5 @@
 <default-header class="top-header"></default-header>
+<toast-notifications></toast-notifications>
```
spadgett: I spoke to @jwforres this morning, and we were both thinking we should just add this to index.html so it's available everywhere. I don't see a reason to wait to use toasts where it makes sense.

spadgett:
[merge]

spadgett:
@jeff-phillips-18 Can you try `hack/install-deps.sh && grunt build`? It's possible a dependency changed. Merge failed from a diff in vendor.js

jeff-phillips-18:
@spadgett done, seems there was a change in the bootstrap-treeview dependency.

spadgett:
@jeff-phillips-18 thanks

[merge]

PR #1475, opened 2017-04-26
Make appearance of provisioned services consistent with that of deployments

luksa:
Previously, the service type was emphasized, whereas the name of the service instance was shown in a small font. I believe this should be reversed (e.g. should be just like deployments are displayed - "DEPLOYMENT" in small font, name of deployment slightly larger and emphasized).

The new look:

screenshot

And the old look:

screenshot

### DIFF
--- a/app/views/overview/_service-instance-row.html
+++ b/app/views/overview/_service-instance-row.html
@@ -6,8 +6,8 @@
     <div class="list-pf-content">
       <div class="list-pf-name">
         <h3>
-          <span ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></span>
-          <div ng-bind-html="row.apiObject.metadata.name | highlightKeywords : row.state.filterKeywords" class="list-row-longname"></div>
+          <div ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords" class="component-label"></div>
+          <span ng-bind-html="row.apiObject.metadata.name | highlightKeywords : row.state.filterKeywords"></span>
         </h3>
         <div class="status-icons">
           <notification-icon ng-if="!row.expanded" alerts="row.notifications"></notification-icon>
--- a/app/views/overview/_service-instance-row.html
+++ b/app/views/overview/_service-instance-row.html
@@ -7,7 +7,10 @@
       <div class="list-pf-name">
         <h3>
           <div ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords" class="component-label"></div>
-          <span ng-bind-html="row.apiObject.metadata.name | highlightKeywords : row.state.filterKeywords"></span>
+          <span ng-if="!row.apiObject.status.dashboardURL" ng-bind-html="row.apiObject.metadata.name | highlightKeywords : row.state.filterKeywords"></span>
+          <span ng-if="row.apiObject.status.dashboardURL">
+            <a ng-href="{{row.apiObject.status.dashboardURL}}" target="_blank">
+              <span ng-bind-html="row.apiObject.metadata.name | highlightKeywords : row.state.filterKeywords"></span></a></span>
         </h3>
         <div class="status-icons">
           <notification-icon ng-if="!row.expanded" alerts="row.notifications"></notification-icon>
--- a/app/styles/_overview.less
+++ b/app/styles/_overview.less
@@ -322,11 +322,6 @@
       }
     }
   }
-  .list-row-longname {
-    .text-muted();
-    font-size: @component-label;
-    margin-top: 3px;
-  }
   .list-row-tabset {
     margin-top: 20px;
     @media (min-width: @screen-sm-min) {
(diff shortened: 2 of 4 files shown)


spadgett:
cc @jwforres @benjaminapetersen @ncameronbritt

app/views/overview/_service-instance-row.html
```diff
@@ -6,8 +6,8 @@
     <div class="list-pf-content">
       <div class="list-pf-name">
         <h3>
-          <span ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></span>
-          <div ng-bind-html="row.apiObject.metadata.name | highlightKeywords : row.state.filterKeywords" class="list-row-longname"></div>
```
spadgett: We should remove the `list-row-longname` styles from _overview.less file since this is the only place it's used.

luksa: Done. BTW: running `grunt build` changes `dist/scripts/vendor.js` and `app/index.html` in ways that aren't the result of my changes. That's why I haven't committed them. (They also change if I perform the build without my changes - on master).

spadgett: Yeah, it looks like there was a change to a dependency. The diff should disappear if you rebase after #1471 merges

luksa: #1471 was merged, so I rebased and squashed. No files changed by `grunt build` anymore.

jwforres:
To explain our original thinking on the inconsistency, today we auto-generate the name of the service instance based on the unique name of the service class.  This means largely the name of the service instance may not be that important to you. There are many service classes that someone would only ever instantiate once, and the pretty display name of that service class is the thing you care about.

There are also cases where the unique name of the service instance does matter though, and the message queue use case is highlighting that.  I might have several queues instantiated into my project.

I think what this is making clear is that we need to optionally let people specify the name for their service instance, which I just saw you have partially implemented in your other commit already.  It needs to be optional and fall back to use generateName when the user doesn't set it.

luksa:
Yeah, I thought that was the reasoning. That said, displaying the name of the instance emphasized even for single-instance services isn't that bad, since, as you say, the autogenerated name of the instance starts with the service class name anyway, so the only difference between the two options is just a few extra characters, which shouldn't bother anyone.

My hack already takes care of cases when the user doesn't enter a name exactly as you say (it only sets the name if it was entered, otherwise it sets the generatedName). Initially, the field was empty by default, but then I changed the default to "my-<service-class-name>". 

Just to clarify why I also set a "name" parameter: EnMasse uses the parameter to give the queue that exact name (as you can see in the video, the name is then also displayed in the EnMasse console). IMHO most users will expect the instance name and the actual queue name to be the same (and also not having to type it in twice). I asked Paul about whether the ServiceCatalog could always send the instance name as a parameter, but he thought that would be too fragile.

spadgett:
> That said, displaying the name of the instance emphasized even for single-instance services isn't that bad, since, as you say, the autogenerated name of the instance starts with the service class name anyway, so the only difference between the two options is just a few extra characters, which shouldn't bother anyone.

This isn't necessarily true since we'll show the service class display name if it has one, which might be different than metadata.name.

We might want to give a way for users to name the instance when provisioning in the UI, although it could be confusing if the service class also has a name parameter.

jwforres:
On Wed, Apr 26, 2017 at 1:35 PM, Marko Lukša <notifications@github.com>
wrote:

> Yeah, I thought that was the reasoning. That said, displaying the name of
> the instance emphasized even for single-instance services isn't that bad,
> since, as you say, the autogenerated name of the instance starts with the
> service class name anyway, so the only difference between the two options
> is just a few extra characters, which shouldn't bother anyone.
>
> My hack already takes care of cases when the user doesn't enter a name
> exactly as you say (it only sets the name if it was entered, otherwise it
> sets the generatedName). Initially, the field was empty by default, but
> then I changed the default to "my-".
>
> Just to clarify why I also set a "name" parameter: EnMasse uses the
> parameter to give the queue that exact name (as you can see in the video,
> the name is then also displayed in the EnMasse console). IMHO most users
> will expect the instance name and the actual queue name to be the same (and
> also not having to type it in twice).
>

Understood, but this is specific to your broker's use case.  Sending a
parameter that a broker doesn't expect could have unforseen consequences.
This means not only that svc catalog can not always send it, but the UI can
not always send it either.

luksa:
I understand. Maybe this could be solved by having the broker include an entry in the ServiceClass metadata. Something like metadata `{"name-is-required-and-must-be-passed-through-parameter": "queue-name"}`.  This could also be used by the UI to decide whether to show the name field at all.

jwforres:
Does reusing the service instance name always make sense though? I see a
common problem with this is going to be imposing kube's name validation
rules onto the naming rules for the system the broker is provisioning into.
My service instance name is limited to letters numbers and hyphens, not
sure what you guys support but I'm guessing the queue name is more flexible
than that?

luksa:
I was just thinking about that myself. In our case, the name of the queue also ends up being included in the name of the pod providing the queue. 

But, in general, I agree. The names should probably be entered separately.

pmorie:
@luksa @jwforres @spadgett 

I think that the name of an instance is a kubernetes-level construct.  I would not expect that name to be available, necessarily, as a parameter, or part of the [request context](https://github.com/duglin/servicebroker/blob/CFisms/context-profiles.md).  If a name is needed for the service instance being provisioned, I think it should be a parameter to provision.

spadgett:
Closing as this repo is now only for bug fixes for 3.x releases.

In 4.0, we allow users to pick the service instance name and we emphasize the instance name more. Any additional changes should be made in the openshift/console repo.

/close

PR #1476, opened 2017-04-26
Fix timing issue showing service class display name on overview

spadgett:
Previously if the service instances loaded before the service classes,
the display name was not correctly displayed in the instance row on the
overview.


jwforres:
[merge]

PR #1478, opened 2017-04-26
[WIP] disambiguate kinds on other resources by showing group

benjaminapetersen:
Getting started on this.  Most of the work is in [this PR](https://github.com/openshift/origin-web-common/pull/51) in `web-common`.

- [ ] bump version in origin-web-common when APIService dedupe is merged
- [x] show groups for resources in the drop down to make it possible to interact with different resources that happen to have the same name.

I'm not sure if simply listing the group under the kind is extremely clear, but its what we typically do:

screen shot 2017-04-26 at 2 23 10 pm

### DIFF
--- a/app/scripts/controllers/otherResources.js
+++ b/app/scripts/controllers/otherResources.js
@@ -57,6 +57,14 @@ angular.module('openshiftConsole')
       }).toString();
     };
 
+    var counts;
+    $scope.isDuplicateKind = function(kind) {
+      if(!counts) {
+        counts = _.countBy($scope.kinds, 'kind');
+      }
+      return counts[kind] > 1;
+    };
+
     // get and clear any alerts
     AlertMessageService.getAlerts().forEach(function(alert) {
       $scope.alerts[alert.name] = alert.data;
--- a/app/views/other-resources.html
+++ b/app/views/other-resources.html
@@ -14,6 +14,7 @@
               <ui-select-match placeholder="Choose a resource to list...">{{$select.selected.kind | humanizeKind : true}}</ui-select-match>
               <ui-select-choices repeat="kind in kinds | filter : {kind: $select.search} : matchKind | orderBy : 'kind'">
                 <div ng-bind-html="(kind.kind | humanizeKind : true) | highlight: $select.search"></div>
+                <small ng-if="isDuplicateKind(kind.kind)" ng-bind-html="kind.group | highlight: $select.search" class="text-muted"></small>
               </ui-select-choices>
             </ui-select>
             <div class="vertical-divider"></div>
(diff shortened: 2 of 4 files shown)


app/views/other-resources.html
```diff
@@ -14,6 +14,7 @@
               <ui-select-match placeholder="Choose a resource to list...">{{$select.selected.kind | humanizeKind : true}}</ui-select-match>
               <ui-select-choices repeat="kind in kinds | filter : {kind: $select.search} : matchKind | orderBy : 'kind'">
                 <div ng-bind-html="(kind.kind | humanizeKind : true) | highlight: $select.search"></div>
+                <small ng-bind-html="kind.group | highlight: $select.search" class="text-muted"></small>
```
jwforres: we shouldn't always show this, most of the time it is unnecessary gorp.  I would only show it when it is necessary to distinguish between two Kinds that are the same.  Similar to what we do in the projects dropdown when we see two projects have the same display name.

benjaminapetersen: Agree, am working on that part now

benjaminapetersen:
Updated, now, if I disable the filter on line `103` I can trigger this behavior:

screen shot 2017-04-26 at 3 49 47 pm

Local Subject Access Review is the only item that appears with a group (both).

jwforres:
works for me, clean it up so we can merge it

app/scripts/controllers/otherResources.js
```diff
@@ -57,6 +58,24 @@ angular.module('openshiftConsole')
       }).toString();
     };
 
+    var countKinds = function() {
+      return _.reduce($scope.kinds, function(result, kind) {
```
spadgett: We should be able to use `_.countBy`

https://lodash.com/docs/3.10.1#countBy

benjaminapetersen:
ok squashed & updated `reduce` to `countBy`.

app/scripts/controllers/otherResources.js
```diff
@@ -57,6 +57,20 @@ angular.module('openshiftConsole')
...
+    var countKinds = function() {
+      return _.countBy($scope.kinds, function(kind) {
+        return kind.kind;
+      });
+    };
+
+    var counts;
+    $scope.isDuplicateKind = function(kind) {
+      if(!counts) {
+        counts = countKinds();
```
jwforres: cant you just do this as a one-liner here using the short form:
```
counts = _.countBy($scope.kinds, "kind")
```

benjaminapetersen: updating

jwforres:
LGTM, just waiting on the bump of origin-web-common

benjaminapetersen:
[merge]

benjaminapetersen:
[merge]

PR #1486, opened 2017-04-28
Fixed Cassandra icon style name

jpkrohling:


app/styles/_openshift-logos-icon.less
```diff
@@ -191,6 +191,6 @@
 .icon-apache:before {
   content: "\f138";
 }
-.icon-casadra:before {
+.icon-casadra:before, .icon-cassandra:before {
```
spadgett: Might be worth a comment here not to remove `icon-casadra` for backwards compatibility.

sg00dwin:
@spadgett I believe it would be fine

spadgett:
[merge]

spadgett:
Merging since @sg00dwin made the other suggested change in #1496

PR #1487, opened 2017-04-28
Bump origin-web-common to v0.0.17

spadgett:

PR #1490, opened 2017-05-01
Remove oscUnique.js from index.html

spadgett:
The directive has moved to origin-web-common. This was an oversight in #1487


spadgett:
No dist changes, only affects development environments

[merge]

PR #1491, opened 2017-05-01
Sort openshift namespace alphabetically with other namespaces in IS tag picker

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1436819


spadgett:
[merge]

PR #1495, opened 2017-05-01
Clean up $rootScope.$on listeners

### LINKED ISSUE #1480: $rootScope.$on handlers not unbound on $destroy
We should be unbinding our `$rootScope.$on` handlers when the scope is destroyed. This is a memory leak and can cause the function to run multiple times.

We have this in a few places like the pod-metrics directive.

cc @jwforres

spadgett:
Fixes a memory leak where we weren't properly unbinding listeners for `$rootScope.$on` events.

Fixes #1480

### DIFF
--- a/app/scripts/directives/deploymentMetrics.js
+++ b/app/scripts/directives/deploymentMetrics.js
@@ -464,7 +464,7 @@ angular.module('openshiftConsole')
           }
         };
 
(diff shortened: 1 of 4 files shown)


spadgett:
@jwforres This is only a problem for `$rootScope.$on` (not `$scope.$on`) since `$rootScope` isn't destroyed.

spadgett:
[merge]

PR #1492, opened 2017-05-01
Support creating templates in a dialog

spadgett:
https://trello.com/c/2A7x5Pbc

Fixes https://github.com/openshift/origin-web-catalog/issues/177
Requires https://github.com/openshift/origin-web-catalog/pull/193

screencast 2017-05-01 13-07-26

Follow-on work:

- [ ] Fix flicker in next steps page. (I don't think this is specific to the dialog.)
- [ ] Preserve entered field values when clicking advanced options
- [x] Don't scroll description with right panel content -- not fixed, but opened https://github.com/openshift/origin-web-catalog/issues/196 to track
- [ ] Support advanced options when creating a new project
- [ ] Update to use angular-patternfly wizard component
- [ ] Improve template dialog styles (I might want your help @rhamilto)

PR #1497, opened 2017-05-02
Fix images displayed when processing template

spadgett:
Fixes bug introduced by #1492


spadgett:
@jwforres Sorry, I had only partially moved the code for displaying images in a template back into the `NewFromTemplate` controller. It's not used in the dialog, so I meant to move it back out of the `process-template` component.

spadgett:
Also fixes a bug where we were showing the advanced options link for full page templates.

@jwforres PTAL

jwforres:
[merge]

PR #1501, opened 2017-05-02
Add flag for disabling templates in the catalog

spadgett:
Add `ENABLE_TECH_PREVIEW_FEATURE.template_service_broker` flag that will use the template broker for templates rather than listing them using the API.

See #1492 and https://trello.com/c/2A7x5Pbc


jwforres:
[merge]

PR #1503, opened 2017-05-02
Bump origin-web-common to v0.0.18

spadgett:


spadgett:
[merge]

spadgett:
> 1) User adds an image stream to a project after creating a new project using the "Import YAML/JSON" tab should process and create the images in the image stream
>  - Expected false to be true.

[merge]

PR #1504, opened 2017-05-02
Add 12 hours time range to metrics time ranges

jfchevrette:
As a console user I have a couple of 'Time Range' settings already at my disposal: 1 hour, 4 hours, 1 day, 3 days and 7 days.

There is however an arguably large gap between 4 hours and 24 hours.

This PR adds a 'Last 12 hours' time range to the metrics tab.


spadgett:
[merge]

PR #1508, opened 2017-05-03
Add landing page tour

jeff-phillips-18:
https://trello.com/c/oveRlqqd

Launched via:
First user visit to the landing page
‘Take Home Page Tour’ button on landing page (less than 2 projects)
‘Tour Home Page’ menu item from the help menu

### DIFF
--- a/app/scripts/controllers/landingPage.js
+++ b/app/scripts/controllers/landingPage.js
@@ -3,17 +3,31 @@
 angular.module('openshiftConsole')
   .controller('LandingPageController',
               function($scope,
+                       $rootScope,
                        AuthService,
                        Catalog,
                        Constants,
                        Navigate,
                        NotificationsService,
-                       RecentlyViewedServiceItems) {
+                       RecentlyViewedServiceItems,
+                       GuidedTourService,
+                       $timeout,
+                       $routeParams,
+                       $location) {
+    var tourConfig = _.get(Constants, 'GUIDED_TOURS.landing_page_tour');
+    var tourEnabled = tourConfig && tourConfig.enabled && tourConfig.steps;
+
     $scope.saasOfferings = Constants.SAAS_OFFERINGS;
 
     $scope.viewMembership = function(project) {
       Navigate.toProjectMembership(project.metadata.name);
     };
+
+    if (tourEnabled) {
+      $scope.startGuidedTour = function (onDone) {
+        GuidedTourService.startTour(tourConfig.steps, onDone);
+      };
+    }
 
     // Currently this is the only page showing notifications, clear any that came previous pages.
     // Once all pages show notifications this should be removed.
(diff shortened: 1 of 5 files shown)


app/scripts/controllers/landingPage.js
```diff
@@ -47,4 +61,31 @@ angular.module('openshiftConsole')
       // add the item to recently-viewed. No-op if the dialog is not open.
       addTemplateToRecentlyViewed();
     });
+
+    function dataLoaded() {
+      if (tourConfig && tourConfig.enabled && tourConfig.steps) {
```
spadgett: Would be good to calculate this once above and remember so we don't have the same logic twice in the same controller.

I think an early return is easier to understand here instead of nested `if` statements.

```js
if (!tourEnabled) {
  return;
}

if ($routeParams.startTour) ...
```

app/scripts/controllers/landingPage.js
```diff
@@ -47,4 +61,31 @@ angular.module('openshiftConsole')
...
+          }
+        }
+      }
+    }
+
+    function resetRouteParams() {
+      // Reset the route params in the next digest cycle
+      $timeout(function() {
+        $location.replace();
+        $location.search('startTour', null);
```
spadgett: Is it necessary to remove the `startTour` query parameter? I might just leave it. This doesn't cause the route to reload, does it?

spadgett: I guess it might be bad if someone bookmarked the page :/

jeff-phillips-18: It does look like it is reloading now that you mention it.

spadgett: @jeff-phillips-18 You should be able to add `reloadOnSearch: false` to the route in app.js

app/scripts/extensions/nav/helpDropdown.js
```diff
@@ -6,19 +6,38 @@ angular.module('openshiftConsole')
...
+          }
+        );
+
+        if (_.get(window, 'OPENSHIFT_CONSTANTS.ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page')) {
+          var tourConfig = _.get(window, 'OPENSHIFT_CONSTANTS.GUIDED_TOURS.landing_page_tour');
+          if (tourConfig && tourConfig.enabled && tourConfig.steps) {
+            options.push(
+              {
+                type: 'dom',
+                node: '<li><a href="./#?startTour=true" >Tour Home Page</a></li>'
```
spadgett: Should this be `/.?startTour=true` with no `#`?

Nit: extra space before `>`

app/scripts/controllers/landingPage.js
jeff-phillips-18: That prevents the tour from launching if it is already on the landing page.

spadgett:
[merge]

PR #1505, opened 2017-05-03
Add delete binding (unbind) action to provisioned services

benjaminapetersen:
Depends on #1578

WIP, still some things to do.  Based on [mock-ups from here](https://openshift.github.io/openshift-origin-design/web-console/3-project-details/binding-in-project).  Progress thus far:

screen shot 2017-05-03 at 12 10 26 pm

screen shot 2017-05-03 at 12 10 34 pm

screen shot 2017-05-03 at 12 10 39 pm

TODO:
- [ ] `<service>` has been unbound from **`<application>`**.  Application name is missing.
- [x] factor `bindService` and `unbindService` duplicate code into separate service, OR collapse `unbindService` into bindService & toggle.

@jwforres @spadgett @ncameronbritt

PR #1514, opened 2017-05-04
[WIP] Use common components for bind pages.

jeff-phillips-18:
remove filters that were added to common (isBindingReady, statusCondition, and isServiceInstanceReady)

Depends on updated common version to include https://github.com/openshift/origin-web-common/pull/58

@benjaminapetersen


benjaminapetersen:
@jeff-phillips-18 heh, we gonna collide re: my PR #1505

app/scripts/directives/bindService.js
```diff
@@ -71,28 +71,24 @@ function BindService($filter,
...
-      });
-    }
-    ctrl.steps.push({
-      label: 'Results',
-      id: 'results',
-      view: 'views/directives/bind-service/results.html'
-    });
+    ctrl.serviceSelection = {};
+    var formStepId = (ctrl.target.kind === 'Instance') ? 'applications' : 'services';
+    var formStepLabel = (ctrl.target.kind === 'Instance') ? 'Applications' : 'Services';
```
benjaminapetersen: For the label, shall we just `_.upperFirst(formStepId)`?

app/scripts/filters/resources.js
```diff
@@ -1090,25 +1090,6 @@ angular.module('openshiftConsole')
       return lastFinishTime;
     };
   })
-  // gets the status condition that matches provided type
```
benjaminapetersen: Makes sense to move these.

app/scripts/directives/bindService.js
spadgett: I think it makes sense as a separate property for when we need to translate

PR #1517, opened 2017-05-04
Remove misleading placeholders from HPA form

spadgett:
Remove the placeholder text for max pods and other inputs when adding an
HPA. It looks too much like either the required field has a value or 4
is a default. This is especially confusing in the add to project form,
where it's easy to miss required fields in a large form.

openshift web console 2017-05-04 15-12-01

cc @liggitt


jwforres:
[merge]

PR #1518, opened 2017-05-04
View all projects should go to projects page, not landing page

### LINKED ISSUE #1484: "View all projects" goes to landing page
When the catalog is turned on, the "View all projects" link in the project sdropdown menu goes to the landing page rather than the projects page. 

screenshot from 2017-04-27 16-41-51

This may be a separate issue, but what do you think about styling that item as a link to distinguish it from the projects? @serenamarie125 @cshinn @beanh66 @jwforres @rhamilto

spadgett:
With the new catalog experience enabled, the view all projects link in
the projects dropdown should go to the projects list rather than the
landing page, which is now a different URL.

Fixes #1484

### DIFF
--- a/app/scripts/directives/nav.js
+++ b/app/scripts/directives/nav.js
@@ -127,7 +127,7 @@ angular.module('openshiftConsole')
           })
           .change(function() {
             var val = $(this).val();
-            var newURL = (val === "") ? "/" : projectOverviewURLFilter(val);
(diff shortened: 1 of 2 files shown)


spadgett:
This doesn't change the styling of the menu item, only where it goes. The style is currently controlled by bootstrap select.

jwforres:
[merge]

PR #1520, opened 2017-05-04
Change button label for "Home" for new landing page

### LINKED ISSUE #1483: "Projects" button goes to landing page when running the catalog
The text "Projects" should be replaced with whatever we want to call the landing page.

screenshot from 2017-04-27 16-34-03

spadgett:
With the new landing page experience enabled, show the label "Home" rather than "Projects" for the home button in the page upper left.

Fixes #1483


jwforres:
[merge]

PR #1525, opened 2017-05-08
Update blacklist to support objects/strings

benjaminapetersen:
Depends on [PR61](https://github.com/openshift/origin-web-common/pull/61) from `origin-web-common`


benjaminapetersen:
Oop, this one was for testing. Don't actually want to PR it.

PR #1526, opened 2017-05-08
Fix inconsistent behavior in kve between a standard row & a ui-select row

benjaminapetersen:
Re [bugzilla #1448830](https://bugzilla.redhat.com/show_bug.cgi?id=1448830)
There are currently two different behaviors with how we are handling rows in the kve:

1. We allow the user to save a default row that has a key w/o a value:
screen shot 2017-05-08 at 5 27 05 pm

2.  We do not allow the user to save if the row includes a ui-select to choose from a config map or secret:
screen shot 2017-05-08 at 5 27 21 pm

In this case the user must manually remove the row if they decide they no longer need to select the env var.  

This PR would allow the user to save the form, but would delete the key_with_no_value row if the user doesn't actually select a `valueFrom` from a `configmap` or `secret`:

screen shot 2017-05-08 at 5 34 40 pm

I'm not 100% sure this is a better experience, though.  By automatically eliminating the incomplete row, we might put the user into an unexpected state. Though the original experience is a bit inconsistent, it might be more correct.

Its also possible I'm missing something.  Thoughts?

@jwforres @spadgett @rhamilto

### DIFF
--- a/app/scripts/services/keyValueEditorUtils.js
+++ b/app/scripts/services/keyValueEditorUtils.js
@@ -141,7 +141,12 @@
                     entries,
                     function(entry) {
                       entry = cleanEntry(entry);
-                      return (entry.name || entry.value || entry.valueFrom) ? entry : undefined;
+                      var hasValueFrom = entry.valueFrom;
+                      if(!hasValueFrom) {
+                        return (entry.name || entry.value) ? entry : undefined;
+                      }
+                      // if the valueFrom is an empty object, this entry is invalid and should be discarded
+                      return _.keys(entry.valueFrom).length ? entry : undefined;
                     }));
         };
 
--- a/app/views/directives/key-value-editor.html
+++ b/app/views/directives/key-value-editor.html
@@ -142,7 +142,7 @@
           </div>
           <div ng-if="!isValueFromReadonly(entry)">
             <div class="ui-select">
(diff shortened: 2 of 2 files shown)


spadgett:
@benjaminapetersen It's possible to set an environment variable to an empty value. We need to keep that behavior.

benjaminapetersen:
Right, so perhaps in this case if the user doesn't choose a secret/configmap value we should keep the `key` with an empty string as the `value` instead of dropping the row?

jwforres:
That's not what I would expect it to do from those inputs.

benjaminapetersen:
Shall we leave it as-is?

options:
1. leave current behavior, if user clicks "add value using config map or secret" the "save" button is disabled until the user selects the value (though we allows "save" on a regular row w/o a value)
2. **this PR**, allow the user to "save", but if the config map/secret isn't selected drop the row
3. update to allow the user to "save", but if the config map/secret isn't selected then save only the key w/no value

Is there another reasonable alt?

spadgett:
I think the current behavior is reasonable. The difference is that an environment variable with an empty value is valid, whereas the `valueFrom` control with nothing selected is not valid. The second case is like trying to add a value with no key, which is invalid and disables the save button.

It's consistent in that the save button is consistently enabled or disabled based on whether the form is valid. It seems worse to silently drop something the user has entered.

benjaminapetersen:
I'm good with that.  I'll close this out then & clear the bugzilla.

PR #1534, opened 2017-05-10
Bump origin-web-common to v0.0.20

spadgett:
@benjaminapetersen FYI, this contains your fix https://github.com/openshift/origin-web-common/pull/61

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -929,6 +929,9 @@ select:invalid {
 .alert {
   margin-bottom: 14px;
   margin-top: 7px;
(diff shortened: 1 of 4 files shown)


jwforres:
@benjaminapetersen when this merges can you open a follow-on PR to switch from "Binding" in the blacklist to `{kind: 'Binding', group: ''}` so it is clear in the source which one we are blacklisting

benjaminapetersen:
Ok will do.

jwforres:
[merge]

PR #1536, opened 2017-05-10
Replacing deprecated grunt-autoprefixer with postcss

### LINKED ISSUE #1337: grunt-autoprefixer is very outdated
We're currently 13 releases behind the most recent release of [grunt-autoprefixer](https://github.com/nDmitry/grunt-autoprefixer).  We should probably update.

rhamilto:
Fixes #1337

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -198,9 +198,11 @@ module.exports = function (grunt) {
     },
 
     // Add vendor prefixed styles
-    autoprefixer: {
+    postcss: {
       options: {
-        browsers: ['last 1 version']
+        processors: [
+          require('autoprefixer')({browsers: ['last 1 version']})
+        ]
       },
       dist: {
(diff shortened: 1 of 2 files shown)


jwforres:
LGTM surprised there are no changes to the dist as a result

rhamilto:
>  LGTM surprised there are no changes to the dist as a result

I was, too!  I even made a less change to ensure it was true.

spadgett:
It's still using the same autoprefixer under the covers, correct? It's just changing the grunt task that runs the prefixer?

rhamilto:
> It's still using the same autoprefixer under the covers, correct? It's just changing the grunt task that runs the prefixer?

Yep

jwforres:
[merge]

benjaminapetersen:
Huh. that was less painful than expected for sure.

PR #1539, opened 2017-05-11
Support for Gitlab and Bitbucket webhooks in the BC editor

jhadvig:
https://trello.com/c/yy3ACVjj

Adding support for view/add/remove GitLab and Bitbucket webhooks.
Screen:
screenshot at 2017-05-11 14-09-39

screenshot at 2017-05-11 14-38-43

Not really sure if we want to display them by default as the screenshot show, but also not sure if we want to hide them in some "Advanced webhooks settings" since we already have advance view in the BC editor.

@spadgett @jwforres thoughts ?

### DIFF
--- a/app/scripts/controllers/edit/buildConfig.js
+++ b/app/scripts/controllers/edit/buildConfig.js
@@ -107,10 +107,29 @@ angular.module('openshiftConsole')
 
     $scope.triggers = {
       githubWebhooks: [],
+      gitlabWebhooks: [],
+      bitbucketWebhooks: [],
       genericWebhooks: [],
       imageChangeTriggers: [],
       builderImageChangeTrigger: {},
       configChangeTrigger: {}
+    };
+
+    $scope.createTriggerSelect = {
+      selectedType: "",
+      options: [{
+        type: 'github',
+        label: 'GitHub'
+      }, {
+        type: 'gitlab',
+        label: 'GitLab'
+      }, {
+        type: 'bitbucket',
+        label: 'Bitbucket'
+      }, {
+        type: 'generic',
+        label: 'Generic'
+      }]
     };
 
     $scope.runPolicyTypes = [
@@ -381,6 +400,12 @@ angular.module('openshiftConsole')
           case "GitHub":
             triggerMap.githubWebhooks.push({disabled: false, data: trigger});
             break;
+          case "GitLab":
+            triggerMap.gitlabWebhooks.push({disabled: false, data: trigger});
+            break;
+          case "Bitbucket":
+            triggerMap.bitbucketWebhooks.push({disabled: false, data: trigger});
+            break;
           case "ImageChange":
             var imageChangeFrom = trigger.imageChange.from;
             if (!imageChangeFrom) {
@@ -460,6 +485,8 @@ angular.module('openshiftConsole')
 
     var updateTriggers = function() {
       var triggers = [].concat($scope.triggers.githubWebhooks,
+                              $scope.triggers.gitlabWebhooks,
+                              $scope.triggers.bitbucketWebhooks,
                               $scope.triggers.genericWebhooks,
                               $scope.triggers.imageChangeTriggers,
                               $scope.triggers.builderImageChangeTrigger,
@@ -520,6 +547,23 @@ angular.module('openshiftConsole')
         sourceMap[key] = true;
       });
       return sourceMap;
+    };
+
+    $scope.addWebhookTrigger = function(webhookLabel) {
+      if (!webhookLabel) {
+        return;
+      }
+      var webhook = {
+        disabled: false,
+        data: {
+          type: webhookLabel
+        }
+      };
+      var webhookType = _.find($scope.createTriggerSelect.options, function(o) {return o.label === webhookLabel}).type
+      webhook.data[webhookType] = {
+        secret: ApplicationGenerator._generateSecret()
+      };
+      $scope.triggers[webhookType + "Webhooks"].push(webhook);
     };
 
     $scope.save = function() {
--- a/app/scripts/directives/editWebhookTriggers.js
+++ b/app/scripts/directives/editWebhookTriggers.js
@@ -2,7 +2,7 @@
 
 angular.module('openshiftConsole')
   // Element directive edit BC webhook triggers
-  .directive('editWebhookTriggers', function(ApplicationGenerator) {
+  .directive('editWebhookTriggers', function() {
     return {
       restrict: 'E',
       scope: {
@@ -13,21 +13,6 @@ angular.module('openshiftConsole')
         projectName: "=",
         form: "="
       },
-      templateUrl: 'views/directives/edit-webhook-triggers.html',
-      controller: function($scope) {
-        $scope.addWebhookTrigger = function(type) {
-          var webhook = {
-            disabled: false,
-            data: {
-              type: type
-            }
-          };
-          webhook.data[(type === "GitHub") ? "github" : "generic"] = {
-            secret: ApplicationGenerator._generateSecret()
-          };
-          $scope.triggers.push(webhook);
-          $scope.form.$setDirty();
-        };
-      }
+      templateUrl: 'views/directives/edit-webhook-triggers.html'
     };
   });
--- a/app/views/directives/edit-webhook-triggers.html
+++ b/app/views/directives/edit-webhook-triggers.html
@@ -1,32 +1,31 @@
-<h5>{{type}} webhooks
-  <span class="help action-inline">
-    <a href="" aria-hidden="true">
+<div class="display-webhooks">
+  <h5>{{type}} Webhooks
+    <span class="help action-inline">
       <span class="sr-only">{{typeInfo}}</span>
-      <i class="pficon pficon-help" data-toggle="tooltip" aria-hidden="true" data-original-title="{{typeInfo}}"></i>
-    </a>
-  </span>
-</h5>
-<div ng-repeat="trigger in triggers">
-  <div class="trigger-info">
-    <span class="trigger-url">
-      <copy-to-clipboard is-disabled="trigger.disabled" clipboard-text="bcName | webhookURL : trigger.data.type : trigger.data[(type === 'GitHub') ? 'github' : 'generic'].secret : projectName"></copy-to-clipboard>
-    </span>
-    <span class="visible-xs-inline trigger-actions">
-      <a href="" ng-if="!trigger.disabled" class="action-icon" ng-click="trigger.disabled = true; form.$setDirty()" role="button">
-        <span class="pficon pficon-close" aria-hidden="true" title="Remove"></span>
-        <span class="sr-only">Remove</span>
-      </a>
-      <a href="" ng-if="trigger.disabled" class="action-icon" ng-click="trigger.disabled = false" role="button">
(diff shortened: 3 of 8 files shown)


jwforres:
what if the webhook types were in a dropdown where you could select one and
hit add, and then if there are in a particular type you show them like you
have there. there might be a better way to layout the webhooks as well
instead of in a separated list like that, might be able to have a table
where the type is one of the columns. definitely think there are other ways
we could represent this

jwforres:
@ncameronbritt

ncameronbritt:
@jwforres Sound like you're suggesting something similar to the way we handle users/roles in Membership?

ncameronbritt:
What about something like this:
screen shot 2017-05-12 at 8 27 37 am
screen shot 2017-05-12 at 8 27 52 am
screen shot 2017-05-12 at 8 28 02 am

ncameronbritt:
Is having multiple (and multiple kinds of) webhooks something that users need?

jwforres:
The most likely scenario is having one of the source webhooks
(github/gitlab/bitbucket) and then having the generic webhook.  You might
trigger the generic webhook from something in your CI infrastructure.  But
you might also have a Jenkins pipeline that pulls source from multiple
repos as part of a larger build workflow, in that case you might want any
one of those repos to trigger webhooks, odds are in this case that you'll
be using the same repo type (e.g. GitHub) and could share the same
webhook.  But its definitely possible for you to have repos in different
systems, or you might just not want to share the same webhook secret across
different repos.

jhadvig:
@jwforres @ncameronbritt  so I was thinking that we could list only existing webhooks, but instead of having a link under each type for `Add {{type}} Webhook` we could have a selectbox to specify the webhook type with a `Add` btn at the end of the wwebhook list.
openshift web console

jwforres:
yeah something like that could work. FYI cameron the user doesn't have to
enter anything in order to add the webhook, the URL input is just for them
to copy once the URL is generated for them.

ncameronbritt:
Here are a few sketches. #1 seems like what @jhadvig proposed. Seems slightly odd that we would have to add a header for a new type. And how would they order? What if you had more than one for a given type?

The others are riffs on using more of a table format. The main difference being how the "add" action would work. Here I'd imagine that adding one would just append another row in the table. 
screen shot 2017-05-12 at 9 51 30 am

jhadvig:
@ncameronbritt so I was thinking something between #1 and #4. Dont think its necessary to have an extra `Add webhook` header. 

openshift web console-33

OR

openshift web console-3

so we align with the style which we already have.
@jwforres thoughts ?

jhadvig:
personally I would go with the the second option from my last comment

sspeiche:
I can't see the different between the two but it looks good to me.

jhadvig:
@sspeiche its the `Add` button... on aligns with the `Remove`/`Undo`style we have in the existing webhooks, and the other is different.

ncameronbritt:
I think I like the second one better, @jhadvig. 

I still wonder about having them in more of a list with the label to the left of the URL and appending the most recently added to the end of the list. I think that might make it easier to keep up with which one is which if, for example, you had two GitHub webhooks. But that might not be a realistic scenario.

jhadvig:
@ncameronbritt well then we would have to think about the way how to display the Remove/Undo logic in the list. I would suggest do that as a followup if we need to and just add the addtional types and the selectbox with the Add link as was suggested. Thoughts ? :)

ncameronbritt:
I'm ok with that.

jhadvig:
PR updated. Added logic for adding webhooks into `editWebhookTriggers.js`. When used the `add-webhook` attribute has to be specified ot the `edit-webhook-triggers` directive.
Attaching screens:
openshift web console-1

openshift web console-2

@jwforres @ncameronbritt PTAL

app/views/directives/edit-webhook-triggers.html
```diff
@@ -1,32 +1,59 @@
...
+      </span>
+      <span class="hidden-xs trigger-actions">
+        <a href="" role="button" ng-if="!trigger.disabled" ng-click="trigger.disabled = true; form.$setDirty()">Remove</a>
+        <a href="" role="button" ng-if="trigger.disabled" ng-click="trigger.disabled = false">Undo</a>
+      </span>
+    </div>
+  </div>
+</div>
+<div class="add-webhook" ng-if="addWebhook">
+  <h5>Add webhook</h5>
```
spadgett: Add Webhook

app/views/directives/edit-webhook-triggers.html
```diff
@@ -1,32 +1,59 @@
...
   <div class="trigger-info">
     <span class="trigger-url">
-      <copy-to-clipboard is-disabled="trigger.disabled" clipboard-text="bcName | webhookURL : trigger.data.type : trigger.data[(type === 'GitHub') ? 'github' : 'generic'].secret : projectName"></copy-to-clipboard>
+      <ui-select ng-model="triggerMeta.typeToCreate"
+        theme="bootstrap"
```
spadgett: I don't think `theme="bootstrap"` is needed

app/views/edit/build-config.html
```diff
@@ -451,21 +451,50 @@ <h3 class="with-divider">Triggers
...
+                                triggers="triggers"
                                 form="form"
                                 bc-name="buildConfig.metadata.name"
                                 project-name="project.metadata.name">
                               </edit-webhook-triggers>
+                              <edit-webhook-triggers
+                                ng-if="triggers.gitlabWebhooks.length"
+                                add-webhook="false"
+                                type="GitLab"
+                                type-info="A GitLab source repository must be configured to use the webhook to trigger a build when source is committed."
```
spadgett: "The GitLab source repository..."

app/views/edit/build-config.html
```diff
@@ -451,21 +451,50 @@ <h3 class="with-divider">Triggers
...
+                                triggers="triggers"
+                                form="form"
+                                bc-name="buildConfig.metadata.name"
+                                project-name="project.metadata.name">
+                              </edit-webhook-triggers>
+                              <edit-webhook-triggers
+                                ng-if="triggers.bitbucketWebhooks.length"
+                                add-webhook="false"
+                                type="Bitbucket"
+                                type-info="A Bitbucket source repository must be configured to use the webhook to trigger a build when source is committed."
```
spadgett: "The Butbucket source repository...."

app/views/directives/edit-webhook-triggers.html
```diff
@@ -1,32 +1,59 @@
...
     <span class="visible-xs-inline trigger-actions">
-      <a href="" ng-if="!trigger.disabled" class="action-icon" ng-click="trigger.disabled = true; form.$setDirty()" role="button">
-        <span class="pficon pficon-close" aria-hidden="true" title="Remove"></span>
-        <span class="sr-only">Remove</span>
-      </a>
-      <a href="" ng-if="trigger.disabled" class="action-icon" ng-click="trigger.disabled = false" role="button">
-        <span class="fa fa-repeat" aria-hidden="true" title="Undo"></span>
-        <span class="sr-only">Undo</span>
+      <a href="" class="action-icon" ng-click="addWebhookTrigger(triggerMeta.typeToCreate)" role="button">
+        <span class="fa fa-plus" aria-hidden="true" title="Add"></span>
```
spadgett: You should give the add buttons here and for mobile disabled styles when a type is not selected.

app/scripts/directives/editWebhookTriggers.js
```diff
@@ -15,18 +16,46 @@ angular.module('openshiftConsole')
...
           };
-          webhook.data[(type === "GitHub") ? "github" : "generic"] = {
+          var webhookType;
+          webhookType = getWebhookType($scope.triggerMeta.typeToCreate);
+          webhook.data[webhookType] = {
             secret: ApplicationGenerator._generateSecret()
           };
-          $scope.triggers.push(webhook);
-          $scope.form.$setDirty();
+          $scope.triggers[webhookType + "Webhooks"].push(webhook);
```
jhadvig: But we have to pass the entire map cause we dont know what kind of webhook will user choose. When he chooses GitHub and creates it, we have to add the new webhook to the `githubwebhooks` array, or `bitbucketwebhooks` if he chooses Bitbucket. For that case we need to pass the entire map, just for the creating webhook action. Other option would be to split the create and view action for the webhooks in the `edit-webhook-triggers` and have separate directives for them. Thoughts ?

app/views/directives/edit-webhook-triggers.html
jhadvig: Not really sure what you mean... dont we want to have the same styling here as we have in the for displaying the WH ? Meaning for non-mobile we would have `Add` link and for mobile an icon, as displayed in https://github.com/openshift/origin-web-console/pull/1539#issuecomment-301728175

spadgett: When I haven't selected anything in dropdown, I can't click the add button yet. It should have disabled classes so it's grayed out, `cursor: not-allowed`, etc.

spadgett: Ah, I see. It's strange to me that `edit-webhooks` only edits webhooks of one type, but there's an add control for ALL types inside the directive. `edit-webhooks` directive should either be one directive for all webhooks or add should be outside the directive. Given the way the code is structured, I think moving add outside the directive is easier.

jhadvig:
Comments addressed. @spadgett PTAL :)

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -522,6 +549,23 @@ angular.module('openshiftConsole')
...
     };
 
+    $scope.addWebhookTrigger = function(webhookLabel) {
+      if (!webhookLabel) {
+        return;
+      }
+      var webhook = {
+        disabled: false,
+        data: {
+          type: webhookLabel
```
spadgett: Ah, I see that this is the API type value. Sorry didn't realize. I suggest:

```js
      options: [{
        type: 'GitHub',
        propterty: 'github'
      }, { ... }]
```

And then this function should take in parameter `type`.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -522,6 +549,23 @@ angular.module('openshiftConsole')
...
+        }
+      };
+      var webhookType = _.find($scope.createTriggerSelect.options, function(o) {return o.label === webhookLabel}).type
```
spadgett: Missing `;`. You can use the shorthand here

```js
var webhookType = _.find($scope.createTriggerSelect, { label: webhookLabel }).type;
...
```

or if you rename the properties as above, this would be

```js
var property = _.find($scope.createTriggerSelect, { type: type }).property;
```

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -522,6 +549,23 @@ angular.module('openshiftConsole')
...
+      webhook.data[webhookType] = {
+        secret: ApplicationGenerator._generateSecret()
+      };
+      $scope.triggers[webhookType + "Webhooks"].push(webhook);
```
spadgett: I wouldn't `"Webhooks"` if we don't need to? Maybe just...?

```js
$scope.triggers[webhookType].push(webhook)
```

spadgett:
[merge]

spadgett:
Looks like test flake #1685

[merge]

PR #1541, opened 2017-05-11
Update config to use vars & reduce repetition

benjaminapetersen:
This is particularly handy when used as a base for `config.local.js` & needing to switch back and forth between a vagrant machine, docker, etc.  Being able to update the `hostIP` to (example)`10.245.2.2` once rather than pick out a half dozen occurrences is handy.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -1,36 +1,43 @@
-// This is the default configuration for the dev mode of the web console.
-// A generated version of this config is created at run-time when running
-// the web console from the openshift binary.
-//
-// To change configuration for local development, copy this file to
-// assets/app/config.local.js and edit the copy.
-window.OPENSHIFT_CONFIG = {
-  apis: {
-    hostPort: "localhost:8443",
-    prefix: "/apis"
-  },
-  api: {
-    openshift: {
-      hostPort: "localhost:8443",
-      prefix: "/oapi"
-    },
-    k8s: {
-      hostPort: "localhost:8443",
-      prefix: "/api"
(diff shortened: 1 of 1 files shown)


app/config.js
```diff
@@ -1,36 +1,45 @@
...
+
+(function() {
+  // This is the default configuration for the dev mode of the web console.
+  // A generated version of this config is created at run-time when running
+  // the web console from the openshift binary.
+  //
+  // To change configuration for local development, copy this file to
+  // assets/app/config.local.js and edit the copy.
+  var hostIP = 'localhost';
+  var hostPort = ':8443';
```
jwforres: are you actually using hostIP or hostPort anywhere? I don't see the point in splitting these out.

app/config.js
```diff
@@ -1,36 +1,45 @@
...
+  var fullHost = hostIP + hostPort;
```
jwforres: masterPublicHostname to be consistent with what it actually is you need to set it to

app/config.js
```diff
@@ -1,36 +1,45 @@
...
+  var devServer = 'localhost:9000';
```
jwforres: devConsoleHostname

app/config.js
benjaminapetersen: Just the next line, but setting it up for anything we might do in the future. I can collapse if we don't think so.

benjaminapetersen:
updated

jwforres:
[merge]

PR #1543, opened 2017-05-11
Make mini pod donut clickable

spadgett:
Make the mini donut in the collapsed row clickable like the larger donut when the row is expanded.

@ncameronbritt @serenamarie125 FYI


jwforres:
[merge]

PR #1544, opened 2017-05-12
Improve delete dialog message

spadgett:
Show a message stating that children will be deleted and remove the old
warnings about running pods, which are no longer accurate. Toast the
"marked for deletion" message instead of using AlertMessageService.

https://trello.com/c/mZuAqafi

### DIFF
--- a/app/scripts/directives/deleteLink.js
+++ b/app/scripts/directives/deleteLink.js
@@ -1,7 +1,17 @@
 'use strict';
 
 angular.module("openshiftConsole")
-  .directive("deleteLink", function ($uibModal, $location, $filter, $q, hashSizeFilter, APIService, DataService, AlertMessageService, Navigate, Logger) {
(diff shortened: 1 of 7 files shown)


app/scripts/directives/deleteLink.js
```diff
@@ -165,9 +173,9 @@ angular.module("openshiftConsole")
               if (!promises.length) {
                 navigateToList();
               } else {
-                // Wait until all promises resolve so that we can add alerts to
-                // AlertMessageService before navigating, otherwise they aren't
-                // displayed.
+                // Wait until all promises resolve so that we can add
+                // notifications to NotificationsService before navigating,
```
jwforres: is this comment still true? the notification service lives outside the view.  Still seems reasonable that you would wait to navigate until the delete is actually successful, i think the comment just needs updating.

jwforres:
[merge]

PR #1434, opened 2017-04-18
Switch to creating a DeploymentRequest to rollout a new deployment of a DC

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1433229

PR #1545, opened 2017-05-12
Add DC paused message to overview

benjaminapetersen:
Fix bugzilla 1449504

- [x] Add warning alerts to list row notifications
- [x] Add paused deployment alert handling
- [x] Disable the "deploy" action from the overview kebab when a dc/deployment is paused 

screen shot 2017-05-12 at 11 59 28 am

screen shot 2017-05-12 at 11 32 02 am

screen shot 2017-05-12 at 2 40 13 pm

screen shot 2017-05-12 at 2 55 24 pm

### DIFF
--- a/app/scripts/services/resourceAlerts.js
+++ b/app/scripts/services/resourceAlerts.js
@@ -4,9 +4,11 @@ angular.module("openshiftConsole")
   .factory("ResourceAlertsService",
            function($filter,
                     AlertMessageService,
+                    DeploymentsService,
                     Navigate,
                     QuotaService) {
     var annotation = $filter('annotation');
+    var humanizeKind = $filter('humanizeKind');
     var deploymentStatus = $filter('deploymentStatus');
     var getGroupedPodWarnings = $filter('groupedPodWarnings');
 
@@ -100,6 +102,34 @@ angular.module("openshiftConsole")
       }
     };
 
+    // deploymentConfig, k8s deployment
+    var getPausedDeploymentAlerts = function(deployment) {
+      var alerts = {};
+      if(_.get(deployment, 'spec.paused')) {
+        alerts[deployment.metadata.uid + '-paused'] = {
+          type: 'info',
+          message: deployment.metadata.name + ' is paused.',
+          detail: 'This will stop any new rollouts or triggers from running until resumed.',
+          links: [{
+            href: "",
+            label: 'Resume Rollouts',
+            onClick: function() {
+              DeploymentsService.setPaused(deployment, false, {namespace: deployment.metadata.namespace}).then(
+                _.noop,
+                function(e) {
+                  alerts[deployment.metadata.uid + '-pause-error'] = {
+                    type: "error",
+                    message: "An error occurred resuming the " + humanizeKind(deployment.kind) + ".",
+                    details: $filter('getErrorDetails')(e)
+                  };
+                });
+            }
+          }]
+        };
+      }
+      return alerts;
+    };
+
     var getDeploymentStatusAlerts = function(deploymentConfig, mostRecentRC) {
       if (!deploymentConfig || !mostRecentRC) {
         return {};
@@ -143,7 +173,6 @@ angular.module("openshiftConsole")
         };
         break;
       }
-
       return alerts;
     };
 
@@ -180,6 +209,7 @@ angular.module("openshiftConsole")
       getPodAlerts: getPodAlerts,
       setGenericQuotaWarning: setGenericQuotaWarning,
       getDeploymentStatusAlerts: getDeploymentStatusAlerts,
+      getPausedDeploymentAlerts: getPausedDeploymentAlerts,
       getServiceInstanceAlerts: getServiceInstanceAlerts
     };
   });
--- a/app/views/overview/_list-row-actions.html
+++ b/app/views/overview/_list-row-actions.html
@@ -15,7 +15,10 @@
               <a href="" ng-click="row.startBuild(row.buildConfigs[0])">Start Build</a>
             </li>
             <li ng-if="'deploymentconfigs/instantiate' | canI : 'create'" role="menuitem">
-              <a href="" ng-click="row.startDeployment()">Deploy</a>
+              <a href="" ng-if="row.canDeploy()" ng-click="row.startDeployment()">Deploy</a>
+              <a href="" ng-if="!(row.canDeploy())" class="disabled-link" aria-disabled="true">
+                Deploy <span ng-if="row.isPaused()">(Paused)</span>
+              </a>
             </li>
             <li ng-if="'deploymentconfigs' | canI : 'update'" role="menuitem">
               <a ng-href="{{row.apiObject | editResourceURL}}">Edit</a>
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -501,7 +501,11 @@ function OverviewController($scope,
 
     // Add any failed / canceled deployment notifications.
     var mostRecent = getMostRecentReplicationController(deploymentConfig);
-    notifications = ResourceAlertsService.getDeploymentStatusAlerts(deploymentConfig, mostRecent);
+    _.assign(
+      notifications,
+      ResourceAlertsService.getDeploymentStatusAlerts(deploymentConfig, mostRecent),
+      ResourceAlertsService.getPausedDeploymentAlerts(deploymentConfig)
+    );
 
     // Roll up notifications like pod warnings for any visible replication controller.
     var visibleReplicationControllers = getVisibleReplicationControllers(deploymentConfig);
@@ -534,7 +538,7 @@ function OverviewController($scope,
   //   key: object UID
   //   value: alerts object
   var updateDeploymentWarnings = function(deployment) {
-    var notifications = {};
+    var notifications = ResourceAlertsService.getPausedDeploymentAlerts(deployment);
 
     // Roll up notifications like pod warnings for any visible replica set.
     var visibleReplicaSets = getVisibleReplicaSets(deployment);
--- a/app/scripts/directives/overview/listRow.js
+++ b/app/scripts/directives/overview/listRow.js
@@ -225,6 +225,30 @@ function OverviewListRow($filter,
       });
   };
 
+  row.canDeploy = function() {
+    if (!row.apiObject) {
+      return false;
+    }
+
+    if (row.apiObject.metadata.deletionTimestamp) {
+      return false;
+    }
+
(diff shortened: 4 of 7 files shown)


spadgett:
I don't think we show call info messages warnings. Then you could have two types of warnings, which is confusing.

@ncameronbritt Any suggest for the label by the info icon in collapsed state (first screenshot above)?

benjaminapetersen:
Yeah thats true, seems weird 🙁

benjaminapetersen:
I could say `1 issue` ~ but is that splitting hairs?

spadgett:
It's not really in an issue, more "nothing is wrong, but you should know this..." Maybe "1 message"?

ncameronbritt:
The first thing that popped in my mind was "alert".  "Heads-up", but maybe that's too colloquial.

benjaminapetersen:
Yeah that makes sense, I like "message".

benjaminapetersen:
screen shot 2017-05-12 at 3 01 13 pm

ncameronbritt:
Seeing it, I like "message"

benjaminapetersen:
`message` vs `notice`?

app/scripts/services/resourceAlerts.js
```diff
@@ -100,6 +101,33 @@ angular.module("openshiftConsole")
       }
     };
 
+    // deploymentConfig, k8s deployment
+    var getPausedDeploymentAlerts = function(alerts, deployment) {
+      if(deployment.spec.paused) {
+        alerts[deployment.metadata.uid + '-paused'] = {
+          type: 'info',
+          message: deployment.metadata.name + ' is paused.',
+          detail: 'This will stop any new rollouts or triggers from running until resumed',
```
spadgett: Nit: add a period at the end

app/scripts/services/resourceAlerts.js
```diff
@@ -100,6 +101,33 @@ angular.module("openshiftConsole")
       }
     };
 
+    // deploymentConfig, k8s deployment
+    var getPausedDeploymentAlerts = function(alerts, deployment) {
```
spadgett: Here you take in the `alerts` object as a parameter, but in `getDeploymentStatusAlerts`, it's a return value. Would be good to be consistent

app/views/overview/_list-row-actions.html
```diff
@@ -15,7 +15,11 @@
               <a href="" ng-click="row.startBuild(row.buildConfigs[0])">Start Build</a>
             </li>
             <li ng-if="'deploymentconfigs/instantiate' | canI : 'create'" role="menuitem">
-              <a href="" ng-click="row.startDeployment()">Deploy</a>
+              <a
+                href="" 
+                ng-attr-aria-disabled="{{canDeploy() ? undefined : 'true'}}"
+                ng-class="{ 'disabled-link': !canDeploy() }"
```
spadgett: I don't think adding the class actually prevents you from clicking the menu item. Won't that still cause an error if you click the disabled item?

app/scripts/services/resourceAlerts.js
benjaminapetersen: Agree!  Second one is modifying the array:

```javascript
    notifications = ResourceAlertsService.getDeploymentStatusAlerts(deploymentConfig, mostRecent);
    notifications = ResourceAlertsService.getPausedDeploymentAlerts(notifications, deploymentConfig);
```
Maybe it would be better like this:
```javascript
var notifications = {};
angular.extend(notifications, ResourceAlertsService.getDeploymentStatusAlerts(deploymentConfig, mostRecent), ResourceAlertsService.getPausedDeploymentAlerts(deploymentConfig))
```
And eliminate the need to pass in the alerts inconsistently.

app/scripts/services/resourceAlerts.js
benjaminapetersen: will do

app/views/overview/_list-row-actions.html
benjaminapetersen: Let me double check.

app/scripts/services/resourceAlerts.js
spadgett: Works for me

spadgett: I have a slight preference for `_.assign`

app/views/overview/_list-row-actions.html
benjaminapetersen: I can do something like this:

```html
            <li ng-if="'deploymentconfigs/instantiate' | canI : 'create'" role="menuitem">
              <a
                href=""
                ng-if="canDeploy()"
                ng-click="row.startDeployment()">Deploy</a>
              <a
                href=""
                ng-if="!(canDeploy())"
                ng-attr-aria-disabled="{{canDeploy() ? undefined : 'true'}}"
                ng-class="{ 'disabled-link': !canDeploy() }">Deploy (paused)</a>
            </li>
```

screen shot 2017-05-15 at 12 53 50 pm

spadgett: Yeah, I think we've done that in other places. Note that "paused" isn't always the reason it's disabled, so if we add that message need to do it for other conditions

spadgett: @ncameronbritt Opinion on putting the reason in parens for the disabled action?

app/scripts/controllers/newOverview.js
```diff
@@ -533,7 +537,8 @@ function OverviewController($scope,
   //   value: alerts object
   var updateDeploymentWarnings = function(deployment) {
     var notifications = {};
-
+    _.assign(notifications, ResourceAlertsService.getPausedDeploymentAlerts(deployment));
```
spadgett: Nit: I'd rather keep it simple here and use

```js
var notifications = ResourceAlertsService.getPausedDeploymentAlerts(deployment);
```

since you're not mixing together multiple objects.

app/views/overview/_list-row-actions.html
```diff
@@ -15,7 +15,15 @@
...
-              <a href="" ng-click="row.startDeployment()">Deploy</a>
+              <a
+                href=""
+                ng-if="row.canDeploy()"
+                ng-click="row.startDeployment()">Deploy</a>
+              <a
+                href=""
+                ng-if="!(row.canDeploy())"
+                ng-attr-aria-disabled="{{row.canDeploy() ? undefined : 'true'}}"
+                ng-class="{ 'disabled-link': !row.canDeploy() }">Deploy (paused)</a>
```
spadgett: No reason for the conditional checks here since you already have the `ng-if`. We can't say "(paused)" unless we check that's the reason it's disabled. It could also be being deleted for instance.

```html
<a
  href=""
  ng-if="!(row.canDeploy())"
  aria-disabled="true"
  class="disabled-link">Deploy</a>
```

benjaminapetersen: I can wrap the `(paused)` with a `span` and `ng-if="isPaused()"` 

Also... should the above `switch` be updated? I was considering updating the `switch` to `DeploymentConfig || Deployment`, but not sure if that is right. Do we want a `Resume Rollouts` button added here?  I initially assumed we should keep this menu slim, but seems like that might actually be pretty helpful in this case.

spadgett: > I can wrap the (paused) with a span and ng-if="isPaused()"

Seems reasonable. Should probably be `(Paused)` with a capital.

> Also... should the above switch be updated? I was considering updating the switch to DeploymentConfig || Deployment, but not sure if that is right.

We don't support manual rollouts for upstream deployments in the web console. @jwforres switched to DeploymentRequest for DeploymentConfigs (#1434)... I suspect the API for upstream deployments is similar, although I haven't looked at it. We probably should add it, but not for this PR.

>  Do we want a Resume Rollouts button added here?

I think the link in the notification is good enough, and what you have currently is consistent with what we do in the deployment configs page.

spadgett: OK I looked and there is no equivalent to "Deploy" that I can find for upstream deployments. `oc rollout latest` only works for deployment configs.

benjaminapetersen: ok neat, thanks!

app/scripts/controllers/newOverview.js
benjaminapetersen: ill have this cleanup in my next update.

benjaminapetersen:
screen shot 2017-05-22 at 1 58 35 pm

benjaminapetersen:
updated/rebased

app/scripts/services/resourceAlerts.js
```diff
@@ -100,6 +101,34 @@ angular.module("openshiftConsole")
...
+          links: [{
+            href: "",
+            label: 'Resume Rollouts',
+            onClick: function() {
+              DeploymentsService.setPaused(deployment, false, {namespace: deployment.metadata.namespace}).then(
+                _.noop,
+                function(e) {
+                  alerts[deployment.metadata.uid + '-pause-error'] = {
+                    type: "error",
+                    message: "An error occurred resuming the deployment config.",
```
spadgett: Need to show the right message depending on the kind.

app/views/overview/_list-row-actions.html
```diff
@@ -15,7 +15,17 @@
...
-              <a href="" ng-click="row.startDeployment()">Deploy</a>
+              <a
+                href=""
+                ng-if="row.canDeploy()"
+                ng-click="row.startDeployment()">Deploy</a>
+              <a
+                href=""
+                ng-if="!(row.canDeploy())"
+                class="disabled-link"
+                disabled>
```
spadgett: `disabled` should probably be `aria-disabled="true"`

app/scripts/services/resourceAlerts.js
benjaminapetersen: ah, true, ill `humanize(kind)`

app/views/overview/_list-row-actions.html
benjaminapetersen: Cool, i'll swap. I did a little research briefly and found some things implying that redundant aria can be dropped, but ill roll with consistent.

spadgett: But `disabled` is not normally an attribute on links like it is for buttons, so I'm not sure if screen readers will respect it. If we really want to be correct, I think we should have 

```
role="button" aria-disabled="true" tabindex="-1"
```

benjaminapetersen: yeah thats true.  I'll add the `role="button"`, that should be there as well.  I don't think we are using tabindex regularly though, correct?

spadgett:
[merge]

PR #1549, opened 2017-05-15
Use image listing from registry-image-widgets

petervo:
Rebase of #1136

Had to move it here because I couldn't get access to update the other one.


petervo:
Replaced by #1136 after all.

PR #1550, opened 2017-05-15
Add inner scroll shadow to responsive tables

rhamilto:
Fixes https://github.com/openshift/origin-web-console/issues/1237

screen shot 2017-05-15 at 3 20 35 pm
screen shot 2017-05-15 at 3 20 42 pm
screen shot 2017-05-15 at 3 20 49 pm

Note:  this PR includes a mixin for adding inner scroll shadows to non-table elements, but the mixin hasn't been applied anywhere as I was unable to find a cross-browser compatible solution for the intended use case (copy-to-clipboard textareas).  But it seems like it could have utility for other elements, so I've included it here.


spadgett:
[merge]

PR #1553, opened 2017-05-17
Bug 1449908 - Group replica sets by owner reference

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1449908

Use owner references to decide if a replica set is owned by a deployment. Change maps to use deployment UID instead of deployment name. This avoids problems when a deployment is deleted and another is created with the same name.

openshift web console 2017-05-17 12-09-50

### DIFF
--- a/app/scripts/controllers/newOverview.js
+++ b/app/scripts/controllers/newOverview.js
@@ -20,6 +20,7 @@ angular.module('openshiftConsole').controller('NewOverviewController', [
   'Logger',
   'MetricsService',
   'Navigate',
+  'OwnerReferencesService',
   'ProjectsService',
   'ResourceAlertsService',
   'RoutesService',
@@ -44,6 +45,7 @@ function OverviewController($scope,
                             Logger,
                             MetricsService,
                             Navigate,
+                            OwnerReferencesService,
                             ProjectsService,
                             ResourceAlertsService,
                             RoutesService) {
@@ -155,7 +157,7 @@ function OverviewController($scope,
   var size = function() {
     return _.size(overview.deploymentConfigs) +
            _.size(overview.vanillaReplicationControllers) +
-           _.size(overview.deployments) +
+           _.size(overview.deploymentsByUID) +
            _.size(overview.vanillaReplicaSets) +
            _.size(overview.statefulSets) +
            _.size(overview.monopods) +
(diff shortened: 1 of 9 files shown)


app/scripts/controllers/newOverview.js
```diff
@@ -1195,13 +1197,13 @@ function OverviewController($scope,
       group: "extensions",
       resource: "deployments"
     }, context, function(deploymentData) {
-      overview.deployments = deploymentData.by('metadata.name');
+      overview.deploymentsByUID = deploymentData.by('metadata.uid');
```
jwforres: this isnt new, but were we not sorting deployments before?  how is it being ordered when you are looking at the by Resource Type view?

app/scripts/services/ownerReferences.js
```diff
@@ -0,0 +1,28 @@
...
+  .factory("OwnerReferencesService", function() {
+    var getOwnerReferences = function(apiObject) {
+      return _.get(apiObject, 'metadata.ownerReferences');
+    };
+    return {
+      getOwnerReferences: getOwnerReferences,
+      groupByControllerUID: function(apiObjects) {
+        return _.groupBy(apiObjects, function(apiObject) {
+          var ownerRefs = getOwnerReferences(apiObject);
+          var controllerRef = _.find(ownerRefs, { controller: true });
```
jwforres: is it possible (in theory) for more than one controller to claim ownership, if that were the case you would want the item to appear under each group and not just the first one you find.   Would be curious if the API prevents you from having more than one ownerRef with controller true.  Could try manually adding an ownerRef on a pod that points to a random controller that exists.

app/scripts/controllers/newOverview.js
spadgett: Looks like an oversight and they weren't sorted.

spadgett:
@jwforres updated

app/scripts/controllers/newOverview.js
jwforres: can you open an issue for this so we dont lose track of it

jwforres:
[merge]

app/scripts/controllers/newOverview.js
spadgett: Opened #1555

PR #1556, opened 2017-05-18
Sort deployments on the overview

spadgett:
Fixes #1555

PR #1557, opened 2017-05-18
Specify list-pf-content as a block element to prevent layout from breaking

sg00dwin:
after updating to latest PF.

Override bug introduced in PF v3.23.1 where list-pf-content was changed to display: flex

### DIFF
--- a/app/styles/_list-pf.less
+++ b/app/styles/_list-pf.less
@@ -17,7 +17,9 @@
 }
 
 .list-pf-content {
-  flex-grow: 1;
+  // Make block element to override bug introduced in PF v3.23.1 where list-pf-content was changed to display: flex
+  // TODO: remove if reverted in PF
+  display: block;
   .alert:first-child {
     margin-top: 0;
   }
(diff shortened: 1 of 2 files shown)


spadgett:
Travis is failing on installing jQuery. You'll probably need my bower.json change from #1556 or wait for that to merge. @jwforres FYI

app/styles/_list-pf.less
```diff
@@ -17,6 +17,9 @@
 }
 
 .list-pf-content {
+  // Make block element to override bug introduced in PF v3.23.1 where list-pf-content was changed to display: flex
+  // TODO: remove if reverted in PF
+  display: block;
   flex-grow: 1;
```
spadgett: Do we still need `flex-grow` with `display: block`?

spadgett:
It looks like the expanded section should have `list-pf-container` instead of `list-pf-content` when I look at the example here:

https://rawgit.com/andresgalante/patternfly/pf-list-dist/dist/tests/list-pf.html

Although it's a little different there since it has multiple levels of expansion. I'd rather just remove that class instead of overriding styles to insulate us from future changes. @sg00dwin @rhamilto Thoughts?

screen shot 2017-05-19 at 8 14 37 am

spadgett:
Well nevermind it has `list-pf-content` too as a child. I still feel like it might be safer to remove that?

spadgett:
[test]

sg00dwin:
@spadgett  We could remove the `.list-pf-content` markup from inside a the row.expanded. Would need to move a couple of alert styles to have a different parent `.list-pf-container`?
https://github.com/openshift/origin-web-console/blob/master/app/styles/_list-pf.less#L21-L26

@rhamilto would you relocate these to another partial?

spadgett:
@sg00dwin I'd suggest just renaming it to `overview-row-expanded` or something. I'm OK with the change you have right now as well, though.

spadgett:
I'm good with this change. We can only look at changing the class name as a follow up if we decide it's better.

[merge]

PR #1558, opened 2017-05-19
Bug Fix 1452018 and 1451708

zherman0:
Fixed disabled button still going to linked page even though it appears disabled.  Fixed logic error that prevent exact quota values from being accepted.

### DIFF
--- a/app/views/storage.html
+++ b/app/views/storage.html
@@ -30,11 +30,13 @@
             <div class="col-md-12">
               <div class="section-header page-header-bleed-right page-header-bleed-left">
                 <div class="hidden-xs pull-right" ng-if="project && ('persistentvolumeclaims' | canI : 'create')">
-                  <a ng-disabled="outOfClaims" ng-href="project/{{project.metadata.name}}/create-pvc" class="btn btn-default">Create Storage</a>
+                  <a ng-if="!outOfClaims" ng-href="project/{{project.metadata.name}}/create-pvc" class="btn btn-default">Create Storage</a>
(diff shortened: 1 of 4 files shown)


app/views/storage.html
```diff
@@ -30,11 +30,13 @@
             <div class="col-md-12">
               <div class="section-header page-header-bleed-right page-header-bleed-left">
                 <div class="hidden-xs pull-right" ng-if="project && ('persistentvolumeclaims' | canI : 'create')">
-                  <a ng-disabled="outOfClaims" ng-href="project/{{project.metadata.name}}/create-pvc" class="btn btn-default">Create Storage</a>
+                  <a ng-if="!outOfClaims" ng-href="project/{{project.metadata.name}}/create-pvc" class="btn btn-default">Create Storage</a>
+                  <a ng-if="outOfClaims" ng-href="" ng-disabled="true" class="btn btn-default">Create Storage</a>
```
spadgett: Just add the `disabled` class, and no need to use `ng-href`.

```html
<a ng-if="outOfClaims" href="" class="btn btn-default disabled">Create Storage</a>
```

Probably a good idea to tell screenreaders it's disabled, too, with `aria-disabled`


```html
<a ng-if="outOfClaims" href="" class="btn btn-default disabled" aria-disabled="true">Create Storage</a>
```

Also make the change to the same button below under `visible-xs-block`

zherman0:
Thanks @spadgett - made the changes as suggested

spadgett:
[merge]

PR #1563, opened 2017-05-22
Resolving issue where mini log viewer clips text with no eliding or a…

### LINKED ISSUE #1559: Mini log viewer clips text with no eliding or ability to scroll
When a log line is too long for the mini log viewer on the overview it truncates the text, but there is no ellipsis which makes it look like you should be able to copy/scroll to get the full content.  However its really just truncated with no way to view the rest of the line.
clipped-log-viewer

rhamilto:
…bility to scroll

Fixes #1559

Bonus:  improves vertical alignment of “View full log” link so it sits
on the same baseline as the text on the left

screen shot 2017-05-22 at 10 17 10 am
screen shot 2017-05-22 at 10 17 17 am

### DIFF
--- a/app/styles/_overview.less
+++ b/app/styles/_overview.less
@@ -43,11 +43,11 @@
   margin-bottom: 5px;
   margin-top: 5px;
   padding: 5px;
-  .truncate();
   .mini-log-content {
     line-height: @mini-log-line-height;
     min-height: (@mini-log-num-lines * @mini-log-line-height);
     .mini-log-line {
+      .truncate();
       white-space: pre;
     }
   }
@@ -457,6 +457,10 @@
     @media (max-width: @screen-xs-max) {
       margin-top: 10px;
     }
+  }
+  .view-full-log {
+    position: relative;
+    top: 2px;
   }
 
   // Deployment animations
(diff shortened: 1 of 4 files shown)


app/styles/_overview.less
```diff
@@ -43,12 +43,11 @@
   margin-bottom: 5px;
   margin-top: 5px;
   padding: 5px;
-  .truncate();
   .mini-log-content {
     line-height: @mini-log-line-height;
     min-height: (@mini-log-num-lines * @mini-log-line-height);
     .mini-log-line {
-      white-space: pre;
```
spadgett: This improves our newline handling, but will also collapse spaces and tabs. Trying to decide if that's a problem.

rhamilto: I incorrectly assumed that would break .truncate()'s [.text-overflow()](https://github.com/twbs/bootstrap/blob/master/less/mixins/text-overflow.less#L7).  It looks like it doesn't, so I've restored it.

rhamilto:
[test]

spadgett:
Thanks @rhamilto 

[merge]

PR #1565, opened 2017-05-22
Removing unused Messenger dependency and related styles

rhamilto:
We are no longer using Messenger, so there is no need to keep it included in our code base.


rhamilto:
[test]

spadgett:
[merge]

PR #1567, opened 2017-05-22
Routes - convert form alerts to toast notifications for creating, editing and deleting 

sg00dwin:
@spadgett ptal

Task
- [x] Modify js to enable persistent error/warning notification to auto dismiss upon successful form submission.

### DIFF
--- a/app/scripts/controllers/createRoute.js
+++ b/app/scripts/controllers/createRoute.js
@@ -17,9 +17,9 @@ angular.module('openshiftConsole')
                        AuthorizationService,
                        DataService,
                        Navigate,
+                       NotificationsService,
                        ProjectsService,
                        keyValueEditorUtils) {
-    $scope.alerts = {};
     $scope.renderOptions = {
       hideFilterWidget: true
     };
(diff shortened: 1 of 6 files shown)


app/scripts/controllers/createRoute.js
```diff
@@ -17,6 +17,7 @@ angular.module('openshiftConsole')
                        AuthorizationService,
                        DataService,
                        Navigate,
+                       NotificationsService,
                        ProjectsService,
                        keyValueEditorUtils) {
     $scope.alerts = {};
```
spadgett: You can remove `$scope.alerts` since it's no longer used and also remove the `alerts` element from create-route.html

app/scripts/controllers/edit/route.js
```diff
@@ -17,6 +17,7 @@ angular.module('openshiftConsole')
                        AuthorizationService,
                        DataService,
                        Navigate,
+                       NotificationsService,
                        ProjectsService,
                        RoutesService) {
     $scope.alerts = {};
```
spadgett: Same comment here: remove `$scope.alerts` and remove `<alerts>` from the view

spadgett:
> Modify js to enable persistent error/warning notification to auto dismiss upon successful form submission.

Let's open a separate issue so we don't forget if we don't do it in this PR. @jwforres FYI

jwforres:
If we open a separate issue for that it needs to be fixed by stagecut.  I don't want to have us in a state in online where you submit a successful form and have to manually close old errors.

spadgett:
@jwforres We might hold off on this PR in that case. It requires changes to `NotificationsService` in origin-web-common. But I do think we could fix the `AlertsMessageService` cases now, which won't have this problem.

spadgett:
@sg00dwin @jwforres Maybe I can knock the common change for this out quickly.

jwforres:
ok lets split the PR then if needed

jwforres:
or that, either way

spadgett:
I opened a PR that adds remove and truncates long notification text:

https://github.com/openshift/origin-web-common/pull/76

spadgett:
@sg00dwin I opened a PR against your branch with the changes, but you'll need the updates I made to common.

https://github.com/sg00dwin/origin-web-console/pull/3

spadgett:
Added [WIP] since this is waiting on a origin-web-common release now.

spadgett:
[merge]

PR #1573, opened 2017-05-23
Only show service catalog resources when available

spadgett:
This lets us show the new landing page without the service catalog running.

See also https://github.com/openshift/origin-web-catalog/pull/241

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -5,6 +5,7 @@ angular.module('openshiftConsole').controller('OverviewController', [
   '$filter',
   '$routeParams',
   'AlertMessageService',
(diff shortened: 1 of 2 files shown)


jwforres:
Just double checking, you don't see Create Binding in the kebab actions for deployments right.  I think we hid it there weren't any services.

spadgett:
Yeah I tested it. We only show it if pod_presets is on and there are service instances (which will be empty if the service catalog isn't running).

spadgett:
https://github.com/openshift/origin-web-console/blob/master/app/views/overview/_list-row-actions.html#L28

jwforres:
[merge]

PR #1574, opened 2017-05-23
Show owner reference links on browse pages

spadgett:
Add links to an object's owners on various browse pages. For instance, here is a replica set owned by a deployment:

openshift web console 2017-05-23 12-52-37

@jwforres Thoughts on this change?

For now, I intentionally left out some relationships we already had special handling for like

build -> bc
pod -> rc -> dc

### DIFF
--- a/app/scripts/filters/navigate.js
+++ b/app/scripts/filters/navigate.js
@@ -1,0 +1,36 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  // Resource is either a resource object, or a name.  If resource is a name, kind and namespace must be specified
+  // Note that builds and deployments can only have their URL built correctly (including their config in the URL)
+  // if resource is an object
+  .filter('navigateResourceURL', function(Navigate) {
+    return function(resource, kind, namespace, apiVersion) {
+      return Navigate.resourceURL(resource, kind, namespace, null, {apiVersion: apiVersion});
+    };
+  })
+  .filter('configURLForResource', function(Navigate) {
+    return function(resource, /* optional */ action) {
+      return Navigate.configURLForResource(resource, action);
+    };
+  })
+  .filter('editResourceURL', function(Navigate) {
+    return function(resource, kind, namespace) {
+      var url = Navigate.resourceURL(resource, kind, namespace, "edit");
+      return url;
+    };
+  })
+  .filter('editYamlURL', function(Navigate) {
+    return function(object, /* optional */ returnURL) {
+      return Navigate.yamlURL(object, returnURL);
+    };
+  })
+  .filter('urlForOwnerRef', function(Navigate) {
+    return function(ownerReference, namespace) {
+      if (!ownerReference) {
+        return null;
+      }
+
+      return Navigate.resourceURL(ownerReference.name, ownerReference.kind, namespace);
+    };
+  });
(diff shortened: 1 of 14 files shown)


app/views/browse/_pod-details.html
```diff
@@ -25,6 +25,7 @@
           <a ng-href="{{rcName | navigateResourceURL : 'ReplicationController' : pod.metadata.namespace}}"
             ><span ng-if="deploymentVersion">#{{deploymentVersion}}</span><span ng-if="!deploymentVersion">{{rcName}}</span></a></span>
         </dd>
+        <owner-references ng-if="!dcName" api-object="pod"></owner-references>
```
spadgett: @jwforres Drawback here is that if the pod has other owner references besides the rc, they won't show up. But switching to always use `<owner-references>` instead of the special handling above changes

```
Deployment: mongodb, #1
```

to

```
Replication Controller: mongodb-1
```

Can't decide what's best.

app/scripts/filters/util.js
```diff
@@ -348,30 +348,6 @@ angular.module('openshiftConsole')
           .replace(/^./, function(str){ return str.toUpperCase(); });
     };
   })
-  // Resource is either a resource object, or a name.  If resource is a name, kind and namespace must be specified
-  // Note that builds and deployments can only have their URL built correctly (including their config in the URL)
-  // if resource is an object
-  .filter('navigateResourceURL', function(Navigate) {
```
spadgett: We were accumulating enough navigation filters that I moved them to a separate file.

app/scripts/filters/navigate.js
```diff
@@ -0,0 +1,36 @@
...
+      return Navigate.yamlURL(object, returnURL);
+    };
+  })
+  .filter('urlForOwnerRef', function(Navigate) {
+    return function(ownerReference, namespace) {
+      if (!ownerReference) {
+        return null;
+      }
+
+      return Navigate.resourceURL(ownerReference.name, ownerReference.kind, namespace);
```
jwforres: If the ownerRef points to a resource that would navigate to the Other Resources page, and the user doesn't have List rights to that resource, I'm not sure what it will do.  Its not likely, but probably something we should at least handle be checking what kinds are available after the filtering when the page loads and showing some kind of generic "Could not list <resource>"

 Its an existing issue but this would make us more likely to hit it i think.

app/views/browse/_pod-details.html
jwforres: i dont want to remove the link to the DeploymentConfig, since thats where you can edit the configuration

jwforres:
The general concern I have with this is its going to show things like the Template Instance object in that list, are we sure we really want to show that without any context?

spadgett:
We could whitelist kinds or only link to certain kinds with dedicated pages.

jwforres:
Not against that idea. But if we did want to show all potential owner refs
and not maintain a whitelist long term, then another option would be to
make the label "Owned by:" and the value the list of "Kind name".  It would
bother me less to see "Owned by: Template instance foo" than to see
"Template instance: foo"  and be wondering why i care about it.

spadgett:
Closing for now in favor of #1732. We might want to revisit a more general solution later.

PR #1576, opened 2017-05-23
Removing orphaned oscKeyValues directive and spec test

### LINKED ISSUE #1209: Remove unused osc-key-values directive
It's been replaced by the key value editor, but the old code is still there.

rhamilto:
Fixes #1209


spadgett:
[merge]

PR #1578, opened 2017-05-23
Update bind service directive to use common components/service. Bump Versions

### LINKED ISSUE #1465: Bind message says it successfully bound the service to itself
screen shot 2017-04-24 at 11 31 25 am

cc @jwforres

jeff-phillips-18:
Bumps version of origin-web-common to v0.0.26
Bumps version of origin-web-catalog to v0.0.15

Uses common pages for binding in the bindService directive
Updates the process template dialog to use the angular patternfly wizard (keeps consistent styling)

Fixes #1465


app/scripts/controllers/landingPage.js
```diff
@@ -51,10 +52,21 @@ angular.module('openshiftConsole')
 
     AuthService.withUser().then(function() {
       var includeTemplates = !_.get(Constants, 'ENABLE_TECH_PREVIEW_FEATURE.template_service_broker');
-      Catalog.getCatalogItems(includeTemplates).then(function(items) {
+      Catalog.getCatalogItems(includeTemplates).then(_.spread(function(items, errorMessage) {
+        if (errorMessage) {
+          var alert = {
+            name: 'error-loading-catalog-items',
+            data: {
+              type: 'warning',
```
spadgett: Suggest an error here rather than a warning.

I'm surprised DataService doesn't show an error if the list fails. Won't there be two errors?

app/scripts/directives/bindService.js
```diff
@@ -70,28 +73,46 @@ function BindService($filter,
...
+    }
+    ctrl.nextTitle = "Close";
+    ctrl.wizardComplete = true;
+
+    ctrl.bindService();
+  };
+
+  ctrl.$onInit = function() {
+    ctrl.serviceSelection = {};
+    var formStepId = (ctrl.target.kind === 'Instance') ? 'applications' : 'services';
```
spadgett: Would be it be simpler to have one step ID and then just change the step indicator label? I don't see where the ID is checked anywhere.

app/scripts/directives/bindService.js
```diff
@@ -151,89 +177,33 @@ function BindService($filter,
...
+
+    BindingService.bindService(context, _.get(svcToBind, 'metadata.name'), _.get(appToBind, 'metadata.name')).then(function(binding){
       ctrl.binding = binding;
+      ctrl.error = null;
 
-      DataService.watchObject({
-        group: 'servicecatalog.k8s.io',
-        resource: 'bindings'
-      }, _.get(ctrl.binding, 'metadata.name'), context, function(binding) {
+      DataService.watchObject(BindingService.bindingResource, _.get(ctrl.binding, 'metadata.name'), context, function(binding) {
```
spadgett: Not a new problem, but don't we need to be removing the watch when on scope destroy?

app/scripts/filters/resources.js
```diff
@@ -1351,8 +1316,21 @@ angular.module('openshiftConsole')
...
   })
   .filter('hasAlternateBackends', function() {
-    return function(route) {
+    return function (route) {
       var alternateBackends = _.get(route, 'spec.alternateBackends', []);
       return !_.isEmpty(alternateBackends);
     };
+  })
+  // TODO: these two filters are defined in origin-web-catalog but are not available, need to figure out why or move to common
+  .filter('applicationHasDeployment', function ($filter) {
```
spadgett: We have both filters already in origin-web-console

https://github.com/openshift/origin-web-console/blob/master/app/scripts/filters/resources.js#L719-L728

Although if we move them to common it might be better to remove the dependency on DeploymentsService so we don't have to move the entire service.

app/views/directives/process-template-dialog.html
```diff
@@ -1,74 +1,49 @@
...
+    <div pf-wizard
+         hide-header="true"
+         hide-sidebar="true"
+         step-class="order-service-wizard-step"
+         wizard-ready="$ctrl.wizardReady"
+         next-title="$ctrl.nextTitle"
+         on-finish="$ctrl.close()"
+         on-cancel="$ctrl.close()"
+         wizard-done="$ctrl.wizardDone">
+      <div pf-wizard-step ng-repeat="step in $ctrl.steps track by $index"
```
spadgett: `track by step.id` might be better?

spadgett:
Template dialog is empty the second time it's opened:

openshift web console 2017-05-23 17-39-56

spadgett:
(Only after instantiating the template.)

spadgett:
I see this runtime error after canceling the template dialog. Appears to be a new problem.

```
angular.js:14199 TypeError: Cannot read property 'metadata' of null
    at Object.createFromTemplateURL (navigate.js:106)
    at resources.js:491
    at fn (eval at compile (angular.js:15126), <anonymous>:4:471)
    at regularInterceptedExpression (angular.js:16354)
    at expressionInputsWatch (angular.js:16282)
    at Scope.$digest (angular.js:17828)
    at Scope.$apply (angular.js:18102)
    at HTMLButtonElement.<anonymous> (angular-patternfly.js:9090)
    at HTMLButtonElement.dispatch (jquery.js:4435)
    at HTMLButtonElement.elemData.handle (jquery.js:4121)
```

spadgett:
Not related to this change, but I also noticed https://github.com/openshift/origin-web-catalog/issues/245 when testing.

spadgett:
The vertical divider extends too far down now (cc @rhamilto)

openshift web console 2017-05-23 17-56-49

spadgett:
We're showing the back button always on wizard now. Previously we were hiding it for two-step wizard because it was always disabled.

spadgett:
Seeing some other layout problems as well :/

openshift web console 2017-05-23 18-01-16

app/scripts/controllers/landingPage.js
jeff-phillips-18: Will update to an error. the Catalog service getCatalogItems does not add any alerts/notifications as it does not set the errorNotification option in its calls to DataService.list

spadgett: Hm, `errorNotification` should default to true.

https://github.com/openshift/origin-web-common/blob/master/src/services/dataService.js#L950

jeff-phillips-18: Ah, OK. I will remove the addNotification call.

app/scripts/filters/resources.js
jeff-phillips-18: That dependency is why I didn't move them to common but added similar to catalog without the dependency. I was hesitant to add to common removing the dependency at this time.

spadgett: OK I think we should look at it as a follow on to avoid having the same filters in multiple places.

spadgett:
Disregard that last screenshot (which I've now deleted). I accidentally switched branches.

spadgett:
I have a fix for the kebab not working:

https://github.com/openshift/origin-web-common/pull/81

app/scripts/controllers/landingPage.js
jeff-phillips-18: Done.

app/scripts/directives/bindService.js
jeff-phillips-18: Done.

app/scripts/directives/bindService.js
jeff-phillips-18: Done.

app/views/directives/process-template-dialog.html
jeff-phillips-18: Done.

jeff-phillips-18:
@spadgett Fixed the issues you reported. Modified the structure a bit so the wizard would get re-created when re-displayed (resolves the blank page issue). Added a check for template in the process-template.html file since it still exists momentarily and updated when the template is set to null on close.

app/scripts/filters/resources.js
jeff-phillips-18: Added issue https://github.com/openshift/origin-web-common/issues/82

app/scripts/directives/bindService.js
```diff
@@ -70,28 +74,45 @@ function BindService($filter,
...
+  var showBind = function() {
+    ctrl.nextTitle = 'Bind';
+
+    validityWatcher = $scope.$watch("ctrl.selectionForm.$valid", function(isValid) {
+      ctrl.steps[0].valid = isValid;
     });
+  };
 
+  var showResults = function() {
+    if (validityWatcher) {
```
benjaminapetersen: Mind explaining the validityWatcher here?  I'm curious.  I don't know that I've seen a watch stored, explicitly called, and nulled out like this before.

app/scripts/directives/bindService.js
```diff
@@ -151,89 +177,36 @@ function BindService($filter,
...
+      ctrl.appToBind = ctrl.target;
     }
     // TODO: handle not having any service instances when binding app to service
-    ctrl.gotoStep(ctrl.steps[0]);
-  };
-
-  var humanizeKind = $filter('humanizeKind');
-  ctrl.groupByKind = function(object) {
-    return humanizeKind(object.kind);
   };
```
benjaminapetersen: nit: I tend to prefer `$onInit` at the beginning and `$onDestroy` at the end to make it easy/consistent to find the setup/tear down code. Not a big issue, just a pattern preference.

app/styles/_catalog.less
```diff
@@ -1,8 +1,37 @@
...
+.bind-service-config {
+  padding: 0 (@grid-gutter-width / 2);
   width: 100%;
+  @media (min-width: @screen-sm-min) {
+    margin-bottom: -(@grid-gutter-width / 2);
+    margin-top: -(@grid-gutter-width / 2);
+    overflow-y: auto;
+    padding-right: 0;
+    .scroll-shadows-vertical(65%, 0.25);
+    _::-webkit-full-page-media, _:future, :root & { // only target Safari
```
benjaminapetersen: gnarly selector :smile:

app/scripts/directives/bindService.js
jeff-phillips-18: I like to keep the ctrl.$... functions together

app/scripts/directives/bindService.js
jeff-phillips-18: The validity watcher is added to know when the form is valid, and we then update the next button enablement. It must be removed when we leave the page so we keep a handle to it and call it when the  page changes or the component is destroyed.

app/styles/_catalog.less
```diff
@@ -1,8 +1,37 @@
 @landing-search-area-height: 52px;
+@bind-service-page-height: 310px;
```
rhamilto: Nit:  alphabetization

app/styles/_catalog.less
```diff
@@ -1,8 +1,37 @@
...
+@bind-service-wizard-min-external-height: 300px;
 
-.bind-service-wizard .wizard-pf-contents {
-  display: block;
+.wizard-pf-footer.wizard-pf-position-override {
```
rhamilto: Nit: alphabetization here and the rest of the rules you are adding

app/scripts/controllers/landingPage.js
```diff
@@ -51,10 +52,20 @@ angular.module('openshiftConsole')
...
+      Catalog.getCatalogItems(includeTemplates).then(_.spread(function(items, errorMessage) {
+        if (errorMessage) {
+          var alert = {
+            name: 'error-loading-catalog-items',
+            data: {
+              type: 'error',
+              message: errorMessage
+            }
+          };
+          AlertMessageService.addAlert(alert);
```
spadgett: Adding the alert to `AlertMessageService` will show the message on the next page the user navigates to, which I don't think is what we want.

I suggest we change the `Catalog` service to pass `{ errorNotification: false }` and then show its formatted message here. In the meantime, we could either not do anything with the message or possibly just show separate alerts until we make the catalog change?

jeff-phillips-18: Probably best to have duplicates rather than none at all. I will add the notification back here, remove the alert here and file an issue for catalog to not add notifications.

jeff-phillips-18:
@rhamilto Fixed the alphabetization issues, thanks.

app/scripts/controllers/landingPage.js
spadgett: Thanks, works for me

spadgett:
The extra space between the buttons here looks odd to me when the back button is hidden.

openshift web console 2017-05-24 12-24-25

spadgett:
Also it looks like the Bind button text is not correctly centered?

jeff-phillips-18:
Fixed the spacing for the buttons, added an override for not until angular patternfly fix can be made (https://github.com/patternfly/angular-patternfly/issues/469)

The button text is also an angular patternfly issue adding an extra space after the title. This we will need to wait for a fix on (https://github.com/patternfly/angular-patternfly/issues/468)

spadgett:
[merge]

PR #1582, opened 2017-05-24
Fixing grunt serve:dist so it works

### LINKED ISSUE #1325: grunt serve:dist doesn't work
We have a `dist` target for `grunt serve`. It doesn't work, however, since the config is not wired up. This would be helpful to test problems that only happen with minified code.

rhamilto:
Fixes #1325


jwforres:
[merge]

Issue #1218, opened 2017-02-03
Role descriptions are not fully visible in dropdown.


benjaminapetersen:
The new role description are nice, but due to length they aren't especially readable in the dropdown:

![screen shot 2017-02-03 at 11 29 47 am](https://cloud.githubusercontent.com/assets/280512/22599239/2285386c-ea04-11e6-8b2c-800fcdff383e.png)

benjaminapetersen:
@jwforres do we want to ask anyone from ux to give some feedback on this?

jwforres:
Did you try letting them wrap? How does that look?

benjaminapetersen:
I didn't because I didn't want 1 entry to end up like 4 lines long, but I can give it a spin & see.

sg00dwin:
@benjaminapetersen @jwforres I went ahead and made a version where the description is wrapped and added some height to the dropdown. I think it works. Thoughts?

<img width="372" alt="screen shot 2017-04-17 at 4 25 42 pm" src="https://cloud.githubusercontent.com/assets/1874151/25103444/9053d9ec-238a-11e7-8723-09110a34926b.png">

jwforres:
can we give that last column a little more width so we can make the input
wider, that way we should get a little less wrapping.  but yes generally i
like this approach

benjaminapetersen:
+1 agree, def better than not reading at all. I think adjusting the column should be doable, just be sure to test w/a few roles on a service account as well as a user (service accounts get especially long).

rhamilto:
@sg00dwin, it looks like you've got a branch underway.  How about a PR?

PR #1585, opened 2017-05-24
Adjustment of membership ui-select role to allow help text to wrap an…

sg00dwin:
…d give select more height than default.

Fixes https://github.com/openshift/origin-web-console/issues/1218

@rhamilto


rhamilto:
[merge]

PR #1587, opened 2017-05-25
Bug Fix 1451013 - Show Events on PVC detail page

zherman0:
In order to explain to users why a PersistentVolumeClaim is not binding, I have added the events information section to the detail page for that specific PVC.

### DIFF
--- a/app/scripts/controllers/persistentVolumeClaim.js
+++ b/app/scripts/controllers/persistentVolumeClaim.js
@@ -41,6 +41,7 @@ angular.module('openshiftConsole')
     .get($routeParams.project)
     .then(_.spread(function(project, context) {
       $scope.project = project;
+      $scope.projectContext = context;
       DataService
         .get("persistentvolumeclaims", $routeParams.pvc, context)
         .then(function(pvc) {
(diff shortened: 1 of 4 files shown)


zherman0:
@jwforres @spadgett @erinboyd - Hey guys, here is the best solution I have for now for bug 1451013.  Let me know your thoughts or ideas.

zherman0:
When no events are present, the event grid will still be there:
pvc-events-noevents
With events, each event will be listed in the standard event grid:
pvc-events_withevents

jwforres:
Events should be on their own tab for consistency with all other detail pages.

jwforres:
The main tab should be "Details"

jwforres:
It is something that will require work at the API level, but @spadgett suggested adding Conditions to PVCs, they have been doing that for other resources that can get stuck in failure states.  Then you can know from looking at the object itself without having to look at events.

Adding the events tab is still a good thing to do, and is a good short term solution.

zherman0:
@jwforres @spadgett - Made the pvc details page use tabs like other places in the console.

spadgett:
[merge][severity: bug]

PR #1593, opened 2017-05-25
Removing form-horizontal so labels don't overlay inputs

rhamilto:
Depends on https://github.com/openshift/origin-web-catalog/pull/253, Fully fixes https://github.com/openshift/origin-web-catalog/issues/250


rhamilto:
[merge][severity: bug]

PR #1599, opened 2017-05-26
Check services for bindability before creating bindings

jeff-phillips-18:


app/views/overview/_list-row-actions.html
```diff
@@ -25,7 +25,7 @@
             </li>
             <!-- FIXME: Can't enable canI checks on svc cat resources until we have aggregation
             <li ng-if="(row.state.serviceInstances | hashSize) > 0 && {resource: 'bindings', group: 'servicecatalog.k8s.io'} | canI : 'create'" role="menuitem"> -->
-            <li ng-if="('pod_presets' | enableTechPreviewFeature) && (row.state.serviceInstances | hashSize) > 0" role="menuitem">
+            <li ng-if="('pod_presets' | enableTechPreviewFeature) && (row.state.bindableServiceInstances | hashSize) > 0" role="menuitem">
```
jwforres: bindableServiceInstances should be an array now, since you used _.filter, so you can just check row.state.bindableServiceInstances.length

app/scripts/directives/bindService.js
```diff
@@ -71,6 +71,7 @@ function BindService($scope,
                                         .concat(replicaSets)
                                         .concat(statefulSets);
       ctrl.applications = _.sortByAll(apiObjects, ['metadata.name', 'kind']);
```
benjaminapetersen: Curious about reason for making these bools into strings `"true" : "false"`?

benjaminapetersen: I see https://github.com/openshift/origin-web-common/blob/master/src/components/binding/bindServiceForm.html#L20, but not sure why a string here as well.

jeff-phillips-18: There are three options: 
- 'true' = bind to a selected application
- 'false' = bind, but not to an application
- 'none' = do not bind
-

app/views/overview/_list-row-actions.html
jeff-phillips-18: Updated, thanks.

app/scripts/directives/bindService.js
benjaminapetersen: Curious if that is something that can be clarified.  "false" meaning "yes, but no" and none meaning "false" (ie, what you thought false would have meant) seems confusing.

spadgett: Maybe something like

bindingType = 'application' | 'secret-only' | 'none'

to avoid confusion around types?

benjaminapetersen: right, something that clearly indicates intent.

jeff-phillips-18: OK, I can update https://github.com/openshift/origin-web-common/blob/master/src/components/binding/bindServiceForm.js to use these values.

jeff-phillips-18: Added https://github.com/openshift/origin-web-common/pull/90

jeff-phillips-18:
Removed WIP status, ready for final reviews.

app/scripts/controllers/overview.js
```diff
@@ -1132,9 +1134,16 @@ function OverviewController($scope,
   // extract & share
   var sortServiceInstances = function() {
     if(!state.serviceInstances && !state.serviceClasses) {
+      state.bindableServiceInstances = null;
       return;
     }
-    state.orderedServiceInstances = _.toArray(state.serviceInstances).sort(function(left, right) {
+
```
benjaminapetersen: Curious, rather than creating a duplicate array `serviceInstancesArray`, `_.filter` works fine on a hash as does `_.sortBy` instead of `[].sort`.  Possible to use the `_` version?

app/scripts/directives/bindService.js
```diff
@@ -168,6 +169,8 @@
         });
       }
       else {
+        ctrl.bindType = 'application';
+        ctrl.appToBind = ctrl.target;
```
benjaminapetersen: Lines `173` and `184` seem to be the same & I believe one can be removed. 

I'd actually prefer to see some of this logic broken out into separate functions.  Its a gnarly big if/else & there might be some long term benefits to breaking it down into smaller pieces.

app/scripts/directives/bindService.js
```diff
@@ -195,13 +198,13 @@
...
       var svcToBind = ctrl.target.kind === 'Instance' ? ctrl.target : ctrl.serviceInstances[ctrl.serviceToBind];
-      var appToBind = ctrl.target.kind !== 'Instance' ? ctrl.target : ctrl.appToBind;
+      var application = ctrl.bindType === 'application' ? _.get(ctrl.appToBind, 'metadata.name') : undefined;
 
       var context = {
         namespace: _.get(svcToBind, 'metadata.namespace')
       };
 
-      BindingService.bindService(context, _.get(svcToBind, 'metadata.name'), _.get(appToBind, 'metadata.name')).then(function(binding){
+      BindingService.bindService(context, _.get(svcToBind, 'metadata.name'), application).then(function(binding){
```
benjaminapetersen: So DataService patterns don't lead with `context`, they tend towards something like `(resource, name, context, opts)`.  We missed this & should prob open an issue in common to keep these consistent.  I'm curious if `application` even ends up as part of context in the end, but I'd have to look.

spadgett:
We should prioritize getting this in because binding is currently broken in the web console. We updated common without making the corresponding the corresponding console change for https://github.com/openshift/origin-web-common/pull/90

app/scripts/directives/bindService.js
jeff-phillips-18: Updated, thanks!

app/scripts/controllers/overview.js
jeff-phillips-18: Updated to use _.sortByAll. Thanks!

jeff-phillips-18:
@spadgett ah yes, sorry about that.

jwforres:
[merge]

PR #1613, opened 2017-05-30
[WIP] Removing prefixes from .flex*()[1], removing invalid .pad-auto-*()

rhamilto:
[1] flexbox no longer requires prefixes; we’re using autoprefixer anyhow

In lieu of https://github.com/openshift/origin-web-console/pull/1614 if for some reason we don't want to remove the flexbox mixin entirely.


rhamilto:
Closing in favor of #1614

PR #1614, opened 2017-05-30
Removing flexbox mixin, removing invalid .pad-auto-*()

rhamilto:
Flexbox no longer requires prefixes for the browsers we support, so this PR removes the flexbox mixin that adds browser-specific prefixes to flexbox properties.

Additionally, this PR removes .pad-auto-* from the spacers mixin as auto is not a valid value for padding.

cc: @benjaminapetersen


benjaminapetersen:
+1 and we use autoprefixer

sg00dwin:
Looks good

spadgett:
@rhamilto I assume you prefer this to #1613?

jwforres:
@rhamilto conflict on dist now, rebase / rebuild please

rhamilto:
> @rhamilto I assume you prefer this to #1613?

Yep.  Just put both up so we could choose.

> @rhamilto conflict on dist now, rebase / rebuild please

Done.

spadgett:
Thanks @rhamilto 

[merge]

PR #1615, opened 2017-05-30
Removing scroll-shadows-horizontal from quota table-responsives

### LINKED ISSUE #1610: White gradient at edges of table header on quota page
openshift web console 2017-05-30 14-54-17

rhamilto:
as scroll-shadows-horizontal is not compatible with table-responsives that have a header or colored background rows.

Fixes #1610

My bad for assuming and not testing!

### DIFF
--- a/app/views/quota.html
+++ b/app/views/quota.html
@@ -63,7 +63,7 @@
                     </div>
                   </div>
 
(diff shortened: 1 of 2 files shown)


spadgett:
[merge][severity: bug]

PR #1616, opened 2017-05-30
Increasing min-width for single row header

### LINKED ISSUE #1606: The `j` is clipped when "My Projects" wraps
The header didn't wrap before since it was just "Projects." Now that it's "My Projects," the header can wrap and the descender gets clipped.

screen shot 2017-05-26 at 3 15 13 pm

rhamilto:
now that the page title is longer

Fixes #1606

screen shot 2017-05-30 at 4 23 03 pm
screen shot 2017-05-30 at 4 23 16 pm

### DIFF
--- a/app/styles/_projects.less
+++ b/app/styles/_projects.less
@@ -74,7 +74,7 @@
 .projects-bar {
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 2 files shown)


spadgett:
[merge][severity: bug]

PR #1617, opened 2017-05-30
Fix inconsistent membership prompt messages

benjaminapetersen:
Refreshes the binding list on a fail (as was already done with a success) to ensure consistency.


jwforres:
[merge][severity:bug]

PR #1623, opened 2017-05-31
Handle editing routes that reference deleted services

### LINKED ISSUE #1273: Editing a route with a missing service preselects a different service
If you edit a route that points to a service that no longer exists, a different service is preselected in the dropdown. We should instead preserve the value and warn the user. Otherwise it's too easy to accidentally change the route to point to another service.

spadgett:
Fixes #1273

openshift web console 2017-05-31 12-50-57

### DIFF
--- a/app/scripts/controllers/edit/route.js
+++ b/app/scripts/controllers/edit/route.js
@@ -75,14 +75,40 @@ angular.module('openshiftConsole')
             };
 
             DataService.list("services", context).then(function(resp) {
+              $scope.loading = false;
+
               var servicesByName = resp.by("metadata.name");
               var to = _.get(route, 'spec.to', {});
-              $scope.loading = false;
-              $scope.services = orderByDisplayName(servicesByName);
+
+              // Make sure there is an option in the route editor for a service
+              // even if it doesn't exist. The editor will warn if the service
+              // is not found.
+              var ensureService = function(objectReference) {
+                if (servicesByName[objectReference.name]) {
+                  return;
+                }
+
+                // Make a copy to avoid mutating the DataService cache.
+                servicesByName = angular.copy(servicesByName);
+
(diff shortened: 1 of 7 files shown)


app/scripts/controllers/edit/route.js
```diff
@@ -75,26 +75,55 @@ angular.module('openshiftConsole')
...
+                  return;
+                }
+
+                // Make a copy to avoid mutating the DataService cache.
+                servicesByName = angular.copy(servicesByName);
+
+                // Add a dummy service since the `osc-routing` directive
+                // expects an object, not just a name. TODO: Update the
+                // osc-routing directive to take in a name rather than an
+                // object.
```
spadgett: @jwforres I might see how hard it is to change the directive rather than making a dummy object since I think that's cleaner, although it's a larger code change.

spadgett:
@jwforres updated to pass in the service name rather than an object. I've left the original commit as well in case we want it

app/views/directives/osc-routing-service.html
```diff
@@ -16,15 +16,21 @@
...
+          Service {{model.name}} does not exist.
+        </span>
+      </div>
+      <div ng-if="(optionNames | size) === 0" class="has-error">
         <span class="help-block">
           There are no <span ng-if="is-alternate">additional</span> services in your project to expose
           with a route.
         </span>
       </div>
+      <!-- FIXME: Not ever displayed! -->
```
jwforres: is there an issue tracking that we need to fix this?

spadgett: No I intended to fix this in this PR, but forgot. Let me look at it.

spadgett:
@jwforres Fixed the unnamed port warning.

openshift web console 2017-06-01 16-37-38

jwforres:
[merge]

Issue #1352, opened 2017-03-16
sweep for .length checks, migrate to _.size()


liggitt:
many API array fields have the potential to be either `null` or `[]`. for safety, we should consistently use _.size() or _.isEmpty() to handle both cases

liggitt:
in 1.6, when we switch to protobuf storage, optional array fields will start tending to be null instead of [] (the `items` field in List objects guarantees `[]`, and objects where `null` vs `[]` is significant will preserve the distinction, but others will prefer null)

spadgett:
@benjaminapetersen Is this complete?

benjaminapetersen:
@spadgett close, I've been through the console & common, I believe the catalog is only partially complete.

jwforres:
@benjaminapetersen think finishing the sweep of catalog is something you could wrap up today? at least identifying whats left and leaving some pointers in here if theres not enough time to get the PRs open

benjaminapetersen:
@jwforres @spadgett I combed through the 3 repos for length checks in JS files, I believe we are good there.  I have not checked over `/views` yet.  It might be a good idea to look through the `html` for `.length` checks that might need an updated to ` | size` filters.

spadgett:
> I have not checked over /views yet. It might be a good idea to look through the html for .length checks that might need an updated to | size filters.

From an API standpoint at least we should be OK. There's no runtime error in the view, and length checks will evaluate to false if the array is undefined.

benjaminapetersen:
Right, the views should handle themselves w/o error.  I think we can close this.

PR #1624, opened 2017-05-31
Update .length checks in controllers to use _.size()

benjaminapetersen:
re #1352 
Updating uses of `.length` in controllers to `_.size()` and/or `_.isEmpty()`. 

Opting for consistency by changing all the references (there are few), rather than being particular about only changing those associated w/an API object field.

### DIFF
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -207,7 +207,7 @@ angular.module('openshiftConsole')
 
         // Sort now to avoid sorting on every digest loop.
         $scope.orderedBuilds = BuildsService.sortBuilds($scope.builds, true);
-        $scope.latestBuild = $scope.orderedBuilds.length ? $scope.orderedBuilds[0] : null;
+        $scope.latestBuild = _.first($scope.orderedBuilds);
       },
       // params object for filtering
       {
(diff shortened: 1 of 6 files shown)


app/scripts/controllers/buildConfig.js
```diff
@@ -207,7 +211,7 @@ angular.module('openshiftConsole')
 
         // Sort now to avoid sorting on every digest loop.
         $scope.orderedBuilds = BuildsService.sortBuilds($scope.builds, true);
-        $scope.latestBuild = $scope.orderedBuilds.length ? $scope.orderedBuilds[0] : null;
+        $scope.latestBuild = latestBuild($scope.orderedBuilds);
```
jwforres: i think we would be fine just doing `_.first($scope.orderedBuilds)` here and dropping the added method.  we dont need it to be specifically set to null, just some falsey / non-existent value.

jwforres:
We have to sweep all of the javascript, this includes directives/components, filters, services -- and across all three repos.  Do we really want to change *every* reference to length across all of those files?

benjaminapetersen:
Yeah, not really.  Thats a lot of changes.  Starting with some small PRs to see how this goes as I fish around.

app/scripts/controllers/buildConfig.js
benjaminapetersen: cool, ill update

jwforres:
[merge]

jwforres:
[merge]

PR #1625, opened 2017-05-31
[WIP] Removing HTMLService and linkify filter as they're now in common

rhamilto:
Requires https://github.com/openshift/origin-web-common/pull/89 and a resulting common release.


rhamilto:
Included in #1658

PR #1628, opened 2017-05-31
Update .length check in directives/podDonut to use _.size()

benjaminapetersen:
re #1352 

So in the above issue:

>  the items field in List objects guarantees []

I'm not sure if `pod.spec.containers` = List object (seems plausible).  If so, we can skip this PR.


jwforres:
Pretty sure that today containers can't be empty, but its not something I trust to be true forever in the future, so I'm fine with going ahead and taking care of this one to be safe

[merge]

PR #1633, opened 2017-06-02
show toast notifications over modals

### LINKED ISSUE #1632: z-index of toast notifications is not high enough to overlap the overlay component
Toast notifications and the overlay component have the same z-index of 1050, we want toasts to pop up over top of the overlay.

z-index

jeff-phillips-18:
Fixes #1632

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -1235,3 +1235,8 @@ pre.clipped {
     width: 100%;
   }
 }
+
+// Override Patternfly z-index for toast notifications list to keep toasts above modals
+.toast-notifications-list-pf {
+  z-index: @zindex-modal + 5;
+}
--- a/dist/styles/main.css
+++ b/dist/styles/main.css
@@ -3867,7 +3867,7 @@ table.dataTable th:active{outline:0}
 .toast-pf-max-width{max-width:31.1%}
 .toast-pf-top-right{left:auto}
 }
-.toast-notifications-list-pf{margin-left:20px;position:fixed;right:20px;top:20px;z-index:1050}
+.toast-notifications-list-pf{margin-left:20px;position:fixed;right:20px;top:20px}
 .toast-notifications-list-pf .toast-pf{clear:both;float:right}
 @media (min-width:768px){.toast-notifications-list-pf{max-width:calc(50%)}
 }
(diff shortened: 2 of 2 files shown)


app/styles/_core.less
```diff
@@ -1235,3 +1235,8 @@ pre.clipped {
     width: 100%;
   }
 }
+
+// Override Patternfly z-index for toast notifications list to keep toasts above modals
+.toast-notifications-list-pf {
+  z-index: @zindex-modal + 5;
```
spadgett: Curious if `+ 5` is arbitrary or if we're using `+ 1`, `+ 2`, `+ 3` ... elsewhere?

spadgett:
Should this go in common?

jeff-phillips-18:
I used +5 as the halfway point between @zindex-modal and @zindex-popover

It could go in common.

spadgett:
> I used +5 as the halfway point between @zindex-modal and @zindex-popover

Makes sense. Do you mind adding that as a comment?

jeff-phillips-18:
Not at all.

spadgett:
@rhamilto @sg00dwin Should the toast be below popover? Trying to understand how that will interact with the dropdowns we have in our masthead.

jeff-phillips-18:
I think it should be, if I interact with something (causing the popover) I would expect it not to be cutoff by the toast.

jeff-phillips-18:
console includes patternfly less files after catalog and common, so any catalog or common overrides will not work. @rhamilto is this what we want or should we re-order?

spadgett:
> I think it should be, if I interact with something (causing the popover) I would expect it not to be cutoff by the toast.

Yeah I think this is probably right, but wanted to ask the question. It think it means the toast will slide in between the menu trigger and the menu itself, but I think that's OK?

We might check that that's really what happens since I know setting `position` on elements changes how `z-index` can behave.

rhamilto:
> console includes patternfly less files after catalog and common, so any catalog or common overrides will not work. @rhamilto is this what we want or should we re-order?

We should reorder if we need to override.  I initially put 'em in there in alpha order.

jeff-phillips-18:
The toast covers both the trigger and the menu. They are childed to the navbar which is specifically meant to be below the toast notifications.

PR #1636, opened 2017-06-02
Import catalog and common less files after patternfly to allow overri…

jeff-phillips-18:
…des.


jeff-phillips-18:
@rhamilto Good idea, added comments.

app/styles/main.less
```diff
@@ -3,16 +3,16 @@
...
-
 // PatternFly
 @import "@{pf-less-path}/patternfly.less";
 @import "@{pf-less-path}/patternfly-additions.less";
 
+// origin-web-catalog - after patternfly to gain any overrides
+@import "@{catalog-less-path}/catalog.less";
+
+// origin-web-common - after patternfly to gain any overrides
+@import "@{common-less-path}/common.less";
```
rhamilto: Thinking about this further, maybe catalog should come last so we could override common without the need to cut releases of both?

jeff-phillips-18: Yeah, makes sense. Done.

spadgett:
[merge]

PR #1638, opened 2017-06-02
Update .length check in directives/popups to use _.size()

benjaminapetersen:
re: 1352

`podWarningsFilter` can return null if there are no warnings, this could result in a runtime error in the `for` loop.


jwforres:
[merge]

PR #1641, opened 2017-06-02
Update .length check in services/ImagesService

benjaminapetersen:
Re: #1352  

@jwforres @spadgett

### DIFF
--- a/app/scripts/services/images.js
+++ b/app/scripts/services/images.js
@@ -197,7 +197,7 @@ angular.module("openshiftConsole")
       resources.push(deploymentConfig);
 
       var service;
-      if (config.ports.length) {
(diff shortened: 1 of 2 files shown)


app/scripts/services/images.js
```diff
@@ -229,7 +229,7 @@ angular.module("openshiftConsole")
                 var value = "";
                 if (ind > 0) {
                   key = entry.substring(0, ind);
-                  if (ind + 1 < entry.length) {
+                  if (ind + 1 < _.size(entry)) {
```
jwforres: this is a string not an array, i don't mind the change for the sake of getting these PRs in, but for future reference I don't think its really necessary to eliminate length checks on strings

jwforres:
[merge]

PR #1643, opened 2017-06-05
Don't modify cached template

spadgett:
Make a copy of the template in the process-template directive before
editing parameters. This fixes a problem in the new experience where you
open the same template more than once and the previous values you
entered are used.


jwforres:
[merge]

PR #1647, opened 2017-06-05
Update .length check in services/routes to use _.isEmpty

benjaminapetersen:
re #1352

### DIFF
--- a/app/scripts/services/routes.js
+++ b/app/scripts/services/routes.js
@@ -33,7 +33,7 @@ angular.module("openshiftConsole")
 
       var targetPort = route.spec.port ? route.spec.port.targetPort : null;
       if (!targetPort) {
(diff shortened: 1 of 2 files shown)


app/scripts/services/routes.js
```diff
@@ -33,7 +33,7 @@ angular.module("openshiftConsole")
 
       var targetPort = route.spec.port ? route.spec.port.targetPort : null;
       if (!targetPort) {
-        if (service.spec.ports.length > 1) {
+        if (!_.isEmpty(service.spec.ports)) {
```
jwforres: this change isnt right, its supposed to be checking if there are multiple ports

benjaminapetersen:
Ah, ok that makes sense, the warning clarifies as well. updated.

jwforres:
[merge]

PR #1654, opened 2017-06-06
Update overview to use toast notifications

spadgett:
https://trello.com/c/6WVJdgjL


spadgett:
[test]

app/scripts/controllers/overview.js
```diff
@@ -1087,16 +1089,36 @@ function OverviewController($scope,
...
+        } else {
+          buildName = build.metadata.name;
+        }
+
+        NotificationsService.addNotification({
+          type: "success",
+          message: _.capitalize(buildType) + " " + buildName + " successfully created.",
+          links: [{
+            href: buildURL,
+            label: 'View ' + _.capitalize(buildType)
```
jwforres: any reason to not have the View Log link here?

spadgett: The reason I linked to the build instead of the log is because the build might not start immediately. But I'm not against changing it.

spadgett:
rebased

jwforres:
[merge]

PR #1646, opened 2017-06-05
Switch deployment alerts to toast notifications

spadgett:
https://trello.com/c/6WVJdgjL

Updates the following to use toast notifications:

- deploy
- delete deployment
- edit deployment config
- edit resource limits
- edit health checks
- edit yaml (of replication controller)
- deployment rollback
- add/delete/edit env variables
- cancelled deployment
- delete horizontal pod autoscaler

@sg00dwin I've kept your original commit and added another with the changes from my review.

@jwforres PTAL, I've made enough changes to Steve's PR that I think it would be good for you to review.

Replaces https://github.com/openshift/origin-web-console/pull/1594

PR #1656, opened 2017-06-06
Prevent cancel from submitting edit dc form

spadgett:
The form element has an `ng-submit`, which causes the form to be submitted when the cancel button is clicked. Cancel was recently changed from a link to a button element in #1646, which caused the regression.


jwforres:
[merge]

PR #1658, opened 2017-06-06
Update origin-web-common to 0.0.31, origin-web-catalog to 0.0.19

jeff-phillips-18:


spadgett:
Do we need the changes from #1625 when we bump origin-web-common?

cc @rhamilto

spadgett:
I think we can remove `applicationHasDeployment` and `applicationHasDeploymentConfig` now as well.

jeff-phillips-18:
@spadgett PR #1599 will remove those filters.

jeff-phillips-18:
Updated dependencies in karam.conf.js to add new dependencies

jeff-phillips-18:
@spadgett should I include the changes from #1625 or close this and @rhamilto can bump in #1625?

rhamilto:
@jeff-phillips-18, in talking with @spadgett, it sounds like its best for you to cherry-pick #1625 and doing the update since this is already open.

spadgett:
> it sounds like its best for you to cherry-pick #1625 and doing the update since this is already open.

Works for me.

I think it's a runtime error if a filter or service is defined twice? Would be good for everything to go in as one PR.

spadgett:
[merge]

PR #1666, opened 2017-06-08
Don't mark YAML editor as immediately changed

spadgett:
This fixes two issues:

* Don't enable the save button before the user modifies the YAML. The intent in the code is for the button to only enable after the user types something.
* Don't prompt about unsaved changes when the user navigates away immediately.

### DIFF
--- a/app/scripts/controllers/edit/yaml.js
+++ b/app/scripts/controllers/edit/yaml.js
@@ -61,7 +61,12 @@ angular.module('openshiftConsole')
       var session = editor.getSession();
       session.setOption('tabSize', 2);
       session.setOption('useSoftTabs', true);
-      session.on('change', onChange);
+
+      // Wait for the editor to initialize before adding the on change handler
+      // so it's not immediately marked as modified.
+      setTimeout(function() {
+        session.on('change', onChange);
+      });
     };
 
     var watches = [];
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/edit/yaml.js
```diff
@@ -61,7 +61,12 @@ angular.module('openshiftConsole')
       var session = editor.getSession();
       session.setOption('tabSize', 2);
       session.setOption('useSoftTabs', true);
-      session.on('change', onChange);
+
+      // Wait for the editor to initialize before adding the on change handler
+      // so it's not immediately marked as modified.
+      setTimeout(function() {
```
jwforres: is this always going to work or are we creating a race condition where it will just work most of the time

spadgett: I can't find a cleaner way. It always works in my testing.

We could do something like this:

```js
    // onChange is called once when the initial content is loaded in the editor.
    var initialized = false;
    $scope.onChange = function() {
      if (initialized) {
        $scope.modified = true;
      }

      initialized = true;
    };
```

I worry that this will break if they ever change the behavior so that initialization doesn't trigger `onChange`, however.

Open to alternatives. The worst that happens is that the editor is incorrectly marked dirty so you're prompted when pressing the back button.

jwforres:
[merge]

PR #1672, opened 2017-06-08
Set up pod presets during binding to applications

jwforres:


benjaminapetersen:
[merge]

jwforres:
@rhamilto should be unblocked you and @benjaminapetersen will need to use the images pmorie/apiserver:pod-preset and pmorie/controller-manager:pod-preset for now for service catalog to test things out

also need to give svc cat service account cluster-admin instead of admin for the moment

```
oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:service-catalog:default
```

PR #1677, opened 2017-06-12
Update import YAML component to use toast notifications

spadgett:
https://trello.com/c/6WVJdgjL

### DIFF
--- a/app/scripts/directives/fromFile.js
+++ b/app/scripts/directives/fromFile.js
@@ -1,18 +1,18 @@
 'use strict';
 
 angular.module("openshiftConsole")
-  .directive("fromFile", function($q,
-                                  $uibModal,
+  .directive("fromFile", function($filter,
                                   $location,
-                                  $filter,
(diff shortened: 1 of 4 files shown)


spadgett:
@jwforres I know this will probably conflict with #1382. We can merge yours first if you like.

This is the last reference to `AlertsMessageService.addAlert`, so I can go through a remove references to `AlertMessageService.getAlerts` from the controllers when this goes in.

jwforres:
so this only takes care of the AlertsMessageService bits, not the alerts that are set on quota errors, was that intentional?

jwforres:
as far as #1382 I wanted the visual fix from @sg00dwin included before we merge it since the dropdown is always there, not just in the new experience

spadgett:
Updated to handle quota errors as toasts:

screen shot 2017-06-12 at 12 01 50 pm

jwforres:
[merge]

jwforres:
flake [merge]

PR #1678, opened 2017-06-12
Show process template errors as toasts

spadgett:
https://trello.com/c/6WVJdgjL

### DIFF
--- a/app/scripts/directives/processTemplate.js
+++ b/app/scripts/directives/processTemplate.js
@@ -8,6 +8,7 @@ angular.module('openshiftConsole').component('processTemplate', {
     '$uibModal',
     'DataService',
     'Navigate',
+    'NotificationsService',
     'ProcessedTemplateService',
     'QuotaService',
     'SecurityCheckService',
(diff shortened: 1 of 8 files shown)


app/scripts/directives/processTemplate.js
```diff
@@ -152,16 +153,32 @@ function ProcessTemplate($filter,
     modalInstance.result.then(createResources);
   };
 
+  var alerts = {};
+  var hideNotificationErrors = function() {
+    NotificationsService.hideNotification("process-template-error");
+    _.each(alerts, function(alert) {
+      if (alert.id) {
```
jwforres: this assumes that we will only ever put id's on the error alerts, should this also check that the alert is a warning or error alert

spadgett:
@jwforres Thanks, updated.

app/scripts/directives/processTemplateDialog.js
```diff
@@ -80,6 +79,8 @@
     };
 
     ctrl.close = function() {
+      $scope.$broadcast('hideTemplateNotificationErrors');
```
jwforres: assuming i'll need to do the same thing from my overlay PR then

jwforres:
[merge]

app/scripts/directives/processTemplateDialog.js
spadgett: Yeah, and something similar is needed for `from-file` when #1677 goes in.

PR #1681, opened 2017-06-12
Show toast notifications in create from builder flow

spadgett:
https://trello.com/c/6WVJdgjL

### DIFF
--- a/app/scripts/controllers/create/createFromImage.js
+++ b/app/scripts/controllers/create/createFromImage.js
@@ -1,29 +1,30 @@
 "use strict";
 
 angular.module("openshiftConsole")
-  .controller("CreateFromImageController", function ($scope,
-      Logger,
-      $q,
-      $routeParams,
-      APIService,
-      DataService,
-      ProjectsService,
-      Navigate,
-      ApplicationGenerator,
-      LimitRangesService,
-      MetricsService,
-      HPAService,
-      QuotaService,
-      SecretsService,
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/create/createFromImage.js
```diff
@@ -1,29 +1,30 @@
...
+                       APIService,
+                       ApplicationGenerator,
+                       DataService,
+                       HPAService,
+                       ImagesService,
+                       LimitRangesService,
+                       Logger,
+                       MetricsService,
+                       Navigate,
+                       NotificationsService,
```
spadgett: Only change to dependencies is `NotificationsService`. But I sorted and reformatted the list.

app/scripts/controllers/create/createFromImage.js
```diff
@@ -313,16 +325,7 @@ angular.module("openshiftConsole")
                   }
                 );
                 return d.promise;
-              },
-              function(result) { // failure
```
jwforres: assume DataService.batch was never resulting in a rejected promise?

spadgett: We were passing an extra argument to `TaskList.add` that it doesn't take :/

jwforres: ah yep didnt have my curly braces matched up right, got it.

jwforres:
[merge]

PR #1683, opened 2017-06-12
Bug Fix 1460167 - View Quota does not work on Overview page

zherman0:
The project name variable was not being passed in the Navigate.quotaURL call which made the "View Quota" link have an undefined link.  Added parameter to the call. @jwforres @spadgett


jwforres:
[merge]

zherman0:
This is for bug 1460167, the title was a typo.

PR #1690, opened 2017-06-12
Consistently format server error messages

spadgett:
Consistently format server errors in toast notifications across pages. Prefix the messages with `"Reason: "` in the filter, which tends to work well with most messages in the server. Remove the `capitalize` property, which is not always correct for some messages and is only used in one place.

One example:

screen shot 2017-06-12 at 6 56 44 pm


jwforres:
[merge]

PR #1705, opened 2017-06-14
Add delete binding action to provisioned services

benjaminapetersen:
https://trello.com/c/JMKxStDt

- This is my `unbind-service` branch w/spadgett's `Update delete binding from service-instance-row to use app names` cherry-pick.  
- Rebased to eliminate the cherry-pick commit

I may close the previous `unbind-service` PR & keep this on to avoid the need to fiddle around with merge conflicts again.

Prev #1505

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -337,6 +337,7 @@
         <script src="scripts/directives/buildStatus.js"></script>
         <script src="scripts/directives/routeServiceBarChart.js"></script>
         <script src="scripts/directives/bindService.js"></script>
+        <script src="scripts/directives/unbindService.js"></script>
         <script src="scripts/directives/processTemplate.js"></script>
         <script src="scripts/directives/processTemplateDialog.js"></script>
         <script src="scripts/directives/deployImageDialog.js"></script>
--- a/app/scripts/directives/bindService.js
+++ b/app/scripts/directives/bindService.js
@@ -55,15 +55,15 @@
       }
     };
 
-
     var deploymentConfigs, deployments, replicationControllers, replicaSets, statefulSets;
     var sortApplications = function() {
       // Don't waste time sorting on each data load, just sort when we have them all
       if (deploymentConfigs && deployments && replicationControllers && replicaSets && statefulSets) {
-        var apiObjects = deploymentConfigs.concat(deployments)
-                                          .concat(replicationControllers)
-                                          .concat(replicaSets)
-                                          .concat(statefulSets);
+        var apiObjects =  [].concat(deploymentConfigs)
+                            .concat(deployments)
+                            .concat(replicationControllers)
+                            .concat(replicaSets)
+                            .concat(statefulSets);
         ctrl.applications = _.sortByAll(apiObjects, ['metadata.name', 'kind']);
         ctrl.bindType = ctrl.applications.length ? "application" : "secret-only";
       }
--- a/app/scripts/directives/unbindService.js
+++ b/app/scripts/directives/unbindService.js
@@ -1,0 +1,106 @@
+'use strict';
+
+(function() {
+
+  angular.module('openshiftConsole').component('unbindService', {
+    controller: [
+      '$scope',
+      '$filter',
+      'DataService',
+      UnbindService
+    ],
+    controllerAs: 'ctrl',
+    bindings: {
+      target: '<',
+      bindings: '<',
+      applicationsByBinding: '<',
+      onClose: '<'
+    },
+    templateUrl: 'views/directives/unbind-service.html'
+  });
+
+  function UnbindService($scope, $filter, DataService) {
+    var ctrl = this;
+    var validityWatcher;
+    var context;
+    var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
+
+    var unbindService = function() {
+      DataService.delete({
+        group: 'servicecatalog.k8s.io',
+        resource: 'bindings'
+      }, ctrl.selectedBinding, context).then(_.noop, function(err) {
+        ctrl.error = err;
+      });
+    };
+
+    var setupValidator = function() {
+      var firstStep = _.first(ctrl.steps);
+      firstStep.valid = false;
+      // TODO: auto-select if one option is kludgy. will follow-on a fix
+      // if(_.size(ctrl.bindings) === 1) {
+      //   firstStep.valid = true;
+      //   ctrl.selectedBinding = _.first(ctrl.bindings);
+      // } else {
+        validityWatcher = $scope.$watch("ctrl.selectedBinding", function(selectedBinding) {
+          firstStep.valid = !!selectedBinding;
+        });
+      // }
+    };
+
+    var tearDownValidator = function() {
+      if(validityWatcher) {
+        validityWatcher();
+        validityWatcher = undefined;
+      }
+    };
+
+    var showDeleteForm = function() {
+      ctrl.nextTitle = 'Delete';
+      setupValidator();
+    };
+
+    var showResults = function() {
+      ctrl.nextTitle = 'Close';
+      ctrl.wizardComplete = true;
+      unbindService();
+      tearDownValidator();
+    };
+
+    ctrl.$onInit = function() {
+      var formStepLabel = (ctrl.target.kind === 'Instance') ? 'Applications' : 'Services';
+      ctrl.displayName = serviceInstanceDisplayName(ctrl.target);
+      ctrl.steps = [{
+        id: 'deleteForm',
+        label: formStepLabel,
+        view: 'views/directives/bind-service/delete-binding-select-form.html',
+        onShow: showDeleteForm
+      }, {
+        id: 'results',
+        label: 'Results',
+        view: 'views/directives/bind-service/delete-binding-result.html',
+        onShow: showResults
+      }];
+
+      context = {
+        namespace: _.get(ctrl.target, 'metadata.namespace')
+      };
+    };
+
+    ctrl.appsForBinding = function(bindingName) {
+      return _.get(ctrl.applicationsByBinding, bindingName);
+    };
+
+    ctrl.closeWizard = function() {
+      if (_.isFunction(ctrl.onClose)) {
+        ctrl.onClose();
+      }
+    };
+
+    ctrl.$onDestroy = function() {
+      tearDownValidator();
+    };
+
+  }
+
+})();
--- a/app/views/directives/bind-service/delete-binding-result.html
+++ b/app/views/directives/bind-service/delete-binding-result.html
@@ -1,0 +1,32 @@
+<div>
+  <div ng-if="!ctrl.error">
+    <h3 class="mar-top-none">
+      Binding for the following has been deleted:
+    </h3>
+
+    <div
+      ng-if="ctrl.appsForBinding(ctrl.selectedBinding)"
+      ng-repeat="appForBinding in ctrl.appsForBinding(ctrl.selectedBinding)">
+      {{appForBinding.metadata.name}} <small class="text-muted">&ndash; {{ appForBinding.kind | humanizeKind : true}}</small>
+    </div>
+    <div ng-if="!(ctrl.appsForBinding(ctrl.selectedBinding))">
+      <!-- TODO: selected a secret name in the form, but see a binding name in the result. -->
+      {{ctrl.selectedBinding}} <small class="text-muted">&ndash; Binding</small>
+    </div>
+
+    <p class="mar-top-lg">
+      <span class="pficon pficon-info"></span> You will need to redeploy your pods for this to take effect.
+    </p>
+  </div>
+  <div ng-if="ctrl.error">
+    <div class="title">Deletion of Binding Failed <span class="fa fa-times text-danger"></span></div>
+    <div class="sub-title">
+      <span ng-if="ctrl.error.data.message">
+        {{ctrl.error.data.message | upperFirst}}
+      </span>
+      <span ng-if="!ctrl.error.data.message">
+        An error occurred deleting the binding.
+      </span>
+    </div>
+  </div>
+</div>
--- a/app/views/overview/_list-row-actions.html
+++ b/app/views/overview/_list-row-actions.html
@@ -25,8 +25,11 @@
             </li>
             <!-- FIXME: Can't enable canI checks on svc cat resources until we have aggregation
             <li ng-if="(row.state.serviceInstances | hashSize) > 0 && {resource: 'bindings', group: 'servicecatalog.k8s.io'} | canI : 'create'" role="menuitem"> -->
-            <li ng-if="('pod_presets' | enableTechPreviewFeature) && row.state.bindableServiceInstances.length" role="menuitem">
+            <li ng-if-start="('pod_presets' | enableTechPreviewFeature) && row.state.bindableServiceInstances.length" role="menuitem">
               <a href="" ng-click="row.showOverlayPanel('bindService', {target: row.apiObject})">Create Binding</a>
+            </li>
+            <li ng-if-end role="menuitem">
+              <a href="" ng-click="row.showOverlayPanel('unbindService', {target: row.apiObject})">Delete Binding</a>
             </li>
             <li ng-if="row.current && ('deploymentconfigs/log' | canI : 'get')" role="menuitem">
               <a ng-href="{{row.current | navigateResourceURL}}?tab=logs">View Logs</a>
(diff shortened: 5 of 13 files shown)


benjaminapetersen:
From a service instance, bindings to apps list `kind` & `app name`, otherwise if a secret lists the binding name:

screen shot 2017-06-13 at 11 50 00 pm

Bindings from apps list the binding name:
screen shot 2017-06-13 at 11 51 29 pm

benjaminapetersen:
@jwforres @spadgett pretty sure I'll close the other one & squash this, just thinking on it a bit.

app/scripts/services/serviceInstances.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .factory('ServiceInstancesService', function() {
+    return {
+      // NOTE: should this be a filter?  or will shared functionality grow here?
```
spadgett: I think this should be a filter. Maybe more than one: `serviceClassDisplayName` and `serviceInstanceDisplayName`?

app/scripts/services/bindingModalUtils.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .factory('BindingModalUtils', [
```
spadgett: Instead of making it specific to the binding modal, this could be a general utility that combines and (optionally?) sorts collections of API objects.

We have this need for config maps and secrets in some places where we combine them in a select.

app/scripts/services/bindingModalUtils.js
```diff
@@ -0,0 +1,24 @@
...
+                                    deployments,
+                                    replicationControllers,
+                                    replicaSets,
+                                    statefulSets) {
+          // Don't waste time sorting on each data load, just sort when we have them all
+          if (deploymentConfigs && deployments && replicationControllers && replicaSets && statefulSets) {
+            var apiObjects = deploymentConfigs.concat(deployments)
+                                              .concat(replicationControllers)
+                                              .concat(replicaSets)
+                                              .concat(statefulSets);
```
spadgett: Probably want to check if these are arrays before calling `concat`. Or at least make a comment that the inputs must be arrays.

app/scripts/directives/bindService.js
```diff
@@ -55,20 +59,6 @@
...
-    var deploymentConfigs, deployments, replicationControllers, replicaSets, statefulSets;
-    var sortApplications = function() {
-      // Don't waste time sorting on each data load, just sort when we have them all
-      if (deploymentConfigs && deployments && replicationControllers && replicaSets && statefulSets) {
-        var apiObjects = deploymentConfigs.concat(deployments)
-                                          .concat(replicationControllers)
-                                          .concat(replicaSets)
-                                          .concat(statefulSets);
-        ctrl.applications = _.sortByAll(apiObjects, ['metadata.name', 'kind']);
-        ctrl.bindType = ctrl.applications.length ? "application" : "secret-only";
```
spadgett: You are no longer setting `ctrl.bindType`

app/scripts/directives/bindService.js
```diff
@@ -55,20 +59,6 @@
       }
     };
 
-
-    var deploymentConfigs, deployments, replicationControllers, replicaSets, statefulSets;
-    var sortApplications = function() {
```
spadgett: Suggest leaving `sortApplications` as a function then calling your new service function from here so you don't have to repeat the parameters in every list callback.

```js
var sortApplication = function() {
  ctrl.applications = BindingModalUtils.sortApplications(deploymentConfigs, deployments, replicationControllers, replicaSets, statefulSets);
  ctrl.bindType = _.isEmpty(ctrl.applications) ? "secret-only" : "application";
};
```

app/scripts/directives/unbindService.js
```diff
@@ -0,0 +1,121 @@
...
+  function UnbindService($filter, BindingModalUtils, DataService, ServiceInstancesService) {
+    var ctrl = this;
+    var context;
+    var getDisplayName = ServiceInstancesService.getDisplayName;
+    // var sortApplications = BindingModalUtils.sortApplications;
+    // TODO: once we have podPreset
+    // var deploymentConfigs, deployments, replicationControllers, replicaSets, statefulSets;
+
+
+    // TODO: once we have podPreset we will be able to associate the
```
spadgett: Remove the TODO comments?

app/scripts/directives/unbindService.js
```diff
@@ -0,0 +1,121 @@
...
+      'BindingModalUtils',
+      'DataService',
+      'ServiceInstancesService',
+      UnbindService
+    ],
+    controllerAs: 'ctrl',
+    bindings: {
+      target: '<',
+      bindings: '<',
+      applications: '<',
```
spadgett: `bindings` and `applications` are maps. Looking at just the names, I thought these were flat lists. Can we use similar names to what we use on the overview so it's clear they are the same maps?

app/scripts/directives/unbindService.js
```diff
@@ -0,0 +1,121 @@
...
+        onShow: showConfirm
+      }];
+
+      context = {
+        namespace: _.get(ctrl.target, 'metadata.namespace')
+      };
+    };
+
+    ctrl.appNameForBinding = function(bindingName) {
+      return _.get(_.first(_.get(ctrl.applications, bindingName)), ['metadata', 'name']) || bindingName;
```
spadgett: I find the nested lodash calls on one line here and just below hard to follow.

spadgett: It looks like we're showing just the first app name? Usually there is only one app, but if unbind is going to affect multiple apps, it's important to know.

app/views/overview/_list-row-actions.html
```diff
@@ -28,6 +28,9 @@
             <li ng-if="('pod_presets' | enableTechPreviewFeature) && row.state.bindableServiceInstances.length" role="menuitem">
               <a href="" ng-click="row.showOverlayPanel('bindService', {target: row.apiObject})">Create Binding</a>
             </li>
+            <li ng-if="('pod_presets' | enableTechPreviewFeature) && row.state.bindableServiceInstances.length" role="menuitem">
```
spadgett: Maybe use `ng-if-start` on the `li` above and `ng-if-end` since the conditions are the same?

app/scripts/services/bindingModalUtils.js
spadgett: Is this only used in the bind modal and no longer needed for unbind? I suggest for now we leave the working bind code alone if we're not reusing it for unbind.

benjaminapetersen:
@spadgett yup! bring on the reviews.  Not exactly ready like I'd like to be, but time is of the essence.  I'll hit these items while I work on cleanup.

app/scripts/services/serviceInstances.js
benjaminapetersen: Agree.  This will become `serviceInstanceDisplayName`.  

For a `serviceClassDisplayName` filter, are you thinking we need it because the `externalMetadata` field is a different than the typical `metadata.name`?

benjaminapetersen: Moved the function to a filter, left a comment for a `TODO` on `serviceClassDisplayName` per my previous comment.

app/scripts/services/bindingModalUtils.js
benjaminapetersen: Yeah, probably a premature refactor on my part, assuming I would need it for `unbind`. I  will return this to its previous state.

app/scripts/services/bindingModalUtils.js
benjaminapetersen: agree, I'll start with `[].concat(deploymentConfigs)` to ensure we don't error here.

app/scripts/directives/bindService.js
benjaminapetersen: I moved it back to `bindService` as it originally was. Def think this is a good call if we end up sharing it in the future.

app/scripts/directives/bindService.js
benjaminapetersen: fixed

app/scripts/directives/unbindService.js
benjaminapetersen: `bindingsByInstanceRef` && `applicationsByBinding`, will do.

app/scripts/services/bindingModalUtils.js
spadgett: I think it's more than this... If you pass an object as an argument to concat, does it just append the object?

Maybe a moot point if you're rolling back the retractor though

benjaminapetersen: Yeah... everything below has a `_.toArray` wrap (original code before my change).  If we do move it out in future, agree, perhaps the `_.toArray` should come with it & happen internally.

app/scripts/directives/unbindService.js
benjaminapetersen: Eliminated this whole section, don't think any of it will ever be needed w/the maps from overview.

app/scripts/directives/unbindService.js
benjaminapetersen: Updating to this:

screen shot 2017-06-14 at 1 19 32 pm

@ncameronbritt for input if this deviates or is new.  In theory, if a binding represents multiple apps, this is a bit wasteful of space as they could be listed horizontally, but I don't know of another pattern off the top of my head so going with consistency for now.  We can iterate if needed.

app/views/overview/_list-row-actions.html
benjaminapetersen: will do

spadgett:
For the radio buttons, wonder if it should be 

```html
{{name}} <small class="text-muted">&ndash; -- {{kind}}</small>`
```

like in the bind application dropdown?

openshift web console 2017-06-14 13-41-19

benjaminapetersen:
@spadgett @jwforres finished round 1 of requested changes. I'm looking it over again myself, but feel free to give it another pass.  Def prefer a goal of pre 3pm to get it in the queue :)

benjaminapetersen:
@spadgett yeah good Q.  We have all of these:
screen shot 2017-06-14 at 1 45 39 pm
screen shot 2017-06-14 at 1 45 43 pm
screen shot 2017-06-14 at 1 45 47 pm

I like your suggestion, I think it would improve density / clarity.

app/views/directives/bind-service/remove-binding-confirm.html
```diff
@@ -0,0 +1,21 @@
+<div>
+  <div ng-if="!ctrl.error">
+    <h3 class="mar-top-none">
+      Binding for {{ctrl.appKindForBinding(ctrl.selectedBinding) | humanizeKind : true}} <strong>{{ctrl.appNameForBinding(ctrl.selectedBinding)}}</strong> has been deleted.
```
spadgett: Looks like you're still calling a method you removed, `appNameForBinding`?

spadgett:
On the last binding dialog, we show the binding name, but we don't show the name anywhere on the overview currently, so I'm not going to know which to delete.

app/views/directives/bind-service/remove-binding-confirm.html
benjaminapetersen: Fixing that now, working on listing here as well.

spadgett:
I think the first wizard step label should either be "Services" or "Applications" depending on what we're showing to be consistent with the Create Binding dialog.

benjaminapetersen:
Density here definitely improves clarity:

screen shot 2017-06-14 at 2 04 41 pm

app/views/directives/bind-service/select-binding.html
```diff
@@ -0,0 +1,26 @@
...
+            <div class="component-label">
+              {{ (appForBinding.kind) | humanizeKind : true}}
+            </div>
+            <p>{{appForBinding.metadata.name || binding.metadata.name}}</p>
+          </div>
+          <div ng-if="!(ctrl.appsForBinding(binding))">
+            <div class="component-label">
+              Secret
+            </div>
+            <p>{{appForBinding.metadata.name || binding.metadata.name}}</p>
```
spadgett: Should this be...

```html
<p>{{binding.spec.secretName}}</p>
```

spadgett:
Need to disable the Delete button when nothing is selected or preselect the first item. I think preselecting the first is OK? @ncameronbritt @jwforres 

openshift web console 2017-06-14 14-17-05

ncameronbritt:
I think preselecting if there is only one makes sense. I think if there is more than one we wouldn't preselect, and the "Delete" button would be disabled. I believe that is consistent with the Bind flow.

benjaminapetersen:
Updating confirmation to list apps & match first step in panel:

screen shot 2017-06-14 at 2 22 32 pm

Thoughts?  (I see I have more comments to loop back on, have not addressed yet)

app/views/directives/bind-service/select-binding.html
benjaminapetersen: Yup will do

benjaminapetersen:
@spadgett @ncameronbritt  if first should be "Applications" or "Services", should last step also be "Results"?  The word "Confirmation" implies to me I would have one more changes to back out of the action.

spadgett:
Yes, should be "Results". I believe that's what we use everywhere else.

benjaminapetersen:
@spadgett @ncameronbritt working on the pre-select / disable now.  I think that requires this `validityWatcher` like bind service.  FWIW, I'd like to look at updating `pf-wizard` to support this is a more standard way, its a little kludgy, may open an issue since validation is pretty common.

app/index.html
```diff
@@ -337,6 +337,7 @@ <h2 class="text-center" id="temporary-loading-message" style="display: none;">Lo
         <script src="scripts/directives/buildStatus.js"></script>
         <script src="scripts/directives/routeServiceBarChart.js"></script>
         <script src="scripts/directives/bindService.js"></script>
+        <script src="scripts/directives/unbindService.js"></script>
```
benjaminapetersen: I prob should rename this file since we no longer call it "unbind"

app/views/overview/_list-row.html
```diff
@@ -18,4 +18,10 @@
     <bind-service target="row.overlay.state.target"
                   on-close="row.closeOverlayPanel"></bind-service>
   </div>
+  <div ng-if="row.overlay.panelName === 'unbindService'">
+    <unbind-service
+      target="row.overlay.state.target"
+      bindings-by-instance="row.bindings"
```
spadgett: Ah, so it's not a map. I'd call the attribute bindings if it's just the array of bindings. Sorry if I misunderstood.

app/views/directives/bind-service/delete-binding-result.html
```diff
@@ -0,0 +1,31 @@
...
+      Binding for the following has been deleted:
+    </h3>
+
+    <div
+      ng-if="ctrl.appsForBinding(ctrl.selectedBinding)"
+      ng-repeat="appForBinding in ctrl.appsForBinding(ctrl.selectedBinding)">
+      {{appForBinding.metadata.name}} <small class="text-muted">&ndash; {{ appForBinding.kind | humanizeKind : true}}</small>
+    </div>
+    <div ng-if="!(ctrl.appsForBinding(ctrl.selectedBinding))">
+      {{ctrl.selectedBinding}} <small class="text-muted">&ndash; Secret</small>
```
spadgett: This is always the binding name here and not a secret correct?

spadgett: I think you can put `Binding` here for now and we can circle back. I have some concerns that we don't show the binding name anywhere in the UI, but we can fix as a follow on to get the basic functionality in.

benjaminapetersen: I'll swap this & leave a `TODO:` because its weird to select "secret" then see "binding" deleted, agree. :/

app/views/overview/_list-row.html
benjaminapetersen: Ok, here will be: 
```html
  <unbind-service
      target="row.overlay.state.target"
      binding="row.bindings"
      on-close="row.closeOverlayPanel"></unbind-service>
```
and in the service instance row template:
```html
    <unbind-service
      target="row.overlay.state.target"
      bindings="row.bindings"
      applications-by-binding="row.state.applicationsByBinding"
      on-close="row.closeOverlayPanel"></unbind-service>
```
So we will have `bindings` and `applicationsByBinding`.

spadgett:
Thanks for the fast turnaround @benjaminapetersen 

[merge]

benjaminapetersen:
woohoo! :yum:

PR #1706, opened 2017-06-14
Hide error notifications on $scope destroy, not just cancel

spadgett:
Hide error notifications for editors on $scope destroy rather than only
in the cancel click handler. This will hide the notifications regardless
of how the user navigates away (back button, etc.).


spadgett:
@jwforres I only updated the full page editors, not components in dialogs, to be safe.

jwforres:
[merge]

jwforres:
@spadgett looks like a dist failure

PR #1719, opened 2017-06-15
Bug 1461346 - Filter events by UID

spadgett:
Use the API object UID rather than kind and name to filter events. This
prevents events for showing for a previous object with the same name.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1461346

### DIFF
--- a/app/scripts/controllers/build.js
+++ b/app/scripts/controllers/build.js
@@ -58,6 +58,8 @@ angular.module('openshiftConsole')
       title: $routeParams.build
     });
 
+    var buildPod;
+    var annotation = $filter('annotation');
     var watches = [];
 
     var setLogVars = function(build) {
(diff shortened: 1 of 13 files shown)


spadgett:
Fixed the build page to show events for both builds and the build pod. @bparees FYI

openshift web console 2017-06-15 12-11-38

app/scripts/controllers/build.js
```diff
@@ -72,45 +74,6 @@ angular.module('openshiftConsole')
       }
     };
 
-    var buildResolved = function(build, action) {
```
spadgett: @jwforres had to move these functions inside the ProjectsService callback

app/scripts/controllers/build.js
```diff
@@ -122,6 +85,66 @@ angular.module('openshiftConsole')
...
+            $scope.breadcrumbs[2].title = "#" + buildNumber;
+          }
+          if (action === "DELETED") {
+            $scope.alerts["deleted"] = {
+              type: "warning",
+              message: "This build has been deleted."
+            };
+          }
+
+          var buildPodName;
```
spadgett: My changes to the function are from here to line 123.

app/scripts/directives/events.js
```diff
@@ -80,64 +78,90 @@ angular.module('openshiftConsole')
...
+            // Assign `$scope.events` here instead of the watch callback to
+            // avoid the "Filter hiding all events" message from flickering
+            // briefly before the debounced filter and update calls.
+            $scope.events = filterForResource(allEvents);
+            update();
+          });
         }, 250, {
           leading: true,
-          trailing: false,
+          trailing: true,
```
spadgett: Not sure why trailing was false. I think it should be true.

https://lodash.com/docs/3.10.1#debounce

> Note: If leading and trailing options are true, func is invoked on the trailing edge of the timeout only if the the debounced function is invoked more than once during the wait timeout.

bparees:
cool!

app/scripts/controllers/build.js
```diff
@@ -122,6 +85,66 @@ angular.module('openshiftConsole')
...
+          if (!buildPod) {
+            buildPodName = annotation(build, 'buildPod');
+            if (buildPodName) {
+              DataService.get("pods", buildPodName, context).then(function(response) {
```
jwforres: do we want to handle errors for this get or let data service show the toast?

jwforres:
[merge][severity:blocker]

spadgett:
[merge][severity:blocker]

PR #1720, opened 2017-06-15
Bug 1461702 - Fix display of build webhooks to support new types

jwforres:
Also fixes outdated documentation links.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1461702


jwforres:
The build-config.html diff looks big because i fixed an indention issue with that section of the template

spadgett:
We had a lot of broken links :(

spadgett:
[merge][severity: blocker]

jwforres:
hmm, travis is fine with the dist but jenkins failed on main.css, it looks right to me though...

lemme kick it again and see what happens

[merge][severity: blocker]

jwforres:
@spadgett concerned that this keeps getting a CSS dist in Jenkins, and only Jenkins...

spadgett:
We might start printing the content of the diff instead of just the files so we know what's really different.

jwforres:
added a commit to switch to use git diff
[test]

Issue #1703, opened 2017-06-14
Enter key no longer submits image search in deploy-image component


spadgett:
Enter used to trigger a search. Now you need to click the magnifying glass.

jwforres:
looks like form and ng-form behave differently when it comes to enter key submission :-\

I know there was a reason we switched it to ng-form in my PR but I can't remember why now

jwforres:
It's not very clean but I can work around it by adding a `ng-keyup="$event.keyCode === 13 && findImage()` on the text input

jwforres:
apparently ng-form is not supposed to be used in place of form but nested underneath form, see https://github.com/angular/angular.js/issues/2513#issuecomment-39399816

spadgett:
We were using it because there is an outer form in the dialog and you can't nest form elements. Enter seems to work OK in the dialog. Maybe we need an outer form for the non-dialog page.

I believe we have the same problem with `process-template` as well.

PR #1722, opened 2017-06-15
Enter key on forms for deploy image and process template

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/1703


spadgett:
LGTM but probably worth a comment saying that `<form>` is needed so the enter key works. Worried someone will remove it.

jwforres:
@spadgett added the comment to both templates

spadgett:
[merge][severity: bug]

PR #1726, opened 2017-06-16
Bug 1460142 - Use $routeChangeStart for confirm-on-exit

spadgett:
Listen for `$routeChangeStart` instead of `$locationChangeStart` in the
`confirm-on-exit` directive. This prevents us from incorrectly prompting
when the user switches tab with `persist-tab-state` on.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1460142


jwforres:
[merge][severity: blocker]

PR #1728, opened 2017-06-16
Fix delete binding modal to show secret names

benjaminapetersen:
Now the results panel will show the secret instead of the binding.  

Aside, I am working on ordering the applications select form, probably ordering by the first app in the binding's associated app list.

screen shot 2017-06-16 at 10 12 12 am
screen shot 2017-06-16 at 10 12 18 am

### DIFF
--- a/app/views/directives/bind-service/delete-binding-result.html
+++ b/app/views/directives/bind-service/delete-binding-result.html
@@ -5,13 +5,12 @@
     </h3>
 
     <div
-      ng-if="ctrl.appsForBinding(ctrl.selectedBinding)"
-      ng-repeat="appForBinding in ctrl.appsForBinding(ctrl.selectedBinding)">
+      ng-if="ctrl.appsForBinding(ctrl.selectedBinding.metadata.name) | size"
+      ng-repeat="appForBinding in ctrl.appsForBinding(ctrl.selectedBinding.metadata.name)">
       {{appForBinding.metadata.name}} <small class="text-muted">&ndash; {{ appForBinding.kind | humanizeKind : true}}</small>
     </div>
-    <div ng-if="!(ctrl.appsForBinding(ctrl.selectedBinding))">
-      <!-- TODO: selected a secret name in the form, but see a binding name in the result. -->
-      {{ctrl.selectedBinding}} <small class="text-muted">&ndash; Binding</small>
+    <div ng-if="!(ctrl.appsForBinding(ctrl.selectedBinding.metadata.name)  | size)">
+      {{ctrl.selectedBinding.spec.secretName}} <small class="text-muted">&ndash; Secret</small>
     </div>
 
     <p class="mar-top-lg">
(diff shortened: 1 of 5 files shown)


app/views/directives/bind-service/delete-binding-result.html
```diff
@@ -5,13 +5,12 @@ <h3 class="mar-top-none">
...
-      ng-if="ctrl.appsForBinding(ctrl.selectedBinding)"
-      ng-repeat="appForBinding in ctrl.appsForBinding(ctrl.selectedBinding)">
+      ng-if="ctrl.appsForBinding(ctrl.selectedBinding.metadata.name)"
+      ng-repeat="appForBinding in ctrl.appsForBinding(ctrl.selectedBinding.metadata.name)">
       {{appForBinding.metadata.name}} <small class="text-muted">&ndash; {{ appForBinding.kind | humanizeKind : true}}</small>
     </div>
-    <div ng-if="!(ctrl.appsForBinding(ctrl.selectedBinding))">
-      <!-- TODO: selected a secret name in the form, but see a binding name in the result. -->
-      {{ctrl.selectedBinding}} <small class="text-muted">&ndash; Binding</small>
+    <div ng-if="!(ctrl.appsForBinding(ctrl.selectedBinding.metadata.name))">
```
spadgett: Probably should use the `size` filter. In case we ever change the overview to set an empty array instead of null/undefined.

benjaminapetersen:
Updating to sort by first associated app name:
screen shot 2017-06-16 at 10 40 01 am
will push shortly.

app/views/directives/bind-service/delete-binding-result.html
benjaminapetersen: done.

app/views/directives/bind-service/delete-binding-select-form.html
```diff
@@ -3,18 +3,18 @@ <h3 class="mar-top-none">
 </h3>
 <form name="ctrl.bindingSelection" class="mar-bottom-lg">
   <fieldset ng-disabled="ctrl.isDisabled">
-    <div ng-repeat="binding in ctrl.bindings" class="radio">
+    <div ng-repeat="binding in ctrl.bindings | orderBy: ctrl.firstAppForBindingName : binding" class="radio">
```
spadgett: `firstAppForBindingName` is a function right?

```html
<div ng-repeat="binding in ctrl.bindings | orderBy: ctrl.firstAppForBindingName() : binding" class="radio">
```

spadgett: Actually I see `orderBy` can be passed a function -- disregard my comment above -- but it looks like `:binding` on the end is wrong. Isn't that the reverse sort flag?

benjaminapetersen: Yeah that was definitely unneeded.  Didn't seem to do a flip, but the binding was passed as an arg anyway.  Removed in last push.

benjaminapetersen:
@spadgett back up and running (finally) w/a catalog & broker.  Any concern w/merging this && I will work on moving that sort to overview as a follow?  (just looking at time till the gate closes again)

spadgett:
[merge][severity:bug]

app/views/directives/bind-service/delete-binding-select-form.html
spadgett: Worries me that the ordering wasn't flipped. We should double check the sorting, but either way this should fix the immediate problem

PR #1736, opened 2017-06-19
Removing openshift-icon as they're no longer needed

rhamilto:
In preparing to move openshift-logos-icon to a separate repo, I noticed we no longer have a need for openshift-icon; as a result, I'm removing it from the repo (hooray for less code!).


app/styles/_openshift-icon.less
```diff
@@ -1,331 +0,0 @@
...
-  position: relative;
-  .font-icon:nth-child(n+2) {
-    position: absolute;
-    left: 0;
-    top: 0;
-    text-shadow: none;
-  }
-}
-
-.font-icon {
```
jwforres: don't we still need .font-icon stuff, that class is used in conjunction with .icon-whatever from the logo icon set

rhamilto: It's still in [_openshift-logos-icon.less](https://github.com/openshift/origin-web-console/blob/master/app/styles/_openshift-logos-icon.less#L11-L24).

jwforres: there are styles here that im not seeing over there though, not needed?

rhamilto: Nope.  I tested a couple context where we use .font-icons from logos (existing catalog and service catalog) and everything is fine without those styles.

jwforres:
[merge][severity:bug]

PR #1738, opened 2017-06-19
Bug 1462205 - Delete binding shouldn't show up when all bindings are pending deletion

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1462205

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -1101,6 +1101,7 @@ function OverviewController($scope,
     // - API objects by binding name
     state.bindingsByApplicationUID = {};
     state.applicationsByBinding = {};
+    state.deleteableBindingsByApplicationUID = {};
 
     // If there are no bindings, nothing to do.
     if (_.isEmpty(state.bindings)) {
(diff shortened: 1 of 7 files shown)


app/scripts/controllers/overview.js
```diff
@@ -1101,6 +1101,7 @@ function OverviewController($scope,
     // - API objects by binding name
     state.bindingsByApplicationUID = {};
     state.applicationsByBinding = {};
+    state.validBindingsByApplicationUID = {};
```
jwforres: im not in love with the variable names here or in service-row, open to other suggestions

jwforres:
This also fixes the text display to say Deleting with the spinner instead of Pending

app/scripts/controllers/overview.js
spadgett: I don't have a great suggestion :(

 `valid` sort of implies ready to me... `deleteableBindingsByApplicationUID` too on the nose?

jwforres: yeah i started to call it undeletedBindingsByApplicationUID, its just... weird

spadgett: This is specifically for the delete binding action, so I don't mind a name that makes that clear even if it doesn't roll off the tongue.

Alternatively we could put the filter in overview-list-row and call it `deleteableBindings`.

jwforres:
@spadgett updated the variable names, I also changed it to only pass the deleteable bindings into the unbind service directive

spadgett:
@benjaminapetersen FYI, probably conflicts with #1742

spadgett:
[merge][severity:blocker]

PR #1739, opened 2017-06-19
Send propagationPolicy null so that instances and bindings can be deleted cleanly

jwforres:
Depends on https://github.com/openshift/origin-web-common/pull/113

### DIFF
--- a/app/scripts/directives/deleteLink.js
+++ b/app/scripts/directives/deleteLink.js
@@ -147,6 +147,11 @@ angular.module("openshiftConsole")
               deleteOptions.gracePeriodSeconds = 0;
             }
 
+            // TODO - remove once this is resolved https://github.com/kubernetes-incubator/service-catalog/issues/942
+            if (scope.group === 'servicecatalog.k8s.io') {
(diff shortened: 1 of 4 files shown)


jwforres:
Waiting on next release of common

spadgett:
LGTM

spadgett:
Should we worry about the other resources page? Now that we have the aggregator, service catalog resources are showing up

jwforres:
Yeah good catch, we should either block them from showing up or handle the
delete correctly there. Main reason I would consider having them not
showing up is the editing experience, I don't know how much of update the
svc catalog is handling properly.

On Jun 19, 2017 9:56 PM, "Sam Padgett" <notifications@github.com> wrote:

Should we worry about the other resources page? Now that we have the
aggregator, service catalog resources are showing up

—
You are receiving this because you authored the thread.
<https://github.com/openshift/origin-web-console/pull/1739#issuecomment-309622367>,
or mute the thread

spadgett:
[merge][severity:blocker]

PR #1742, opened 2017-06-20
Sort bindingsByInstanceRef by first associated application

benjaminapetersen:
Sort bindingsByInstanceRef by first associated application (from sorted applications)

Per @spadgett suggestion, sorting these in the overview controller so that they do not need to be resorted elsewhere.

screen shot 2017-06-20 at 10 28 53 am

screen shot 2017-06-20 at 10 28 59 am

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -1110,9 +1110,9 @@ function OverviewController($scope,
 
     // All objects that can be a target for bindings.
     var objectsByKind = [
+      overview.deployments,
       overview.deploymentConfigs,
       overview.vanillaReplicationControllers,
-      overview.deployments,
       overview.vanillaReplicaSets,
       overview.statefulSets
     ];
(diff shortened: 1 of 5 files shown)


app/scripts/controllers/overview.js
```diff
@@ -1153,6 +1153,18 @@ function OverviewController($scope,
         });
       });
     });
+
+    // sorting the bindingList for each instance by the first associated app for each binding.
+    // doing it once here eliminates the need to resort them throughout the console.
+    overview.bindingsByInstanceRef = _.reduce(overview.bindingsByInstanceRef, function(result, bindingList, key) {
+      result[key] = _.sortBy(bindingList, function(binding) {
+        var apps =  _.get(state.applicationsByBinding, [binding.metadata.name]);
+        var sorted = _.sortBy(apps, 'metadata.name');
```
spadgett: It would be better to sort the applicationsByBinding list above instead of here. Then it's also sorted when expanding the provisioned service row in the overview.

benjaminapetersen:
@spadgett rebased, removed the extra `_.sort()`.  returning the binding `name` if no apps sorts the secrets.  That minor adjustment to `objectsByKind` is included.

I still am a little uncertain if the kind is handled/should be handled per doing this:
screen shot 2017-06-20 at 10 28 53 am
Or if these just happen to appear neat and tidy due to the naming conventions.

spadgett:
[merge][severity:bug]

PR #1743, opened 2017-06-20
Utilizing openshift-logos-icon dependency

rhamilto:
Instead of local copy.


jwforres:
were we not using the dist/styles/fonts/openshift-logos-icon.eot  font file at all? looks like it was completely removed from the final dist output

did you run the grunt server in the dist mode to see if the logos still loaded correctly on all the browsers?

rhamilto:
> were we not using the dist/styles/fonts/openshift-logos-icon.eot font file at all? looks like it was completely removed from the final dist output

No as the .eot is for IE6-8 and IE9 in compatibility mode.

> did you run the grunt server in the dist mode to see if the logos still loaded correctly on all the browsers?

I did.  But let me double check as you asking the question has cast doubt.  ;-)

jwforres:
Ok :)  the changes LGTM but I want to hold this till monday given the risk

rhamilto:
> I did. But let me double check as you asking the question has cast doubt. ;-)

Double check is good.

screen shot 2017-06-20 at 1 34 04 pm

screen shot 2017-06-20 at 1 34 55 pm

jwforres:
[merge]

PR #1744, opened 2017-06-20
Add requester-username parameter when needed on bind

### LINKED ISSUE #1737: Need to send username param for template broker provision and bind
We should hide the username param coming from the template broker schemas so that the user doesn't need to type it in.  This is temporary until we do something differently for username in 3.6

Binding for template broker is currently broken from the console since we don't support parameters and username is required.

spadgett:
Requires https://github.com/openshift/origin-web-common/pull/115
See also https://github.com/openshift/origin-web-catalog/pull/314

Fixes #1737


spadgett:
[merge][severity:blocker]

PR #1747, opened 2017-06-20
Bug fix where pod quota warning was misaligned on browse pages

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1463116

After:
screen shot 2017-06-20 at 1 52 59 pm

### DIFF
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -100,6 +100,9 @@ code.command {
 
 .deployment-donut {
   justify-content: center;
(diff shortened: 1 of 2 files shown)


spadgett:
Not caused by this change, but check events link is broken on this page since we're not watching for changes to the `tab` query parameter. Will open a separate issue.

spadgett:
[merge][severity:bug]

spadgett:
Opened #1748

PR #1749, opened 2017-06-20
Watch for changes to the tab query parameter

### LINKED ISSUE #1748: Quota check events link doesn't work for donut on replication controller page
The link goes to the events tab on the same page, but doesn't trigger a tab change when clicked.

openshift web console 2017-06-20 14-03-22

@jwforres fyi

spadgett:
In the `persist-tab-state` directive, watch for changes to the `tab`
query parameter so we can update the selected tab. This lets links to
the same page, but a different tab, work (like the check events link in
the `deployment-donut` quota warning).

Fixes #1748


jwforres:
[merge][severity:bug]

PR #1752, opened 2017-06-20
Bug 1462781 - should show the image stream from reference when it exists instead of pushed image

jwforres:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1462781


jwforres:
[merge][severity:blocker]

PR #1754, opened 2017-06-21
Bump origin-web-catalog to v0.0.25

spadgett:
Fix binding result message to not show JSON blob

https://github.com/openshift/origin-web-catalog/pull/320


spadgett:
[merge][severity:blocker]

PR #1756, opened 2017-06-21
Show average pod metrics for all containers on overview

### LINKED ISSUE #1668: Overview metrics only show first container
We should either show total for all containers or let users pick the container. (Total usage always in collapsed view?)

@smarterclayton @jwforres

spadgett:
Currently, the overview only shows the average metrics for the first container in each pod. Instead, include usage for all containers (but still average by pod).

Fixes #1668

cc @smarterclayton


jwforres:
LGTM but should probably wait till monday to merge

jwforres:
[merge]

PR #1759, opened 2017-06-22
Change create project redirect for new experience

### LINKED ISSUE #1485: Create project from projects page takes you to the old "add to project" page
When the catalog is turned on, creating a project from the projects list page takes the user to the old (current) add to project page. User should probably be directed back to the projects list page, or into the project with the empty state.

screenshot from 2017-04-27 16-46-28
screenshot from 2017-04-27 16-47-26

spadgett:
Don't redirect users to the old catalog after creating a project when the new experience is enabled. Instead go back to the project list.

Closes #1485

### DIFF
--- a/app/scripts/controllers/createProject.js
+++ b/app/scripts/controllers/createProject.js
@@ -9,7 +9,24 @@
  */
 angular.module('openshiftConsole')
   .controller('CreateProjectController',
-              function($scope, AuthService) {
+              function($scope,
+                       $location,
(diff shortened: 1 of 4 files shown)


jwforres:
Seems like a good compromise fix for 3.6, I do agree we should probably switch to the popup create dialog in 3.7

should wait to merge till monday

jwforres:
[merge]

PR #1762, opened 2017-06-23
Fix delete binding result message

spadgett:
Fix a timing issue where the app names changed to the secret name on the delete binding results step. This happened because deleting the binding updated the `bindingsByApplication` map that was passed in.

Also changes the results step to only show the redeploy pods message if the service was bound to an app.

### DIFF
--- a/app/scripts/directives/unbindService.js
+++ b/app/scripts/directives/unbindService.js
@@ -26,11 +26,16 @@
     var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
 
     var unbindService = function() {
+      var bindingName = ctrl.selectedBinding.metadata.name;
+      // Make sure to get the unbound apps now. Otherwise they don't appear on
+      // the result page since deleting the binding will remove them from the
+      // map that is passed in.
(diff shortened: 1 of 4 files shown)


app/scripts/directives/unbindService.js
```diff
@@ -26,11 +26,16 @@
     var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
 
     var unbindService = function() {
+      var bindingName = ctrl.selectedBinding.metadata.name;
+      // Make sure to get the unbound apps now. Otherwise they don't appear on
+      // the result page since deleting the binding will remove them from the
+      // map that is passed in.
+      ctrl.unbdoundApps = ctrl.appsForBinding(bindingName);
```
jwforres: typo unboundApps, here and below

spadgett:
Thanks, fixed.

spadgett:
@benjaminapetersen we didn't see this before because of the bug that prevented the binding from being deleted

jwforres:
can merge when queue reopens

benjaminapetersen:
Prob related to what I just did for #1764 as well.

jwforres:
[merge]

PR #1764, opened 2017-06-23
Fix issue #1709, autoselect single binding for deletion

### LINKED ISSUE #1709: Auto select a single binding in the delete binding flow
Per [this comment](https://github.com/openshift/origin-web-console/pull/1705#issuecomment-308516632) from @ncameronbritt, if there is a single binding, it should be automatically selected & the "delete" button should be enabled.

@spadgett fyi, will fast follow this as a bug.

benjaminapetersen:
Fixes #1709 

Boo `ng-model` & `ng-value` not playing nice.

### DIFF
--- a/app/scripts/directives/unbindService.js
+++ b/app/scripts/directives/unbindService.js
@@ -26,11 +26,13 @@
     var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
 
     var unbindService = function() {
+      ctrl.deletedBinding = _.first(_.filter(ctrl.bindings, {metadata: {name: ctrl.selectedBinding}}));
+      ctrl.appsForDeletedBinding = ctrl.appsForBinding(ctrl.selectedBinding);
       DataService.delete({
         group: 'servicecatalog.k8s.io',
         resource: 'bindings'
       },
-      ctrl.selectedBinding.metadata.name,
+      ctrl.selectedBinding,
       context,
       { propagationPolicy: null })
       .then(_.noop, function(err) {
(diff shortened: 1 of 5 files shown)


spadgett:
> Boo ng-model & ng-value not playing nice.

I'd like to find out why this isn't working because I think it should

app/scripts/directives/unbindService.js
```diff
@@ -26,11 +26,13 @@
     var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
 
     var unbindService = function() {
+      ctrl.deletedBinding = _.first(_.filter(ctrl.bindings, {metadata: {name: ctrl.selectedBinding}}));
+      ctrl.appsForDeletedBinding = angular.copy(ctrl.appsForBinding(ctrl.selectedBinding));
```
spadgett: I believe the copy is unnecessary since the entire map is replaced in the overview controller

jwforres:
@spadgett FWIW i hit the same issue as Ben on the ng-model and ng-value not working well together

spadgett:
We could have a problem where a watch is updating the bindings list so that the objects are not equal.

app/scripts/directives/unbindService.js
benjaminapetersen: right, I think thats why I'm copying it.  Otherwise the related apps don't map by the time the results panel appears.

spadgett: You still shouldn't need a copy because you have a reference to the old array, which is not mutated by the overview controller. The entire map is replaced, so the previous array is unchanged. It should work without the copy.

spadgett: By "copying" I mean the `angular.copy` call specifically

benjaminapetersen: True! Overly cautious on my part, updated & removed the `angular.copy`.

benjaminapetersen:
@spadgett @jwforres from what I remember, `ng-model` and `ng-view` simply don't communicate & stay in sync. Its been long enough that I'm fuzzy on the details, however.

Issue #1751, opened 2017-06-20
Project viewer shouldn't see create / delete binding or delete instance actions


jwforres:
In the overview I should only see the option to create and delete bindings if I have access to do so. Same with delete on service instances.

PR #1768, opened 2017-06-26
Should only see actions for service cat resources in overview that the user can do

jwforres:
Fixes https://github.com/openshift/origin-web-console/issues/1751

### DIFF
--- a/app/scripts/directives/overview/listRow.js
+++ b/app/scripts/directives/overview/listRow.js
@@ -1,6 +1,6 @@
 'use strict';
 
-(function() {  
+(function() {
   angular.module('openshiftConsole').component('overviewListRow', {
     controller: [
       '$filter',
@@ -40,6 +40,7 @@
     var canI = $filter('canI');
     var deploymentIsInProgress = $filter('deploymentIsInProgress');
     var isBinaryBuild = $filter('isBinaryBuild');
+    var enableTechPreviewFeature = $filter('enableTechPreviewFeature');
 
     var updateTriggers = function(apiObject) {
       var triggers = _.get(apiObject, 'spec.triggers');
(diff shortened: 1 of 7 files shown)


jwforres:
Long term these are the ACL checks we want, in the short term these wont actually matter because we need to disable viewer access to instances/bindings, see https://github.com/openshift/origin-web-console/pull/1769 related to that

app/scripts/directives/overview/listRow.js
```diff
@@ -150,6 +151,8 @@
 
     row.canIDoAny = function() {
```
benjaminapetersen: So the purpose of this function seems to be "if I can do anything at all, show the dropdown".  Its fairly lengthy, and a lot of the logic is repeated in the template with all the `ng-if` statements.  I'm wondering if there is a way to break it down?  

I pause at the `switch` with nested `if` that have `//comments` that are needed to reveal the true intention.  I wonder if the details of this could be broken up into smaller functions and/or pushed up into the related `Pod`, & `Deployment` services (ex:`DeploymentService.canIDoAny()`).

jwforres: Not new, I'm just adding to the existing behavior.  If we want to come up with some alternative we'll have to do it later, its too late to make a significant change here.

And no pushing this up to DeploymentService.canIDoAny would not be right, this is specific to the actions that we choose to show on the overview.

jwforres:
[merge]

app/scripts/directives/overview/listRow.js
benjaminapetersen: Sounds good, yeah I'm not sure either, just thinking out loud.

PR #1769, opened 2017-06-26
Don't request instances and bindings if the user doesn't have watch rights

jwforres:


jwforres:
FYI @spadgett viewers won't be able to see instances/bindings for tech preview because of the concerns around params containing secret info

app/scripts/controllers/overview.js
```diff
@@ -1349,7 +1349,9 @@ function OverviewController($scope,
       updateQuotaWarnings();
     }, {poll: true, pollInterval: DEFAULT_POLL_INTERVAL}));
 
-    if (SERVICE_CATALOG_ENABLED) {
+    var canI = $filter('canI');
+    // The canI check on watch should be temporary until we have a different solution for handling secret parameters
```
benjaminapetersen: Curious (my own ignorance) how changing what we do for handling secret parameters will help with this watch on `instances`?  Not saying the comment needs more necessarily, I just don't know that I follow.

jwforres: We want instances to be watchable by project viewers, once they are then this check isn't really necessary, we dont have this check for any of the other watches today.  We can't make them watchable (or listable) by project viewers yet because of the way parameters are are currently stored on instances and bindings.  Viewers today should not be allowed to see potentially confidential information.

jwforres:
travis passed, the dist is correct, the local variable minification in the overview code changed because i added a new variable

[merge]

app/scripts/controllers/overview.js
benjaminapetersen: Awesome, thx!

PR #1776, opened 2017-06-28
Fixes several bugs in service instance row

jwforres:
bindable was set incorrectly if service classes loaded after instances
displayName was never correct if the service class had a pretty name
fixed several references to osbMetadata to be externalMetadata


jwforres:
[merge]

PR #1777, opened 2017-06-28
Upgrade versions of catalog (0.0.27) and common (0.0.39)

jeff-phillips-18:
Added project display name to common bind service form component.


app/scripts/directives/bindService.js
```diff
@@ -20,11 +21,13 @@
...
                        $filter,
                        DataService,
-                       BindingService) {
+                       BindingService,
+                       ProjectsService) {
     var ctrl = this;
     var validityWatcher;
     var bindingWatch;
     var statusCondition = $filter('statusCondition');
+    console.dir(ctrl.target);
```
dtaylor113: Do we want this in production code?

app/scripts/directives/bindService.js
```diff
@@ -178,10 +180,13 @@
       });
 
       if (ctrl.target.kind === 'Instance') {
-        ctrl.bindType = "secret-only";
-        ctrl.appToBind = null;
-        ctrl.serviceToBind = ctrl.target.metadata.name;
-        loadApplications();
+        ProjectsService.get(ctrl.target.metadata.namespace).then(_.spread(function(project) {
```
jwforres: This is going to cause another request for the project when we already have the project loaded

jeff-phillips-18: where is the project?

app/scripts/directives/bindService.js
jwforres: No, needs to be removed

jeff-phillips-18: sorry, left over...

app/scripts/directives/bindService.js
jwforres: You should update the bindService component to take the project as a param, anywhere we are using that component the project has already been loaded by the main controller for the page

jwforres: in both list-row and service-instance-row its available as row.state.project

jeff-phillips-18: Awesome. Thanks @jwforres !

jeff-phillips-18: Updated. So much better. Thanks again.

jwforres:
[merge]

PR #1779, opened 2017-06-29
Fix service instance object dump in bind app to service dialog

jwforres:


jwforres:
the two dialog directions svc to app vs app to svc where behaving differently

jwforres:
[merge]

PR #1781, opened 2017-06-30
[WIP] Don't show in-page alerts for create project

spadgett:
Fixes an issue where both an in-page alert and toast notification can appear when there is an error creating a project.


spadgett:
Checking to see if this breaks our integration tests.

[test]

spadgett:
Replaced by #1787 which has the common repo change

Issue #1684, opened 2017-06-12
Test flake: Timed out waiting for the WebDriver server


spadgett:
https://ci.openshift.redhat.com/jenkins/job/test_pull_requests_origin_web_console/1500/console

```
Error: Timed out waiting for the WebDriver server at http://172.18.8.133:56366/wd/hub
    at Error (native)
    at onResponse (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/http/util.js:87:11)
    at /data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/http/util.js:42:21
    at /data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/http/http.js:96:5
    at ClientRequest.<anonymous> (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/http/index.js:145:7)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at Socket.socketErrorListener (_http_client.js:309:9)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
Fatal error: protractor exited with code: 1
```

spadgett:
https://ci.openshift.redhat.com/jenkins/job/test_pull_requests_origin_web_console/1617/

```
Selenium standalone server started at http://172.18.5.43:37253/wd/hub
ERROR - Unable to start a WebDriver session.

/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/atoms/error.js:113
  var template = new Error(this.message);
                 ^
UnknownError: Unable to connect to host 127.0.0.1 on port 7055 after 45000 ms. Firefox console output:
Error: cannot open display: :10
```

spadgett:
https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_origin_web_console/16

PR #1792, opened 2017-07-03
Bug 1462067 - Don't set propagationPolicy when deleting pod immediately

spadgett:
This change removes the propagation policy from delete options when force deleting a pod from the web console. It makes the web console DELETE request consistent with `oc delete --now --force` (only when delete immediately is checked).

@jwforres WIP because I haven't been able to verify against a pod that is really stuck in terminating. I'm concerned we might have a GC bug with propagation policy foreground and grace period 0.

@derekwaynecarr Do you know how I can purposely get a pod stuck in Terminating state to test this fix?

cc @deads2k 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1462067

### DIFF
--- a/app/scripts/directives/deleteLink.js
+++ b/app/scripts/directives/deleteLink.js
@@ -145,6 +145,7 @@ angular.module("openshiftConsole")
             var deleteOptions = {};
             if (scope.options.deleteImmediately) {
               deleteOptions.gracePeriodSeconds = 0;
+              deleteOptions.propagationPolicy = null;
             }
 
             // TODO - remove once this is resolved https://github.com/kubernetes-incubator/service-catalog/issues/942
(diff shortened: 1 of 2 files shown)


jwforres:
We've been trying to avoid making any special exceptions in the console though, so far every weird case we have hit has been a bug with GC.

spadgett:
If you're trying to force delete something, foreground propagation seems wrong. But this comment worries me:

> However console (v3.6.133) doesn't work, it still put pod in Terminating status. **Then try again by CLI command and now pod can't be removed.**

https://bugzilla.redhat.com/show_bug.cgi?id=1462067#c12

jwforres:
Agreed this change is OK then and we need to spawn a separate issue for the fact that GC + Terminating can get pods permanently stuck in storage.

jwforres:
looks like the webdriver start flake [merge]

spadgett:
Flake

```
Selenium standalone server started at http://172.18.5.43:37253/wd/hub
ERROR - Unable to start a WebDriver session.

/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/atoms/error.js:113
  var template = new Error(this.message);
                 ^
UnknownError: Unable to connect to host 127.0.0.1 on port 7055 after 45000 ms. Firefox console output:
Error: cannot open display: :10
```

[merge]

spadgett:
Flake #1684

PR #1794, opened 2017-07-05
Let users add labels when creating a config map

### LINKED ISSUE #1185: Should be able to specify config map labels
The create and edit forms for config maps don't let you set label values.

spadgett:
Fixes #1185

@jwforres It looks like none of our editors let you change labels, so I only added it to the config map create form for now.

screen shot 2017-07-05 at 12 12 14 pm


jwforres:
The code changes LGTM but I know we've been having discussions in general about how we are presenting labels to users. Wondering if we should take an overall look at how we push label usage in the console before adding them to more screens?

cc @ncameronbritt

PR #1796, opened 2017-07-05
Use config.local.js if present for grunt serve:dist

### LINKED ISSUE #1795: grunt serve:dist doesn't use config.local.js
This prevents log in from working with our default config.js and `oc cluster up` for one.

Related to #1325

@jwforres @rhamilto FYI

spadgett:
Fixes #1795

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -157,14 +157,21 @@ module.exports = function (grunt) {
           open: true,
           livereload: false,
           middleware: function (connect) {
+            var rewriteRules = [
+              '^/styles/extensions\.css$ /' + contextRoot + '/extensions/extensions.css [R]',
+              '^/scripts/extensions\.js$ /' + contextRoot + '/extensions/extensions.js [R]',
+              '^/$ /' + contextRoot + '/ [R=302]',
+              '^/' + contextRoot + '(.*)$ $1',
+              '!^/(config.js|(bower_components|scripts|images|styles|views|extensions)(/.*)?)$ /index.html [L]'
+            ];
+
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -157,14 +157,21 @@ module.exports = function (grunt) {
...
+              '^/styles/extensions\.css$ /' + contextRoot + '/extensions/extensions.css [R]',
+              '^/scripts/extensions\.js$ /' + contextRoot + '/extensions/extensions.js [R]',
+              '^/$ /' + contextRoot + '/ [R=302]',
+              '^/' + contextRoot + '(.*)$ $1',
+              '!^/(config.js|(bower_components|scripts|images|styles|views|extensions)(/.*)?)$ /index.html [L]'
+            ];
+
+            // If config.local.js exists, use that instead of config.js.
+            if (grunt.file.exists('app/config.local.js')) {
+              rewriteRules.unshift('^/config.js$ /config.local.js [L]');
```
jwforres: is this how we are doing it for the dev server?  the only problem i see with this is that if you delete your config.local.js file you have to restart your grunt server

spadgett: No, we copy the `config.local.js` to `.tmp` for `grunt serve`, but I don't think that works here.

https://github.com/openshift/origin-web-console/blob/master/Gruntfile.js#L552-L563

I suspect copying has the same problem when you delete the file.

jwforres: Ok, not a big deal since we rarely create/delete them

jwforres:
[merge]  going ahead and merging since this doesn't affect shipped code

PR #1797, opened 2017-07-05
Allow duplicate keys when importing YAML

### LINKED ISSUE #1629: Template can be imported via CLI, but fails to be imported on the web UI
It appears that there's a bogus YAML validation happening on the UI, preventing templates that can be imported via `oc create -f` from being imported via the web UI:

jaeger-production-template

Ref: https://github.com/jaegertracing/jaeger-openshift/pull/8#issuecomment-305474587

spadgett:
For compatibility with `oc create`, accept YAML with duplicate keys when
importing YAML. Duplicate values will be overridden.

Still count it as an error when editing YAML, however, since the initial
YAML should not have duplicate keys editing an existing resource. This
gives the user a chance to correct mistakes.

Fixes #1629

### DIFF
--- a/app/scripts/directives/fromFile.js
+++ b/app/scripts/directives/fromFile.js
@@ -39,6 +39,13 @@ angular.module("openshiftConsole")
           $scope.editorErrorAnnotation = _.some(editorAnnotations, { type: 'error' });
         };
 
+        var parseYAML = function() {
+          return jsyaml.safeLoad($scope.editorContent, {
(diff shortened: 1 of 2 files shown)


jwforres:
Should wait for 3.7

spadgett:
I'm working on adding annotations by the line numbers on problems (errors and warnings).

screen shot 2017-07-05 at 4 27 10 pm

spadgett:
Closing in favor of #1798, which enables this with better validation.

PR #1798, opened 2017-07-05
Improve YAML editor validation and feedback

spadgett:
- Create a common YAML editor component.
- Show errors and warnings as Ace editor annotations by the line number.
- For compatibility with `oc create`, accept YAML with duplicate keys
  when importing YAML. Duplicate keys when importing YAML are considered
  a warning.

screen shot 2017-07-05 at 5 59 49 pm

Replaces #1797

@jwforres

### DIFF
--- a/app/scripts/directives/uiAceYaml.js
+++ b/app/scripts/directives/uiAceYaml.js
@@ -1,0 +1,124 @@
+'use strict';
+
+(function() {
+  angular.module('openshiftConsole').component('uiAceYaml', {
+    controller: [
+      '$scope',
+      UIAceYAML
+    ],
+    controllerAs: '$ctrl',
+    bindings: {
+      resource: '=',
+      ngRequired: '<?',
+      showFileInput: '<?'
+    },
+    templateUrl: 'views/directives/ui-ace-yaml.html'
+  });
+
+  function UIAceYAML($scope) {
+    var ctrl = this;
+    var aceEditor;
+
+    var parseYAML = function(strict) {
+      // https://github.com/nodeca/js-yaml#safeload-string---options-
+      return jsyaml.safeLoad(ctrl.model, {
+        // If `strict` is false, allow duplicate keys in the YAML for
+        // compability with `oc create` using the js-yaml `json` option. This
+        // also lets us parse JSON as YAML, where duplicate keys are allowed.
+        json: !strict
+      });
+    };
+
+    var clearAnnotations = function() {
+      aceEditor.getSession().clearAnnotations();
+      $scope.$evalAsync(function() {
+        ctrl.annotations = {};
+      });
+    };
+
+    var setAnnotation = function(e, severity) {
+      var session = aceEditor.getSession();
+      var length = session.getLength();
+      var row = _.get(e, 'mark.line', 0);
+      var col = _.get(e, 'mark.column', 0);
+      var message = e.message || 'Could not parse content.';
+
+      // If the error line is reported as being after the last line, use the
+      // last line. This can happen when the error is at the very end of the
+      // document and the user doesn't have a final newline.
+      if (row >= length) {
+        row = length - 1;
+      }
+
+      var annotation = {
+        row: row,
+        column: col,
+        text: message,
+        type: severity
+      };
+      session.setAnnotations([ annotation ]);
+
+      $scope.$evalAsync(function() {
+        ctrl.annotations = {};
+        ctrl.annotations[severity] = [ annotation ];
+      });
+    };
+
+    var setValid = function(valid) {
+      $scope.$evalAsync(function() {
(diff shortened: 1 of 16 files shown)


spadgett:
[test]

app/scripts/directives/uiAceYaml.js
```diff
@@ -0,0 +1,124 @@
...
+        ctrl.annotations = {};
+      });
+    };
+
+    var setAnnotation = function(e, severity) {
+      var session = aceEditor.getSession();
+      var length = session.getLength();
+      var row = _.get(e, 'mark.line', 0);
+      var col = _.get(e, 'mark.column', 0);
+      var message = e.message || 'Could not parse YAML.';
```
jwforres: if i dont *know* that JSON is YAML, and i pasted in JSON and saw "Could not parse YAML", would that confuse me?  We specifically call it Import JSON / YAML as the action

app/scripts/directives/uiAceYaml.js
```diff
@@ -0,0 +1,124 @@
...
+      $scope.$evalAsync(function() {
+        ctrl.form.$setValidity('yamlValid', valid);
+      });
+    };
+
+    var updated = function() {
+      // Check for errors, then check for warnings.
+      try {
+        ctrl.resource = parseYAML(false);
+        setValid(true);
```
jwforres: does resetting it here cause unnecessary flashing of the disabled state of the Save button, if going from invalid to invalid?  should we wait to reset it until the try and catch blocks for the warnings test?

app/views/directives/ui-ace-yaml.html
```diff
@@ -0,0 +1,44 @@
+<ng-form name="$ctrl.form">
+  <div class="form-group" id="yaml-file">
+    <osc-file-input
+      ng-if="$ctrl.showFileInput"
+      model="$ctrl.fileUpload"
+      drop-zone-id="yaml-file"
+      help-text="Upload file by dragging & dropping, selecting it, or pasting from the clipboard."
```
jwforres: not new, but should this be "Upload a file"

app/views/directives/ui-ace-yaml.html
```diff
@@ -0,0 +1,44 @@
...
+      ng-disabled="false"
+      hide-clear="true"></osc-file-input>
+
+    <div class="edit-yaml-errors">
+      <!--
+        Since the YAML parser quits after the first error, don't give a count.
+        Just show an indicator and link to the first problem line.
+      -->
+      <div ng-if="firstError = $ctrl.annotations.error[0]">
```
jwforres: if the whole file contents fits in the display window, is it weird to have this link?

app/scripts/directives/uiAceYaml.js
spadgett: No flashing (warnings are still valid)

app/scripts/directives/uiAceYaml.js
spadgett: Good point. "Could not parse content." maybe?

Hopefully you never see the default message

app/views/directives/ui-ace-yaml.html
spadgett: I think it's subtle enough? Opinion? It might be stranger for it to appear and disappear when you scroll

openshift web console 2017-07-11 15-15-00

spadgett: Or disappear when you click it

app/scripts/directives/uiAceYaml.js
jwforres: nevermind, dumb moment, realizing angular wouldnt have the chance to run a digest loop in the middle

app/views/directives/ui-ace-yaml.html
jwforres: I don't feel too strongly about it @ncameronbritt ?

spadgett: It will move the cursor to the right line, so it's still useful even if the error was visible

jwforres: ok, if its always doing something obvious im fine with it

jwforres: Can it also move the cursor to the right column within the line?

spadgett: It's possible yeah, although I'm not doing it. One problem is that the parser often reports the column just after the error, so the cursor will be moved past where you need to make the change.

jwforres:
Ok that's what I was worried about, nvm then

spadgett:
@jwforres Another look? I think I've addressed the comments so far.

jwforres:
[merge]

spadgett:
[merge]

PR #1803, opened 2017-07-06
Let users save env changes with enter

### LINKED ISSUE #1651: Enter key doesn't save changes on Environment tab
The enter key does not submit the form when editing environment variables for a deployment config.

cc @benjaminapetersen

spadgett:
The enter key now submits the form for in-page env variable editors.
Previously we were using `ng-form` without a surrounding `form` element,
which prevented the enter key from working.

Fixes #1651

### DIFF
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -463,7 +463,7 @@
                       cannot-sort
                       cannot-delete
                       show-header></key-value-editor>
-                    <ng-form name="forms.bcEnvVars" class="mar-bottom-xl block">
+                    <form name="forms.bcEnvVars" class="mar-bottom-xl block">
                       <div ng-if="'buildconfigs' | canI : 'update'">
                         <confirm-on-exit dirty="forms.bcEnvVars.$dirty"></confirm-on-exit>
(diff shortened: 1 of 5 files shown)


app/views/browse/stateful-set.html
```diff
@@ -148,26 +148,24 @@
                 <uib-tab active="selectedTab.environment" ng-if="statefulSet">
                   <uib-tab-heading>Environment</uib-tab-heading>
                   <div class="resource-environment">
-                    <ng-form name="forms.statefulSetEnvVars">
```
jwforres: i wonder if statefulsets will let us edit env vars yet

spadgett: Probably worth reevaluating the items in https://trello.com/c/M4fAYrUT

spadgett: Looks like it. I have a separate WIP PR now for stateful sets (#1815). Will remove stateful sets changes from this one.

spadgett:
Rolled back the stateful set changes so as not to conflict with #1815

spadgett:
Replaced by #1821

PR #1806, opened 2017-07-07
Let users edit app label directly for "add to project"

### LINKED ISSUE #1422: Not obvious that the app label can be changed
The app label in the UI is not modifiable and this is a bug (it should be) as it is from the CLI.

image

As explained by @bparees in an email:

> here's the label applying logic for new-app:
> 
> for non-template deployments:
> 
> 1) if the user provides labels on the command line, assign those labels
> 2) assign the default app=name to all objects, unless (1) already set the "app" label.
> 
> for template deployments:
> 
> 1) if the template defines labels, assign them
> 2) if the user provides labels on the command line, assign those labels (possibly overwriting values from (1)
> 3) assign the default app=name to all objects unless (1) or (2) already set the "app" label or any of the objects otherwise already has an app label defined (in which case we don't set it on any object).
> 
> So there should always be *an* app label available unless your template explicitly defines the app label on only a subset of objects in the template.

spadgett:
Prefill the label values instead of having a separate "system labels"
section that is read-only. Fixes a usability bug where it wasn't obvious
the app label could be overridden.

Fixes #1422

### DIFF
--- a/app/scripts/controllers/create/createFromImage.js
+++ b/app/scripts/controllers/create/createFromImage.js
@@ -116,8 +116,7 @@ angular.module("openshiftConsole")
             include: true,
             portOptions: []
           };
-          scope.userDefinedLabels = [];
-          scope.systemLabels = [appLabel];
+          scope.labelArray = [appLabel];
           scope.annotations = {};
           scope.scaling = {
             replicas: 1,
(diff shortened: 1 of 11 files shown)


spadgett:
@jwforres opinion on this change?

spadgett:
@bparees This change would let users optionally remove template labels from generated objects when instantiating templates from the web console. Is that a problem?

It might be possible to make the name readonly, but let users edit the value directly if we don't want users to remove the labels altogether.

bparees:
> @bparees This change would let users optionally remove template labels from generated objects when instantiating templates from the web console. Is that a problem?

@spadgett no, i don't think so.  Especially since this has no effect when a user instantiates a template via the service catalog provision flow, right?

spadgett:
> Especially since this has no effect when a user instantiates a template via the service catalog provision flow, right?

It won't, but users can still instantiate templates directly from YAML in the web console.

bparees:
> It won't, but users can still instantiate templates directly from YAML in the web console.

sure, i'm just noting that the one case where we might rely on labels at all is in the provision flow, so if you're not changing that, i don't think we care.

spadgett:
@jwforres PTAL

I need to rebuild to fix the scripts.js conflict, but should be ready for review

spadgett:
[merge]

Issue #1689, opened 2017-06-12
Overview should warn me when I get evicted


smarterclayton:
I have a deployment with one pod that gets frequently evicted due to memory pressure.  I had 2 evicted pods and the overview when collapsed showed 3 pods, but when expanded showed only 1 in the donut.  The 3 pod collapsed ring showed completely blue, but should have showed 2/3 something else (or an alert).

I expected to know about the evicted pod from overview, and see something in the ring or next to it (I had 1/1 pod, but the two evicted are important).

@derekwaynecarr

spadgett:
@derekwaynecarr Is there something in pod status that tells us if a pod has been evicted?

spadgett:
Two issues here:

1. The pod count in the collapsed overview row includes Failed pods, but we (intentionally) don't include Failed pods in the donut center text when the row is expanded. This is fairly easy to fix.
2. We were purposely filtering out Failed pods from the donut statuses. I'm not sure the best way to convey this since you might have hundreds of failed pods and only a few running, so the donut will turn mostly red. At least that is what I see in my environment when I set up an eviction-hard threshold. Maybe just an alert that you have failed / evicted pods with a link to the pods page?

@jwforres @smarterclayton @derekwaynecarr Opinion?

spadgett:
Something like...

![openshift web console 2017-07-07 16-09-58](https://user-images.githubusercontent.com/1167259/27975094-cbde8870-632e-11e7-9f54-da29c46ded4a.png)

spadgett:
![screen shot 2017-07-07 at 4 11 14 pm](https://user-images.githubusercontent.com/1167259/27975122-ee67aae8-632e-11e7-9db5-e17c159233ef.png)

PR #1808, opened 2017-07-07
Don't include Failed pods in count beside mini donut

spadgett:
The mini donut in a collapsed overview row includes Failed pods, but the
center text when the row is expanded does not. Don't include Failed pods
in the collapsed state count so that the numbers match.

See #1689


jwforres:
[merge][severity:blocker]

PR #1809, opened 2017-07-07
Warn on overview about evicted pods

### LINKED ISSUE #1689: Overview should warn me when I get evicted

spadgett:
openshift web console 2017-07-07 16-09-58

Closes #1689

cc @smarterclayton @derekwaynecarr

### DIFF
--- a/app/scripts/services/resourceAlerts.js
+++ b/app/scripts/services/resourceAlerts.js
@@ -18,6 +18,18 @@ angular.module("openshiftConsole")
       }
 
       var alerts = {};
+      // Handle evicted pod warnings specially so that evicted pods don't appear in the donut.
+      if (_.some(pods, { status: { phase: 'Failed', reason: 'Evicted' } })) {
+        alerts['pod_evicted'] = {
+          type: 'warning',
+          message: 'Pods have been evicted.',
+          links: [{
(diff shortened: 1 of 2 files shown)


spadgett:
@jwforres PTAL

app/scripts/services/resourceAlerts.js
```diff
@@ -18,6 +18,18 @@ angular.module("openshiftConsole")
       }
 
       var alerts = {};
+      // Handle evicted pod warnings specially so that evicted pods don't appear in the donut.
+      if (_.some(pods, { status: { phase: 'Failed', reason: 'Evicted' } })) {
+        alerts['pod_evicted'] = {
+          type: 'warning',
+          message: 'Pods have been evicted.',
```
jwforres: feel like the message could be a bit more helpful than that, @derekwaynecarr said pods are typically evicted because they are using more resources than they requested

PR #1815, opened 2017-07-10
[WIP] Let users update env vars for stateful sets

spadgett:
https://trello.com/c/M4fAYrUT

TODO:

- [ ] Support `valueFrom`


spadgett:
@jwforres FYI, I think we need to refactor some of env var code. There is too much copied code between different controllers :(

spadgett:
Replaced by #1821

PR #1818, opened 2017-07-10
Route table tweaks

spadgett:
Recommendations from @ncameronbritt 

- Show external link icon for route links
- "Routes To" -> "Service" in table header

### DIFF
--- a/app/styles/_typography.less
+++ b/app/styles/_typography.less
@@ -70,6 +70,10 @@ h1 {
 
 // Misc
 // --------------------------------------------------
+a > .fa-external-link {
+  font-size: 75%;
+}
+
 .action-inline {
   margin-left: @action-link-inline-margin;
   font-size: @font-size-small;
--- a/app/views/browse/route.html
+++ b/app/views/browse/route.html
@@ -64,7 +64,10 @@
                 <div ng-repeat="ingress in route.status.ingress" ng-init="admittedCondition = (ingress | routeIngressCondition : 'Admitted')" class="route-status">
                   <div class="h3">
                     <span ng-if="(route | isWebRoute)">
-                      <a ng-href="{{route | routeWebURL : ingress.host}}" target="_blank">{{route | routeLabel : ingress.host : true}}</a>
+                      <a ng-href="{{route | routeWebURL : ingress.host}}" target="_blank">
+                        {{route | routeLabel : ingress.host : true}}
+                        <i class="fa fa-external-link" aria-hidden="true"></i>
+                      </a>
                     </span>
                     <span ng-if="!(route | isWebRoute)">
                       {{route | routeLabel : ingress.host}}
--- a/app/views/browse/routes.html
+++ b/app/views/browse/routes.html
@@ -43,7 +43,11 @@
                   <tr>
                     <th>{{customNameHeader || 'Name'}}</th>
                     <th>Hostname</th>
-                    <th>Routes To</th>
+                    <!--
+                      Right now only services are supported. If other types are
+                      added, we'll need to change the table header.
+                     -->
+                    <th>Service</th>
                     <th>Target Port</th>
                     <th>TLS Termination</th>
                   </tr>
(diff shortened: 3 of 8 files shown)


spadgett:
The icon can wrap by itself to the next line unfortunately. I'm not sure how to prevent this and still let long routes wrap.

openshift web console 2017-07-10 16-04-39

ncameronbritt:
You had that done before I put up the issue!

If you append the icon without whitespace, doest it still wrap? How do we handle long URLs currently?

Also, should the icon be styled like a link? In this older screenshot from the Overview, it's not.
screen shot 2017-07-10 at 4 11 21 pm

spadgett:
> Also, should the icon be styled like a link? In this older screenshot from the Overview, it's not.

I was trying to be consistent with other links like these. I like it better when it's not part of the link, though :/

openshift web console 2017-07-10 16-20-35

ncameronbritt:
I think I actually like it better as part of the link. No strong preference, but we should be consistent.

Any preferences @openshift/team-ux-review ?

spadgett:
> If you append the icon without whitespace, doest it still wrap?

Still wraps given how we break URLs.

> How do we handle long URLs currently?

We break anywhere in the value so that long URLs don't overflow the column. There might be other ways to handle it, though.

ncameronbritt:
I know those URLs can be long... But breaking just for the icon isn't very good.

spadgett:
For comparison, here is how the icon looks as part of the link on the overview:

openshift web console 2017-07-10 16-26-13

spadgett:
> I know those URLs can be long... But breaking just for the icon isn't very good.

We have the same problem in other places with our warning icon unforunately.

@rhamilto @sg00dwin I wonder if

```css
  overflow: hidden;
  text-overflow: ellipsis;
```

isn't a better way of handling long unbroken strings than word break.

https://css-tricks.com/handling-long-unexpected-content-css/

spadgett:
> I think I actually like it better as part of the link.

I think what bothers me is the icon seems too large for the rest of the text.

beanh66:
I actually like it as part of the link too @ncameronbritt. Good point on the size though @spadgett, it wouldn't hurt to be a little smaller.

ncameronbritt:
I think @beanh66 and I agree.

sg00dwin:
Alternatively, since all links in that column will be external, what about denoting that by placing the icon next to the column heading.

screen shot 2017-07-11 at 10 24 27 am

spadgett:
Icon as part of link with font-size 75%

screen shot 2017-07-11 at 10 33 23 am

screen shot 2017-07-11 at 10 33 40 am

ncameronbritt:
I think that looks better @spadgett 

It's an interesting thought @sg00dwin. To me the icon indicates to the user what will happen when they click that link, so I'm not sure that it makes sense to include it in the table header.

spadgett:
@ncameronbritt I looked at this again with @rhamilto, and we can't find a good way to keep the route URL wrapping but prevent the icon from ever wrapping on its own. I think we have to live with the wrapping if we want the icon.

ncameronbritt:
I think that's ok @spadgett. Your initial comment about it gave me the impression that it happened all the time. If it's only with certain URL lengths/screen widths then it seems reasonable.

spadgett:
Yeah, it's uncommon. It only happens if your window is exactly the wrong width.

app/styles/_typography.less
```diff
@@ -70,6 +70,10 @@ h1 {
 
 // Misc
 // --------------------------------------------------
+a > .fa-external-link {
+  font-size: 75%;
```
jwforres: how does this change affect the look of the Learn More links?

spadgett: It makes them look better (IMO :)

There's one in the last screenshot above

jwforres:
LGTM for 3.7

spadgett:
[merge]

PR #1819, opened 2017-07-11
[WIP] Init Container Display & Link

### LINKED ISSUE #1790: Show link to full pod template on overview if resource has init containers
We should show a message on the overview and link to the configuration tab with the full pod template if resource has init containers.

Follow on to https://trello.com/c/kXzDqvXG

@jwforres FYI

cdcabrera:
Show count of init containers with a detail link.

Closes #1790 

initcontainers-jul-10-2017 20-35-06

### DIFF
--- a/app/scripts/directives/resources.js
+++ b/app/scripts/directives/resources.js
@@ -1,6 +1,33 @@
 'use strict';
 
 angular.module('openshiftConsole')
+  .component('initContainersSummary', {
+    bindings: {
+      apiObject: '<'
+    },
+    templateUrl: 'views/_init-containers-summary.html',
+    controller: function ($filter) {
+      var ctrl = this;
+
+      ctrl.$onChanges = function (changesObj) {
+        var apiObject = _.get(changesObj.apiObject, 'currentValue');
+
+        if (apiObject) {
+          ctrl.podTemplate = $filter('podTemplate')(apiObject);
+
+          switch (apiObject.kind) {
+            case 'DeploymentConfig':
+            case 'Deployment':
+              ctrl.tab = 'configuration';
+              break;
+            default:
+              ctrl.tab = 'details';
+              break;
+          }
+        }
+      };
+    }
+  })
   .directive('containerStatuses', function($filter) {
     return {
       restrict: 'E',
--- a/app/scripts/filters/util.js
+++ b/app/scripts/filters/util.js
@@ -356,6 +356,12 @@ angular.module('openshiftConsole')
       return Navigate.resourceURL(resource, kind, namespace, null, {apiVersion: apiVersion});
     };
   })
+  // Resource must be the resource object itself, it can NOT be a name.
+  .filter('navigateToTabURL', function (Navigate) {
+    return function(resource, tab) {
+      return Navigate.resourceURL(resource, null, null, null, {tab: tab});
+    };
+  })
   .filter('configURLForResource', function(Navigate) {
     return function(resource, /* optional */ action) {
       return Navigate.configURLForResource(resource, action);
--- a/app/scripts/services/navigate.js
+++ b/app/scripts/services/navigate.js
@@ -258,6 +258,11 @@ angular.module("openshiftConsole")
               group: rgv.group
             });
         }
+
+        if (_.get(opts, "tab")) {
+          url.setSearch("tab", opts.tab);
+        }
+
         return url.toString();
       },
 
--- a/app/views/_init-containers-summary.html
+++ b/app/views/_init-containers-summary.html
@@ -1,0 +1,7 @@
+<span ng-if="$ctrl.podTemplate.spec.initContainers.length" class="text-muted small">
+  <ng-pluralize class="mar-right-sm"
+                count="$ctrl.podTemplate.spec.initContainers.length"
+                when="{'1': '&nbsp;{} init container','other': '&nbsp;{} init containers'}">
+  </ng-pluralize>
+  <a ng-href="{{$ctrl.apiObject | navigateToTabURL:$ctrl.tab}}">View Details</a>
+</span>
--- a/app/views/overview/_list-row-expanded.html
+++ b/app/views/overview/_list-row-expanded.html
@@ -17,6 +17,8 @@
             builds="row.state.builds"
             class="hide-ng-leave">
           </pod-template>
+
+          <init-containers-summary api-object="row.apiObject"></init-containers-summary>
         </div><!-- /overview-pod-template -->
 
       <div class="overview-animation-block" ng-class="{
@@ -119,6 +121,8 @@
               pod-template="row.current | podTemplate"
               images-by-docker-reference="row.state.imagesByDockerReference"
               builds="row.state.builds"></pod-template>
+
+            <init-containers-summary api-object="row.apiObject"></init-containers-summary>
           </uib-tab>
           <uib-tab ng-if="row.current && row.state.showMetrics && row.state.breakpoint === 'xxs'" active="row.selectedTab.metrics">
             <uib-tab-heading>Metrics</uib-tab-heading>
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -1608,7 +1608,7 @@ kind: t,
 group: p.group
 });
 }
-return c.toString();
+return _.get(r, "tab") && c.setSearch("tab", r.tab), c.toString();
 },
 toResourceURL: function(t) {
 e.url(this.resourceURL(t));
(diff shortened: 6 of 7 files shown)


cdcabrera:
@spadgett still WIP, yet to implement the "tab" parameter referenced in #1790 for the navigation filter / service

cdcabrera:
@spadgett we have a couple of quick ways to add tab filtering into the existing use of ```navigateResourceURL```

### We could... 
Add it in like so... not necessarily fantastic typing out all the parameters
```javascript
{{pod | navigateResourceURL:null:null:null:'details'}}
```

Or simply place a $scope method as a wrapper to the filter... 
```javascript
{{ addInTabsMethod(pod) }}
```
connected to
```javascript
.filter('navigateResourceURL', function(Navigate) {
  return function(resource, kind, namespace, apiVersion, tab) {
    return Navigate.resourceURL(resource, kind, namespace, null, {apiVersion: apiVersion, tab:tab});
  };
})
```


### Or we could...

Aim for an "additional" filter that doesn't interfere with the existing setup at all...
```javascript
{{pod | navigateResourceURL | openTab:'details'}}
```

spadgett:
How about something like...

```
{{pod | navigateToTab : 'logs'}}
```

We have a need for this in a lot of places (often for logs). If we only support passing an object to the filter, we don't need most of the params.

spadgett:
`navigateToTabURL` might be better

cdcabrera:
@spadgett tab linking now works, per request the filter is called ```navigateToTabURL ```

app/scripts/filters/util.js
```diff
@@ -356,6 +356,11 @@ angular.module('openshiftConsole')
       return Navigate.resourceURL(resource, kind, namespace, null, {apiVersion: apiVersion});
     };
   })
+  .filter('navigateToTabURL', function (Navigate) {
+    return function(resource, tab) {
```
spadgett: Worth a comment here that `resource` must be the the API object itself since it can be either the object or name using the other `navigateResourceURL` filter

app/scripts/services/navigate.js
```diff
@@ -258,6 +258,11 @@ angular.module("openshiftConsole")
               group: rgv.group
             });
         }
+
+        if (_.get(opts, "tab")) {
+          url.removeSearch("tab").addSearch("tab", opts.tab);
```
spadgett: Is

```js
url.setSearch("tab", opts.tab);
```

sufficient?

app/views/_init-count.html
```diff
@@ -0,0 +1,11 @@
+<!--
+  Expects the following variables:
+  podTemplate
```
spadgett: The comment looks wrong. (I'm OK just taking the comment out.)

app/scripts/directives/resources.js
```diff
@@ -1,6 +1,15 @@
 'use strict';
 
 angular.module('openshiftConsole')
+  .directive('initCount', function() {
+    return {
+      restrict: 'E',
+      scope: {
+        pod: '='
```
spadgett: Call it `podTemplate`. It's not always a pod.

app/scripts/directives/resources.js
```diff
@@ -1,6 +1,15 @@
 'use strict';
 
 angular.module('openshiftConsole')
+  .directive('initCount', function() {
```
spadgett: Something like `initContainersSummary` is a little clearer

app/views/_init-count.html
```diff
@@ -0,0 +1,11 @@
...
+ -->
+
+<span ng-if="pod.spec.initContainers.length" class="text-muted small">
+  <ng-pluralize count="pod.spec.initContainers.length"
+                when="{'1': '&nbsp;Init container','other': '&nbsp;{} Init containers'}">
```
spadgett: Let's always show a count, even if one like `1 Init Container`.

For case: either `<count> init containers` or `<count> Init Containers`? (The way you have it, `2 Init containers`, looks odd to me.)

app/scripts/directives/resources.js
spadgett: Do you mind making this an Angular component instead of directive? We've switched to components for new code.

app/views/_init-count.html
```diff
@@ -0,0 +1,11 @@
...
+  </ng-pluralize>
+  <a ng-href="{{pod | navigateToTabURL:'details'}}">View Details</a>
```
spadgett: `details` is not always the right tab depending on what kind of resource it is :/  Some resources need `configuration`

`pod` is not a pod, but a pod template. This link will only work for pods not other types like deployment configs.

cdcabrera: Gotcha

app/scripts/services/navigate.js
cdcabrera: Had trouble with just using ```.setSearch``` overriding all search parameters, we'll double check though

spadgett: Ah OK, this works then.

app/views/_init-count.html
spadgett: For this one, you'll have to take in the entire object, not just the pod template, to create the link. So I'd suggest changing the attribute to `apiObject` / `api-object` and the call the `podTemplate` filter from within the view to get the template.

cdcabrera: @spadgett need some input about ```details``` not being the right tab. 

Can place a check/condition that alters the tab between ```details``` and ```configuration```, but unsure of what property to check against?

spadgett: You need to pass `apiObject` in from the list row and pass that to the navigateToTabURL filter. Then check `apiObject.kind` to see what tab it should be. For `DeploymentConfig` and `Deployment`, you should use `configuration`. (I think, need to double check.)

app/scripts/services/navigate.js
cdcabrera: annnnnnd of course it works now, updated to ```setSearch```

spadgett: OK just confirming it doesn't remove any of query params if set?

cdcabrera: It doesn't appear to. That's what it initially appeared like on testing. Appears as if you have to be careful when testing it in the console since it doesn't give the appearance of updating the returned object used by URI js, that is until it's expanded. Based on that understanding went back through and did a few tests involving...
- replacing an existing "tab" query 
- adding a "tab" parameter in addition to an existing querystring
- and adding, removing, then adding the "tab" query back

they came back as expected.

cdcabrera:
[test]

app/views/_init-containers-summary.html
```diff
@@ -0,0 +1,6 @@
+<span ng-if="$ctrl.podTemplate.spec.initContainers.length" class="text-muted small">
+  <ng-pluralize count="$ctrl.podTemplate.spec.initContainers.length"
+                when="{'1': '&nbsp;{} Init container','other': '&nbsp;{} Init containers'}">
```
spadgett: Can you fix the case in the messages here?

app/views/overview/_list-row-expanded.html
```diff
@@ -17,6 +17,8 @@
             builds="row.state.builds"
             class="hide-ng-leave">
           </pod-template>
+
+          <init-containers-summary api-object="row.current"></init-containers-summary>
```
spadgett:
```html
<init-containers-summary api-object="row.apiObject"></init-containers-summary>
```

app/views/overview/_list-row-expanded.html
```diff
@@ -119,6 +121,8 @@ <h4 class="h5">Average Usage <small>Last 15 Minutes</small></h4>
               pod-template="row.current | podTemplate"
               images-by-docker-reference="row.state.imagesByDockerReference"
               builds="row.state.builds"></pod-template>
+
+            <init-containers-summary api-object="row.current"></init-containers-summary>
```
spadgett:
```html
<init-containers-summary api-object="row.apiObject"></init-containers-summary>
```

app/views/_init-containers-summary.html
```diff
@@ -0,0 +1,6 @@
+<span ng-if="$ctrl.podTemplate.spec.initContainers.length" class="text-muted small">
+  <ng-pluralize count="$ctrl.podTemplate.spec.initContainers.length"
```
spadgett: Nit: can you add something like `class="mar-sm-right"` or `mar-xs-right` so there's a small amount of margin between the text and link?

spadgett:
[merge]

PR #1823, opened 2017-07-11
Consistent add storage links

spadgett:
- Use "Add Storage" instead of "Add Storage to ..." for deployment
  configs to be consistent with other pages and other actions like edit
  health checks and set resource limits
- Show the name of the object being edited on the add storage and add
  config files forms to be consistent with other editors

Feedback from @ncameronbritt. Let me know if you agree with these changes.

### DIFF
--- a/app/views/add-config-volume.html
+++ b/app/views/add-config-volume.html
@@ -33,7 +33,7 @@
                   </div>
 
                   <div ng-if="configMaps.length || secrets.length || ('configmaps' | canI : 'create') || ('secrets' | canI : 'create')" class="mar-top-xl">
-                    <h1>Add Config Files</h1>
+                    <h1>Add Config Files to {{name}}</h1>
                     <div class="help-block">
                       Add values from a config map or secret as volume. This will make the data available as files for {{kind | humanizeKind}} {{name}}.
                     </div>
--- a/app/views/attach-pvc.html
+++ b/app/views/attach-pvc.html
@@ -36,7 +36,7 @@
                 </div>
 
                 <div ng-show="pvcs && pvcs.length && attach.resource" class="mar-top-xl">
-                  <h1>Add Storage</h1>
+                  <h1>Add Storage to {{name}}</h1>
                   <div class="help-block">
                     Add an existing persistent volume claim to the template of {{kind | humanizeKind}} {{name}}.
                   </div>
--- a/app/views/browse/_replica-set-details.html
+++ b/app/views/browse/_replica-set-details.html
@@ -128,9 +128,9 @@
         <div ng-if="deploymentConfigName">
           <volumes volumes="replicaSet.spec.template.spec.volumes" namespace="project.metadata.name"></volumes>
           <div ng-if="'deploymentconfigs' | canI : 'update'">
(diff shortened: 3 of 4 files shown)


spadgett:
openshift web console 2017-07-11 12-17-52

openshift web console 2017-07-11 12-18-10

app/views/add-config-volume.html
```diff
@@ -33,7 +33,7 @@ <h2 class="text-center">No config maps or secrets.</h2>
                   </div>
 
                   <div ng-if="configMaps.length || secrets.length || ('configmaps' | canI : 'create') || ('secrets' | canI : 'create')" class="mar-top-xl">
-                    <h1>Add Config Files</h1>
+                    <h1>Add Config Files: {{name}}</h1>
```
jwforres: actually not sure I like this, which editors were we doing this already?

ncameronbritt:
I don't have a really strong opinion about "Add Storage" vs. "Add Storage to..." The question I was asking in my review was, is that overly pedantic?

With binding we do call out the thing you're creating the binding for. Not sure about the other editors. Again, no really strong opinion about adding the {{name}}. 

The drastic change of context is a bigger usability concern, imho. But I also recognize that it's not something with a quick and easy fix.

spadgett:
> I don't have a really strong opinion about "Add Storage" vs. "Add Storage to..." The question I was asking in my review was, is that overly pedantic?

That's fair. I guess my point is we don't do that for other actions that work on the deployment config (edit health checks, edit limits, add autoscaler).

spadgett:
@jwforres 

> actually not sure I like this, which editors were we doing this already?

Add storage is the only page where we don't have the name somewhere in the header.

@ncameronbritt Inconsistent breadcrumbs from the header in most of these :/

openshift web console 2017-07-11 14-31-00

openshift web console 2017-07-11 14-32-31

Different header text in these, but they have the name.

openshift web console 2017-07-11 14-32-57

openshift web console 2017-07-11 14-33-52

spadgett:
openshift web console 2017-07-11 14-56-43

spadgett:
> The drastic change of context is a bigger usability concern, imho. But I also recognize that it's not something with a quick and easy fix.

Right. Trying to get some band aid fixes in until we can make the bigger changes.

spadgett:
So circling back on this: what should we do? I don't have a strong opinion except I'd like the editor to make it more obvious what resource you are editing. Right now the add storage page doesn't say except in the breadcrumbs, which is easy to miss.

jwforres:
"Add storage to mariadb" would make more sense to me but I'm not strongly opposed to what you have already.

spadgett:
I updated it to `Add Storage to <name>`

jwforres:
LGTM but looks like dist needs updating

jwforres:
actually looks like the spec tests bombed

spadgett:
Looks like it might be a flake?

jwforres:
[merge]

spadgett:
Removing merge tag until I can rebase.

spadgett:
[merge][severity: bug]

PR #1830, opened 2017-07-11
Update mini donut total in `$evalAsync` block

spadgett:
Since the function is called in response to chart `onrendered`, it's
possible for it to execute outside of a digest loop. This causes the UI
to display the wrong number of pods beside the mini donut until the next
digest loop.


spadgett:
@jwforres I think we should consider for 3.6.

jwforres:
[merge][severity:blocker]

PR #1832, opened 2017-07-12
Bug 1470010 - Use owner references to find deployment on replica set page

spadgett:
- Avoids a problem where the label selector from another deployment might match a replica set, showing the wrong deployment link on the browse replica set page.
- Fixes a bug where an old replica set that becomes active while the user is on the page cannot be scaled. (This can happen if an old replica set revision is reused because it matches the current deployment spec.)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1470010


jwforres:
[merge][severity:blocker]

PR #1835, opened 2017-07-12
Don't fade "0 pods" text in mini donut

spadgett:
The 0 pods text in the overview mini donut is too light to be legible with the default `disabled-link` styles.


spadgett:
openshift web console 2017-07-12 12-19-17

jwforres:
[merge][severity:blocker]

PR #1839, opened 2017-07-12
Edit the YAML of deployments in the `apps` group

spadgett:


app/scripts/filters/canI.js
```diff
@@ -17,6 +17,7 @@ angular
       ],
       'deployments': [
         {group: 'autoscaling', resource: 'horizontalpodautoscalers', verbs: ['create', 'update']},
+        {group: 'apps',        resource: 'deployments',              verbs: ['update']},
         {group: 'extensions',  resource: 'deployments',              verbs: ['create', 'update']}
```
spadgett: Looking at this, I don't think "create" should be here.

spadgett:
[merge][severity: blocker]

PR #1843, opened 2017-07-13
Fix missing builds with no build config

jwforres:


app/views/builds.html
```diff
@@ -135,6 +131,51 @@
                       </span>
                     </td>
                   </tr>
+
+                  <!-- Builds that were not spawned from a build config -->
+                  <tr ng-repeat="latestBuild in buildsNoConfig">
```
spadgett: I know this was copied from above, but should use `track by (latestBuild | uid)`

jwforres:
@spadgett updated track by on both ng-repeats

app/views/builds.html
```diff
@@ -49,11 +49,11 @@
...
                   </tr>
                 </thead>
-                <tbody ng-if="!(latestByConfig | hashSize)">
+                <tbody ng-if="!(latestByConfig | hashSize) && !(buildsNoConfig | hashSize)">
                   <tr><td colspan="7"><em>{{emptyMessage}}</em></td></tr>
                 </tbody>
-                <tbody ng-if="(latestByConfig | hashSize)">
-                  <tr ng-repeat="(buildConfigName, latestBuild) in latestByConfig">
+                <tbody ng-if="(latestByConfig | hashSize) || (buildsNoConfig | hashSize)">
+                  <tr ng-repeat="(buildConfigName, latestBuild) in latestByConfig track by (latestBuild | uid)" ng-if="buildConfigName !== ''">
```
spadgett: Nit: might be better as `track by buildConfigName` since the row is for the BC?

jwforres: hmm good point, do we actually even need track by in this case then? will ng-repeat track by the key for a hash? looking...

jwforres:
fixed [merge][severity:blocker]

PR #1847, opened 2017-07-14
Bug 1470976 - Support number input on iOS

spadgett:
For the iOS keyboard to prompt for numbers, the inputs must have a `pattern` attribute of digits.

screen shot 2017-07-14 at 9 14 33 am

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1470976

- [x] Test number inputs in IE

### DIFF
--- a/app/views/_compute-resource.html
+++ b/app/views/_compute-resource.html
@@ -9,7 +9,9 @@
                ng-attr-id="{{id}}"
                ng-model="input.amount"
                min="0"
+               pattern="\d*"
                ng-attr-placeholder="{{placeholder}}"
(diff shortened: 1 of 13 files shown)


spadgett:
> did you double check IE/Edge at all, since we have some funky issues with number inputs there already

I'll do some testing. I hope it fixes those issues actually.

spadgett:
@jwforres It's marginally better in IE 11. It doesn't show an error, but now at least it clears the invalid value when you blur the input.

Edge works fine.

spadgett:
[merge]

PR #1850, opened 2017-07-14
Make tags wider in image stream listing table

### LINKED ISSUE #1848: Image stream tags cropped too short
In building images for origin inside origin, I need to create image tags that have the name of the origin image. However, when displaying them in the web console, the tags are cropped too small:
image

spadgett:
Doubles the width of the tags in the table. There is a `title` attribute if it still gets clipped.

Fixes #1848

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -217,14 +217,20 @@ registry-imagestream-push {
   margin-top: @grid-gutter-width;
 }
 
-/* No expandable listing when shown in the Openshift Web Console */
-registry-imagestream-listing .listing-ct-toggle {
-  display: none;
+registry-imagestream-listing {
+  /* No expandable listing when shown in the Openshift Web Console */
+  .listing-ct-toggle {
+    display: none;
+  }
+  .registry-image-tag {
+    max-width: 115px;
+  }
+  table {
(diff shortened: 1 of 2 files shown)


spadgett:
Before:

openshift web console 2017-07-14 12-31-11

After:

openshift web console 2017-07-14 12-31-34

jwforres:
how is that behaving at mobile? i know that is where we had some issues related to those tags

jwforres:
I also don't love how it looks at that width :-\  most tags are actually short

jwforres:
maybe somewhere in the middle?

spadgett:
Here is somewhere in between:

openshift web console 2017-07-14 13-25-02

Really I think the table column should grow/shrink to fit the content, but that change probably needs to happen upstream in registry-image-widths

spadgett:
> how is that behaving at mobile? i know that is where we had some issues related to those tags

Not great. I'm not sure if we can fix that without an upstream fix though.

openshift web console 2017-07-14 13-27-19

jwforres:
[merge][severity:blocker]

spadgett:
[merge][severity: blocker]

PR #1855, opened 2017-07-16
autocapitalize="off" -> autocapitalize="none"

### LINKED ISSUE #1854: autocapitalize="off" should be autocapitalize="none"
`autocapitalize="off"` is deprecated.

See

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input
https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html#//apple_ref/doc/uid/TP40008058-autocapitalize

spadgett:
Closes #1854

### DIFF
--- a/app/views/_templateopt.html
+++ b/app/views/_templateopt.html
@@ -38,7 +38,7 @@
           ng-blur="cleared = false"
           ng-trim="false"
           autocorrect="off"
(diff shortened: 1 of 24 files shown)


spadgett:
[merge]

PR #1864, opened 2017-07-20
Lock firefox version

jwforres:


jwforres:
@spadgett going to try 52.2 to start with since that is working in jenkins, if that won't work i'll see if it will let me go back to 31.0 which is the last one that worked

jwforres:
We also have an option to lock to version aliases https://docs.travis-ci.com/user/firefox#Version-aliases

Maybe we should lock to latest-esr since that would be what RHEL is typically using?

spadgett:
+1 to `latest-esr` if it works

jwforres:
it couldn't resolve 52.2 correctly so going to try latest-esr

jwforres:
well no luck, its not the firefox version causing the problem.  I've tried "latest" which is 54, "latest-esr" which is 52, and 31.0 which is what it was using before the update.  There is something else going on here that happens to have coincided with firefox getting updated.

jwforres:
attempted to switch to PhantomJS but it seems to just get stuck and not actually run the tests, will debug more tmrw

jwforres:
FWIW i confirmed that there was a *significant* update to the images being used in travis, it went from ubuntu 12 to ubuntu 14

spadgett:
#1872 switches to PhantomJS (at least for now), which makes our tests pass again

PR #1872, opened 2017-07-24
Use PhantomJS for Travis tests

spadgett:
Fixes a problem where our tests fail in Firefox after Travis switched to Ubuntu Trusty. Also removes the node_modules and bower_components cache since that has caused problems in the past.

See also #1864

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,8 @@
 language: node_js
 node_js:
   - "4"
-cache:
-  directories:
-  - node_modules
-  - bower_components
 before_script:
   - make build
 script:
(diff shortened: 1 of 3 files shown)


spadgett:
https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming

benjaminapetersen:
We can always look at [nightmare](https://github.com/segmentio/nightmare) / [nightmare](http://www.nightmarejs.org/) (corny name?) over phantom, it claims to be a lot faster, though is still not a real browser.

spadgett:
[merge][severity: bug]

PR #1877, opened 2017-07-25
YAML ace editor fix

rhamilto:
Fixes bug where ace-editor for yaml didn't have a height inside of an overlay/panel

### DIFF
--- a/app/styles/_editor.less
+++ b/app/styles/_editor.less
@@ -10,6 +10,10 @@
   }
 }
 
+.ace_editor.editor {
+  height: 350px;
+}
+
 .ace-inline {
   height: 300px;
 }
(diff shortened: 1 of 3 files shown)


app/styles/_editor.less
```diff
@@ -50,7 +50,9 @@
   text-align: right;
 }
 
-#from-file .editor {
-  height: 350px;
+#from-file,
+#yaml-file {
```
spadgett: It looks like this rule with apply to Edit YAML action, not just Import YAML? Edit YAML currently grows with the page.

rhamilto:
> It looks like this rule with apply to Edit YAML action, not just Import YAML? Edit YAML currently grows with the page.

Yeah.  And Add to Project > From File also still has no height.  :/  Will fix afternoon.

rhamilto:
This should do it, @spadgett

spadgett:
Thanks @rhamilto 

[merge][severity: blocker]

PR #1880, opened 2017-07-25
Fix edit YAML

spadgett:
The wrong object was passed to DataService.update as part of the refactoring in #1798


spadgett:
Thanks @jeff-phillips-18 

[merge][severity: blocker]

PR #1882, opened 2017-07-26
Skip digest when watching events

spadgett:
The events directive already debounces updates because it filters and
sorts. Pass the `skipDigest` flag to `DataService` to avoid unnecessary
Angular digest loops.


benjaminapetersen:
[merge]

PR #1884, opened 2017-07-27
Bump origin-web-common to 0.0.34, patternfly to 3.26.0, angular-patternfly to 4.2.0

spadgett:
cc @rhamilto @benjaminapetersen


spadgett:
Thanks, will update. That catalog changes are only the patternfly bump, but you're right: I should mention it.

spadgett:
Blocker since it containers @jeff-phillips-18's fix for the broken dialog next buttons.

[merge][severity: blocker]

PR #1887, opened 2017-07-27
Fix runtime error on deploy image page

spadgett:
The ng-form was not correctly initialized, resulting in a runtime error.

Some whitespace changes in the view.

### DIFF
--- a/app/views/directives/deploy-image.html
+++ b/app/views/directives/deploy-image.html
@@ -2,7 +2,7 @@
   <p>
     Deploy an existing image from an image stream tag or docker pull spec.
   </p>
-  <ng-form name="forms.imageSelection">
+  <ng-form name="imageSelection">
     <fieldset ng-disabled="loading">
       <div class="radio">
         <label>
@@ -62,6 +62,7 @@
     <div ng-if="!loading" class="h2">Select an image stream tag or enter an image name.</div>
     <div ng-if="loading" class="h2">Loading image metadata for <span class="word-break">{{imageName | stripSHA}}</span>...</div>
   </div>
+
   <div class="row mar-bottom-md" ng-if-start="!loading && import.image">
     <div class="col-sm-12 mar-top-lg mar-bottom-lg">
       <div class="separator"></div>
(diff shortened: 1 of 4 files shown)


spadgett:
To see the diff without whitespace:

https://github.com/openshift/origin-web-console/pull/1887/files?w=1

app/views/directives/deploy-image.html
```diff
@@ -62,137 +62,139 @@
...
-    <div class="col-sm-2 hidden-xs text-right h2">
-      <span class="fa fa-cube text-muted" style="font-size: 100px;" aria-hidden="true"></span>
-    </div>
-    <div class="col-sm-10">
-      <div ng-if="runsAsRoot" class="alert alert-warning">
-        <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
-        Image <strong>{{import.name}}</strong> runs as the
-        <strong>root</strong> user which might not be permitted by your cluster
-        administrator.
+  <div ng-show="!loading && import.image">
```
spadgett: Changed this to an `ng-show` so that the "Deploy" button isn't enabled in the dialog before the form below is displayed.

spadgett:
@benjaminapetersen Definitely would like some help validating the fix! Note this is the 3.6 branch, so you'll need to clean install deps.

benjaminapetersen:
@spadgett ok will do. pulling it down now.

spadgett:
[test]

spadgett:
@rhamilto

spadgett:
Looks like the `ng-show` change breaks test automation.

spadgett:
Hold on review, I'll need to look at this some more due to the test failure.

benjaminapetersen:
This is the error I see, same as the one you found?
screen shot 2017-07-27 at 3 47 58 pm

spadgett:
Yes, that's the one.

benjaminapetersen:
This change does fix the bug, though I haven't figured out why `$scope.forms` didn't work.

spadgett:
@benjaminapetersen Best as I can tell, dots in the name aren't working with `ng-form` (although I'm not sure why).

Pushed an update that hopefully fixes the test failure. It also gets rid of the big diff in the view.

spadgett:
Blocker due to the runtime error.

[merge][severity: blocker]

Issue #1889, opened 2017-07-28
Upgrade grunt-ng-annotate


spadgett:
We should upgrade grunt-ng-annotate. Our current version is very old and does not work with Angular components.

spadgett:
It looks like ng-annotate is no longer maintained, which is concerning.

https://www.npmjs.com/package/ng-annotate

@jeff-phillips-18 @benjaminapetersen FYI

benjaminapetersen:
Hmm. Yeah, this is problematic. ☹️ 
It doesn't even seem like the babel plugin that replaces it will make feature parity.

benjaminapetersen:
Since we are not using babel, we can't migrate to the babel plugin.  The correct path forward then is likely to drop the dependency & update to use the `['dep', function(dep) {}]` format.

Do we want to consider other options, or should I update this issue?

PR #1890, opened 2017-07-28
Bug 1476157 - Fix minification problems due to dependency injection

spadgett:
Our version of ngAnnotate does not handle components, which causes
runtime errors in the built binary due to a recent component. Switch the
component to use the array notation for dependency injection and put it
in its own file.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1476157
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1476216

Opened tech debt issue #1889 for upgrading grunt-ng-annotate.


spadgett:
[merge][severity: blocker]

PR #1892, opened 2017-07-28
Update stepChanged listener for angular-patternfly 4.2.0

spadgett:
Pass `on-step-changed` to the `pf-wizard` component rather than using `$scope.$on('wizard:stepChanged')`, which no longer works. Fixes the button label on the last page of the deploy image and import YAML dialogs.

cc @rhamilto


spadgett:
[test]

spadgett:
[merge][severity: blocker]

PR #1901, opened 2017-08-02
Edit/Create Projects on page in a popup

jeff-phillips-18:
https://trello.com/c/djoHeMob

Requires https://github.com/openshift/origin-web-common/pull/135

### DIFF
--- a/app/scripts/controllers/projects.js
+++ b/app/scripts/controllers/projects.js
@@ -109,6 +109,35 @@ angular.module('openshiftConsole')
       onSortChange: update
     };
 
+    $scope.newProjectPanelShown = false;
+
+    $scope.createProject = function() {
+      $scope.newProjectPanelShown = true;
+    };
+
+    $scope.closeNewProjectPanel = function() {
+      $scope.newProjectPanelShown = false;
+    };
+
+    $scope.onNewProject = function() {
+      $scope.newProjectPanelShown = false;
+    };
+
+    $scope.editProjectPanelShown = false;
+
+    $scope.editProject = function(project) {
+      $scope.editingProject = project;
+      $scope.editProjectPanelShown = true;
+    };
+
+    $scope.closeEditProjectPanel = function() {
+      $scope.editProjectPanelShown = false;
+    };
+
+    $scope.onEditProject = function() {
+      $scope.editProjectPanelShown = false;
+    };
+
     $scope.$watch('search.text', _.debounce(function(searchText) {
       $scope.keywords = filterKeywords = KeywordService.generateKeywords(searchText);
       $scope.$apply(filterProjects);
--- a/app/styles/_buttons.less
+++ b/app/styles/_buttons.less
@@ -41,3 +41,7 @@
     text-decoration: none;
   }
 }
+
+.icon-button-text {
+  padding-left: 4px;
+}
--- a/app/views/projects.html
+++ b/app/views/projects.html
@@ -25,9 +25,13 @@
                         <h1>My Projects</h1>
                         <div class="projects-options">
                           <div class="projects-add" ng-if="canCreate">
-                            <a href="create-project" class="btn btn-md btn-primary">
-                              Create Project
-                            </a>
+                            <button ng-click="createProject()" class="btn btn-md btn-primary">
+                              <span class="fa fa-plus" aria-hidden="true"></span>
+                              <span class="icon-button-text">Create Project</span>
+                            </button>
+                            <origin-modal-popup shown="newProjectPanelShown" modal-title="Create Project" on-close="closeNewProjectPanel">
+                              <create-project is-dialog="true" redirect-action="onNewProject" on-cancel="closeNewProjectPanel"></create-project>
+                            </origin-modal-popup>
                           </div>
                           <div class="projects-search">
                             <form role="form" class="search-pf has-button">
@@ -101,7 +105,7 @@
                                 </a>
                               </li>
                               <li role="menuitem">
-                                <a ng-href="project/{{project.metadata.name}}/edit?then=./">
+                                <a href="" ng-click="editProject(project)">
                                   Edit Project
                                 </a>
                               </li>
@@ -120,6 +124,11 @@
                             </ul>
                           </div>
                         </div>
+                        <origin-modal-popup shown="editProjectPanelShown && editingProject === project"
+                                            ng-class="{'with-description': (project | description | size)}"
+                                            modal-title="Edit Project" on-close="closeEditProjectPanel">
+                          <edit-project project="project" is-dialog="true" redirect-action="onEditProject" on-cancel="closeEditProjectPanel"></edit-project>
+                        </origin-modal-popup>
                       </div>
                     </div>
                     <p class="projects-instructions" ng-if="canCreate === false" ng-include="'views/_cannot-create-project.html'"></p>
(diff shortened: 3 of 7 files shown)


jeff-phillips-18:
I have not removed the create project and/or edit project pages. Not sure if we want them removed as part of this PR or not.

app/views/projects.html
```diff
@@ -25,9 +25,13 @@ <h2 class="text-center">Loading...</h2>
                         <h1>My Projects</h1>
                         <div class="projects-options">
                           <div class="projects-add" ng-if="canCreate">
-                            <a href="create-project" class="btn btn-md btn-primary">
+                            <button ng-click="createProject()" class="btn btn-md btn-primary">
+                              <span class="fa fa-plus"></span>
                               Create Project
-                            </a>
+                            </button>
+                            <origin-modal-popup shown="showNewProjectPanel" modal-title="Create Project" on-close="closeNewProjectPanel()">
```
spadgett: As much as possible, I'd like to be consistent in how we handle callbacks for our components. It looks like `shown` and `on-close` are handled differently here (as well as `on-cancel` for `create-project`).

app/views/projects.html
```diff
@@ -101,7 +105,7 @@ <h2 class="h1">
                                 </a>
                               </li>
                               <li role="menuitem">
-                                <a ng-href="project/{{project.metadata.name}}/edit?then=./">
+                                <a ng-href="#" ng-click="editProject(project)">
```
spadgett:
```
<a href="" ng-click="editProject(project)">
```

app/scripts/controllers/projects.js
```diff
@@ -109,6 +109,35 @@ angular.module('openshiftConsole')
...
+    };
+
+    $scope.onNewProject = function() {
+      $scope.showNewProjectPanel = false;
+    };
+
+    $scope.showEditProjectPanel = false;
+
+    $scope.editProject = function(project) {
+      $scope.edittingProject = project;
```
spadgett: `editingProject`

app/views/projects.html
jeff-phillips-18: I could change either of the components but that would need to be done there and not here. Most of the components bind functions using < or =, I used & for the new origin-modal-popup component. Should I change that?

app/views/projects.html
jeff-phillips-18: fixed

app/scripts/controllers/projects.js
jeff-phillips-18: fixed

app/views/projects.html
spadgett: Right, understood. I suggest we should use `<` or `=` like the other components. I'd at least like to be consistent with callbacks inside the same component. Here `shown` is `<` and `onClose` is `&` in the same component.

Sorry, I missed this reviewing https://github.com/openshift/origin-web-common/pull/135

(As an aside, I find callbacks in Angular components generally weird and confusing...)

spadgett: Wondering if `onShown` is not better as well for consistent naming.

jeff-phillips-18: shown is a flag not a function.

jeff-phillips-18: I can rename the controller variable to reduce confusion.

spadgett: Ah OK please disregard then :)

jeff-phillips-18: Fixed and renamed the variables.

spadgett:
> I have not removed the create project and/or edit project pages. Not sure if we want them removed as part of this PR or not.

If we remove them, we might think about redirects for the old URLs (maybe just /create-project) in case someone has a link / bookmark. Not too worried about it for this PR.

app/views/projects.html
```diff
@@ -25,9 +25,13 @@ <h2 class="text-center">Loading...</h2>
                         <h1>My Projects</h1>
                         <div class="projects-options">
                           <div class="projects-add" ng-if="canCreate">
-                            <a href="create-project" class="btn btn-md btn-primary">
+                            <button ng-click="createProject()" class="btn btn-md btn-primary">
+                              <span class="fa fa-plus"></span>
```
spadgett: `aria-hidden="true"`

app/views/projects.html
```diff
@@ -101,7 +105,7 @@ <h2 class="h1">
                                 </a>
                               </li>
                               <li role="menuitem">
-                                <a ng-href="project/{{project.metadata.name}}/edit?then=./">
+                                <a ng-href="" ng-click="editProject(project)">
```
spadgett: nit: `href=""`

spadgett:
If we remove the create project page, we'll also need to update our e2e tests.

https://github.com/spadgett/origin-web-console/blob/f6611831bc6b9ba10b0b7a7c8b90e66e7ddc940b/test/integration/page-objects/createProject.js

jeff-phillips-18:
Created issue #1917 to remove the pages.

jeff-phillips-18:
Removed WIP status now that we have updated the version of origin-web-common.
Ready for final reviews/merge.

spadgett:
Hey @jeff-phillips-18. I pulled the changes down, and everything works well. Thanks.

A couple of small style questions for @openshift/team-ux-review. I won't block the PR on any of these, particularly if they require origin-web-common changes. Just small things I noticed.

1. Should there be any margin between the `+` and the text on the button? I'm not sure if there is a standard style for this in Patternfly.

No margin (current):

openshift web console 2017-08-08 08-15-56

2px margin:

openshift web console 2017-08-08 08-17-48

2. Should we remove `input-lg` and `btn-lg` styles now that the form is only in a dialog?

`input-lg` (current):

openshift web console 2017-08-08 08-18-59

Without `input-lg`:

openshift web console 2017-08-08 08-20-42

3. I noticed the dialog is spaced farther from the kebab than the actions dropdown and feels a little too far away. I assume the spacing should be about the same.

spadgett:
Added Trello card https://trello.com/c/djoHeMob for QE testing.

spadgett:
[test] to make sure this doesn't break our e2e tests

spadgett:
It looks like we do have padding on the button in the landing page, so we should be consistent here so they look the same.

screen shot 2017-08-08 at 9 54 35 am

app/views/projects.html
jeff-phillips-18: Fixed

app/views/projects.html
jeff-phillips-18: Fixed

jeff-phillips-18:
Fixed the create button spacing.

Added issue for input and button sizes: https://github.com/openshift/origin-web-common/issues/142

app/styles/_buttons.less
```diff
@@ -41,3 +41,7 @@
     text-decoration: none;
   }
 }
+
+.icon-button-text {
+  padding-left: 4px;
```
spadgett: 👍  thanks for making this a generic style

app/views/projects.html
```diff
@@ -101,7 +105,7 @@ <h2 class="h1">
                                 </a>
                               </li>
                               <li role="menuitem">
-                                <a ng-href="project/{{project.metadata.name}}/edit?then=./">
+                                <a g-href="" ng-click="editProject(project)">
```
spadgett: typo: `g-href`

jeff-phillips-18: ugh

jeff-phillips-18: Fixed

jeff-phillips-18:
Fixed popup vertical positioning for projects with descriptions.

spadgett:
[merge]

PR #1903, opened 2017-08-02
Don't minify angular-patternfly.js

### LINKED ISSUE #1900: Catalog modals are broken when using dist
screen shot 2017-08-01 at 3 18 32 pm

Known issues:

* button labels are showing angular expressions
* the modals don't actually work

spadgett:
Minifying angular-patternfly.js as part of vendor.js breaks dependency
injection for angular-patternfly. We use ngAnnotate, but that doesn't
work for Angular 1.5 components that don't have the "ngInject" prologue.
Instead use angular-patternfly.min.js as-is.

Fixes #1900


spadgett:
@jeff-phillips-18 Opinion on this change? It's the least invasive fix I could find.

I think we should look at this in angular-patternfly (and origin-web-common as well) since it might be a problem for other downstream projects.

spadgett:
[test]

jeff-phillips-18:
@spadgett This is purely an Angular Patternfly issue. I believe I have a fix in that repo.

spadgett:
Ok thanks Jeff. That sounds a lot cleaner.

PR #52, opened 2016-05-31
Fix login redirect

spadgett:
Fixes #50

@jwforres @liggitt PTAL

Issue #1685, opened 2017-06-12
Test flake: User adds an image stream to a project after creating a new project using the Import YAML tab should process and create the images in the image stream


spadgett:
https://ci.openshift.redhat.com/jenkins/job/test_pull_requests_origin_web_console/1495/console

```
1) User adds an image stream to a project after creating a new project using the Import YAML tab should process and create the images in the image stream
  - Expected false to be true.
      at catalogPage.processImageStream.then (/data/src/github.com/openshift/origin-web-console/test/integration/features/user_adds_imagestream_to_project.spec.js:38:79)
    at /data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/goog/base.js:1582:15
    at webdriver.promise.ControlFlow.runInNewFrame_ (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:1654:20)
    at notify (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:465:12)
    at notifyAll (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:442:7)
    at resolve (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:420:7)
    at fulfill (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:535:5)
    at Object.webdriver.promise.asap (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:776:5)
    at newFrame.onComplete (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:1670:25)
```

spadgett:
Related error in https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_origin_web_console/13/console

1) User adds a template to a project after creating a new project using the Import YAML tab should save the template in the project catalog
  - Error: URL hasn't changed to project/console-test-project-1500489744656/overview(is currently: undefined)
Wait timed out after 5089ms

PR #1906, opened 2017-08-03
Service Instance Details Pages

cdcabrera:
Prep for update around service instances and bindings. Page for service instances.

Related to [Origin Design PR #52](https://github.com/openshift/openshift-origin-design/pull/52)
Dependency on [Origin-Web-Common PR 164](https://github.com/openshift/origin-web-common/pull/164)

aug-17-2017 11-42-16-640

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -272,6 +272,16 @@ angular
         controller: 'ServiceController',
         reloadOnSearch: false
       })
+      .when('/project/:project/browse/service-instances', {
+        templateUrl: 'views/service-instances.html',
+        controller: 'ServiceInstancesController',
+        reloadOnSearch: false
+      })
+      .when('/project/:project/browse/service-instances/:instance', {
+        templateUrl: 'views/browse/service-instance.html',
+        controller: 'ServiceInstanceController',
+        reloadOnSearch: false
+      })
       .when('/project/:project/browse/storage', {
         templateUrl: 'views/storage.html',
         controller: 'StorageController',
--- a/app/scripts/constants.js
+++ b/app/scripts/constants.js
@@ -261,6 +261,18 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
                 "/create-route",
                 "/edit/routes/"
               ]
+            },
+            {
+              label: "Provisioned Services",
+              href: "/browse/service-instances",
+              prefixes: [
+                "/browse/service-instances/"
+              ],
+              canI: {
+                resource: 'serviceinstances',
+                group: 'servicecatalog.k8s.io',
+                verb: 'list'
+              }
             }
           ]
         }
--- a/app/scripts/controllers/serviceInstance.js
+++ b/app/scripts/controllers/serviceInstance.js
@@ -1,0 +1,107 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .controller('ServiceInstanceController', function ($scope,
+                                                     $filter,
+                                                     $routeParams,
+                                                     DataService,
+                                                     ProjectsService,
+                                                     ServiceInstancesService) {
+    $scope.alerts = {};
+    $scope.projectName = $routeParams.project;
+    $scope.serviceInstance = null;
+    $scope.serviceClass = null;
+    $scope.serviceClasses = null;
+
+    $scope.breadcrumbs = [
+      {
+        title: "Provisioned Services",
+        link: "project/" + $routeParams.project + "/browse/service-instances"
+      }
+    ];
+
+    $scope.deprovision = function() {
+      ServiceInstancesService.deprovision($scope.serviceInstance);
+    };
+
+    var watches = [];
+
+    var updateBreadcrumbs = function() {
+      if(!$scope.serviceInstance || !$scope.serviceClasses) {
+        return;
+      }
+
+      $scope.breadcrumbs.push({
+        title: $filter('serviceInstanceDisplayName')($scope.serviceInstance, $scope.serviceClasses)
+      });
+    };
+
+    var updateServiceClassMetadata = function() {
+      if(!$scope.serviceInstance || !$scope.serviceClasses) {
+        return;
+      }
+
+      var serviceClassName = _.get($scope.serviceInstance.spec, 'serviceClassName');
+      $scope.serviceClass = _.get($scope.serviceClasses, [serviceClassName]);
+      $scope.plan = _.find(_.get($scope.serviceClass, 'plans'), {name: $scope.serviceInstance.spec.planName });
+    };
+
+    var serviceResolved = function(service, action) {
+      $scope.loaded = true;
+      $scope.serviceInstance = service;
+
+      if (action === "DELETED") {
+        $scope.alerts["deleted"] = {
+          type: "warning",
+          message: "This provisioned service has been deleted."
+        };
+      }
+
+      updateServiceClassMetadata();
+    };
+
+    ProjectsService
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+        $scope.projectContext = context;
+
+        DataService
+          .get({
+            group: 'servicecatalog.k8s.io',
+            resource: 'serviceinstances'
+          }, $routeParams.instance, context, { errorNotification: false })
+          .then(function(service) {
+
+            serviceResolved(service);
+            updateBreadcrumbs();
+
+            watches.push(DataService.watchObject({
+              group: 'servicecatalog.k8s.io',
+              resource: 'serviceinstances'
+            }, $routeParams.instance, context, serviceResolved));
+
+          }, function(error) {
+            $scope.loaded = true;
+            $scope.alerts["load"] = {
+              type: "error",
+              message: "The service details could not be loaded.",
+              details: $filter('getErrorDetails')(error)
+            };
+          });
+
+        DataService.list({
+          group: 'servicecatalog.k8s.io',
+          resource: 'serviceclasses'
+        }, context, function(serviceClasses) {
+          $scope.serviceClasses = serviceClasses.by('metadata.name');
+          updateServiceClassMetadata();
+          updateBreadcrumbs();
+        });
+
+        $scope.$on('$destroy', function(){
+          DataService.unwatchAll(watches);
+        });
+
+    }));
+  });
--- a/app/scripts/controllers/serviceInstances.js
+++ b/app/scripts/controllers/serviceInstances.js
@@ -1,0 +1,98 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .controller('ServiceInstancesController', function ($scope,
+                                                      $filter,
+                                                      $routeParams,
+                                                      APIService,
+                                                      BindingService,
+                                                      Constants,
+                                                      DataService,
+                                                      LabelFilter,
+                                                      Logger,
+                                                      ProjectsService) {
+    $scope.alerts = {};
+    $scope.bindingsByInstanceRef = {};
+    $scope.emptyMessage = "Loading...";
+    $scope.labelSuggestions = {};
+    $scope.projectName = $routeParams.project;
+    $scope.serviceClasses = {};
+    $scope.serviceInstances = {};
+    $scope.unfilteredServiceInstances = {};
+
+    var watches = [];
+
+    var updateFilter = function() {
+      $scope.serviceInstances = LabelFilter.getLabelSelector().select($scope.unfilteredServiceInstances);
+    };
+
+    var sortServiceInstances = function() {
+      $scope.unfilteredServiceInstances = BindingService.sortServiceInstances($scope.unfilteredServiceInstances, $scope.serviceClasses);
+    };
+
+    ProjectsService
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+        $scope.projectContext = context;
+
+        watches.push(DataService.watch({
+          group: 'servicecatalog.k8s.io',
+          resource: 'serviceinstancecredentials'
+        }, context, function(bindings) {
+          var bindingsByName = bindings.by('metadata.name');
+          $scope.bindingsByInstanceRef = _.groupBy(bindingsByName, 'spec.instanceRef.name');
+        }));
+
+        watches.push(DataService.watch({
+          group: 'servicecatalog.k8s.io',
+          resource: 'serviceinstances'
+        }, context, function(serviceInstances) {
+          $scope.emptyMessage = "No provisioned services to show";
+          $scope.unfilteredServiceInstances = serviceInstances.by('metadata.name');
+
+          sortServiceInstances();
+          updateFilter();
+          updateFilterWarning();
+
+          LabelFilter.addLabelSuggestionsFromResources($scope.unfilteredServiceInstances, $scope.labelSuggestions);
+          LabelFilter.setLabelSuggestions($scope.labelSuggestions);
+
+          Logger.log("provisioned services (subscribe)", $scope.unfilteredServiceInstances);
+        }));
+
+        DataService.list({
+          group: 'servicecatalog.k8s.io',
+          resource: 'serviceclasses'
+        }, context, function(serviceClasses) {
+          $scope.serviceClasses = serviceClasses.by('metadata.name');
+          sortServiceInstances();
+          updateFilter();
+        });
+
+        function updateFilterWarning() {
+          if (!LabelFilter.getLabelSelector().isEmpty() && _.isEmpty($scope.serviceInstances)  && !_.isEmpty($scope.unfilteredServiceInstances)) {
+            $scope.alerts["all-instances-filtered"] = {
+              type: "warning",
+              details: "The active filters are hiding all provisioned services."
+            };
+          }
+          else {
+            delete $scope.alerts["all-instances-filtered"];
+          }
+        }
+
+        LabelFilter.onActiveFiltersChanged(function(labelSelector) {
+          // trigger a digest loop
+          $scope.$evalAsync(function() {
+            $scope.serviceInstances = labelSelector.select($scope.unfilteredServiceInstances);
+            updateFilterWarning();
+          });
+        });
+
+        $scope.$on('$destroy', function(){
+          DataService.unwatchAll(watches);
+        });
+
+      }));
+  });
--- a/app/scripts/filters/canI.js
+++ b/app/scripts/filters/canI.js
@@ -29,6 +29,9 @@ angular
       'imageStreams': [
         {group: '', resource: 'imagestreams', verbs: ['update', 'delete']}
       ],
+      'serviceInstances': [
+        {group: 'servicecatalog.k8s.io', resource: 'serviceinstances', verbs: ['update', 'delete']}
+      ],
       'persistentVolumeClaims': [
         {group: '', resource: 'persistentvolumeclaims', verbs: ['update', 'delete']}
       ],
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -1328,4 +1328,44 @@ angular.module('openshiftConsole')
       var serviceClassDisplayName = _.get(serviceClasses, [serviceClassName, 'externalMetadata', 'displayName']);
       return serviceClassDisplayName || serviceClassName || instanceName;
     };
+  })
+  .filter('serviceInstanceStatus', function(isServiceInstanceReadyFilter) {
+    return function(instance) {
+      var status = 'Pending';
+      var conditions = _.get(instance, 'status.conditions');
+      var instanceError = _.find(conditions, {type: 'Failed', status: 'True'});
+
+      if (instanceError) {
+        status = 'Failed';
+      } else if (isServiceInstanceReadyFilter(instance)) {
+        status = 'Ready';
+      }
+
+      return status;
+    };
+  })
+  .filter('readyConditionMessage', function(statusConditionFilter) {
+    return function(instance) {
+      return _.get(statusConditionFilter(instance, 'Ready'), 'message');
+    };
+  })
+  .filter('failedConditionMessage', function(statusConditionFilter) {
+    return function(instance) {
+      return _.get(statusConditionFilter(instance, 'Failed'), 'message');
+    };
+  })
+  .filter('serviceInstanceConditionMessage', function(serviceInstanceStatusFilter, statusConditionFilter) {
+    return function(instance) {
+      var serviceInstanceStatus = serviceInstanceStatusFilter(instance);
+      var serviceInstanceMessage = null;
+
+      switch(serviceInstanceStatus) {
+        case 'Failed':
+        case 'Ready':
+          serviceInstanceMessage = _.get(statusConditionFilter(instance, serviceInstanceStatus), 'message');
+          break;
+      }
+
+      return serviceInstanceMessage;
+    };
   });
--- a/app/scripts/services/serviceInstances.js
+++ b/app/scripts/services/serviceInstances.js
@@ -1,0 +1,69 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .factory("ServiceInstancesService", function($filter,
+                                               $uibModal,
+                                               DataService,
+                                               NotificationsService) {
+
+    function deprovision(apiObject) {
+      var getErrorDetails = $filter('getErrorDetails');
+      var modalScope = {
+        alerts: {
+          deprovision: {
+            type: 'error',
+            message: 'Service \'' + apiObject.spec.serviceClassName + '\' will be deleted and no longer available.'
+          }
+        },
+        detailsMarkup: 'Delete Service?',
+        okButtonText: 'Delete',
+        okButtonClass: 'btn-danger',
+        cancelButtonText: 'Cancel'
+      };
+
+      // TODO: we probably have to handle bindings in here.
+      // either:
+      // - automatically remove the bindings
+      // - tell the user they must manually unbind before continue
+      return $uibModal.open({
+        animation: true,
+        templateUrl: 'views/modals/confirm.html',
+        controller: 'ConfirmModalController',
+        resolve: {
+          modalConfig: function() {
+            return modalScope;
+          }
+        }
+      })
+      .result.then(function() {
+        NotificationsService.hideNotification("deprovision-service-error");
+
+        return DataService.delete({
+            group: 'servicecatalog.k8s.io',
+            resource: 'serviceinstances'
+          },
+          apiObject.metadata.name,
+          { namespace: apiObject.metadata.namespace },
+          { propagationPolicy: null
+          }) // TODO - remove once this is resolved https://github.com/kubernetes-incubator/service-catalog/issues/942
+          .then(function() {
+            NotificationsService.addNotification({
+              type: "success",
+              message: "Successfully deleted provisioned service " + apiObject.metadata.name + "."
+            });
+          }, function(err) {
+            NotificationsService.addNotification({
+              id: "deprovision-service-error",
+              type: "error",
+              message: "An error occurred while deleting provisioned service " + apiObject.metadata.name + ".",
+              details: getErrorDetails(err)
+            });
+          });
+      });
+    }
+
+    return {
+      deprovision: deprovision
+    };
+  });
+
--- a/app/views/browse/service-instance.html
+++ b/app/views/browse/service-instance.html
@@ -1,0 +1,88 @@
+<project-header class="top-header"></project-header>
+<project-page>
+
+  <!-- Middle section -->
+  <div class="middle-section">
+    <div class="middle-container">
+      <div class="middle-header">
+        <div class="container-fluid">
+          <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
+          <alerts alerts="alerts"></alerts>
+          <div ng-if="!loaded" class="mar-top-xl">Loading...</div>
+          <div ng-if="serviceInstance">
+            <h1 class="contains-actions">
+              <div class="pull-right dropdown" ng-hide="!('serviceInstances' | canIDoAny)">
+                <button type="button" class="dropdown-toggle btn btn-default actions-dropdown-btn hidden-xs" data-toggle="dropdown">
+                  Actions
+                  <span class="caret"></span>
+                </button>
+                <a href=""
+                   class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
+                   data-toggle="dropdown"><i class="fa fa-ellipsis-v" aria-hidden="true"></i><span class="sr-only">Actions</span></a>
+                <ul class="dropdown-menu dropdown-menu-right actions action-button">
+                  <li ng-if="{resource: 'serviceinstances', group: 'servicecatalog.k8s.io'} | canI : 'update'">
+                    <a ng-href="{{serviceInstance | editYamlURL}}" role="button">Edit YAML</a>
+                  </li>
+                  <li ng-if="{resource: 'serviceinstances', group: 'servicecatalog.k8s.io'} | canI : 'delete'">
+                    <a href="" ng-click="deprovision()" role="button">Delete</a>
+                  </li>
+                </ul>
+              </div>
+              {{serviceInstance | serviceInstanceDisplayName:serviceClasses}}
+              <small class="meta">created <span am-time-ago="serviceInstance.metadata.creationTimestamp"></span></small>
+            </h1>
+            <labels labels="serviceInstance.metadata.labels" clickable="true" kind="service-instances" project-name="{{serviceInstance.metadata.namespace}}" limit="3"></labels>
+          </div>
+        </div>
+      </div><!-- /middle-header-->
+      <div class="middle-content" persist-tab-state>
+        <div class="container-fluid">
+          <div class="row" ng-if="serviceInstance">
+            <div class="col-md-12">
+              <uib-tabset>
+                <uib-tab active="selectedTab.details">
+                  <uib-tab-heading>Details</uib-tab-heading>
+                  <div class="resource-details">
+                    <div class="row">
+                      <div class="col-lg-6">
+                        <p ng-bind-html="plan.description | linkify : '_blank'"></p>
+                        <dl class="dl-horizontal left">
+                          <dt ng-if-start="serviceClass.description || serviceClass.externalMetadata.longDescription">Description:</dt>
+                          <dd ng-if-end>
+                            <p class="pre-wrap" ng-bind-html="serviceClass.description | linkify : '_blank'"></p>
+                            <p class="pre-wrap" ng-bind-html="serviceClass.externalMetadata.longDescription | linkify : '_blank'"></p>
+                          </dd>
+                          <dt>Status:</dt>
+                          <dd>
+                            <status-icon status="serviceInstance | serviceInstanceStatus" disable-animation></status-icon>
+                            <span flex>{{serviceInstance | serviceInstanceStatus | sentenceCase}}</span>
+                          </dd>
+                          <dt ng-if-start="serviceInstance | serviceInstanceConditionMessage">Status Reason:</dt>
+                          <dd ng-if-end>
+                            {{serviceInstance | serviceInstanceConditionMessage}}
+                          </dd>
+                        </dl>
+                      </div>
+                      <div class="col-lg-6">
+                        <resource-service-bindings
+                          project="project"
+                          project-context="projectContext"
+                          api-object="serviceInstance">
+                        </resource-service-bindings>
+                      </div>
+                    </div>
+                    <annotations annotations="serviceInstance.metadata.annotations"></annotations>
+                  </div>
+                </uib-tab>
+                <uib-tab active="selectedTab.events" ng-if="'events' | canI : 'watch'">
+                  <uib-tab-heading>Events</uib-tab-heading>
+                  <events api-objects="[ serviceInstance ]" project-context="projectContext" ng-if="selectedTab.events"></events>
+                </uib-tab>
+              </uib-tabset>
+            </div><!-- /col-* -->
+          </div>
+        </div>
+      </div><!-- /middle-content -->
+    </div><!-- /middle-container -->
+  </div><!-- /middle-section -->
+</project-page>
--- a/app/views/directives/resource-service-bindings.html
+++ b/app/views/directives/resource-service-bindings.html
@@ -1,12 +1,13 @@
 <div class="mar-bottom-xl" ng-if="$ctrl.showBindings">
-  <h3>Service Bindings</h3>
-  <overview-service-binding
+  <h3>Bindings</h3>
+  <service-binding
     ng-repeat="binding in $ctrl.bindings track by (binding | uid)"
     namespace="$ctrl.projectContext.projectName"
     binding="binding"
+    ref-api-object="$ctrl.apiObject"
     service-classes="$ctrl.serviceClasses"
     service-instances="$ctrl.serviceInstances">
-  </overview-service-binding>
+  </service-binding>
   <div ng-if="($ctrl.bindableServiceInstances | size) && ({resource: 'serviceinstancecredentials', group: 'servicecatalog.k8s.io'} | canI : 'create')">
     <a href="" ng-click="$ctrl.createBinding()" role="button">
       <span class="pficon pficon-add-circle-o" aria-hidden="true"></span>
--- a/app/views/overview/_service-instance-row.html
+++ b/app/views/overview/_service-instance-row.html
@@ -6,7 +6,7 @@
     <div class="list-pf-content">
       <div class="list-pf-name">
         <h3>
-          <span ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></span>
+          <a ng-href="{{row.apiObject | navigateResourceURL}}" ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></a>
           <div ng-bind-html="row.apiObject.metadata.name | highlightKeywords : row.state.filterKeywords" class="list-row-longname"></div>
         </h3>
         <div class="status-icons">
--- a/app/views/service-instances.html
+++ b/app/views/service-instances.html
@@ -1,0 +1,84 @@
+<project-header class="top-header"></project-header>
+<project-page>
+
+  <!-- Middle section -->
+  <div class="middle-section">
+    <div class="middle-container">
+      <div class="middle-header header-toolbar">
+        <div class="container-fluid">
+          <div class="page-header page-header-bleed-right page-header-bleed-left">
+            <h1>
+              Provisioned Services
+            </h1>
+          </div>
+          <div class="data-toolbar">
+            <div class="data-toolbar-filter">
+              <project-filter></project-filter>
+            </div>
+          </div>
+        </div>
+      </div><!-- /middle-header-->
+      <div class="middle-content">
+        <div class="container-fluid">
+          <alerts alerts="alerts"></alerts>
+          <div class="row">
+            <div class="col-md-12">
+              <table class="table table-bordered table-mobile table-layout-fixed">
+                <colgroup>
+                  <col class="col-sm-3">
+                </colgroup>
+                <thead>
+                  <tr>
+                    <th>Name</th>
+                    <th>Status</th>
+                    <th>Created</th>
+                    <th>Bindings</th>
+                  </tr>
+                </thead>
+                <tbody ng-if="(serviceInstances | size) === 0">
+                  <tr>
+                    <td colspan="4"><em>{{emptyMessage}}</em></td>
+                  </tr>
+                </tbody>
+                <tbody ng-if="(serviceInstances | size) > 0">
+                  <tr ng-repeat="serviceInstance in serviceInstances track by (serviceInstance | uid)">
+                    <td data-title="Name"><a ng-href="{{serviceInstance | navigateResourceURL}}">{{serviceInstance | serviceInstanceDisplayName:serviceClasses}}</a></td>
+
+                    <td data-title="Status">
+                      <div row class="status">
+                        <status-icon status="serviceInstance | serviceInstanceStatus" disable-animation></status-icon>
+                        <span flex>{{serviceInstance | serviceInstanceStatus | sentenceCase}}</span>
+                      </div>
+                    </td>
+                    <td data-title="Created">
+                      <span am-time-ago="serviceInstance.metadata.creationTimestamp" am-without-suffix="true"></span> ago
+                    </td>
+                    <td data-title="Bindings">
+                      <div ng-if="bindingsByInstanceRef[serviceInstance.metadata.name].length">
+                        <p ng-if="firstBinding = bindingsByInstanceRef[serviceInstance.metadata.name][0]">
+                          <span ng-if="application = applicationsByBinding[firstBinding.metadata.name][0]">
+                            {{application.metadata.name}}
+                          </span>
+                          <span ng-if="!application">
+                            {{firstBinding.spec.secretName}}
+                          </span>
+                          <ng-pluralize count="bindingsByInstanceRef[serviceInstance.metadata.name].length"
+                                        when="{'0':'', '1':'', '2':'and {} other', 'other':'and {} others'}"
+                                        offset="1">
+                          </ng-pluralize>
+                        </p>
+                      </div>
+                      <div ng-if="!bindingsByInstanceRef[serviceInstance.metadata.name].length">
+                        No bindings
+                      </div>
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+            </div><!-- /col-* -->
+          </div>
+        </div>
+      </div><!-- /middle-content -->
+    </div><!-- /middle-container -->
+  </div><!-- /middle-section -->
+</project-page>
--- a/app/index.html
+++ b/app/index.html
@@ -224,6 +224,7 @@
         <script src="scripts/services/resourceAlerts.js"></script>
         <script src="scripts/services/listRowUtils.js"></script>
         <script src="scripts/services/ownerReferences.js"></script>
+        <script src="scripts/services/serviceInstances.js"></script>
         <script src="scripts/controllers/landingPage.js"></script>
         <script src="scripts/services/events.js"></script>
         <script src="scripts/controllers/projects.js"></script>
@@ -248,6 +249,8 @@
         <script src="scripts/controllers/statefulSet.js"></script>
         <script src="scripts/controllers/services.js"></script>
         <script src="scripts/controllers/service.js"></script>
+        <script src="scripts/controllers/serviceInstances.js"></script>
+        <script src="scripts/controllers/serviceInstance.js"></script>
         <script src="scripts/controllers/secrets.js"></script>
         <script src="scripts/controllers/secret.js"></script>
         <script src="scripts/controllers/createSecret.js"></script>
@@ -359,6 +362,7 @@
         <script src="scripts/directives/fromFileDialog.js"></script>
         <script src="scripts/directives/create/nextSteps.js"></script>
         <script src="scripts/directives/imageNames.js"></script>
+        <script src="scripts/directives/serviceBinding.js"></script>
 
         <!-- New Overview -->
         <script src="scripts/directives/buildCounts.js"></script>
@@ -371,7 +375,6 @@
         <script src="scripts/directives/overview/networking.js"></script>
         <script src="scripts/directives/overview/pipelines.js"></script>
         <script src="scripts/directives/overview/serviceBindings.js"></script>
-        <script src="scripts/directives/overview/serviceBinding.js"></script>
 
         <script src="scripts/directives/istagSelect.js"></script>
         <script src="scripts/directives/deployImage.js"></script>
--- a/app/scripts/directives/overview/serviceInstanceRow.js
+++ b/app/scripts/directives/overview/serviceInstanceRow.js
@@ -4,12 +4,10 @@
   angular.module('openshiftConsole').component('serviceInstanceRow', {
     controller: [
       '$filter',
-      '$uibModal',
-      'DataService',
+      'AuthorizationService',
       'BindingService',
       'ListRowUtils',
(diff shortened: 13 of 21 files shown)


spadgett:
@cdcabrera Let me know when you'd like a review, thanks!

cdcabrera:
@spadgett going through and generating screenshots along with questions

app/scripts/app.js
```diff
@@ -271,6 +271,16 @@ angular
         controller: 'ServiceController',
         reloadOnSearch: false
       })
+      .when('/project/:project/browse/provisioned', {
```
spadgett: I think it would be good to match the API kind in the code even if we prefer to it as "Provisioned Service" in the UI. So I'd suggest `/browse/service-instances/` and `/browser/services-instances/:instance` with corresponding names to the controller and views.

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
...
+ * @name openshiftConsole.controller:ProvisionedServiceController
+ * @description
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('ProvisionedServiceController', function ($scope,
+                                             $routeParams,
+                                             DataService,
+                                             ProjectsService,
+                                             $filter) {
```
spadgett: Nit: I like to keep the lists alphabetized to keep them more manageable.

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
...
+    $scope.projectName = $routeParams.project;
+    $scope.displayName = null;
+    $scope.serviceInstance = null;
+    $scope.serviceInstances = null;
+    $scope.serviceClasses = {};
+    $scope.alerts = {};
+    //$scope.renderOptions = $scope.renderOptions || {};
+    //$scope.renderOptions.hideFilterWidget = true;
```
spadgett: The two commented out lines can be removed.

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
...
+        link: "project/" + $routeParams.project + "/browse/provisioned"
+      },
+      {
+        title: $routeParams.service
+      }
+    ];
+
+    $scope.podFailureReasons = {
+      "Pending": "This pod will not receive traffic until all of its containers have been created."
+    };
```
spadgett: Looks like you have a fair amount of code from the service controller that is not needed here.

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
...
+
+    //var allPods = {};
+    var watches = [];
+
+    // receives routes for the current service and maps service ports to each route name
+    var getPortsByRoute = function() {
```
spadgett: This entire function can be removed.

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
...
+
+      $scope.displayName = $filter('serviceInstanceDisplayName')($scope.serviceInstance, $scope.serviceClasses);
+    };
+
+    var serviceResolved = function(service, action) {
+      $scope.loaded = true;
+      $scope.serviceInstance = service;
+
+      //getPodsForService();
+      getPortsByRoute();
```
spadgett: These lines can be removed.

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
...
+
+      if (action === "DELETED") {
+        $scope.alerts["deleted"] = {
+          type: "warning",
+          message: "This service has been deleted."
```
spadgett: "This provisioned service has been deleted."

app/scripts/constants.js
```diff
@@ -190,6 +190,13 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
               prefixes: [
                 "/browse/routes/"
               ]
+            },
+            {
+              label: "Provisioned Services",
+              href: "/browse/provisioned",
+              prefixes: [
+                "/browse/provisioned/"
+              ]
```
spadgett: We should only add this menu item if service catalog is enabled and the user has authority.

You can add an `isValid` item for the tech preview feature check, and there is a `canI` check for Membership below that you can copy.

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
...
+          _.each(svcEndpoint.subsets, function(subset) {
+            _.each(subset.addresses, function(address) {
+              if (_.get(address, "targetRef.kind") === "Pod") {
+                $scope.podsWithEndpoints[address.targetRef.name] = true;
+              }
+            });
+          });
+        }));*/
+
+        watches.push(DataService.watch("routes", context, function(routes) {
```
spadgett: This can be removed.

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+                                                         $scope,
+                                                         APIService,
+                                                         BindingService,
+                                                         Constants,
+                                                         DataService,
+                                                         ProjectsService,
+                                                         ResourceAlertsService,
+                                                         $filter,
+                                                         LabelFilter,
+                                                         Logger) {
```
spadgett: nit: alpha

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+    //var limitWatches = $filter('isIE')() || $filter('isEdge')();
+    //var DEFAULT_POLL_INTERVAL = 60 * 1000; // milliseconds
+
+    // Enable service catalog features if the new experience is enabled and the
+    // servicecatalog.k8s.io resources are available.
+    var SERVICE_CATALOG_ENABLED =
+      _.get(Constants, 'ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page') &&
+      APIService.apiInfo({ group: 'servicecatalog.k8s.io', resource: 'serviceclasses' }) &&
+      APIService.apiInfo({ group: 'servicecatalog.k8s.io', resource: 'instances' }) &&
+      APIService.apiInfo({ group: 'servicecatalog.k8s.io', resource: 'bindings' });
```
spadgett: FYI, @jeff-phillips-18 added something for this in #1912 where you don't have to copy the code. But I think we should disable the left nav menu entirely if service catalog is not enabled (per my comment above).

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
...
+ * @description
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('ProvisionedServiceController', function ($scope,
+                                             $routeParams,
+                                             DataService,
+                                             ProjectsService,
+                                             $filter) {
+    $scope.projectName = $routeParams.project;
```
spadgett: Is `$scope.projectName` used?

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,190 @@
+'use strict';
+
+/**
+ * @ngdoc function
+ * @name openshiftConsole.controller:ProvisionedServiceController
+ * @description
+ * Controller of the openshiftConsole
+ */
```
spadgett: I know we have these comments in other files, but I'd prefer just to remove them if we don't have any real description here.

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
+'use strict';
+
+/**
+ * @ngdoc function
+ * @name openshiftConsole.controller:ProvisionedServicesController
+ * @description
+ * # ProjectController
+ * Controller of the openshiftConsole
+ */
```
spadgett: Same comment here: we can probably just get rid of this comment.

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+    //$scope.routesByService = {};
+    //$scope.routes = {};
+    $scope.labelSuggestions = {};
+    $scope.alerts = $scope.alerts || {};
+    $scope.emptyMessage = "Loading...";
+    $scope.emptyMessageRoutes = "Loading...";
+
+    var watches = [];
+
+    var state = {
```
spadgett: This is an overview-specific variable. We shouldn't need it here.

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+      _.get(Constants, 'ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page') &&
+      APIService.apiInfo({ group: 'servicecatalog.k8s.io', resource: 'serviceclasses' }) &&
+      APIService.apiInfo({ group: 'servicecatalog.k8s.io', resource: 'instances' }) &&
+      APIService.apiInfo({ group: 'servicecatalog.k8s.io', resource: 'bindings' });
+
+    $scope.projectName = $routeParams.project;
+    $scope.serviceInstances = {};
+    $scope.unfilteredServiceInstances = {};
+    //$scope.routesByService = {};
+    //$scope.routes = {};
```
spadgett: Yeah, you can remove these two lines.

app/scripts/filters/canI.js
```diff
@@ -29,6 +29,9 @@ angular
       'imageStreams': [
         {group: '', resource: 'imagestreams', verbs: ['update', 'delete']}
       ],
+      'instances': [
+        {group: 'servicecatalog.k8s.io', resource: 'instances', verbs: ['update', 'create', 'delete']}
```
spadgett:
```js
{group: 'servicecatalog.k8s.io', resource: 'instances', verbs: ['update', 'delete']}
```

Eventually we will need to check bindings too if we let users create bindings from the actions menu.

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+    };
+
+    ProjectsService
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+
+        var canI = $filter('canI');
+
+        if (SERVICE_CATALOG_ENABLED && canI({resource: 'instances', group: 'servicecatalog.k8s.io'}, 'watch')) {
```
spadgett: Better just to hide the nav item so you can't even get here.

I'd remove the `SERVICE_CATALOG_ENABLED` and `canI` checks, otherwise the page will be forever "Loading..." if you don't have authority. Better to see the error when the watch fails.

Same for the other `SERVICE_CATALOG_ENABLED` and `canI` checks below.

app/views/browse/provisioned-service.html
```diff
@@ -0,0 +1,101 @@
...
+                  Actions
+                  <span class="caret"></span>
+                </button>
+                <a href=""
+                   class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
+                   data-toggle="dropdown"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
+                <ul class="dropdown-menu dropdown-menu-right actions action-button">
+                  <li ng-if="'routes' | canI : 'create'">
+                    <a ng-href="project/{{project.metadata.name}}/create-route?instance={{serviceInstance.metadata.name}}"
+                       role="button">Create Route</a>
```
spadgett: Remove this action. It's not relevant here.

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+
+    $scope.projectName = $routeParams.project;
+    $scope.serviceInstances = {};
+    $scope.unfilteredServiceInstances = {};
+    //$scope.routesByService = {};
+    //$scope.routes = {};
+    $scope.labelSuggestions = {};
+    $scope.alerts = $scope.alerts || {};
+    $scope.emptyMessage = "Loading...";
+    $scope.emptyMessageRoutes = "Loading...";
```
spadgett: You shouldn't need to show any routes on this page.

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+            sortServiceInstances();
+            updateFilter();
+          });
+        }
+
+        function updateFilterWarning() {
+          if (!LabelFilter.getLabelSelector().isEmpty() && _.isEmpty($scope.serviceInstances)  && !_.isEmpty($scope.unfilteredServiceInstances)) {
+            $scope.alerts["services"] = {
+              type: "warning",
+              details: "The active filters are hiding all services."
```
spadgett: "all provisioned services."

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+          }, context, function(serviceClasses) {
+            state.serviceClasses = serviceClasses.by('metadata.name');
+            sortServiceInstances();
+            updateFilter();
+          });
+        }
+
+        function updateFilterWarning() {
+          if (!LabelFilter.getLabelSelector().isEmpty() && _.isEmpty($scope.serviceInstances)  && !_.isEmpty($scope.unfilteredServiceInstances)) {
+            $scope.alerts["services"] = {
```
spadgett: Let's give the alert a more specific ID like `"all-instances-filtered"`

app/views/overview/_service-instance-row.html
```diff
@@ -6,7 +6,8 @@
     <div class="list-pf-content">
       <div class="list-pf-name">
         <h3>
-          <span ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></span>
+          <!--<span ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></span>-->
+          <a href="{{row.apiObject | navigateResourceURL}}" ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></a>
```
spadgett: Nit: use `ng-href`

app/views/provisioned.html
```diff
@@ -0,0 +1,156 @@
...
+  <!-- Middle section -->
+  <div class="middle-section">
+    <div class="middle-container">
+      <div class="middle-header header-toolbar">
+        <div class="container-fluid">
+          <div class="page-header page-header-bleed-right page-header-bleed-left">
+            <h1>
+              Provisioned Services
+              <span class="page-header-link">
+                <a ng-href="{{'services' | helpLink}}" target="_blank">
```
spadgett: We need to link to the real service catalog topic, not to the services topic, which is something different. You might just remove this for now.

app/views/provisioned.html
```diff
@@ -0,0 +1,156 @@
...
+                <colgroup>
+                  <col class="col-sm-3">
+                </colgroup>
+                <thead>
+                  <tr>
+                    <th>Name</th>
+                    <th>Bindings</th>
+                    <th>Status</th>
+                    <th>Created</th>
+                    <th>Web Interface Url</th>
```
spadgett: Shouldn't need this column.

app/views/provisioned.html
```diff
@@ -0,0 +1,156 @@
...
+          <div class="row">
+            <div class="col-md-12">
+              <table class="table table-bordered table-hover table-mobile table-layout-fixed">
+                <colgroup>
+                  <col class="col-sm-3">
+                </colgroup>
+                <thead>
+                  <tr>
+                    <th>Name</th>
+                    <th>Bindings</th>
```
spadgett: I think this might be better after status and created columns.

app/views/provisioned.html
```diff
@@ -0,0 +1,156 @@
...
+
+                    <!--<th>Name</th>
+                    <th>Cluster IP</th>
+                    <th>External IP</th>
+                    <th>Ports</th>
+                    <th>Selector</th>
+                    <th>Age</th>-->
+                  </tr>
+                </thead>
+                <tbody ng-if="(serviceInstances | hashSize) === 0">
```
spadgett: Use `| size` instead of `| hashSize`

app/views/provisioned.html
```diff
@@ -0,0 +1,156 @@
...
+                  <tr>
+                    <td colspan="5"><em>{{emptyMessage}}</em></td>
+                  </tr>
+                </tbody>
+                <tbody ng-if="(serviceInstances | hashSize) > 0">
```
spadgett: `| size`

app/views/provisioned.html
```diff
@@ -0,0 +1,156 @@
...
+                  <tr ng-repeat="serviceInstance in serviceInstances | orderObjectsByDate : true">
```
spadgett: Better to sort in the controller, not the view.

Use `track by (serviceInstance | uid)`

app/views/provisioned.html
```diff
@@ -0,0 +1,156 @@
...
+                    <td data-title="Name"><a href="{{serviceInstance | navigateResourceURL}}">{{serviceInstance | serviceInstanceDisplayName:serviceClasses}}</a></td>
```
spadgett: nit: `ng-href`

app/views/provisioned.html
```diff
@@ -0,0 +1,156 @@
...
+                          <a
+                            ng-if="row.bindings.length > 1"
+                            ng-click="row.toggleExpand($event, true)">
+                            {{row.bindings.length -1}} other<span ng-if="row.bindings.length > 2">s</span></a>
+                        </span>
+                      </p>-->
+                    </td>
+                    <td data-title="Status">
+                      <div row class="status">
+                        <status-icon status="serviceInstance | deploymentStatus" disable-animation></status-icon>
```
spadgett: `deploymentStatus` is not correct here. You'll need to look at the overview service instance row to see how we determine status.

app/scripts/controllers/provisionedService.js
cdcabrera: Sure thing

app/scripts/controllers/provisionedService.js
cdcabrera: It isn't, there are a few bits here and there like that where wasn't sure about them keeping around... part of my questions that need to filter these bits out

app/scripts/controllers/provisionedService.js
cdcabrera: Indeed, used the service controller as the basis. Anything we can wipe out... keep those comments coming

app/scripts/controllers/provisionedServices.js
cdcabrera: Nice! That was one of the questions, can we centralize it seems silly to keep repeating it, I'll take a look at his PR

app/views/provisioned.html
cdcabrera: Sure thing on cleaning the columns up. 

@spadgett  Had a question about how we handle the bindings piece and if there's a chance that display would be better as a directive/component? @jeff-phillips-18 may have solved this already

app/scripts/app.js
cdcabrera: @spadgett what about matching file names too?

spadgett: Yes, let's be consistent with the filenames also

app/scripts/controllers/provisionedServices.js
jeff-phillips-18: Yes. CatalogService. SERVICE_CATALOG_ENABLED

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+      return _.get(state, ['notificationsByObjectUID', uid], {});
+    };*/
+
+    var updateFilter = function() {
+      $scope.serviceInstances = LabelFilter.getLabelSelector().select($scope.unfilteredServiceInstances);
+    };
+
+    // TODO: code duplicated from directives/bindService.js
+    // extract & share
+    var sortServiceInstances = function() {
```
jeff-phillips-18: See https://github.com/openshift/origin-web-console/pull/1912/files#diff-7878fa660740a71bf5dbfa99b53c8b43R1152

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,265 @@
...
+          return _.get(state.serviceClasses, [item.spec.serviceClassName, 'externalMetadata', 'displayName']) || item.spec.serviceClassName;
+        },
+        function(item) {
+          return _.get(item, 'metadata.name', '');
+        }
+      );
+    };
+
+    var groupBindings = function() {
+      // Build two maps:
```
jeff-phillips-18: See https://github.com/jeff-phillips-18/origin-web-console/blob/e4dbd7f0fa9729e77d7d734650286d75bb800c1c/app/scripts/controllers/overview.js#L1084

app/views/provisioned.html
jeff-phillips-18: Seems like you should use the overview-service-binding component. Might need a modification or three to be able to show the small number of differences between showing it for an application or a service, but otherwise there is a lot of overlap.

app/scripts/controllers/provisionedService.js
cdcabrera: It appears used by the "overview" navigation, when removed the project name becomes "undefined" and the navigation redirects towards an error page

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,209 @@
...
+      $scope.notificationsByObjectUID[uid] = notifications || {};
+    };
+
+    var getNotifications = function(apiObject) {
+      var uid = getUID(apiObject);
+      if (!uid) {
+        return {};
+      }
+      return _.get(state, ['notificationsByObjectUID', uid], {});
+    };*/
```
spadgett: This can go.

app/scripts/controllers/provisionedServices.js
```diff
@@ -0,0 +1,209 @@
...
+    var sortServiceInstances = function() {
+      $scope.bindableServiceInstances = BindingService.filterBindableServiceInstances($scope.unfilteredServiceInstances, $scope.serviceClasses);
+      $scope.unfilteredServiceInstances = BindingService.sortServiceInstances($scope.unfilteredServiceInstances, $scope.serviceClasses);
+    };
+
+    /*var refreshSecrets = _.debounce(function(context) {
+      DataService.list("secrets", context, null, { errorNotification: false }).then(function(secretData) {
+        state.secrets = secretData.by("metadata.name");
+      });
+    }, 300);*/
```
spadgett: This can go.

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,96 @@
...
+            resource: 'instances'
+          }, $routeParams.instance, context, { errorNotification: false })
+          .then(function(service) {
+            serviceResolved(service);
+            setDisplayName();
+            watches.push(DataService.watchObject({
+              group: 'servicecatalog.k8s.io',
+              resource: 'instances'
+            }, $routeParams.instance, context, serviceResolved));
+
```
spadgett:
```js
        watches.push(DataService.watch({
          group: 'servicecatalog.k8s.io',
          resource: 'bindings'
        }, context, function(bindings) {
	  $scope.bindings = _.filter(bindings.by('metadata.name'), { spec: { instanceRef: { name: instance.metadata.name } } });
        }));
```

app/scripts/controllers/provisionedService.js
```diff
@@ -0,0 +1,96 @@
...
+        DataService.list({
+          group: 'servicecatalog.k8s.io',
+          resource: 'serviceclasses'
+        }, context, function(serviceClasses) {
+          $scope.serviceClasses = serviceClasses.by('metadata.name');
+          setDisplayName();
+        });
+
+        watches.push(DataService.watch({
```
spadgett: This watch can go since you're already watching the instance for this page above.

cdcabrera:
@spadgett & @jeff-phillips-18 additional updates... 

Short checklist of things that may need to be altered
- [ ] the [Plan name/description display on service instance detail](https://github.com/openshift/origin-web-console/pull/1906/files#diff-bc751d1d01f9cbcce7a12073f1606ed0R55) and [the controller aspect](https://github.com/openshift/origin-web-console/pull/1906/files#diff-44e681d6033bb7edf7fcdf36facb5392R43)
- [ ] filters, [added a couple of simple filters](https://github.com/openshift/origin-web-console/pull/1906/files#diff-7e6acf93d418261406d06472945ed7c2R1332) to help with the status around service instance, not keen on them but they did help simplify the process for display
  - [service-instance.html](https://github.com/openshift/origin-web-console/pull/1906/files#diff-bc751d1d01f9cbcce7a12073f1606ed0R65)
  - [service-instances.html](https://github.com/openshift/origin-web-console/pull/1906/files#diff-6ca4f62b569dd42d4128a674f496a119R54)
- [ ] additions around [resource-service-bindings component line 42](https://github.com/openshift/origin-web-console/pull/1906/files#diff-436ebc1046bfa813c8773243b1187374R42) that haven't been moved to the ```origin-web-common``` repo
- [ ] the event tab around the service instance detail view is non-functioning

app/scripts/filters/resources.js
```diff
@@ -1328,4 +1328,20 @@ angular.module('openshiftConsole')
...
+      var status = 'Failed';
+
+      if (isServiceInstanceReadyFilter(instance)) {
+        status = 'Succeeded';
+      }
+
+      return status;
+    };
+  })
+  .filter('serviceInstanceMessage', function(statusConditionFilter) {
```
spadgett: Maybe `'readyConditionMessage'`

app/views/browse/service-instance.html
```diff
@@ -0,0 +1,102 @@
...
+                </button>
+                <a href=""
+                   class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
+                   data-toggle="dropdown"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
+                <ul class="dropdown-menu dropdown-menu-right actions action-button">
+                  <li ng-if="{resource: 'instances', group: 'servicecatalog.k8s.io'} | canI : 'update'">
+                    <a ng-href="{{serviceInstance | editYamlURL}}" role="button">Edit YAML</a>
+                  </li>
+                  <li ng-if="{resource: 'instances', group: 'servicecatalog.k8s.io'} | canI : 'delete'">
+                    <delete-link
```
spadgett: Probably want the same dialog as used here:

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/overview/serviceInstanceRow.js#L84

Let's make it a common component that we can reuse. We'll eventually need to handle bindings.

app/views/browse/service-instance.html
```diff
@@ -0,0 +1,102 @@
...
+                  <uib-tab-heading>Details</uib-tab-heading>
+                  <div class="resource-details">
+                    <div class="row">
+                      <div class="col-lg-6">
+                        <h3>Plan</h3>
+                        <p>
+                          {{plan.description}}
+                        </p>
+                        <dl class="dl-horizontal left">
+                          <dt ng-if-start="serviceClass.description">Description:</dt>
```
spadgett: Probably want this above the plan details

spadgett: We should show `externalMetadata.longDescription` if it's there even if `description` is not.

app/views/browse/service-instance.html
```diff
@@ -0,0 +1,102 @@
...
+                          </dd>
+                          <dt>Status:</dt>
+                          <dd>
+                            <status-icon status="serviceInstance | serviceInstanceStatus" disable-animation></status-icon>
+                            <span flex>{{serviceInstance | serviceInstanceStatus | sentenceCase}}</span>
+                          </dd>
+                          <dt ng-if-start="serviceInstance | serviceInstanceMessage">Status Reason:</dt>
+                          <dd ng-if-end>
+                            {{serviceInstance | serviceInstanceMessage}}
+                          </dd>
```
spadgett: We've used a somewhat different layout for status on the route page and the build config page

app/views/browse/service-instance.html
```diff
@@ -0,0 +1,102 @@
...
+                          <!--<dd>
+                            <status-icon status="serviceInstance | serviceInstanceStatus" disable-animation></status-icon>
+                            <span flex>
+                              {{serviceInstance | serviceInstanceStatus}}<span ng-if="serviceInstance | serviceInstanceMessage">. {{serviceInstance | serviceInstanceMessage}}</span>
+                            </span>
+                          </dd>-->
+                        </dl>
+                      </div>
+                      <div class="col-lg-6">
+                        <resource-service-bindings
```
spadgett: We might rename this component to just `service-bindings` if we're using it for service instances too.

spadgett:
[test]

serenamarie125:
@beanh66 Can you take a look at this, based on your design?

spadgett:
@beanh66 @serenamarie125 Just a heads up that the PR is still WIP, so some things are missing.

beanh66:
Okay no problem @spadgett! @cdcabrera let me know when this is ready for UI review and we can meet if that's easiest.

app/views/overview/_service-binding.html
```diff
@@ -3,7 +3,12 @@
     <div class="col-sm-5 col-md-6">
       <h3 ng-if="$ctrl.serviceClass">
         <span ng-if="$ctrl.serviceClass">
-          {{$ctrl.serviceClass.externalMetadata.displayName || $ctrl.serviceClass.metadata.name}}
+          <span ng-if="$ctrl.refApiObject.kind !== 'Instance'">
+            {{$ctrl.serviceClass.externalMetadata.displayName || $ctrl.serviceClass.metadata.name}}
+          </span>
+          <span ng-if="$ctrl.refApiObject.kind === 'Instance'">
+            {{$ctrl.binding.spec.secretName}}
+          </span>
```
cdcabrera: So two options currently, probably more... 
- first this version, where the ```resource-service-bindings``` component wraps the ```overview-service-binding ``` component. Reflected in how the current code is extending  the display for ```secret-name```. 
- or second, simply place the ```secret-name``` display logic within the ```resource-service-bindings``` component directly

Either way it works for me. But may have to port over additional display logic if we're looking to separate the "overview" namespace from subsequent partials/views. 

And of course this may be an example in support of a componentized directory structure.

cdcabrera:
Reference point for PR #1961 Consider looking at breaking out another component for the status from ```service instance row```.

app/scripts/controllers/serviceInstance.js
```diff
@@ -0,0 +1,104 @@
...
+    };*/
+
+    var setDescription = function() {
+      if(!$scope.serviceInstance || !$scope.serviceClasses) {
+        return;
+      }
+
+      var serviceClassName = _.get($scope.serviceInstance.spec, 'serviceClassName');
+      $scope.serviceClass = _.get($scope.serviceClasses, [serviceClassName]);
+      $scope.plan = _.find(_.get($scope.serviceClass, 'plans'), {name: 'default'});
```
spadgett:
```js
 $scope.plan = _.find(_.get($scope.serviceClass, 'plans'), {name: $scope.serviceInstance.spec.planName });
```

app/views/browse/service-instance.html
cdcabrera: In converting this over to a service, did a comparison on the current directive/component delete link looks like there are obvious "copy" differences, but the core of it is the same. Passing in the additional parameter/option/attribute for "group" and wrapping some of the display logic/ng-ifs with a service instance could solve this while separating out the DOM manipulation from the service

Purpose we either modify the existing, or create a new component, for the delete link for "service instances".

cdcabrera: @spadgett after going over the modal designs, talked with @ncameronbritt about the style. The style for... 

screen shot 2017-08-25 at 4 34 31 pm

fits more inline with the design aspect as opposed to the current...

screen shot 2017-08-25 at 4 34 49 pm

Since this seems to fit partially with the "Service Instances Service" I can do a few of things..
- leave the new "Service Instances Service" as is, free floating and convert the existing Provisioned Service "delete" links over to the preferred modal design
- leave the new "Service Instances Service" as is and leave the delete links/buttons alone
- just convert the new "Service Instances Service" by removing the DOM manipulation/modal
- or leave everything as is


@beanh66

spadgett:
@cdcabrera Let me know when it's a good time to take another look.

app/views/browse/service-instance.html
cdcabrera: I can do this, it's going to affect a handful of files.

cdcabrera:
@spadgett rebased, though the -bot seems to think there's still a ```dist``` conflict. However should be good for a review minus the last review comments. Also setup the Origin-Web-Common branch related to the ```resource-service-bindings``` component and service, but left the code within ```-console``` to avoid having to update Bower.

cdcabrera:
@beanh66 if you're up for it, a review of the differences would be good. Aside from a slight rebase  what's functioning currently:

 - Should be able to step through from Overview.
 - The Provisioned Services link is still underneath Applications on the left hand
 - Columns for the Provisioned Services table display were modified based on relevant data. @spadgett may be able to provide additional details
 - Provisioned Service details available for display, should incorporate @jeff-phillips-18 component

app/scripts/app.js
```diff
@@ -271,6 +271,16 @@ angular
         controller: 'ServiceController',
         reloadOnSearch: false
       })
+      .when('/project/:project/browse/service-instances', {
```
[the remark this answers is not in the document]
spadgett: Following up to say the name upstream will be `ServiceInstance`, so what you have is good 👍

app/scripts/controllers/serviceInstance.js
```diff
@@ -0,0 +1,120 @@
...
+
+    var watches = [];
+
+    /*var setDisplayName = function() {
+      if(!$scope.serviceInstance || !$scope.serviceClasses) {
+        return;
+      }
+
+      $scope.displayName = $filter('serviceInstanceDisplayName')($scope.serviceInstance, $scope.serviceClasses);
+    };*/
```
spadgett: We should either add back or remove commented out code before we're ready to merge

app/scripts/controllers/serviceInstance.js
```diff
@@ -0,0 +1,120 @@
...
+
+    var updateBreadCrumbs = function() {
```
spadgett: nit: `updateBreadcrumbs` to be consistent with the scope var case

app/scripts/controllers/serviceInstance.js
```diff
@@ -0,0 +1,120 @@
...
+            /*watches.push(DataService.watchObject({
+              group: 'servicecatalog.k8s.io',
+              resource: 'instances'
+            }, $routeParams.instance, context, serviceResolved));*/
+            /*watches.push(DataService.watch({
+              group: 'servicecatalog.k8s.io',
+              resource: 'bindings'
+            }, context, function(bindings) {
+              $scope.bindings = _.filter(bindings.by('metadata.name'), { spec: { instanceRef: { name: instance.metadata.name } } });
+            }));*/
```
spadgett: Any reason this block is commented out? We should watch the objects for live page updates.

app/scripts/controllers/serviceInstances.js
```diff
@@ -0,0 +1,185 @@
...
+          group: 'servicecatalog.k8s.io',
+          resource: 'instances'
+        }, context, function(serviceInstances) {
+          $scope.emptyMessage = "No provisioned services to show";
+          $scope.unfilteredServiceInstances = serviceInstances.by('metadata.name');
+
+          //_.each($scope.unfilteredServiceInstances, function (instance) {
+          //  var notifications = ResourceAlertsService.getServiceInstanceAlerts(instance);
+          //  setNotifications(instance, notifications);
+          //});
```
spadgett: Let's remove for now.

app/scripts/controllers/serviceInstances.js
```diff
@@ -0,0 +1,185 @@
...
+      $scope.serviceInstances = LabelFilter.getLabelSelector().select($scope.unfilteredServiceInstances);
+    };
+
+    var sortServiceInstances = function() {
+      $scope.bindableServiceInstances = BindingService.filterBindableServiceInstances($scope.unfilteredServiceInstances, $scope.serviceClasses);
+      $scope.unfilteredServiceInstances = BindingService.sortServiceInstances($scope.unfilteredServiceInstances, $scope.serviceClasses);
+    };
+
+    var groupBindings = function() {
+      // Build two maps:
```
spadgett: I suggest for now just creating a map of bindings by service instance name. Then you can show those in the table. We'll eventually want to show application name, but we can't do that until pod presets are enabled in a later release.

@jeff-phillips-18 FYI

app/scripts/filters/canI.js
spadgett: Remove `'create'` from the array here. You can create instances on this page.

app/scripts/filters/resources.js
```diff
@@ -1328,4 +1328,14 @@ angular.module('openshiftConsole')
       var serviceClassDisplayName = _.get(serviceClasses, [serviceClassName, 'externalMetadata', 'displayName']);
       return serviceClassDisplayName || serviceClassName || instanceName;
     };
+  })
+  .filter('serviceInstanceStatus', function(isServiceInstanceReadyFilter) {
+    return function(instance) {
+      return isServiceInstanceReadyFilter(instance) ? 'Ready' : 'Pending';
```
spadgett: We'll need to check for the `Failed` condition as well as @jeff-phillips-18 added in #1961

app/scripts/services/serviceInstances.js
```diff
@@ -0,0 +1,69 @@
...
+        okButtonText: 'Delete',
+        okButtonClass: 'btn-danger',
+        cancelButtonText: 'Cancel'
+      };
+
+      // TODO: we probably have to handle bindings in here.
+      // either:
+      // - automatically remove the bindings
+      // - tell the user they must manually unbind before continue
+      $uibModal.open({
```
spadgett: You probably want to return the promise here so callers can know when it's complete and if it failed.

app/views/browse/service-instance.html
spadgett: We probably want something closer to the first dialog, but I'd rather do that in a separate PR. Let's reuse what we have for now and revisit the design for both overview and this page later.

app/views/directives/resource-service-bindings.html
```diff
@@ -1,9 +1,10 @@
 <div class="mar-bottom-xl" ng-if="$ctrl.showBindings">
-  <h3>Service Bindings</h3>
+  <h3>Bindings</h3>
   <overview-service-binding
```
spadgett: I'm not sure we should reuse the `overview-service-binding` component here. If we do, it should be renamed.

app/views/overview/_service-binding.html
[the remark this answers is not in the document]
spadgett: `resource-service-bindings` wraps `overview-service-binding`? Let's discuss. At the very least, we should rename `overview-service-binding` if it's used outside of the overview.

app/views/overview/_service-instance-row.html
```diff
@@ -6,7 +6,8 @@
     <div class="list-pf-content">
       <div class="list-pf-name">
         <h3>
-          <span ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></span>
+          <!--<span ng-bind-html="row.displayName | highlightKeywords : row.state.filterKeywords"></span>-->
```
spadgett: Should be able to remove this line.

app/views/directives/resource-service-bindings.html
cdcabrera: Sure thing, np renaming it. It just happens to be spread out over several files, little worried I'd end up with more merge conflicts =). However since it sounds like we're good with it I'll go ahead and do the search and replace.

app/scripts/controllers/serviceInstance.js
cdcabrera: Nope, just a test that wasn't reverted. Can do

app/scripts/constants.js
```diff
@@ -257,6 +257,23 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
...
+              prefixes: [
+                "/browse/service-instances/"
+              ],
+              isValid: function(){
+                return _.get(window.OPENSHIFT_CONSTANTS, 'ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page');
+              },
+              canI: {
+                resource: {
+                  group: 'servicecatalog.k8s.io',
+                  resource: 'instances'
```
spadgett: This will need to change to `'serviceinstances'` when https://github.com/openshift/origin/pull/16150 is merged.

app/scripts/controllers/serviceInstance.js
```diff
@@ -0,0 +1,106 @@
...
+    ProjectsService
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+        $scope.projectContext = context;
+
+        DataService
+          .get({
+            group: 'servicecatalog.k8s.io',
+            resource: 'instances'
```
spadgett: This will need to change to `'serviceinstances'` when https://github.com/openshift/origin/pull/16150 is merged.

app/scripts/controllers/serviceInstance.js
```diff
@@ -0,0 +1,106 @@
...
+          }, $routeParams.instance, context, { errorNotification: false })
+          .then(function(service) {
+
+            serviceResolved(service);
+
+            watches.push(DataService.watchObject({
+              group: 'servicecatalog.k8s.io',
+              resource: 'instances'
```
spadgett: This will need to change to `'serviceinstances'` when https://github.com/openshift/origin/pull/16150 is merged.

app/scripts/controllers/serviceInstances.js
```diff
@@ -0,0 +1,98 @@
...
+
+    ProjectsService
+      .get($routeParams.project)
+      .then(_.spread(function(project, context) {
+        $scope.project = project;
+        $scope.projectContext = context;
+
+        watches.push(DataService.watch({
+          group: 'servicecatalog.k8s.io',
+          resource: 'bindings'
```
spadgett: This will need to change to `'serviceinstancecredentials'` when https://github.com/openshift/origin/pull/16150 is merged.

app/scripts/controllers/serviceInstances.js
```diff
@@ -0,0 +1,98 @@
...
+        }, context, function(bindings) {
+          var bindingsByName = bindings.by('metadata.name');
+          $scope.bindingsByInstanceRef = _.groupBy(bindingsByName, 'spec.instanceRef.name');
+        }));
+
+        watches.push(DataService.watch({
+          group: 'servicecatalog.k8s.io',
+          resource: 'instances'
```
spadgett: This will need to change to `'serviceinstances'` when https://github.com/openshift/origin/pull/16150 is merged.

app/scripts/filters/canI.js
```diff
@@ -29,6 +29,9 @@ angular
       'imageStreams': [
         {group: '', resource: 'imagestreams', verbs: ['update', 'delete']}
       ],
+      'instances': [
+        {group: 'servicecatalog.k8s.io', resource: 'instances', verbs: ['update', 'delete']}
```
spadgett: This will need to change to `'serviceinstances'` when https://github.com/openshift/origin/pull/16150 is merged.

app/views/browse/service-instance.html
```diff
@@ -0,0 +1,95 @@
...
+              <div class="pull-right dropdown" ng-hide="!('instances' | canIDoAny)">
+                <button type="button" class="dropdown-toggle btn btn-default actions-dropdown-btn hidden-xs" data-toggle="dropdown">
+                  Actions
+                  <span class="caret"></span>
+                </button>
+                <a href=""
+                   class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
+                   data-toggle="dropdown"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
+                <ul class="dropdown-menu dropdown-menu-right actions action-button">
+                  <li ng-if="{resource: 'instances', group: 'servicecatalog.k8s.io'} | canI : 'update'">
```
spadgett: This will need to change to `'serviceinstances'` when https://github.com/openshift/origin/pull/16150 is merged.

app/views/browse/service-instance.html
```diff
@@ -0,0 +1,95 @@
...
+                    <a ng-href="{{serviceInstance | editYamlURL}}" role="button">Edit YAML</a>
+                  </li>
+                  <li ng-if="{resource: 'instances', group: 'servicecatalog.k8s.io'} | canI : 'delete'">
```
spadgett: This will need to change to `'serviceinstances'` when https://github.com/openshift/origin/pull/16150 is merged.

app/views/directives/resource-service-bindings.html
```diff
@@ -1,12 +1,13 @@
...
+  <service-binding
     ng-repeat="binding in $ctrl.bindings track by (binding | uid)"
     namespace="$ctrl.projectContext.projectName"
     binding="binding"
+    ref-api-object="$ctrl.apiObject"
     service-classes="$ctrl.serviceClasses"
     service-instances="$ctrl.serviceInstances">
-  </overview-service-binding>
+  </service-binding>
   <div ng-if="($ctrl.bindableServiceInstances | size) && ({resource: 'bindings', group: 'servicecatalog.k8s.io'} | canI : 'create')">
```
cdcabrera: @spadgett this ```bindings``` looks like it falls into the category of... replace with ```serviceinstancecredentials``` ? looks like there's another ```bindings``` reference around line 23ish too.

If replace is the case for those ```bindings``` do we have a related issue, or is there a PR that's already in progress? Found a couple of more instances in other files...

spadgett: Yes.

We have #2048 to update the names in other places.

app/scripts/filters/canI.js
cdcabrera: going to update line 32 to match with ```serviceinstances```,  currently still retains the "instances" string while I walk/run through the changes again

spadgett:
@cdcabrera FYI, renames are there upstream. See https://github.com/openshift/origin-web-console/pull/2048#issuecomment-329326035

app/scripts/directives/overview/serviceBinding.js
```diff
@@ -1,14 +1,15 @@
 'use strict';
 
 (function() {
-  angular.module('openshiftConsole').component('overviewServiceBinding', {
+  angular.module('openshiftConsole').component('serviceBinding', {
```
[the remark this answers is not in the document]
cdcabrera: @spadgett moved the HTML and JS to the associated ```directives``` directories

app/views/service-instances.html
```diff
@@ -0,0 +1,84 @@
...
+            </div>
+          </div>
+        </div>
+      </div><!-- /middle-header-->
+      <div class="middle-content">
+        <div class="container-fluid">
+          <alerts alerts="alerts"></alerts>
+          <div class="row">
+            <div class="col-md-12">
+              <table class="table table-bordered table-hover table-mobile table-layout-fixed">
```
spadgett: We've removed `table-hover` from our tables.

cdcabrera: Sure thing, think that's the only spot... but scanning back through the templates/views on this PR

cdcabrera:
@spadgett looks like there are 2 additional pieces that popped up again while running back through the review updates... can ignore or implement:
- First adding the green status check mark in there for "Ready" under the ```status-icon``` directive
  - [app/views/directives/_status-icon.html](https://github.com/openshift/origin-web-console/blob/master/app/views/directives/_status-icon.html)
```html
<span ng-switch-when="Ready" class="fa fa-check text-success" aria-hidden="true"></span>
```
- And the breadcrumbs seem to be back to generating one to many "crumbs" on the detail view on ```service-instance```... have a fix but it feels kind of gross... do a check for the breadcrumb array length in addition to the load check... up for a better way
  - [app/scripts/controllers/serviceInstance.js](https://github.com/openshift/origin-web-console/pull/1906/files#diff-44e681d6033bb7edf7fcdf36facb5392R29)
```js
var updateBreadcrumbs = function() {
  if(!$scope.loaded || $scope.breadcrumbs.length >= 2) {
    return;
  }
...
```

app/scripts/controllers/serviceInstance.js
```diff
@@ -0,0 +1,106 @@
...
+        $scope.project = project;
+        $scope.projectContext = context;
+
+        DataService
+          .get({
+            group: 'servicecatalog.k8s.io',
+            resource: 'serviceinstances'
+          }, $routeParams.instance, context, { errorNotification: false })
+          .then(function(service) {
+
```
spadgett: Call `updateBreadcrumbs` here and remove it from `serviceResolved`.

cdcabrera: Partially worked here... since the "crumb" display title/name is both partially dependent on ```serviceinstances``` and ```serviceclasses``` located an additional fire underneath the associated ```serviceclasses``` service call.

spadgett: Ah, I see. Then let's not set the breadcrumb until both load. Then we can call it safely from both callbacks.

cdcabrera: Removed the load check, instead now checking for ```serviceinstances``` and ```serviceclasses```. Similar to the check within ```updateServiceClassMetadata```.

spadgett:
@cdcabrera I'm seeing these errors on the overview in the console. Do you see this as well? Doesn't happen on master.

```
angular.js:14199 TypeError: resource.indexOf is not a function
    at normalizeResource (origin-web-common.js:2162)
    at Object.toResourceGroupVersion (origin-web-common.js:2150)
    at Object.canI (origin-web-common.js:2901)
    at ChildScope.$scope.show (nav.js:85)
    at fn (eval at compile (angular.js:15126), <anonymous>:4:297)
    at Scope.$digest (angular.js:17828)
    at angular.js:18033
    at completeOutstandingRequest (angular.js:6045)
    at angular.js:6324
```

app/scripts/constants.js
```diff
@@ -255,6 +255,23 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
...
+              isValid: function(){
+                return _.get(window.OPENSHIFT_CONSTANTS, 'ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page');
+              },
+              canI: {
+                resource: {
+                  group: 'servicecatalog.k8s.io',
+                  resource: 'serviceinstances'
+                },
+                verb: 'list'
+              }
```
spadgett: This should be

```js
canI: {
  resource: 'serviceinstances',
  group: 'servicecatalog.k8s.io',
  verb: 'list'
}
```

app/scripts/constants.js
```diff
@@ -255,6 +255,23 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
...
+            },
+            {
+              label: "Provisioned Services",
+              href: "/browse/service-instances",
+              prefixes: [
+                "/browse/service-instances/"
+              ],
+              isValid: function(){
+                return _.get(window.OPENSHIFT_CONSTANTS, 'ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page');
+              },
```
spadgett: Should be able to remove `isValid` now that #2078 is in the merge queue. The `canI` check below should hide the menu item when service catalog is not enabled.

cdcabrera:
@spadgett, yeah looks like I need to verify something and possible rebase

spadgett:
Yeah you'll want to rebase to pick up #2048 to properly test

app/scripts/constants.js
spadgett: #2078 has merged

cdcabrera:
@spadgett rebased, don't see any errors on my end.

Do we still need the [line in ```constants.js``` around ```isValid``` for the navigation](https://github.com/cdcabrera/origin-web-console/blob/5989859b8feb227b71c4eae34e8b762b7297d711/app/scripts/constants.js#L266)?

```js
isValid: function(){
  return _.get(window.OPENSHIFT_CONSTANTS, 'ENABLE_TECH_PREVIEW_FEATURE.service_catalog_landing_page');
},
```

spadgett:
You can remove `isValid` now

spadgett:
> don't see any errors on my end.

See my comment above on the `canI` check for the nav item. That's the cause.

https://github.com/openshift/origin-web-console/pull/1906#discussion_r139464874

app/views/browse/service-instance.html
```diff
@@ -0,0 +1,88 @@
...
+  <div class="middle-section">
+    <div class="middle-container">
+      <div class="middle-header">
+        <div class="container-fluid">
+          <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
+          <alerts alerts="alerts"></alerts>
+          <div ng-if="!loaded" class="mar-top-xl">Loading...</div>
+          <div ng-if="serviceInstance">
+            <h1 class="contains-actions">
+              <div class="pull-right dropdown" ng-hide="!('instances' | canIDoAny)">
```
spadgett: Needs to be `serviceInstances` here or the menu will never show up.

```html
<div class="pull-right dropdown" ng-if="'serviceInstances' | canIDoAny">
```

spadgett:
You need to update resourceServiceBindings.js with

```js
  ctrl.showBindings = CatalogService.SERVICE_CATALOG_ENABLED &&
                      (_.get(ctrl, 'apiObject.kind') === 'ServiceInstance' || enableTechPreviewFeature('pod_presets'));
```

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/resourceServiceBindings.js#L30

app/scripts/constants.js
cdcabrera: that would break, sure thing

spadgett:
[merge]

spadgett:
[merge]

spadgett:
> Origin Web Console Test Results: FAILURE (https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_web_console/232/) (Base Commit: 181d896) (PR Branch Commit: 0fd1a62)

Flake #1685

[test]

spadgett:
@cdcabrera You have a dist mismatch. Try rebase, `bower install`, and `grunt build`

cdcabrera:
Can do @spadgett

spadgett:
[merge]

PR #1929, opened 2017-08-11
[WIP] [DO_NOT_MERGE] pf header, vertical nav and secondary nav

sg00dwin:
@spadgett 
cc @rhamilto

PR #1932, opened 2017-08-14
Patternfly vertical navigation and project bar

spadgett:
https://trello.com/c/ZcLBaLK7

screen shot 2017-08-31 at 8 23 31 am

This is #1929, rebased

@sg00dwin I've left you as the commit author.
@rhamilto FYI

TODOs:

- [x] [P0] Navigating to Membership for the first time from catalog dropdown "View Membership" link results in permissions error.  But if you have previously visited the Membership page using the nav within the project and then navigate back using the catalog dropdown option, the bug doesn't occur.  Additionally, if you hard refresh on the Membership page, the bug also occurs.
- [x] UXD design feedback needed:
  - [x] product branding is always present
  - [x] project bar is always present if in a project
     - [x] remove project name from breadcrumb
  - [x] pages that didn't previously include sidebar nav should
- [x] make nav keyboard accessible --> see https://github.com/patternfly/patternfly/issues/723
- [x] persist collapsed nav state in `localStorage`
- [x] project bar permanently disappears when navigating away from chromeless logs page (fixed)
- [x] no bottom margin on add autoscaler page
- [x] fix mismatched tags in edit/build-config.html
- [x] clean up indentation in views
- [x] remove orphaned code
  - [x] look for orphaned LESS variables
- [x] Don't request self-subject rules reviews twice for a project -- fixed by https://github.com/openshift/origin-web-common/pull/165

Follow on:

- [ ] Update application launcher to use secondary nav flyout at mobile
- [ ] Update [Customizing the logo](https://docs.openshift.org/latest/install_config/web_console_customization.html#customizing-the-logo) documentation to include max width and height values documented in _navbar-vertical.less.
- [ ] Update branding for OCP per https://gist.github.com/rhamilto/c5c671d4dba9da2b8b62ff3fe0ad9f41#file-gistfile1-txt-L5-L20
 - [ ] Use HTML5 fullscreen for chromeless
 - [ ] Move YAML editor to a modal
 - [ ] determine if pages currently utilizing .surface-shaded should continue to do so --> Jenny to provide design direction
    - [ ] about, command line interface, projects, catalog, etc.
    - [ ] edit pages


app/scripts/extensions/nav/dropdownMobile.js
```diff
@@ -24,38 +24,38 @@ angular.module('openshiftConsole')
         dropdownItems = dropdownItems.concat([{
           type: 'dom',
           node: [
-            '<li>',
+            '<li class="list-group-item">',
```
spadgett: This will break existing extensions if we require a new class on menu items.

@rhamilto @jeff-phillips-18 Any ideas on how we can avoid that?

rhamilto: We've also got a bug that's related.  As it stands, primary nav links need an ng-click="nav.showMobileNav = false;" to correctly close the nav at mobile.

spadgett: I have this fixed except for active styles.

spadgett:
@dtaylor113 FYI, we should be able to update the application launcher menu to match the mobile design when this lands. I suggest we do it as a follow on to this PR.

app/styles/_variables.less
```diff
@@ -7,7 +7,15 @@
 // Patternfly overrides
 @dropdown-divider-margin: 4px 0;
 @list-view-hover-bg: #fafafa;
-@navbar-pf-alt-navbar-brand-padding: 20px 0;
+@navbar-pf-height:
```
spadgett: Needs a value?

app/styles/_variables.less
```diff
@@ -44,17 +52,16 @@
...
 // bottom margin when resizing the log in JavaScript.
 @middle-content-bottom-margin: @grid-gutter-width;
 @middle-content-container-padding: floor((@grid-gutter-width / 2)); // 20px bootstrap default
 @middle-content-container-padding-lg: (@middle-content-container-padding + 10px);
 @navbar-header-offset: 2px;
-@navbar-os-header-height-desktop: 60px;
-@navbar-os-header-height-mobile: 46px;
+@navbar-header-right-margin: 10px;
+@navbar-os-header-desktop: 59px;
+@navbar-os-header-mobile: 39px;
```
spadgett: Why remove `height` from the var name?

app/styles/_variables.less
sg00dwin: actually not used,  needs to be removed

app/styles/_variables.less
sg00dwin: can't recall why  🤔  but shouldn't have been.

app/scripts/extensions/nav/dropdownMobile.js
spadgett: The active styles work now too

spadgett:
I've added some basic accessibility support to the nav. Tested using the keyboard and iOS simulator.

app/views/_sidebar.html
```diff
@@ -1,58 +1,75 @@
...
-        <a ng-if="!primaryItem.href" href="" data-toggle="dropdown" class="dropdown-toggle">
-          <span class="{{primaryItem.iconClass}}"></span> {{primaryItem.label}} <span class="fa fa-angle-right"></span>
+        <a
+          ng-if="!primaryItem.href"
+          href=""
+          ng-click="itemClicked(primaryItem)">
+          <span class="{{primaryItem.iconClass}}"></span> <span class="list-group-item-value">{{primaryItem.label}}</span>
         </a>
+
+        <!-- mobile secondary nav -->
```
spadgett: I don't think this is just for mobile? It's secondary nav for desktop, too, right?

If so, let's update the comment.

spadgett:
@jeff-phillips-18 Let me know if you'd like to review this. It's a large PR, but the views are mostly whitespace changes and removed elements. The important changes are in nav.js, header.html, and _sidebar.html -- and then of course many changes to the less styles.

@rhamilto Is working on a few last items to show active nav for pages that didn't used to have a left nav. We are showing this at the design review today to verify what we've done with @openshift/team-ux-review 

We opted to go with base Patternfly to keep compatibility with our extension points. We have about 5 extension points in the masthead and left nav that we must keep supporting without breaking existing extensions.

I'd like to give this PR review priority if possible. For one, if we merge it earlier in the sprint, it has more soak time. It's also difficult to rebase since it touches many files.

@rhamilto Can you review the less changes from @sg00dwin?

@openshift/team-ux-review I'm really happy with the UX of the new nav. Thanks! This PR also removes more code than it adds, while adding function, which is great.

@rhamilto @sg00dwin Thanks for all your work on this

spadgett:
@abhgupta Heads up that nav changes are likely coming this sprint. You should not need to change any of the online extensions, but we should verify.

app/styles/_vertical-nav.less
```diff
@@ -0,0 +1,169 @@
...
+  }
+  .nav-item-pf-header {
+    font-size: (@font-size-base + 1);
+  }
+  .nav-pf-secondary-nav .list-group-item > a {
+    padding: 4px 0 2px 0;
+    margin-left: 20px;
+  }
+  .nav-vertical-primary {
+    display: none;
```
spadgett: Ideally we should use an `ng-if` for this rather than hiding the vertical nav in CSS

rhamilto:
> @abhgupta Heads up that nav changes are likely coming this sprint. You should not need to change any of the online extensions, but we should verify.

https://github.com/openshift/online/pull/1408 is required for the new nav to work correctly with the new nav.

spadgett:
Cleaned up indentation in the views, squashed, and added dist

spadgett:
[test]

spadgett:
Looks like we have some test failures to work through :)

spadgett:
@sg00dwin We should try to avoid the double scrollbars. I'm not sure the sidebar needs to scroll separately. (Can be a follow on.)

screen shot 2017-09-01 at 12 23 09 pm

spadgett:
Tests are passing. Thanks everyone.

[merge]

PR #1940, opened 2017-08-15
Prefill service labels when creating route

### LINKED ISSUE #1934: Adding app label to routes should be a default, not optional.
In transition from 1.4 to 1.5, the web console create route page allowed labels to be specified when creating the route. In doing that though, it stopped automatically adding the ``app`` label that it did previously. In order to have an ``app`` label added, you had to go through an additional step of clicking on _Copy Service Labels_, but that has actually been broken and doesn't work (https://bugzilla.redhat.com/show_bug.cgi?id=1477933).

The lack of an automatically added ``app`` label means that it is much harder to later delete all the components related to an application because a command of the form using a label selector:

```
oc delete all --selector app=name
```

will ignore the route.

It would be much better to auto populate the list of labels to include on the route by copying them from the service, and then allow the user to delete them or change them. This would bring it into alignment with how labels work on creating an application from the web console from an existing image or using a template.

image

This would also realign behaviour with what it was prior to 1.5, and also with how ``oc expose service`` works, which always automatically adds an ``app`` label.

spadgett:
Prefill the labels from the service by default when creating a route without making the user click "Copy Service Labels."

Fixes #1934

### DIFF
--- a/app/scripts/controllers/createRoute.js
+++ b/app/scripts/controllers/createRoute.js
@@ -76,16 +76,25 @@ angular.module('openshiftConsole')
           name: $scope.serviceName,
           weight: 1
         };
-        DataService.list("services", context).then(function(serviceData) {
-          servicesByName = serviceData.by("metadata.name");
-          $scope.services = orderByDisplayName(servicesByName);
-        });
 
-        $scope.copyServiceLabels = function() {
+        var serviceLabels;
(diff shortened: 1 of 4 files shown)


spadgett:
[merge][severity: bug]

spadgett:
[merge][severity: bug]

PR #1954, opened 2017-08-21
[WIP] Remove create project and edit project pages

jeff-phillips-18:
Removes the Create Project and Edit Project pages. 
Update the Create button on the projects page when there are no projects to use the create project popup.
Update e2e tests (in progress).


jeff-phillips-18:
[test]

app/scripts/controllers/projects.js
```diff
@@ -111,7 +111,12 @@ angular.module('openshiftConsole')
 
     $scope.newProjectPanelShown = false;
 
-    $scope.createProject = function() {
+    $scope.createProject = function(event) {
+      var button =_.get(event, 'target');
+      while (button && !$(button).hasClass('btn')) {
```
dtaylor113: Why is this a while loop?

jeff-phillips-18: It is searching for the button as a parent of the clicked element.

I will change to use angular.element rather than jquery.

spadgett:
Need to fix this empty state button before we can remove the new project page:

openshift web console 2017-08-22 13-19-50

spadgett:
Nevermind, I see you've already done that

benjaminapetersen:
Tests failing:

```
Fatal error: protractor exited with code: 1
```

Our protractor tests, this one:

```
origin-web-console/test/integration/features/user_creates_from_url.spec.js:45:7
```

Not sure why its having trouble with login, though we have issue with the login flow while running locally on mac.  The `create_from_url` is prob our most valuable e2e test as we rarely/never manually test this flow.

benjaminapetersen:
Re: #2083

PR #1956, opened 2017-08-21
Bug 1481127 - More robust handling of large numbers of projects

### LINKED ISSUE #1915: Project list filtering sluggish when you have hundreds of projects
If you have several hundred of projects, typing into the project list filter input stutters and feels unresponsive.

### LINKED ISSUE #1923: Web console is unusable with cluster-reader beyond a certain size.
It looks like the project list is being fetched in the background no matter what, but it's a massive list (10k projects in this case).  Above a certain size, it would probably be better to give up and not fetch it.

spadgett:
Updates origin-web-common to 0.0.47
Updates origin-web-catalog to 0.0.39
Updates uri.js to 1.18.12

- Don't watch extremely large project lists
- Don't request projects on every page load
- Increase debounce wait when filtering projects
- Use `$applyAsync` instead of `$apply` to avoid digest in progress errors
- Only show the first 250 projects on the project list. Users can filter or sort to see other projects
- Don't show more than 100 projects in the header dropdown. Users can navigate back to the project list if they have more.
- Let users type in the name of a project if the list fails.
- Add `track by` to project list `ng-repeat`

Requires https://github.com/openshift/origin-web-common/pull/150
Requires https://github.com/openshift/origin-web-catalog/pull/391

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1481127
Closes #1915
Closes #1923 

TODO:

- [x] Limit number of items in header dropdown
- [x] Limit number of items on projects page?
- [x] Disable project list watch is above a certain size?
- [x] Prompt user for project name on project list when the list can't be loaded
- [x] Increase debounce interval when filtering project list

### DIFF
--- a/app/views/projects.html
+++ b/app/views/projects.html
@@ -64,12 +64,46 @@
                         </div>
                       </div>
                     </div>
-                    <div ng-if="!projects.length" class="h3">
+                    <div ng-if="isProjectListIncomplete">
+                      <div class="alert alert-warning">
+                        <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
+                        <span class="sr-only">Warning:</span>
+                        The complete list of your projects could not be loaded.
+                        Type a project name to go to that project.
+                      </div>
+                      <form>
+                        <div class="form-group">
+                          <label for="typed-project-name">Project Name</label>
+                          <div class="input-group">
+                            <input
+                              class="form-control"
+                              type="text"
+                              id="typed-project-name"
+                              required
+                              minlength="2"
+                              ng-model="input.typedProjectName"
+                              autocorrect="off"
+                              autocapitalize="none"
+                              spellcheck="false">
+                            <span class="input-group-btn">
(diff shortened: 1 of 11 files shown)


spadgett:
@jeff-phillips-18 Still testing, but PTAL

app/views/projects.html
```diff
@@ -110,16 +144,14 @@ <h2 class="h1">
...
                                   label="Delete Project"
-                                  resource-name="{{project.metadata.name}}"
-                                  project-name="{{project.metadata.name}}"
-                                  display-name="{{(project | displayName)}}"
+                                  project="project"
                                   type-name-to-confirm="true"
                                   stay-on-current-page="true"
-                                  alerts="alerts">
-                                </delete-link>
+                                  alerts="alerts"
```
jeff-phillips-18: The deleteProject directive does not use the alerts parameter.

spadgett:
[merge][severity: blocker]

PR #1957, opened 2017-08-21
add bower link instructions to README              

benjaminapetersen:
Adds a block under `Contributing` for using `bower link` to work in our other repos.

@jeff-phillips-18 
@spadgett

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,44 @@ If Jenkins complains that the built dist files are different than the committed 
 3. Run `grunt build`
 4. If anything under dist or dist.java has changed, add it to your commit and re-push
 
+#### Contributing to the primary repositories
+
+The web console is currently split into three repositories.  The two dependency repos are
+[origin-web-common](https://github.com/openshift/origin-web-common) and
+[origin-web-catalog](https://github.com/openshift/origin-web-catalog).
+To make changes to one of these repositories while working in the web console, it is recommended that you clone down the
+repository and create a `bower link`.  The following example assumes you clone your forks to `~/git-repos`:
+
+```bash
+# fork the origin-web-console & clone:
+$ cd ~/git-repos
+$ git clone git@github.com:<your-fork>/origin-web-console.git
+# fork origin-web-common & clone:
+$ cd ~/git-repos
+$ git clone git@github.com:<your-fork>/origin-web-common.git
+# fork origin-web-catalog & clone:
+$ cd ~/git-repos
+$ git clone git@github.com:<your-fork>/origin-web-catalog.git
+#
+# Now, using bower link you can:
+$ cd ~/git-repos/origin-web-common
+$ bower link
+$ cd ~/git-repos/origin-web-catalog
+$ bower link
+$ cd ~/git-repos/origin-web-console
+$ bower link origin-web-common
+$ bower link origin-web-catalog
+#
+# NOTE:
+# When you make changes in the linked repos you will need to rebuild
+# as origin-web-console pulls from the /dist.  `grunt build` or `grunt watch`
+# should take care of this.
+```
+
+When finished, be sure to `hack/clean-deps.sh` and `hack/install-deps.sh` to remove
+the links & avoid having issues with `/dist` conflicts the next time you build.
+
+
 Architecture
 ------------
 


README.md
```diff
@@ -92,6 +92,40 @@ If Jenkins complains that the built dist files are different than the committed
...
+# fork origin-web-catalog & clone:
+$ cd ~/git-repos
+$ git clone git@github.com:<your-fork>/origin-web-catalog.git
+#
+# Now, using bower link you can:
+$ cd ~/git-repos/origin-web-common
+$ bower link
+$ cd ~/git-repos/origin-web-catalog
+$ bower link
+$ cd ~?git-repos/origin-web-console
```
jeff-phillips-18: ?  should be /

benjaminapetersen: +1 got it

README.md
```diff
@@ -92,6 +92,40 @@ If Jenkins complains that the built dist files are different than the committed
...
+$ git clone git@github.com:<your-fork>/origin-web-catalog.git
+#
+# Now, using bower link you can:
+$ cd ~/git-repos/origin-web-common
+$ bower link
+$ cd ~/git-repos/origin-web-catalog
+$ bower link
+$ cd ~/git-repos/origin-web-console
+$ bower link origin-web-common
+$ bower link origin-web-catalog
```
spadgett: Do we need to tell users how to unlink? Will you have built dist conflicts if you don't?

@jeff-phillips-18 may know better

benjaminapetersen: `bower link` is just a symlink bower maintains so it shouldn't blow up across `bower install`s, etc.

You can definitely have `/dist` conflicts if one of the `link`ed repos is on another branch, etc.

`bower unlink` is an alias of `bower uninstall`, which just removes the dependency for a re-install the normal way.  

We can outline all of this if you think its necessary.

benjaminapetersen: Updated, [how about this?](https://github.com/benjaminapetersen/origin-web-console/blob/c9b88764661a881166145d8836ee915a0a50271a/README.md#contributing-to-the-primary-repositories)

benjaminapetersen: Working out the kinks of it right now on other branches 😄

spadgett: > You can definitely have /dist conflicts if one of the linked repos is on another branch, etc.

Even if master is ahead of the tagged version that origin-web-console uses, wouldn't it  cause a dist mismatch? So we'd need instructions for dealing with that

benjaminapetersen: Hmm, yeah this is problematic. Options:
- Clone `origin-web-console` twice, have a copy with the `bower links` to work on the other repos.  Wheeeee.....
- See if we can improve on unit/integration testing instead of doing all this link stuff to manually verify things work as intended
- Setup a little demo app in the secondary repos to manually test against
- ???

All the options have a little bit of janky workflow in them.

README.md
```diff
@@ -92,6 +92,47 @@ If Jenkins complains that the built dist files are different than the committed
...
+# When  you make changes in the `/src` directories,
+# you will likely have to `grunt build` to see the updates.
+#
+# Reduce your complexity and clean up:
+# when you have finished your work, `bower unlink <repo>` & `bower install` to return to using
+# the version of the repo that has been committed to the package.json file. If you don't, you
+# may end up with frequent `/dist` conflicts as `grunt build` will use the linked local copy
+# in whatever state it happens to be in.  
+```
+Alternatively, you could `cd` into `/origin-web-console/bower_components` and create symlinks for `origin-web-catalog` and `origin-web-common`, but you will likely have to recreate these each time you `bower install`.
```
spadgett: I think we can just leave this out since `bower link` just seems better

README.md
```diff
@@ -92,6 +92,47 @@ If Jenkins complains that the built dist files are different than the committed
...
+$ bower link
+$ cd ~/git-repos/origin-web-console
+$ bower link origin-web-common
+$ bower link origin-web-catalog
+#
+# When  you make changes in the `/src` directories,
+# you will likely have to `grunt build` to see the updates.
+#
+# Reduce your complexity and clean up:
+# when you have finished your work, `bower unlink <repo>` & `bower install` to return to using
```
spadgett: Does that work? I'm thinking it might remove the dependency entirely from bower.json.

Let's just recommend

```
$ hack/clean-assets.sh
$ hack/install-assets.sh
```

to clean up.

benjaminapetersen:
updated

spadgett:
[merge][severity: lowrisk]

PR #1959, opened 2017-08-22
Fix problem where UI appears unresponsive while waiting for websocket update

### LINKED ISSUE #1875: Resume rollouts link appears to do nothing if watch updates are slow or websocket is disconnected
The resume rollouts action on the overview and browse pages doesn't disappear on click until after we get a watch update saying that `spec.paused` is set to false. This means the link appears to do nothing if the websocket update doesn't happen immediately.

@smarterclayton Resume rollouts is working for me in current master for both deployments and deployment configs, so I suspect it is in fact slow watches.

benjaminapetersen:
Fix #1875 
@spadgett 

(aside, i think the dist having something to the effect of `return !0` for `return true` is kinda funny... optimize out those 2 characters 😄 )

### DIFF
--- a/app/scripts/services/resourceAlerts.js
+++ b/app/scripts/services/resourceAlerts.js
@@ -101,6 +101,7 @@ angular.module("openshiftConsole")
                     details: $filter('getErrorDetails')(e)
                   };
                 });
+              return true;
             }
           }]
         };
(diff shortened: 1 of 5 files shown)


spadgett:
We also have the issue on the deployment config and deployment pages

benjaminapetersen:
Looking at these.

benjaminapetersen:
Updated.
Interested in reducing the code duplication & reusing `alerts` here, but prob not in scope for a bug fix.  Something like:

```html
  <alerts 
    ng-if="deployment.spec.paused && !updatingPausedState"
    alerts="pausedAlert"></alerts>
```

benjaminapetersen:
This can go today as well, if we can wake up the merge queue...

spadgett:
[merge]

PR #1963, opened 2017-08-23
Partial fix issue 620 kve sluggish

benjaminapetersen:
This is likely a very limited improvement.  I think the actual fix for #620 would be to disable the KVE whenever something is changing.

### DIFF
--- a/app/views/directives/key-value-editor.html
+++ b/app/views/directives/key-value-editor.html
@@ -20,7 +20,7 @@
       class="key-value-editor-entry"
       ng-class-odd="'odd'"
       ng-class-even="'even'"
-      ng-repeat="entry in entries"
+      ng-repeat="entry in entries track by $index"
       as-sortable-item>
 
       <!-- The name/key block -->
--- a/dist/scripts/templates.js
+++ b/dist/scripts/templates.js
@@ -7279,7 +7279,7 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
     "</div>\n" +
     "</div>\n" +
     "</div>\n" +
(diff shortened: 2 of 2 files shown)


benjaminapetersen:
@spadgett

benjaminapetersen:
I tried incrementally increasing the # of lines rendered via `$interval`. Less than 50 was ok.  Once it got in the range of 100-500 env vars renders became pretty slow. I think its the typical "too many two-way bindings in angular" problem.

I can do some more tinkering like what you mentioned to see if we can get a certain fix.

spadgett:
Did the `track by` make it noticeably better?

Try capturing performance in Chrome dev tools to see where the time is spent.

benjaminapetersen:
I don't believe `track by` gives us any value.  As I have been testing w/100 rows I haven't found it to make any difference.

Doing some profiling:

Page load is a little gnarly, but I think we expect that:
profile_kve_page_load

Deleting a value:
profile_kve_delete_valuefrom

Drag & Drop a value:
profile_kve_drag_drop

We get a few flags, but at 100 key-value pairs on the page, the UI is still responsive.  I think what you mentioned about `debouncing` is worth a look, I'm digging into that as well.

benjaminapetersen:
Testing a debounce by updating:

```JavaScript 
var updateAPIObject = _.debounce(function(currentValue, previousValue) { 
  // delay...
}, 200,400,800,1000);  // one at a time, obviously :)
```
The longer the debounce the better, but UI is sluggish no matter what when drag/drop during an update.  

Testing a simple overlay as well that basically disables the KVE if an update is in progress (simple case: if `updateAPIObject` is called, block the controls for debounced 1 second & enable again).  Not sure if this is a great solution either, but worth some exploration.

benjaminapetersen:
Not confident in this one yet. Holding off.

PR #1966, opened 2017-08-25
Add 'Select from Project' wizard to allow project templatesto be imported

jeff-phillips-18:
https://trello.com/c/NFhQawHD

Requires https://github.com/openshift/origin-web-catalog/pull/394

image

image

image

image

cc @openshift/team-ux-review


app/scripts/directives/processTemplateDialog.js
```diff
@@ -4,20 +4,49 @@
...
+                                 Catalog,
+                                 DataService,
+                                 KeywordService,
+                                 NotificationsService,
+                                 ProjectsService,
+                                 RecentlyViewedProjectsService) {
     var ctrl = this;
     var validityWatcher;
+    var projectListWatcher;
+    var MAX_PROJECTS_TO_WATCH = 250;
```
benjaminapetersen: Interested in this constant, wondering if it should be configurable via constants.js?

Curious if we will actually be watching 250 projects at one time.  Thats a lot of watches 😄

app/scripts/directives/processTemplateDialog.js
```diff
@@ -4,20 +4,49 @@
...
+
+    ctrl.selectStep = {
+      id: 'projectTemplates',
+      label: 'Selection',
+      view: 'views/directives/process-template-dialog/process-template-select.html',
+      hidden: ctrl.useProjectTemplate !== true,
```
benjaminapetersen: `ctrl.useProjectTemplate !== true` vs `!ctrl.useProjectTemplate`?  Non-boolean options possible?

app/scripts/directives/processTemplateDialog.js
```diff
@@ -124,5 +234,82 @@
     function instantiateTemplate() {
       $scope.$broadcast('instantiateTemplate');
     }
+
+    function filterForKeywords (searchText, items) {
+      return KeywordService.filterForKeywords(items, ['name', 'tags'], KeywordService.generateKeywords(searchText));
+    }
+
+    function filterChange (filters) {
```
benjaminapetersen: Tiny nit: some inconsistent parens spacing in this file `function filterChange (filters) ` vs `function filterItems()`

app/scripts/directives/processTemplateDialog.js
```diff
@@ -124,5 +234,82 @@
...
+      updateFilterControls();
+
+      // Deselect the currently selected template if it was filtered out
+      if (!_.includes(ctrl.filteredItems, ctrl.selectedTemplate)) {
+        ctrl.templateSelected();
+      }
+    }
+
+    function updateFilterControls() {
+      $timeout(function() {
```
benjaminapetersen: Always have to ask if `$` is needed/should be used or if an "angular native" way can be achieved via template bindings.

app/scripts/directives/processTemplateDialog.js
```diff
@@ -124,5 +234,82 @@
...
+    }
+
+    function filterItems() {
+      ctrl.filteredItems = ctrl.catalogItems;
+      if (ctrl.filterConfig.appliedFilters && ctrl.filterConfig.appliedFilters.length > 0) {
+        _.each(ctrl.filterConfig.appliedFilters, function(filter) {
+          ctrl.filteredItems = filterForKeywords(filter.value, ctrl.filteredItems);
+        });
+      }
+
```
benjaminapetersen: Curious if order of flow here matters, since what is inside `updateFilterControls()` is wrapped in `$timeout`...ie, could this call be the last line in this fn since its contexts will execute later?

app/scripts/directives/processTemplateDialog.js
jeff-phillips-18: I followed the lead from https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/projects.js#L22

We should probably define a constant somewhere. I will add a comment for now.

app/scripts/directives/processTemplateDialog.js
jeff-phillips-18: It's a passed in parameter, I am defaulting to false unless explicitly set to true.

app/scripts/directives/processTemplateDialog.js
jeff-phillips-18: I've removed the $timeout as it doesn't appear to be necessary.

app/scripts/directives/processTemplateDialog.js
jeff-phillips-18: Fixed

app/scripts/directives/processTemplateDialog.js
jeff-phillips-18: Adjusted, not sure what you were getting at but the timeout was unnecessary.

benjaminapetersen: Great!  I just have a tendency to pause and ask a question when I see `$timeout`, they are good at hiding bugs.  Wasn't sure about your code here, but figured I'd mention just in case.

jeff-phillips-18:
I've updated the less to fix the spacing issue for the horizontal rule

jeff-phillips-18:
image

app/scripts/directives/processTemplateDialog.js
```diff
@@ -124,5 +234,78 @@
...
+      updateProjects();
+    };
+
+    function listProjects() {
+      if (!projectListWatcher) {
+        ProjectsService.list().then(function(projectData) {
+          updateProjectData(projectData);
+
+          if (!ProjectsService.isProjectListIncomplete() && _.size(ctrl.unfilteredProjects) <= MAX_PROJECTS_TO_WATCH) {
+            projectListWatcher = ProjectsService.watch($scope, updateProjectData);
```
spadgett: I'm not sure it's worth a watch here. I'd probably just list projects to populate the dropdown. This is consistent with what we do in other project selects in the console.

app/scripts/directives/processTemplateDialog.js
```diff
@@ -43,6 +72,34 @@
 
     ctrl.$onInit = function() {
       ctrl.loginBaseUrl = DataService.openshiftAPIBaseUrl();
+      ctrl.preSelectedProject = ctrl.selectedProject = ctrl.project;
+      listProjects();
+
+      ctrl.projectEmptyState = {
+        icon: 'pficon pficon-info',
+        title: 'No Project Selected',
+        info: 'Please select a project from the dropdown to load Templates and Images from that project'
```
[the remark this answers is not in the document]
jeff-phillips-18: OK, fixed.

app/scripts/directives/processTemplateDialog.js
jeff-phillips-18: Removed the watch.

jeff-phillips-18:
@spadgett Updated to not pull the image streams into the selection list. This dialog will only handle the templates at this point.

benjaminapetersen:
How does it look with lots of templates & images?  If you give that user access to `openshift` namespace & can see a huge list?

Asking because it is a little awkward feeling with just one item, but I bet it looks better when there is more.

jeff-phillips-18:
@benjaminapetersen I can add a bunch of fake templates if you'd like to see.

app/scripts/directives/processTemplateDialog.js
```diff
@@ -43,6 +68,34 @@
...
+      ctrl.projectEmptyState = {
+        icon: 'pficon pficon-info',
+        title: 'No Project Selected',
+        info: 'Please select a project from the dropdown to load Templates and Images from that project.'
+      };
+
+      ctrl.templatesEmptyState = {
+        icon: 'pficon pficon-info',
+        title: 'No Templates or Images',
+        info: 'The selected project has no templates or images available to import.'
```
spadgett: We'll want to remove "images" from these messages if there are only templates. Maybe

"The selected project has no templates available to instantiate."

(When we add images back, we'll want to say specifically "builder images.")

app/scripts/directives/processTemplateDialog.js
```diff
@@ -43,6 +68,34 @@
 
     ctrl.$onInit = function() {
       ctrl.loginBaseUrl = DataService.openshiftAPIBaseUrl();
+      ctrl.preSelectedProject = ctrl.selectedProject = ctrl.project;
+      listProjects();
+
+      ctrl.projectEmptyState = {
+        icon: 'pficon pficon-info',
+        title: 'No Project Selected',
+        info: 'Please select a project from the dropdown to load Templates and Images from that project.'
```
spadgett: 'Please select a project from the dropdown to load templates from that project.'

app/scripts/directives/processTemplateDialog.js
```diff
@@ -124,5 +227,70 @@
...
+    function filterForKeywords(searchText, items) {
+      return KeywordService.filterForKeywords(items, ['name', 'tags'], KeywordService.generateKeywords(searchText));
+    }
+
+    function filterChange(filters) {
+      // only use filters which have a filter criteria 'value'
+      // prevents applying an empty keyword filter
+      // TODO: can remove the following line of code after angular-patternfly issue is fixed:
+      //   https://github.com/patternfly/angular-patternfly/issues/509
+      filters = _.filter(filters, 'value');
```
spadgett: I think this can be removed now?

jeff-phillips-18:
Removed WIP status, ready for final reviews.

app/scripts/directives/processTemplateDialog.js
```diff
@@ -124,5 +231,60 @@
...
+      if (ctrl.totalCount <= 1) {
+        $('.filter-pf.filter-fields input').attr('disabled', '');
+      } else {
+        $('.filter-pf.filter-fields input').removeAttr("disabled");
+      }
+    }
+
+    var updateProjects = function() {
+      var filteredProjects = _.reject(ctrl.unfilteredProjects, 'metadata.deletionTimestamp');
+      var projects = _.sortBy(filteredProjects, $filter('displayName'));
```
dtaylor113: Hi, this ```_.sortBy(...)``` is unnecessary as ```RecentlyViewedProjectsService.orderByMostRecentlyViewed(projects)``` handles a case insensitive sort of displayName.

spadgett:
[merge]

PR #1967, opened 2017-08-25
Bump origin-web-common to 0.0.48

spadgett:
Fixes a runtime error when the project list doesn't load.

@rhamilto @sg00dwin This also will fix the problem we were seeing with the resize listener in #1932


spadgett:
[test]

spadgett:
[merge][severity: blocker]

PR #1968, opened 2017-08-25
bug 1480988. Use project annotation to determine kibana url for ops namespaces

jcantrill:
This PR provides a fix to BZ1480988 to direct users to an operations instance of Kibana when the logging stack is deployed to have the ops logging cluster.  This PR depends on openshift/openshift-ansible#5219

### DIFF
--- a/app/scripts/directives/logViewer.js
+++ b/app/scripts/directives/logViewer.js
@@ -431,7 +431,7 @@ angular.module('openshiftConsole')
             // Kibana archives -------------------------------------------------
 
             APIDiscovery
-              .getLoggingURL()
+              .getLoggingURL($scope.context.project)
               .then(function(url) {
                 var projectName = _.get($scope.context, 'project.metadata.name');
                 var containerName = _.get($scope.options, 'container');
--- a/app/scripts/services/discovery.js
+++ b/app/scripts/services/discovery.js
@@ -1,12 +1,18 @@
 'use strict';
 
 angular.module('openshiftConsole')
-  .factory('APIDiscovery', ['LOGGING_URL', 'METRICS_URL', '$q', function(LOGGING_URL, METRICS_URL, $q) {
+  .factory('APIDiscovery', ['LOGGING_URL', 'METRICS_URL', '$q', '$filter', function(LOGGING_URL, METRICS_URL, $q, $filter) {
     return {
       // Simulate asynchronous requests for now. If these are ever updated to call to a discovery
       // endpoint, we need to make sure to trigger a digest loop using (or update all callers).
-      getLoggingURL: function() {
-        return $q.when(LOGGING_URL);
+      getLoggingURL: function(project) {
+        var loggingURL = LOGGING_URL;
+        var loggingUIHostname = $filter('annotation')(project, 'loggingUIHostname');
+        if(loggingUIHostname) {
+           loggingURL = 'https://' + loggingUIHostname;
+        }
+
+        return $q.when(loggingURL);
       },
       getMetricsURL: function() {
         return $q.when(METRICS_URL);
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -3,7 +3,7 @@
 function OverviewController(e, t, n, a, r, o, i, s, c, l, u, d, m, p, g, f, h, v, y, b, C, S, w, k, j) {
 var P = this, R = t("isIE")() || t("isEdge")();
 e.projectName = n.project;
-var E, T, N = t("annotation"), I = t("buildConfigForBuild"), D = t("deploymentIsInProgress"), A = t("imageObjectRef"), B = t("isJenkinsPipelineStrategy"), L = t("isNewerResource"), U = t("label"), O = t("podTemplate"), x = {}, F = {}, M = {}, V = P.state = {
+var E, T, I = t("annotation"), N = t("buildConfigForBuild"), D = t("deploymentIsInProgress"), A = t("imageObjectRef"), B = t("isJenkinsPipelineStrategy"), L = t("isNewerResource"), U = t("label"), O = t("podTemplate"), x = {}, F = {}, M = {}, V = P.state = {
 alerts: {},
 builds: {},
 clusterQuotas: {},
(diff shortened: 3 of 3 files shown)


app/scripts/directives/logViewer.js
```diff
@@ -401,6 +401,10 @@ angular.module('openshiftConsole')
                   return;
                 }
```
benjaminapetersen: Spelling of `metadata`:`context.project.metdata` vs `context.project.metadata`.

So most of our annotations are on a `filter` that, unfortunately, is now pulled in via a dependency from [origin-web-common](https://github.com/openshift/origin-web-common/blob/master/src/filters/annotation.js#L8).  We should probably add it there...

jcantrill: @benjaminapetersen fixed the spelling

benjaminapetersen: thanks!  so by annotation, i mean making a [PR here](https://github.com/openshift/origin-web-common/blob/master/src/filters/annotation.js#L8) simply to add the line:

```JavaScript 
var annotationMap = {
  // truncated
  loggingKibanaUrl: ['openshift.io/logging.kibana.url']
}
```
And then when that PR merges, can update this PR to use the filter:

```JavaScript
url = "https://" + $scope.context.project.metadata.annotations[$filter('annotationName')('loggingKibanaUrl')];
```

jcantrill:
Depends on openshift/origin-web-common/pull/158

app/scripts/directives/logViewer.js
```diff
@@ -401,6 +401,10 @@ angular.module('openshiftConsole')
...
 
+                if($scope.context.project.metadata.annotations[$filter('annotationName')('loggingVisualURL') {
+                    url = 'https://' + $scope.context.project.metadata.annotations[$filter('annotationName')('loggingVisualURL');
+                }
```
spadgett: @jcantrill It would be better to change `APIDiscovery.getLoggingURL` to take the project as a parameter and move the logic there. Then if we need the logging URL anywhere else, we don't forget to check the project annotation.

I'd suggest this code:

```js
var projectLoggingURL = $filter('annotation')('loggingVisualURL');
if (projectLoggingURL) {
  url = 'https://' + projectLoggingURL;
}

**Comment by benjaminapetersen** [comment]
@jcantrill `grunt build` & commit the `dist` as well & travis should pass.

**Comment by jcantrill** [comment]
@spadgett Updated correctly?

File: app/scripts/services/discovery.js [hunk 3]
```diff
@@ -1,11 +1,15 @@
...
-  .factory('APIDiscovery', ['LOGGING_URL', 'METRICS_URL', '$q', function(LOGGING_URL, METRICS_URL, $q) {
+  .factory('APIDiscovery', ['LOGGING_URL', 'METRICS_URL', '$q', '$filter', function(LOGGING_URL, METRICS_URL, $q, $filter) {
     return {
       // Simulate asynchronous requests for now. If these are ever updated to call to a discovery
       // endpoint, we need to make sure to trigger a digest loop using (or update all callers).
-      getLoggingURL: function() {
+      getLoggingURL: function(project) {
+        if(project.metadata.annotations[$filter('annotationName')('loggingUIHostname')]) {
+           return 'https://' + project.metadata.annotations[$filter('annotationName')('loggingUIHostname')];
+        }
```
spadgett: I think this is easier to read (untested)

```js
var loggingUIHostname = $filter('annotation')(project, 'loggingUIHostname');
if (loggingUIHostname) {
  return 'https://' + loggingUIHostname;
}
```

spadgett:
Will require an origin-web-common release and `dist` files

app/scripts/services/discovery.js
jcantrill: agree.

app/scripts/services/discovery.js
```diff
@@ -1,11 +1,16 @@
...
 
 angular.module('openshiftConsole')
-  .factory('APIDiscovery', ['LOGGING_URL', 'METRICS_URL', '$q', function(LOGGING_URL, METRICS_URL, $q) {
+  .factory('APIDiscovery', ['LOGGING_URL', 'METRICS_URL', '$q', '$filter', function(LOGGING_URL, METRICS_URL, $q, $filter) {
     return {
       // Simulate asynchronous requests for now. If these are ever updated to call to a discovery
       // endpoint, we need to make sure to trigger a digest loop using (or update all callers).
-      getLoggingURL: function() {
+      getLoggingURL: function(project) {
+        loggingUIHostname = $filter('annotationName')(project, 'loggingUIHostname');
```
spadgett: Missing `var`.

```js
var loggingUIHostname = $filter('annotationName')(project, 'loggingUIHostname');
```

app/scripts/services/discovery.js
```diff
@@ -1,11 +1,16 @@
...
+        if(loggingUIHostname) {
+           return 'https://' + loggingUIHostname;
```
spadgett: Sorry @jcantrill missed this on last review, but this needs to be wrapped in a `$q.when` since the method returns a promise.

```js
return $q.when('https://' + loggingUIHostname);
```

Or you can change it to have one return.

```js
var loggingURL = LOGGING_URL;
var loggingUIHostname = $filter('annotationName')(project, 'loggingUIHostname');
if(loggingUIHostname) {
  loggingURL = 'https://' + loggingUIHostname;
}

return $q.when(loggingURL);
```

app/scripts/services/discovery.js
jcantrill: fixed

app/scripts/services/discovery.js
jcantrill: fixed

jcantrill:
@spadgett I made the requested changes but should I expect all the changes found in dist and dist.java ?

spadgett:
No, you shouldn't have that. Try

```
$ hack/clean-asssets.sh
$ hack/install-assets.sh
$ grunt build
```

spadgett:
@liggitt We post the user's access token to the logging URL. Just confirming that the annotation can only be changed by admins and that this is safe.

app/scripts/services/discovery.js
```diff
@@ -1,12 +1,18 @@
...
-  .factory('APIDiscovery', ['LOGGING_URL', 'METRICS_URL', '$q', function(LOGGING_URL, METRICS_URL, $q) {
+  .factory('APIDiscovery', ['LOGGING_URL', 'METRICS_URL', '$q', '$filter', function(LOGGING_URL, METRICS_URL, $q, $filter) {
     return {
       // Simulate asynchronous requests for now. If these are ever updated to call to a discovery
       // endpoint, we need to make sure to trigger a digest loop using (or update all callers).
-      getLoggingURL: function() {
-        return $q.when(LOGGING_URL);
+      getLoggingURL: function(project) {
+        var loggingURL = LOGGING_URL;
+        var loggingUIHostname = $filter('annotationName')(project, 'loggingUIHostname');
```
spadgett: $filter('annotation')

jcantrill:
@spadgett tested.... works like a champ...

spadgett:
[merge]

spadgett:
@jeff-phillips-18 Looks like this needs to be rebased again :/

Issue #1438, opened 2017-04-18
Viewers see empty env var input rather than message when no env vars


spadgett:
I thought this was a little odd. I logged in as a user with `view` role in the project. Instead of seeing a message saying there are no env vars, I see two readonly inputs.

![screen shot 2017-04-18 at 12 47 15 pm](https://cloud.githubusercontent.com/assets/1167259/25142536/36ab70a2-2435-11e7-99b0-e03e1919838c.png)

benjaminapetersen:
hmm. yeah that is weird. ill take a look

benjaminapetersen:
I can reproduce this.  So Its probably something we need to either 1. put inside KVE, or 2. update all the templates where KVE appears with an `ng-if` & an alt message.   @rhamilto 

Do we have a preference either way? I lean towards the `ng-if` since we are not limited to just env vars.

spadgett:
For replication controllers we do this:

https://github.com/spadgett/origin-web-console/blob/6c0257867c68da506b49a3e42fd6476e035c2863/app/views/browse/replica-set.html#L79

![screen shot 2017-04-18 at 1 35 43 pm](https://cloud.githubusercontent.com/assets/1167259/25144488/f4241318-243b-11e7-87f1-103bd3c98e07.png)

rhamilto:
> Do we have a preference either way? I lean towards the ng-if since we are not limited to just env vars.

Agreed.  No point in showing the KVE if there is nothing to display (i.e., read only env vars).

benjaminapetersen:
Ok, I can find 10 instances of KVE in readonly mode, I believe each one should have a message like this, but will check them over.

PR #1970, opened 2017-08-25
Ensure kve does not render on browser builds if no envs to show

benjaminapetersen:
related to #1438 

There is already a good message `The build strategy had no environment variables defined.`, this just ensures the `kve` doesn't render at all (rather than renders but is empty).

@spadgett

### DIFF
--- a/app/views/browse/build.html
+++ b/app/views/browse/build.html
@@ -99,6 +99,7 @@
                   Environment variables can be edited on the <a ng-href="{{build | configURLForResource}}?tab=environment">build configuration</a>.
                 </p>
                 <key-value-editor
+                  ng-if="(build | buildStrategy).env | size"
                   entries="(build | buildStrategy).env"
                   key-placeholder="Name"
                   value-placeholder="Value"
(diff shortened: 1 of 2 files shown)


app/views/browse/build.html
```diff
@@ -104,6 +104,7 @@ <h1 class="contains-actions">
                         Environment variables can be edited on the <a ng-href="{{build | configURLForResource}}?tab=environment">build configuration</a>.
                       </p>
                       <key-value-editor
+                        ng-if="(build | buildStrategy).env"
```
spadgett: Should be checking size to catch empty arrays?

```
ng-if="(build | buildStrategy).env | size"
```

app/views/browse/build.html
```diff
@@ -104,6 +104,7 @@ <h1 class="contains-actions">
                         Environment variables can be edited on the <a ng-href="{{build | configURLForResource}}?tab=environment">build configuration</a>.
                       </p>
                       <key-value-editor
```
benjaminapetersen: @spadgett thx, updated

spadgett:
Hey @benjaminapetersen I think if you rebase this, we should be set to merge

benjaminapetersen:
thanks @spadgett , rebased.

spadgett:
[merge][severity: bug]

PR #1972, opened 2017-08-28
[3.5] Bug 1485892 - Fix overview hang when name only digits

spadgett:


spadgett:
[test]

spadgett:
Jenkins failure is because the job is using `cluster up --version=latest`, which is missing the old `extensions/horizontalpodautoscalers` API group.

spadgett:
@stevekuznetsov Is it possible for PRs against older release branches to cluster up with the corresponding version?

master -> latest
enterprise-3.6 -> v3.6.0
enterprise-3.5 -> v1.5.0

Right now latest is hard-coded:

https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_branch_origin_web_console.yml#L13

spadgett:
@jeff-phillips-18 PTAL. The vendor diff is a little concerning.

spadgett:
> Looks like some bower package got updated?

Yeah looks like at least c3.js was updated

PR #1975, opened 2017-08-28
Fix 1974 build page TypeError

### LINKED ISSUE #1974: Builds page may throw a TypeError if object not found.
url: 
`https://localhost:9000/dev-console/project/myproject/browse/builds`

Error:
screen shot 2017-08-28 at 11 50 46 am

Problem line `66`
screen shot 2017-08-28 at 11 51 40 am

benjaminapetersen:
fixes #1974

@spadgett

### DIFF
--- a/app/scripts/controllers/build.js
+++ b/app/scripts/controllers/build.js
@@ -63,7 +63,7 @@ angular.module('openshiftConsole')
     var watches = [];
 
     var setLogVars = function(build) {
-      $scope.logCanRun = !(_.includes(['New', 'Pending', 'Error'], build.status.phase));
+      $scope.logCanRun = !(_.includes(['New', 'Pending', 'Error'], _.get(build, 'status.phase')));
     };
 
     var updateCanBuild = function() {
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/build.js
```diff
@@ -63,7 +63,7 @@ angular.module('openshiftConsole')
     var watches = [];
 
     var setLogVars = function(build) {
-      $scope.logCanRun = !(_.includes(['New', 'Pending', 'Error'], build.status.phase));
+      $scope.logCanRun = !(_.includes(['New', 'Pending', 'Error'], _.get(build, 'status.phase')));
```
spadgett: I'm confused how `buildResolved` is called with a null value? Just worried the problem runs deeper.

benjaminapetersen: I noticed it due to a timeout with an API response (I think).

spadgett: Shouldn't that have gone to the `buildRejected` callback?

benjaminapetersen: That would make sense.   Hmm. Not sure the cause then.

PR #1976, opened 2017-08-28
Add scripts block to package.json to match common & catalog

benjaminapetersen:
For consistency with our other `package.json` files in the dependency repos.

@spadgett

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,13 @@
     "node": ">=0.10.0"
   },
   "scripts": {
+    "clean-deps": "hack/clean-deps.sh",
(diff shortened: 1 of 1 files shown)


spadgett:
Hm, looks like travis failed. @benjaminapetersen can you take a look and see if it's because of the package.json changes?

benjaminapetersen:
@spadgett yup, looking at this. Not sure why `package.json` would break things.

```bash 
The command "eval npm install  " failed. Retrying, 2 of 3.
```

spadgett:
[merge]

PR #1984, opened 2017-08-29
Use circle add icon for overview create route link

spadgett:
This makes it consistent with create binding on the same page.

openshift web console 2017-08-29 08-16-05

cc @beanh66


spadgett:
[merge]

PR #1987, opened 2017-08-29
Support binding parameters

### LINKED ISSUE #1979: Service names can get cut off at certain viewport sizes.
Service name is cut off when looking in a category at just over 1200px.

@rhamilto 

screen shot 2017-08-28 at 3 34 15 pm

### LINKED ISSUE #2002: Collapse long lines in resources inside my screen
https://api.ci.openshift.org/console/project/ci/browse/config-maps/config

Some configs like the above have some nasty long lines and it would be great if we could collapse those inside the screen. It was also strange at start that I couldn't see the "See All" button and I had to scroll all the way through to the end.

spadgett:
https://trello.com/c/Sws4p5jU

- [x] Put parameters in a secret
- [x] Add binding parameters to in-context flows in origin-web-catalog

Requires https://github.com/openshift/origin-web-common/pull/163

Bumps origin-web-common 0.0.52 -> 0.0.53
Bumps origin-web-catalog 0.0.42 -> to 0.0.43
which also
Fixes #1979
Fixes #2002

### DIFF
--- a/app/scripts/directives/bindService.js
+++ b/app/scripts/directives/bindService.js
@@ -23,7 +23,11 @@
                        DataService,
                        BindingService) {
     var ctrl = this;
-    var validityWatcher;
+    var bindFormStep;
+    var bindParametersStep;
+    var resultsStep;
+    var selectionValidityWatcher;
+    var parametersValidityWatcher;
     var bindingWatch;
     var statusCondition = $filter('statusCondition');
     var enableTechPreviewFeature = $filter('enableTechPreviewFeature');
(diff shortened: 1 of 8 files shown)


app/scripts/directives/bindService.js
```diff
@@ -139,37 +157,64 @@
...
+
+      var instance = ctrl.target.kind === 'Instance' ? ctrl.target : ctrl.serviceToBind;
+      if (!instance) {
+        return;
+      }
+
+      ctrl.serviceClass = ctrl.serviceClasses[instance.spec.serviceClassName];
+      ctrl.serviceClassName = instance.spec.serviceClassName;
+      ctrl.plan = BindingService.getPlanForInstance(instance, ctrl.serviceClass);
+      ctrl.parameterSchema = _.get(ctrl.plan, 'alphaBindingCreateParameterSchema');
```
benjaminapetersen: They don't come up too often, but do you think we ever need to doc these `alpha` kinds of properties with a `TODO` of `FIXME`?  Last I remember was the `osb` prefixes.

app/scripts/directives/bindService.js
```diff
@@ -69,17 +73,31 @@
     };
 
     var showBind = function() {
-      ctrl.nextTitle = 'Bind';
+      ctrl.nextTitle = bindParametersStep.hidden ? 'Bind' : 'Next >';
+      if (!selectionValidityWatcher) {
+        selectionValidityWatcher = $scope.$watch("ctrl.selectionForm.$valid", function(isValid) {
```
benjaminapetersen: I'm tempted to suggest we start using a pattern of `var scopeWatches = []` and a loop in `$onDestroy` to deregister, just to make all of our components consistent.

spadgett: Technically the scope watches clean themselves up on destroy and don't need to be explicitly deregistered. I wonder if it's worth tracking them except in the cases where we want to disable / enable them on certain steps.

app/scripts/directives/bindService.js
spadgett: Definitely reasonable. We've been trying to track the changes in the card because there are so many coming:

https://trello.com/c/y1lbFrbp

benjaminapetersen: Nice, that works for me.  Something like `FIXME: API ALPHA` could be a nice way to search, but might not be worth it if already tracked like the Trello.

app/scripts/directives/bindService.js
benjaminapetersen: Yeah prob not so much.  I think I'm pointlessly tracking one in the drawer, should prob nix it.

spadgett:
[merge]

spadgett:
[merge]

PR #1989, opened 2017-08-30
update image registry widget to v0.13

petervo:
This separates the imagestream and image listings.

Other fixes include:

 * Build date display fixes
 * Use maintainer as author
 * Fix empty table rendering
 * Add tag to imagestream push command

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -29,7 +29,7 @@
     "c3": "0.4.11",
     "kubernetes-label-selector": "2.0.0",
     "kubernetes-container-terminal": "2.1.0",
-    "registry-image-widgets": "0.0.11",
+    "registry-image-widgets": "https://github.com/cockpit-project/registry-image-widgets.git#0.0.13",
     "layout.attrs": "2.1.1",
     "kubernetes-object-describer": "1.0.4",
     "bootstrap-hover-dropdown": "2.1.3",
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -217,7 +217,7 @@ registry-imagestream-push {
   margin-top: @grid-gutter-width;
 }
 
-registry-imagestream-listing {
+registry-image-listing {
   /* No expandable listing when shown in the Openshift Web Console */
   .listing-ct-toggle {
     display: none;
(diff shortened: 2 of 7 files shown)


bower.json
```diff
@@ -29,7 +29,7 @@
     "c3": "0.4.11",
     "kubernetes-label-selector": "2.0.0",
     "kubernetes-container-terminal": "2.1.0",
-    "registry-image-widgets": "0.0.11",
+    "registry-image-widgets": "https://github.com/cockpit-project/registry-image-widgets.git#0.0.13",
```
spadgett: I'm not sure what's involved in this, but is it possible to update bower to use the new repository so we don't need the full URL?

petervo:
I would love to, but I don't have the creds to do it. @jwforres or @stefwalter do you?

petervo:
Alternatively @spadgett you could just push master from cockpit-project to openshift.

spadgett:
Unfortunately I don't have access to update the repository in bower. @stefwalter do you?

stefwalter:
> Unfortunately I don't have access to update the repository in bower. @stefwalter do you?

According to this https://bower.io/docs/creating-packages/ we need to open a ticket manually. Probably worth going with the workaround for now. I won't be around to work through this next week.

spadgett:
> According to this https://bower.io/docs/creating-packages/ we need to open a ticket manually. 

It sounds like you only need open a ticket if `bower unregister` doesn't work?

> Probably worth going with the workaround for now. I won't be around to work through this next week.

We can do that. @petervo can you open an issue to track registering the new bower repository?

spadgett:
[merge][severity: bug]

spadgett:
@petervo There's a dist mismatch. Can you try

```
$ hack/clean-assets.sh
$ hack/install-assets.sh
$ grunt build
```

petervo:
@spadgett, there was no ```hack/clean-assets.sh``` so I ran

```
hack/clean-deps.sh
hack/install-deps.sh
grunt build
```

and pushed the results.

spadgett:
[merge][severity: bug]

spadgett:
@petervo Can you mark any of the open issues this fixes as closed? I think we had some in the origin repo.

Issue #1993, opened 2017-08-31
Long project names in Notifications header overlap link


sg00dwin:
<img width="364" alt="screen shot 2017-08-31 at 9 42 57 am" src="https://user-images.githubusercontent.com/1874151/29926655-133855ba-8e32-11e7-971b-efccd0d376e6.png">

spadgett:
@jeff-phillips-18 @benjaminapetersen FYI

benjaminapetersen:
Nice catch. Hmm. Truncate or just drop the link down a line?

spadgett:
Truncate seems find to me since you can already see the project name in the project bar. Maybe add a `title` attribute for hover.

PR #1997, opened 2017-09-01
Notification panel transitions and visual enhancements

sg00dwin:
Update to PatternFly 3.27.1

Fixes https://github.com/openshift/origin-web-console/issues/1993

screen shot 2017-09-01 at 11 49 37 am
screen shot 2017-08-31 at 9 42 57 am

### DIFF
--- a/app/styles/_navbar-vertical.less
+++ b/app/styles/_navbar-vertical.less
@@ -59,8 +59,8 @@
     .badge {
       margin: 0 !important;
       position: absolute;
-      top: 9px;
       right: 13px;
+      top: 9px;
       @media(min-width: @screen-sm-min) {
         top: 20px;
       }
--- a/app/styles/_notifications.less
+++ b/app/styles/_notifications.less
@@ -5,8 +5,11 @@
 notification-drawer-wrapper {
   .drawer-pf {
     height: calc(100vh ~"-" (@navbar-os-header-height-mobile + @project-bar-height-mobile));
+    opacity: 1;
     position: fixed;
+    right: 0;
     top: (@navbar-os-header-height-mobile + @project-bar-height-mobile);
+    transition: top 150ms ease-in-out, opacity 150ms;
     z-index: @zindex-navbar-fixed - 2; // two less than navbar, one less than project-bar
     @media(max-width: 350px) {
       left: 0;
@@ -21,6 +24,14 @@ notification-drawer-wrapper {
       @media(min-width: @screen-sm-min) {
         left: 270px;
       }
+    }
+    &.hide {
+      opacity: 0;
+      top: 38px;
+    }
+    &.hide-add,
+    &.hide-remove {
+      display: block !important;
     }
     .tech-preview & {
       height: calc(100vh ~"-" (@navbar-os-header-height-mobile + @project-bar-height-mobile + @tech-preview-banner-height));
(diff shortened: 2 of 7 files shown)


app/styles/_notifications.less
```diff
@@ -48,34 +62,69 @@ notification-drawer-wrapper {
...
+    color: @color-pf-black-700;
+    line-height: normal;
     // TODO: hack to eliminate side-to-side wobble
     // There is a hard-coded <h4> that wraps the header:
     //   - https://github.com/patternfly/angular-patternfly/issues/539
     // I'm putting quite a bit of markup inside the header,
     // need to tinker to eliminate this rule.
-    overflow: hidden;
+    overflow:hidden;
+    font-size: @font-size-h4;
```
rhamilto: Nit:  alpha

app/styles/_navbar-vertical.less
```diff
@@ -87,6 +87,12 @@
       @media(max-width: @screen-xs-max) {
         padding: 10px 15px 10px 5px;
       }
+      .badge {
```
rhamilto: I had absolutely positioned .badge in one of my PRs.  It looks like it got rebased away.  :(  Can we redo the absolute positioning?  Otherwise the width of the nav item changes with the coming and going of the badge..m

app/styles/_navbar-vertical.less
```diff
@@ -87,6 +87,12 @@
...
+        background-color: @color-pf-blue-300;
+        border: 2px solid @navbar-os-bg-color;
+        width: 12px;
+        height: 12px;
```
rhamilto: Nit:  alpha

app/styles/_notifications.less
```diff
@@ -48,34 +62,69 @@ notification-drawer-wrapper {
...
+  .drawer-pf-title h3 {
+    font-size: @font-size-base - 1;
+  }
+  .expanded-notification {
+    .drawer-pf-notification-message-expanded {
+      font-size: @font-size-base - 1;
+      padding: 10px 10px 10px 42px;
+      .text-muted;
+    }
+    .drawer-pf-notification-info {
```
rhamilto: Nit:  alpha

app/styles/_notifications.less
```diff
@@ -48,34 +62,69 @@ notification-drawer-wrapper {
...
+      .text-muted;
+    }
+    &.unread {
+      .drawer-pf-notification-message-expanded {
+        color: inherit;
+        font-weight: normal;
+        .word-break();
+      }
+      .drawer-pf-notification-info {
```
rhamilto: Nit:  alpha

app/styles/_navbar-vertical.less
rhamilto: Nevermind.  The changes are there.  But perhaps you should combine the two rules so they're not in two different files?

serenamarie125:
@sg00dwin this looks much better IMO!

sg00dwin:
changes updated

app/styles/_notifications.less
```diff
@@ -31,6 +42,9 @@ notification-drawer-wrapper {
       }
     }
   }
+  .drawer-pf-close, .drawer-pf-toggle-expand {
+    padding: 2px 10px;
+  }
```
jeff-phillips-18: We should give feedback to patternfly on things like this. Shouldn't have to be overriding  these settings.

app/styles/_navbar-vertical.less
```diff
@@ -56,11 +56,15 @@
     @media(min-width: @screen-sm-min) {
       margin-right: @navbar-header-right-margin;
     }
-    .badge {
+    .nav-item-iconic .badge {
+      background-color: @color-pf-blue-300;
+      border: 2px solid @navbar-os-bg-color;
+      height: 12px;
```
jeff-phillips-18: I don't think this should be straying from the Patternfly design, @openshift/team-ux-review ?

app/styles/_notifications.less
serenamarie125: @sg00dwin why are we doing this?  Agree that we shouldn't be overriding things.  If we need PF to change something let me know.  If they cannot do it in time, then we can override with a plan to change later.

app/styles/_navbar-vertical.less
spadgett: @sg00dwin What is the before and after on this change?

If @openshift/team-ux-review agrees it's better, we could look at contributing it upstream in Patternfly.

serenamarie125: @sg00dwin again, why are we doing this?  Was this suggested by @jhaines?  

If there's a good reason, let's get this into PF.  LOTS of products are using the notification drawer, we want this to be consistent if it's indeed something that improves the component.

sg00dwin: discussed with rhamilto, done to improve identification of unread notification. 
Will revert

before 
screen shot 2017-09-01 at 2 42 11 pm

after
screen shot 2017-09-01 at 2 41 18 pm

app/styles/_notifications.less
spadgett: @jeff-phillips-18 @serenamarie125 Makes sense. For context, @sg00dwin added a few pixels of extra padding between the icon and the edge of the drawer.

@sg00dwin It might just be easier to remove this.

Before:

openshift web console 2017-09-01 14-40-35

After:

openshift web console 2017-09-01 14-45-18

app/styles/_navbar-vertical.less
spadgett: @serenamarie125 Opinion? I think the badget stands out better with the black around it. Let us know, and we can add submit this in Patternfly.

Before:

openshift web console 2017-09-01 14-48-41

After:

openshift web console 2017-09-01 14-47-54

app/styles/_notifications.less
sg00dwin: Felt the click target as too small,  5px padding between icon and edge.
I'll revert and suggest issue upstream.

before
screen shot 2017-09-01 at 2 45 35 pm

after
screen shot 2017-09-01 at 2 46 28 pm

app/styles/_navbar-vertical.less
rhamilto: I think it's worth noting the inspiration for the border around the badge came from Github.

app/styles/_notifications.less
rhamilto: > We should give feedback to patternfly on things like this. Shouldn't have to be overriding these settings.

While I don't disagree, sometimes it's infinitely easier just to fix a problem in the code that you control.

serenamarie125: Totally agree that what you have done looks much better!  

I do think that if we take those liberties, please call them out so that UX can review and we can make a plan get them into PatternFly for consistency :D  Understand your point @rhamilto but when our customers are dealing with a portfolio of products, everyone doing their own things just causes further issues.

app/styles/_navbar-vertical.less
serenamarie125: again, imo @sg00dwin @rhamilto I think your proposal looks great, and i did immediately recognize that it was consistent with GitHub.  

Thoughts on putting up a PR in Core PF for this ?

app/styles/_notifications.less
benjaminapetersen: +1 prob agree to both...temp fix here then ensure an issue upstream to track/migrate it.

spadgett:
@jeff-phillips-18 @rhamilto Another look?

rhamilto:
[merge]

spadgett:
Flake #2016

[merge][severity: bug]

PR #1326, opened 2017-03-08
Notifications drawer

benjaminapetersen:
https://trello.com/c/o82aiHhh

WIP Prototype of the notification drawer.  

- using sessionStorage to store read/unread (though we may eliminate this)
- using a service to process events & reduce watch thrashing on nagivation

screen shot 2017-03-08 at 11 09 27 am

TODO:
- [x] Decide on best way to filter/curate notifications, though we don't necessarily want an open drawer with practically nothing in it.
- [x] ~~Collapse history of objects, if `mongodb-1-deploy`, `mongodb-2-deploy` and `mongodb-3-delpoy` all have events, only show what is current & relevant (`mongodb-3-deploy`).~~
- [x] need build events (blocker)
- [x] other text tweaks possible? some events are not terribly friendly.
- [x] ~~click to dismiss notifications, if all dismissed for a group, remove group.~~

Designs for reference:
+ [openshift.github.io drawer design](https://openshift.github.io/openshift-origin-design/web-console/4-patterns/notifications)
+ [drawer design PR](https://github.com/openshift/openshift-origin-design/pull/45/files)
+ [collapsed screenshot](https://raw.githubusercontent.com/openshift/openshift-origin-design/ee538bd61d473b669056c9d7f187fa3d66cc84b3/web-console/4-patterns/img/Phase1-10.png)
+ [expanded screenshot](https://raw.githubusercontent.com/openshift/openshift-origin-design/ee538bd61d473b669056c9d7f187fa3d66cc84b3/web-console/4-patterns/img/Phase1-11.png)

@jwforres @spadgett opening up for conversation.

Issue #1911, opened 2017-08-04
When performing multiple actions, toaster hides the drop down menu in upper right


smarterclayton:
If I'm trying to delete several things, I've found that I often end up with two toasters, which covers the drop down menus so I can't delete the third thing.  The toaster shouldn't end up covering the place I'm likely to click next.

Also, deleting many things was frustrating because none of the tables had a delete action / dropdown associated with them.  I was surprised that we didn't have the actions per row.

spadgett:
> If I'm trying to delete several things, I've found that I often end up with two toasters, which covers the drop down menus so I can't delete the third thing. The toaster shouldn't end up covering the place I'm likely to click next.

@openshift/team-ux-review Ideas? The action button on most pages is about where the toast notifications appear.

We've had feedback that the action button is often too far from the content when the window is maximized as well, so maybe the easy fix is to just move the button.

Issue #1121, opened 2017-01-16
Style guide: Add conventions for formatting angular module/function declarations


jawnsy:
Please add something to the style guide about formatting for `angular.module` declarations. For example:

[directives/buildPipeline.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/buildPipeline.js#L3
):

```js
angular.module('openshiftConsole')
  .directive('buildPipeline', function($filter, Logger) {
```

`angular.module` on same line?

[directives/actionChip.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/actionChip.js#L21):

```js
angular
  .module('openshiftConsole')
  .directive('actionChip', function() {
```

`.module` (or other method calls) on their own line, indented, below `angular`

[directives/deploymentMetrics.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/deploymentMetrics.js#L3
):

```js
angular.module('openshiftConsole')
  .directive('deploymentMetrics', function($interval,
                                           $parse,
                                           $timeout,
                                           $q,
                                           $rootScope,
                                           ChartsService,
                                           ConversionService,
                                           MetricsCharts,
                                           MetricsService,
                                           ModalsService) {
```

imports aligned to function call, to the right of directive declaration.

[directives/labels.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/labels.js#L3
):

```js
angular.module('openshiftConsole')
  .directive('labels', function($location, $timeout, LabelFilter) {
```

imports directly inline, not wrapped.

[controllers/attachPVC.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/attachPVC.js#L12):

```js
angular.module('openshiftConsole')
  .controller('AttachPVCController',
              function($filter,
                       $routeParams,
                       $scope,
                       $window,
                       APIService,
                       AuthorizationService,
                       BreadcrumbsService,
                       DataService,
                       Navigate,
                       ProjectsService,
                       StorageService) {
```

imports below the controller name, function declaration aligned to controller name

spadgett:
Related to this, I think we should leave off ngdoc comments above these declarations when they have no meaningful descriptions. We have a lot of these copied throughout the code, and some don't even have the right controller names (see #1204).

benjaminapetersen:
agree!

spadgett:
Closing as 3.x is in maintenance only mode. In 4.x, we use prettier.

PR #2001, opened 2017-09-02
Custom notifications in notification drawer

benjaminapetersen:
Integrates our internal notifications into the notification drawer.
Enhancement to #1326 

screen shot 2017-09-01 at 11 03 13 pm

@spadgett

### DIFF
--- a/app/scripts/directives/notifications/notificationDrawerWrapper.js
+++ b/app/scripts/directives/notifications/notificationDrawerWrapper.js
@@ -16,7 +16,6 @@
         '$rootScope',
         'Constants',
         'DataService',
-        'NotificationsService',
         'EventsService',
         NotificationDrawerWrapper
       ]
@@ -31,7 +30,6 @@
       $rootScope,
       Constants,
       DataService,
-      NotificationsService,
       EventsService) {
 
       // kill switch if watching events is too expensive
@@ -45,23 +43,26 @@
       // this one is treated separately from the rootScopeWatches as
       // it may need to be updated outside of the lifecycle of init/destroy
       var notificationListener;
-      // our internal notifications
-      // var clientGeneratedNotifications = [];
-
-      var eventsWatcher;
-      var eventsMap = {};
-
-      // TODO:
-      // include both Notifications & Events,
-      // rather than destroying the map each time maintain it & add new items
-
-      // final Processed set of notification groups for UI
-      // IF POSSIBLE, avoid having to convert back to an array.
-      // var notificationGroupsMap = {};
-      var notificationGroups = [];
-
+      var apiEventsWatcher;
+      // data
+      var apiEventsMap = {
+        // projName: { events }
+      };
+      var notificationsMap = {
+        // projName: { notifications }
+      };
 
       var projects = {};
+
+      var hideIfNoProject = function(projectName) {
+        if(!projectName) {
+          drawer.drawerHidden = true;
+        }
+      };
+
+      var projectChanged = function(next, current) {
+        return _.get(next, 'params.project') !== _.get(current, 'params.project');
+      };
 
       var getProject = function(projectName) {
         return DataService
@@ -72,99 +73,71 @@
                 });
       };
 
-      var ensureProjectGroupExists = function(groups, projectName) {
-        if(projectName && !groups[projectName]) {
-          groups[projectName] = {
-            heading: $filter('displayName')(projects[projectName]) || projectName,
-            project: projects[projectName],
-            notifications: []
-          };
-        }
-      };
-
-      var deregisterEventsWatch = function() {
-        if(eventsWatcher) {
-          DataService.unwatch(eventsWatcher);
-        }
-      };
-
-      var watchEvents = function(projectName, cb) {
-        deregisterEventsWatch();
-        if(projectName) {
-          eventsWatcher = DataService.watch('events', {namespace: projectName}, _.debounce(cb, 400), { skipDigest: true });
-        }
-      };
-
-      // NotificationService notifications are minimal, they do no necessarily contain projectName info.
-      // ATM tacking this on via watching the current project.
-      // var watchNotifications = function(projectName, cb) {
-      //   deregisterNotificationListener();
-      //   if(!projectName) {
-      //     return;
-      //   }
-      //   notificationListener = $rootScope.$on('NotificationsService.onNotificationAdded', cb);
-      // };
-
-      var deregisterNotificationListener = function() {
-        notificationListener && notificationListener();
-        notificationListener = null;
+      var makeProjectGroup = function(projectName, notifications) {
+        return {
+          heading: $filter('displayName')(projects[projectName]),
+          project: projects[projectName],
+          notifications: notifications
+        };
       };
 
       var unread = function(notifications) {
         return _.filter(notifications, 'unread');
       };
 
-      // returns a count for each type of notification, example:
-      // {Normal: 1, Warning: 5}
-      // TODO: eliminate this $rootScope.$applyAsync,
-      // there is a quirk here where the values are not picked up the
-      // first time the function runs, despite the same $applyAsync
-      // in the render() function
-      var countUnreadNotificationsForGroup = function(group) {
-        $rootScope.$applyAsync(function() {
+
+      var countUnreadNotifications = function() {
+        _.each(drawer.notificationGroups, function(group) {
           group.totalUnread = unread(group.notifications).length;
           group.hasUnread = !!group.totalUnread;
-          $rootScope.$emit('NotificationDrawerWrapper.count', group.totalUnread);
+          $rootScope.$emit('NotificationDrawerWrapper.onUnreadNotifications', group.totalUnread);
         });
       };
 
-      // currently we only show 1 at a time anyway
-      var countUnreadNotificationsForAllGroups = function() {
-        _.each(notificationGroups, countUnreadNotificationsForGroup);
-      };
-
-      var sortNotifications = function(notifications) {
-        return _.orderBy(notifications, ['event.lastTimestamp', 'event.firstTimestamp'], ['desc', 'desc']);
-      };
-
-      var sortNotificationGroups = function(groupsMap) {
-        // convert the map into a sorted array
-        var sortedGroups = _.sortBy(groupsMap, function(group) {
-          return group.heading;
+      var formatAPIEvents = function(apiEvents) {
+        return _.map(apiEvents, function(event) {
+          return {
+            actions: null,
+            uid: event.metadata.uid,
+            trackByID: event.metadata.uid,
+            unread: !EventsService.isRead(event.metadata.uid),
+            type: event.type,
+            lastTimestamp: event.lastTimestamp,
+            firstTimestamp: event.firstTimestamp,
+            event: event
+          };
         });
-        // and sort the notifications under each one
-        _.each(sortedGroups, function(group) {
-          group.notifications = sortNotifications(group.notifications);
-          group.counts = countUnreadNotificationsForGroup(group);
-        });
-        return sortedGroups;
       };
 
-      var formatAndFilterEvents = function(eventMap) {
-        var filtered = {};
-        ensureProjectGroupExists(filtered, $routeParams.project);
-        _.each(eventMap, function(event) {
-          if(EventsService.isImportantEvent(event) && !EventsService.isCleared(event)) {
-            ensureProjectGroupExists(filtered, event.metadata.namespace);
-            filtered[event.metadata.namespace].notifications.push({
-              unread:  !EventsService.isRead(event),
-              trackByID: event.metadata.uid,
-              event: event,
-              actions: null
-            });
+      var filterAPIEvents = function(events) {
+        return _.reduce(events, function(result, event) {
+          if(EventsService.isImportantAPIEvent(event) && !EventsService.isCleared(event.metadata.uid)) {
+            result[event.metadata.uid] = event;
           }
+          return result;
+        }, {});
+      };
+
+      // we have to keep notifications & events separate as
+      // notifications are ephemerial, but events have a time to live
+      // set by the server.  we can merge them right before we update
+      // the UI.
+      var mergeMaps = function(firstMap, secondMap) {
+        var proj = $routeParams.project;
+        return _.assign({}, firstMap[proj], secondMap[proj]);
+      };
+
+      var sortMap = function(map) {
+        return _.orderBy(map, ['event.lastTimestamp', 'event.firstTimestamp'], ['desc', 'desc']);
+      };
+
+      var render = function() {
+        $rootScope.$evalAsync(function() {
+            drawer.notificationGroups = [
+              makeProjectGroup($routeParams.project, sortMap( mergeMaps(apiEventsMap, notificationsMap )))
+            ];
+            countUnreadNotifications();
         });
-        return filtered;
       };
 
       var deregisterRootScopeWatches = function() {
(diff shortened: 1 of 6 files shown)


app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -174,51 +153,66 @@
...
+      var notificationWatchCallback = function(event, notification) {
+        var uid = notification.id || _.uniqueId('notification-');
+        var project = notification.namespace || $routeParams.project;
+        notificationsMap[project] = notificationsMap[project] || {};
+        notificationsMap[project][uid] = {
+          actions: null,
+          unread: true,
+          uid: uid,
+          // emulating an API event to simplify the template
+          event: {
```
spadgett: I'd rather not add a dummy event. I'd rather copy necessary properties like `uid` and `timestamp` from the event to top-level properties, and use those in the view.

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -174,51 +153,66 @@
...
+            lastTimestamp: notification.lastTimestamp || moment.parseZone(new Date()).utc().format(),
```
spadgett: Better to add the timestamp in NotificationsService

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -174,51 +153,66 @@
...
+      };
+
+      // NotificationsService notifications are minimal, they do no necessarily contain projectName info.
+      // ATM tacking this on via watching the current project.
+      var watchNotifications = function(projectName, cb) {
+        deregisterNotificationListener();
+        if(!projectName) {
+          return;
+        }
+        notificationListener = $rootScope.$on('NotificationsService.onNotificationAdded', cb);
```
spadgett: We should be able to add this once instead of registering and removing on project changes.

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -45,25 +45,38 @@
...
-      var notificationGroups = [];
+      // internal notifications have different types than API events
+      var notificationEventTypeMap = {
+        success: 'Normal',
+        error: 'Warning'
+      };
 
+      var iconClassByEventSeverity = {
+        Normal: 'pficon pficon-info',
+        Warning: 'pficon pficon-warning-triangle-o'
```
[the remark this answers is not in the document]
benjaminapetersen: Looks like they use `alertIcon` : 

```html 
<span class="{{notification.type | alertIcon}}" aria-hidden="true"></span>
```
I'll update & use that

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -73,98 +86,64 @@
...
+                  if(EventsService.isImportantEvent(event) && !EventsService.isCleared(event)) {
+                    result[event.metadata.uid] = {
+                      actions: null,
+                      uid: event.metadata.uid,
+                      unread: !EventsService.isRead(event),
+                      event: event
+                    };
+                  }
+                  return result;
+                }, {});
```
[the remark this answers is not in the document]
benjaminapetersen: I'll update this.

Ref #1121 for when we have time.

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -73,98 +86,64 @@
...
-        // and sort the notifications under each one
-        _.each(sortedGroups, function(group) {
-          group.notifications = sortNotifications(group.notifications);
-          group.counts = countUnreadNotificationsForGroup(group);
-        });
-        return sortedGroups;
+      // Since the events map & internal notifications map are separate
+      // data and updated asyncronously, they need to be copied/merged
+      // at some point before being sent to the UI to render.
+      var mergeAndSortMaps = function(firstMap, secondMap) {
```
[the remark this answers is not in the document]
benjaminapetersen: I tend to agree & almost just inlined this into `render()` since the two separate functions will basically be wrappers around lodash one-liners.

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -73,98 +86,64 @@
...
+        var proj = $routeParams.project;
+        return _.orderBy(
+                  _.assign({}, firstMap[proj], secondMap[proj]),
+                  ['event.lastTimestamp', 'event.firstTimestamp'],
```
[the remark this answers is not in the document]
benjaminapetersen: I can follow-up a PR to `origin-web-common` to do some work on NotificationsService.  [This line](https://github.com/openshift/origin-web-console/pull/2001/files#diff-e13814006e3a1d3d1f4fa5f45b6ebc5dR182) should just be a fallback.

app/scripts/directives/notifications/notificationDrawerWrapper.js
benjaminapetersen: ok ill update

app/scripts/directives/notifications/notificationDrawerWrapper.js
benjaminapetersen: True. no need to update this one like the events watch.

app/scripts/directives/notifications/notificationDrawerWrapper.js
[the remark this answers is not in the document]
spadgett: I'm fine inlining as well, whichever works :)

app/scripts/directives/notifications/notificationDrawerWrapper.js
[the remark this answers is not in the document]
spadgett: Sure, OK as a follow on

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -45,25 +45,38 @@
...
+      };
 
-      // final Processed set of notification groups for UI
-      // IF POSSIBLE, avoid having to convert back to an array.
-      // var notificationGroupsMap = {};
-      var notificationGroups = [];
+      // internal notifications have different types than API events
+      var notificationEventTypeMap = {
+        success: 'Normal',
+        error: 'Warning'
```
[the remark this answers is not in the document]
benjaminapetersen: Eliminating both maps via PR to common

benjaminapetersen:
Updated.  Will need a release of `origin-web-common` with PR 170 before this can go.

dtaylor113:
Hi @benjaminapetersen, The 'Clear All' doesn't appear to be working correctly.  I start out with 3 notifications:

image

I think click on an 'X' to remove one:

image

Then click on 'Clear All' and all three re-appear instead of removing them all as I was expecting:

image

dtaylor113:
Hi, the 'View Build' link in this notification doesn't seem to do anything/navigation anywhere:

image

benjaminapetersen:
Thanks @dtaylor113! I made a few updates last night & have another push coming this morn that addresses these items.  I updated `EventsService` a bit & am reconciling the `id`s.

benjaminapetersen:
btw, Id's are still janky w/o the update to `origin-web-common`.  I'll PR a version bump.

benjaminapetersen:
Rebased, also put a fallback `id` in otherwise the notifications just act too weird (until common version bump happens).

app/scripts/services/events.js
```diff
@@ -12,31 +12,30 @@ angular.module('openshiftConsole')
...
-        return EVENTS_TO_SHOW_BY_REASON[reason];
+      var isImportantAPIEvent = function(event) {
+        return EVENTS_TO_SHOW_BY_REASON[event.reason];
       };
 
-      var markRead = function(event) {
-        _.set(cachedEvents, [event.metadata.uid, READ], true);
+      var markRead = function(id) {
+        _.set(cachedEvents, [id, READ], true);
         BrowserStore.saveJSON('session','events', cachedEvents);
```
spadgett: If we're saving read and clear status by ID in session storage, we should really change NotificationService to generated a UID instead of using `_.uniqueId`. With `_.uniqueId` the IDs will be reused if you refresh the page, and notification might be incorrectly hidden from the drawer.

benjaminapetersen: This is true.  I could pull the `generateSecret` function from `ApplicationGenerator`:

```JavaScript
    scope._generateSecret = function(){
        //http://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript
        function s4() {
          return Math.floor((1 + Math.random()) * 0x10000)
              .toString(16)
              .substring(1);
          }
        return s4()+s4()+s4()+s4();
      };
```

spadgett:
Looks like the sort is still wrong. (I should have latest common changes with the timestamps.)

screen shot 2017-09-06 at 1 27 35 pm

app/scripts/services/events.js
benjaminapetersen: Simpler could just be ` _.uniqueId('notification_') + Date.now() `.

benjaminapetersen:
@openshift/team-ux-review

Need some feedback!  We have our notifications integrated into the drawer with the curated api events.  This creates some quirks that we think need to be addressed before we can merge this feature.  

1.  As a user, if I click `deploy` I will get 1. a toast indicating the action (deploy) and a notification indicating the same action.  I know I initiated the action, so double affirmation is a bit much.

2.  If I close the toast, I might expect that the notification in the drawer should be marked `read`.  It is no longer news to me.

3. The API Events have only 2 types, `Warning` and `Normal`.  Our notifications have 4: `Warning`, `Error`, `Success`, `Info`.   This can cause mismatch, like with the icons here:

screen shot 2017-09-06 at 3 11 10 pm

These two events should be the same level (actually the "Deployment Created" event might be more weighty than "Deployment Started"), but the icons seem to indicate that "Deployment Created" is of lesser value.  

4.  There is a difference in the way we format the two event types.  I'll add another screenshot after `Start Build` to illustrate this further:

screen shot 2017-09-06 at 3 12 15 pm

With the API event we link the name of the object, with the internal notification we add a `view build` link as a separate component.

5.  I really think we should add the counts back to the header, it is definitely easy to miss a "warning" event that gets pushed below the fold:

screen shot 2017-09-06 at 3 14 41 pm

However, since internal events have 4 states (explained above) we probably don't want 4 icons with counts in the header. 

6.  We have discussed that the large toast for a success message is perhaps more obtrusive than it needs to be:

screen shot 2017-09-06 at 3 16 51 pm

It might be ideal to make these smaller similar to how gmail gives a centered `sent message` toast.  This could solve the problem of toasts frequently covering action buttons:

screen shot 2017-09-06 at 3 16 27 pm

And save the large toasts for bad things:

screen shot 2017-09-06 at 3 19 13 pm 1

Thoughts?

Observations gathered from earlier discussion with @ncameronbritt @spadgett

ncameronbritt:
The toast notifications are helpful when a user takes an action in the sense that they provide feedback to the user. But obviously toasts aren't the only way we could provide the user feedback that the system registered their input. The little message like gmail provides is another option. Having the button itself do something is yet another. May be worth thinking about this from a PF perspective too for actions that don't obviously change a user's context.

benjaminapetersen:
Right! @ncameronbritt you suggested something like this for a button effect:



I dunno why github made the gif huge 😕

ncameronbritt:
Right on @benjaminapetersen! I like that that communicates that something is happening right in the place where the user took an action.

serenamarie125:
Looks cool, but not consistent with what we do in PatternFly.  I'd suggest we wait on that behavior and do some
Research to see if that's what we want to consistently throughout Openshift and the rest of the portfolio.  Or in what situations we want to do it :)

spadgett:
@serenamarie125 We've had some feedback from @smarterclayton and @sspeiche that some of the confirmation toasts are annoying. ("I know I just clicked that button. Why are you telling me?") See also #1911. I remember you and @jeff-phillips-18 had similar concerns.

The problem is that on some pages, depending on what tab is selected, nothing on the page changes in response to an action. So it can look like the button didn't do anything. One example is clicking "Start Build" on the build config page with the "Configuration" tab selected. Nothing changes, and it seems like the action didn't work. The user might try clicking the button several times and accidentally creating several builds.

What we'd really like is a way to let users know an action succeeded that is a little less in-your-face than a toast. A more subtle, "Yep, that worked" message. The Gmail screenshot from  @benjaminapetersen is good example. The button changing is also interesting, although I'd like something that works in contexts where there isn't a button.

Do you think this make sense to explore as a pattern in Patternfly?

serenamarie125:
@spadgett I agree that the toasts can be annoying.  IMO, we should be identifying a strategy for default usage of toasts & availability in the notification drawer.  Ultimately the user should be able to configure what they would like to see where as well.

That being said, I do agree that we need a solution so that users are aware that they have performed an action.  @spadgett is this something we can do for 3.8?  If so, i think that will give us the ability to perform some research and come up with a solution which should be used for similar use cases across products.  (thus my answer is yes i think we should explore a PatternFly solution 😄 )

serenamarie125:
@jeff-phillips-18 shouldn't the toast be UNDER the top banner?  Maybe this is a bug in PatternFly, although I thought we had fixed it?

jeff-phillips-18:
@serenamarie125 Patternfly puts the toasts 20px from the top. The pattern design does not specify the placement but shows it as:

image

spadgett:
> is this something we can do for 3.8?

@serenamarie125 Enough people have commented on it that I'd like to try to fix for 3.8.

beanh66:
@benjaminapetersen I wanted to respond back regarding the full list of quirks! It sounds like (1) and (6) from your list are being discussed above but here are my thoughts:

>1. As a user, if I click deploy I will get 1. a toast indicating the action (deploy) and a notification indicating the same action. I know I initiated the action, so double affirmation is a bit much.

Agreed, I think it would be a great thing to explore for 3.8 as @spadgett mentioned.

>2. If I close the toast, I might expect that the notification in the drawer should be marked read. It is no longer news to me.

In the documentation we initially said "When a user interacts with the actions inside of a toast notification, the message should automatically be marked as read inside the drawer." But to your point, this should be expanded to also be the case when users close a toast!

>3. The API Events have only 2 types, Warning and Normal. Our notifications have 4: Warning, Error, Success, Info. This can cause mismatch, like with the icons.

Agreed, that may be strange. Wondering if we should stick to just the two icons in the drawer for now, info and warning. So any info or success toasts would just look like info notifications in the drawer and any warnings or errors would look like warnings. Is that a possibility? What do you think would be the best way around this? Curious others thoughts! @openshift/team-ux-review 

>4. There is a difference in the way we format the two event types. I'll add another screenshot after Start Build to illustrate this further

Is there any way around this? Open to suggestions if we could get the formatting to match, but I understand this may not be possible, especially for 3.7. 

>5. I really think we should add the counts back to the header, it is definitely easy to miss a "warning" event that gets pushed below the fold. However, since internal events have 4 states (explained above) we probably don't want 4 icons with counts in the header.

Agreed we would only want the two icons, not four. I wouldn't be against adding the counts back in, but it sounded like they caused confusion previously. Curious if others have objections to adding these back in?

>6. We have discussed that the large toast for a success message is perhaps more obtrusive than it needs to be

Agreed, I think it would be a great thing to explore for 3.8.

>7. We have also discussed adding a way to configure what internal events end up as a toast, and which can just go directly into the drawer.

Agreed, that would be ideal behavior in the future as @serenamarie125  mentioned.

benjaminapetersen:
Great feedback! We have one more sprint, I think if we reach consensus for what to tweak for 3.7 and what to save for 3.8 we will be in good shape.

dtaylor113:
Hi, just wondering the status of this PR?  My PR builds on this one.  Is there a way to merge this PR and file issues for the remaining work?  -thanks!

spadgett:
@dtaylor113 Now that https://github.com/openshift/origin-web-common/pull/183 has merged, I think we just need to bump common and this can go in.

dtaylor113:
Thanks @spadgett, created PR to bump common [common#190](https://github.com/openshift/origin-web-common/pull/190)

benjaminapetersen:
Rebased, with the `notification.showInDrawer` flag.

benjaminapetersen:
Does it make sense to migrate the open questions about notifications/toasts to an issue in [openshift design](https://github.com/openshift/openshift-origin-design)?

dtaylor113:
Hi, common was bumped to 0.0.57

benjaminapetersen:
@dtaylor113 yup think this PR will wait on all the version bumps in #2103.

benjaminapetersen:
Reviewed & rebased.

[test]

benjaminapetersen:
```
########## FINISHED STAGE: FAILURE: PROVISION CLOUD RESOURCES [00h 02m 39s] 
##########
```

[merge]

benjaminapetersen:
wake up...

[merge]

benjaminapetersen:
Another flake

```
########## FINISHED STAGE: FAILURE: PROVISION CLOUD RESOURCES [00h 02m 39s] ##########
Build step 'Execute shell' marked build as failure
[PostBuildScript] - Execution post build scripts.
```

benjaminapetersen:
flakes, and travis keeps hanging...

[merge]

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -174,50 +147,68 @@
...
+        var project = notification.namespace || $routeParams.project;
+        var id = notification.id || _.uniqueId('notification_') + Date.now();
+        notificationsMap[project] = notificationsMap[project] || {};
+        notificationsMap[project][id] = {
+          actions: null,
+          unread: !EventsService.isRead(id),
+          // using uid to match API events and have one filed to pass
+          // to EventsService for read/cleared, etc
+          trackByID: notification.trackByID,
+          uid: id,
```
spadgett: I don't think we need to set this anymore? Technically it's not a unique ID since `notification.id` is not necessarily unique.

spadgett:
Needs rebase

benjaminapetersen:
rebased again.

PR #2009, opened 2017-09-05
Do not allow binding to an application when pod_presets is false

jeff-phillips-18:
Depends on origin-web-common release with openshift/origin-web-common#167


spadgett:
@jeff-phillips-18 Looks like a dist mismatch

jeff-phillips-18:
Added commit to bump origin-web-common to 0.0.52.

Removed WIP status.

spadgett:
@jeff-phillips-18 Just confirming that creating the binding from the service instance row still works. I thought we had blocked that when pod_presets was false previously

jeff-phillips-18:
@spadgett you are able to create the binding with pod_presets false. Not sure if this was in error before, but its working now ;)

jeff-phillips-18:
`  ENABLE_TECH_PREVIEW_FEATURE: {
    // Enable the new landing page and service catalog experience
    service_catalog_landing_page: true,
    // Set to `true` when the template service broker is enabled for the cluster in master-config.yaml
    template_service_broker: false,
    pod_presets: false
  },
`
image

spadgett:
[merge]

spadgett:
Flake #2016

[merge]

PR #2011, opened 2017-09-05
Removing table-hover from tables to address inconsistency in hover color

### LINKED ISSUE #1904: Use of .table-hover is confusing because we use a blue hover color to indicate clickable
On browse page tables, we use [.table-hover](http://getbootstrap.com/css/#tables-hover-rows).  Given the hover color is blue, this is a bit confusing because in places like the project list page and &lt;registry-imagestream-listing&gt; we use a blue hover color to indicate the item is clickable.  I think we should only use .table-hover if the item is clickable OR change the clickable item hover color.  Granted the cursor changes when the item is clickable, but I think the hover color change dominates the cursor change affordance.

Row is not clickable:
screen shot 2017-08-03 at 9 04 39 am

Row is clickable:
screen shot 2017-08-03 at 9 04 54 am
screen shot 2017-08-03 at 9 05 05 am

thoughts, @openshift/team-ux-review?

rhamilto:
Fixes #1904

Until https://github.com/patternfly/patternfly-design/issues/314 is
addressed, I think we should avoid using .table-hover in our tables.

### DIFF
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -178,7 +178,7 @@
                     </div>
                   </div>
                 </div>
-                <table ng-if="!(buildConfig | isJenkinsPipelineStrategy)" class="table table-bordered table-hover table-mobile">
+                <table ng-if="!(buildConfig | isJenkinsPipelineStrategy)" class="table table-bordered table-mobile">
                   <thead>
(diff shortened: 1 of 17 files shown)


spadgett:
[merge][severity: bug]

PR #2012, opened 2017-09-05
Not allowing creation date to wrap to eliminate bug

rhamilto:
Fixes https://github.com/openshift/origin-web-catalog/issues/412

Note:  this fix is imperfect in that it only applies to the catalog when in the context of the console.  Why?  Because the bug also occurs in the console on the projects list page, and I figured it should be fixed in both places.  The "proper" fix would be to put these additions in origin-web-common, but that necessitates a release of common, and that seems like a lot of overhead for a "simple" bug.  Agree/disagree @spadgett?

### DIFF
--- a/app/styles/_projects.less
+++ b/app/styles/_projects.less
@@ -19,6 +19,13 @@
   }
 }
 
+.project-date, // from origin-web-catalog
+.projects-list {
+  [am-time-ago] {
+    white-space: nowrap;
+  }
+}
+
 .project-description.list-group-item-text {
   width: 100%;
   .highlighted-content,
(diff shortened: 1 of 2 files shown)


app/styles/_projects.less
```diff
@@ -19,6 +19,13 @@
   }
 }
```
benjaminapetersen: Fix looks good to me. I have a PR to common that we will need tomorrow anyway, if that makes you feel better about migrating it up.

spadgett:
[merge][severity: bug]

spadgett:
[merge][severity: bug]

Issue #2016, opened 2017-09-06
Test Flake: Element not found: by.cssContainingText(".add-to-project .dropdown-toggle", "Add to Project")


spadgett:
```
1) User adds an image stream to a project after creating a new project using the Import YAML tab should process and create the images in the image stream
  - Error: Element not found: by.cssContainingText(".add-to-project .dropdown-toggle", "Add to Project")
Wait timed out after 5166ms
```

https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_origin_web_console/110/console

spadgett:
The project bar with the add to project button isn't showing up on the overview page. This only seems to happen in the automated tests and only sometimes.

![user adds an image stream to a project after creating a new project using the import yaml tab should process and create the images in the image stream](https://user-images.githubusercontent.com/1167259/30162956-fccb73ac-93a4-11e7-88d3-88649adef362.png)

spadgett:
Reopening since we saw this again in #2055

spadgett:
It looks like `$onRouteChangeSuccess` is not always getting triggered on first page load based on my test in #2062

PR #2020, opened 2017-09-06
Fixing newline-limit attribute so it actually works

rhamilto:


spadgett:
[merge][severity: bug]

spadgett:
Flake #2016

[merge][severity: bug]

PR #2021, opened 2017-09-06
Add the ability to add a secret to an application

jeff-phillips-18:
image

@beanh66

### DIFF
--- a/app/scripts/directives/addSecretToApplication.js
+++ b/app/scripts/directives/addSecretToApplication.js
@@ -1,0 +1,180 @@
+"use strict";
+(function() {
+  angular.module("openshiftConsole").component('addSecretToApplication', {
+    controller: [
+      '$filter',
+      '$scope',
+      'APIService',
+      'DataService',
+      'Navigate',
+      'NotificationsService',
+      'StorageService',
+      AddSecretToApplication
+    ],
+    controllerAs: 'ctrl',
+    bindings: {
+      project: '<',
+      secret: '<',
+      onComplete: '<',
+      onCancel: '<'
+    },
+    templateUrl: 'views/directives/add-secret-to-application.html'
+  });
+
+  function AddSecretToApplication($filter, $scope, APIService, DataService, Navigate, NotificationsService, StorageService) {
+    var ctrl = this;
+    var deploymentConfigs;
+    var deployments;
+    var replicationControllers;
+    var replicaSets;
+    var statefulSets;
+
+    var sortApplications = function() {
+      // Don't waste time sorting on each data load, just sort when we have them all
+      if (deploymentConfigs && deployments && replicationControllers && replicaSets && statefulSets) {
+        var apiObjects = deploymentConfigs.concat(deployments)
+          .concat(replicationControllers)
+          .concat(replicaSets)
+          .concat(statefulSets);
+        ctrl.applications = _.sortBy(apiObjects, ['metadata.name', 'kind']);
+        ctrl.updating = false;
+      }
+    };
+
+    var getApplications = function() {
+      var hasDeploymentFilter = $filter('hasDeployment');
+      var hasDeploymentConfigFilter = $filter('hasDeploymentConfig');
+
+      ctrl.updating = true;
+      var context = {
+        namespace: ctrl.project.metadata.name
+      };
+      // Load all the "application" types
+      DataService.list('deploymentconfigs', context).then(function(deploymentConfigData) {
+        deploymentConfigs = _.toArray(deploymentConfigData.by('metadata.name'));
+        sortApplications();
+      });
+      DataService.list('replicationcontrollers', context).then(function(replicationControllerData) {
+        replicationControllers = _.reject(replicationControllerData.by('metadata.name'), hasDeploymentConfigFilter);
+        sortApplications();
+      });
+      DataService.list({
+        group: 'apps',
+        resource: 'deployments'
+      }, context).then(function(deploymentData) {
+        deployments = _.toArray(deploymentData.by('metadata.name'));
+        sortApplications();
+      });
+      DataService.list({
+        group: 'extensions',
+        resource: 'replicasets'
+      }, context).then(function(replicaSetData) {
+        replicaSets = _.reject(replicaSetData.by('metadata.name'), hasDeploymentFilter);
+        sortApplications();
+      });
+      DataService.list({
+        group: 'apps',
+        resource: 'statefulsets'
+      }, context).then(function(statefulSetData) {
+        statefulSets = _.toArray(statefulSetData.by('metadata.name'));
+        sortApplications();
+      });
+    };
+
+    ctrl.$onInit = function() {
+      ctrl.addType = 'env';
+      ctrl.disableInputs = false;
+      getApplications();
+    };
+
+    ctrl.$postLink = function() {
+      $scope.$watch(function() {
+        return ctrl.application;
+      }, function() {
+        // Look at the existing mount paths so that we can warn if the new value is not unique.
+        var podTemplate = _.get(ctrl.application, 'spec.template');
+        ctrl.existingMountPaths = StorageService.getMountPaths(podTemplate);
+      });
+    };
+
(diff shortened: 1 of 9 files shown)


app/scripts/directives/addSecretToApplication.js
```diff
@@ -0,0 +1,171 @@
...
+        sortApplications();
+      });
+      DataService.list({
+        group: 'apps',
+        resource: 'statefulsets'
+      }, context).then(function(statefulSetData) {
+        statefulSets = _.toArray(statefulSetData.by('metadata.name'));
+        sortApplications();
+      });
+    };
```
spadgett: We're doing this now in enough places now that it would be good to put this logic into a separate service (ApplicationsService?).

Not for this PR, though.

app/styles/_secrets.less
```diff
@@ -98,3 +98,51 @@ dl.secret-data {
     margin-bottom: 10px;
   }
 }
+
+.add-secret-to-application {
```
rhamilto: Nit:  alpha.  This rule should come first.

app/scripts/directives/addSecretToApplication.js
```diff
@@ -0,0 +1,171 @@
...
+
+      ctrl.disableInputs = true;
+
+      if (ctrl.addType === 'env') {
+        var newEnvFrom = {
+          secretRef: ctrl.secret.metadata
+        };
+
+        // For each container, add the new volume mount.
+        _.each(podTemplate.spec.containers, function(container) {
```
spadgett: We should open a follow on for letting the user pick containers if there's more than one.

app/styles/_secrets.less
```diff
@@ -98,3 +98,51 @@ dl.secret-data {
...
+    .button-group {
+      .btn {
+        margin-left: 10px;
+        &:first-of-type {
+          margin-left: 0;
+        }
+      }
+    }
+    .form-group > label {
+      font-size: 14px;
```
rhamilto: Use variable (e.g., @font-size-base + 1) so font sizes stay customizable.

app/scripts/directives/addSecretToApplication.js
```diff
@@ -0,0 +1,171 @@
...
+          container.envFrom = container.envFrom || [];
+          container.envFrom.push(newEnvFrom);
```
spadgett: We should call `angular.copy` before we start changing the resource. Otherwise it mutates a value that's in the DataService cache

app/scripts/directives/addSecretToApplication.js
```diff
@@ -0,0 +1,171 @@
...
+
+    ctrl.addToApplication = function() {
+      var podTemplate = _.get(ctrl.application, 'spec.template');
+
+      ctrl.disableInputs = true;
+
+      if (ctrl.addType === 'env') {
+        var newEnvFrom = {
+          secretRef: ctrl.secret.metadata
+        };
```
spadgett: I'd prefer only adding valid API fields, even if other metdata is ignored. So

```js
var newEnvFrom = {
  secretRef: {
    name: ctrl.secret.metadata.name
  }
};
```

app/views/directives/add-secret-to-application.html
```diff
@@ -0,0 +1,90 @@
+<div>
+  <div class="dialog-title">
+    <h3>Add to Application</h3>
+  </div>
+  <div class="dialog-body">
+    <form name="addToApplicationForm" novalidate>
+      <fieldset ng-disabled="disableInputs">
+        <div class="form-group">
+          <label>Add this secret to application:</label>
```
rhamilto: Ideally label elements get a for attribute to associate them with the input they are for (here and throughout the file).

app/scripts/directives/addSecretToApplication.js
```diff
@@ -0,0 +1,171 @@
...
+
+        // For each container, add the new volume mount.
+        _.each(podTemplate.spec.containers, function(container) {
+          container.envFrom = container.envFrom || [];
+          container.envFrom.push(newEnvFrom);
+        });
+      } else {
+        var generateName = $filter('generateName');
+        var name = generateName('volume-');
```
spadgett: It might be better to use the secret name as the generateName prefix so it's more obvious what the volume is for.

app/views/directives/add-secret-to-application.html
```diff
@@ -0,0 +1,90 @@
...
+              <ui-select-choices
+                repeat="application in (ctrl.applications) | filter : { metadata: { name: $select.search } } track by (application | uid)"
+                group-by="ctrl.groupByKind">
+                <span ng-bind-html="application.metadata.name | highlight : $select.search"></span>
+              </ui-select-choices>
+            </ui-select>
+          </div>
+        </div>
+        <div class="form-group">
+          <label>Add secret as:</label>
```
rhamilto: Semantically, this would be better as a legend element inside of a fieldset.  Not sure we have styling for that...

app/views/browse/secret.html
```diff
@@ -32,6 +38,8 @@ <h1 class="contains-actions">
               </li>
             </ul>
           </div>
+          <div class="pull-right" ng-hide="!('secrets' | canIDoAny)">
```
spadgett: We should probably just remove the `canIDoAny` check if "Add to Application" is always enabled. Otherwise we need to update the `canIDoAny` filter for secrets.

app/scripts/directives/addSecretToApplication.js
```diff
@@ -0,0 +1,171 @@
...
+      var context = {
+        namespace: ctrl.project.metadata.name
+      };
+
+      DataService.update(APIService.kindToResource(ctrl.application.kind), ctrl.application.metadata.name, ctrl.application, context).then(
+        function() {
+          NotificationsService.addNotification({
+            type: "success",
+            message: "Successfully added " + sourceKind + " " + ctrl.secret.metadata.name + " to " + targetKind + " " + ctrl.application.metadata.name + "."
+          });
```
spadgett: Might be nice to link to the deployment config here.

jeff-phillips-18:
@rhamilto Addressed your issues.

Also noticed the placement of the volume input did not match the design, updated:

image

app/scripts/directives/addSecretToApplication.js
jeff-phillips-18: Heh, I thought the same and started working on it but I ran out of time to make the change.

Added issue: https://github.com/openshift/origin-web-common/issues/175

app/scripts/directives/addSecretToApplication.js
jeff-phillips-18: Added https://github.com/openshift/origin-web-console/issues/2024

app/views/browse/secret.html
jeff-phillips-18: Sorry this was a left over empty div.

jeff-phillips-18:
@spadgett I believe I have addressed your issues. Thanks!

app/scripts/directives/addSecretToApplication.js
```diff
@@ -0,0 +1,180 @@
...
+      };
+
+      DataService.update(APIService.kindToResource(applicationToUpdate.kind), applicationToUpdate.metadata.name, applicationToUpdate, context).then(
+        function() {
+          NotificationsService.addNotification({
+            type: "success",
+            message: "Successfully added " + sourceKind + " " + ctrl.secret.metadata.name + " to " + targetKind + " " + applicationToUpdate.metadata.name + ".",
+            links: [{
+              href: Navigate.resourceURL(applicationToUpdate),
+              label: "View " + targetKind
```
spadgett: Nit: better to call `humanizeKind(applicationToUpdate.kind, true)` to get title case. This is consistent with our other notification links.

spadgett:
[merge]

app/scripts/directives/addSecretToApplication.js
jeff-phillips-18: Fixed

PR #2034, opened 2017-09-08
Use new ApplicationsService.getApplications utility

jeff-phillips-18:


jeff-phillips-18:
Removed WIP status

spadgett:
[merge][severity: bug]

app/scripts/directives/addSecretToApplication.js
```diff
@@ -21,63 +22,16 @@
...
-        sortApplications();
-      });
-      DataService.list({
-        group: 'apps',
-        resource: 'statefulsets'
-      }, context).then(function(statefulSetData) {
-        statefulSets = _.toArray(statefulSetData.by('metadata.name'));
-        sortApplications();
+      ApplicationsService.getApplications(context).then(function(applications) {
+        ctrl.applications = _.sortBy(applications, ['metadata.name', 'kind']);
```
spadgett: ApplicationService sorts, so we shouldn't need to sort here.

https://github.com/openshift/origin-web-common/blob/master/src/services/applicationsService.js#L28

app/scripts/directives/bindService.js
```diff
@@ -106,41 +94,13 @@
...
-        sortApplications();
-      });
-      DataService.list({
-        group: 'apps',
-        resource: 'statefulsets'
-      }, context).then(function(statefulSetData) {
-        statefulSets = _.toArray(statefulSetData.by('metadata.name'));
-        sortApplications();
+      ApplicationsService.getApplications(context).then(function(applications) {
+        ctrl.applications = _.sortBy(applications, ['metadata.name', 'kind']);
```
spadgett: Shouldn't need to sort here.

app/scripts/directives/addSecretToApplication.js
jeff-phillips-18: Fixed

app/scripts/directives/bindService.js
jeff-phillips-18: Fixed

spadgett:
[merge]

PR #2039, opened 2017-09-08
Warn when secret keys are not valid env var names

### LINKED ISSUE #2029: Should warn when secret keys are not valid env var names
In the "Add to Application" dialog, we should warn when some secret keys aren't valid environment variable names. These keys will be skipped when adding env variables.

cc @beanh66

jeff-phillips-18:
fixes #2029 

image

@beanh66

### DIFF
--- a/app/scripts/directives/addSecretToApplication.js
+++ b/app/scripts/directives/addSecretToApplication.js
@@ -84,7 +84,13 @@
     ctrl.$onInit = function() {
       ctrl.addType = 'env';
       ctrl.disableInputs = false;
+
       getApplications();
+
+      var keyValidator = new RegExp("^[A-Za-z_]{1}[A-Za-z0-9_]*$");
+      ctrl.hasInvalidEnvVars = _.some(ctrl.secret.data, function(value, key) {
+        return !keyValidator.test(key);
+      });
     };
 
     ctrl.$postLink = function() {
--- a/app/views/directives/add-secret-to-application.html
+++ b/app/views/directives/add-secret-to-application.html
@@ -26,18 +26,24 @@
         <legend>Add secret as:</legend>
         <div class="form-group">
           <div class="radio">
-            <label class="add-choice" for="envFrom">
-              <input id="envFrom" type="radio" ng-model="ctrl.addType" value="env" ng-disabled="ctrl.disableInputs">
-              Environment variables
-            </label>
+            <div class="add-choice">
+              <label>
+                <input type="radio" ng-model="ctrl.addType" value="env" ng-disabled="ctrl.disableInputs">
+                Environment variables
+              </label>
+              <div class="alert alert-warning  env-warning" ng-show="ctrl.hasInvalidEnvVars">
+                <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
+                <span>Some of the keys for secret <strong>{{ctrl.secret.metadata.name}}</strong> are not valid environment variable names and will not be added.</span>
+              </div>
+            </div>
             <div>
-              <label class="add-choice" for="mountVolume">
-                <input type="radio" ng-model="ctrl.addType" value="volume" ng-disabled="ctrl.disableInputs">
+              <label class="add-choice" for="volume">
+                <input id="volume" type="radio" ng-model="ctrl.addType" value="volume" ng-disabled="ctrl.disableInputs">
                 Volume
               </label>
             </div>
             <div class="volume-options">
-              <div ng-class="{'has-error': (addToApplicationForm.mountVolume.$error.pattern && addToApplicationForm.mountVolume.$touched)}">
+              <div ng-class="{'has-error': (addToApplicationForm.mountVolume.$error.oscUnique || (addToApplicationForm.mountVolume.$error.pattern && addToApplicationForm.mountVolume.$touched))}">
                 <input class="form-control"
                        name="mountVolume"
                        id="mountVolume"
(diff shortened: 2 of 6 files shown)


app/views/directives/add-secret-to-application.html
```diff
@@ -26,18 +26,24 @@
...
+              <div class="alert alert-warning  env-warning" ng-show="ctrl.invalidEnvVars.length">
+                <span class="pficon pficon-warning-triangle-o"></span>
+                <span>Some of the keys for secret <strong>{{ctrl.secret.metadata.name}}</strong> are not valid environment variable names and will not be added.</span>.
+              </div>
+            </div>
             <div>
-              <label class="add-choice" for="mountVolume">
-                <input type="radio" ng-model="ctrl.addType" value="volume" ng-disabled="ctrl.disableInputs">
+              <label class="add-choice" for="volume">
+                <input id="volume" type="radio" ng-model="ctrl.addType" value="volume" ng-disabled="ctrl.disableInputs">
```
spadgett: If the `<label>` element wraps the `<input>`, you don't need the `for` and `id` attributes.

app/views/directives/add-secret-to-application.html
```diff
@@ -26,18 +26,24 @@
...
-              Environment variables
-            </label>
+            <div class="add-choice">
+              <label for="envFrom">
+                <input id="envFrom" type="radio" ng-model="ctrl.addType" value="env" ng-disabled="ctrl.disableInputs">
+                Environment variables
+              </label>
+              <div class="alert alert-warning  env-warning" ng-show="ctrl.invalidEnvVars.length">
+                <span class="pficon pficon-warning-triangle-o"></span>
+                <span>Some of the keys for secret <strong>{{ctrl.secret.metadata.name}}</strong> are not valid environment variable names and will not be added.</span>.
```
spadgett: Looks like you have two periods

app/views/directives/add-secret-to-application.html
```diff
@@ -26,18 +26,24 @@
...
-              <input id="envFrom" type="radio" ng-model="ctrl.addType" value="env" ng-disabled="ctrl.disableInputs">
-              Environment variables
-            </label>
+            <div class="add-choice">
+              <label for="envFrom">
+                <input id="envFrom" type="radio" ng-model="ctrl.addType" value="env" ng-disabled="ctrl.disableInputs">
+                Environment variables
+              </label>
+              <div class="alert alert-warning  env-warning" ng-show="ctrl.invalidEnvVars.length">
+                <span class="pficon pficon-warning-triangle-o"></span>
```
spadgett: `aria-hidden="true"`

app/scripts/directives/addSecretToApplication.js
```diff
@@ -84,7 +84,17 @@
...
+
       getApplications();
+
+      var keyValidator = new RegExp("[A-Za-z_][A-Za-z0-9_]*");
+
+      _.each(_.keys(ctrl.secret.data), function(nextKey) {
+        if (!keyValidator.test(nextKey)) {
+          ctrl.invalidEnvVars.push(nextKey);
+        }
+      });
```
spadgett: If you only care if some key is invalid, this could be

```js
ctrl.hasInvalidEnvVars = _.some(ctrl.secret.data, function(value, key) {
  return !keyValidator.test(key);
});
```

Otherwise, I'd probably avoid the extra `keys` call by using

```js
_.each(ctrl.secret.data, function(value, key) {
  if (!keyValidator.test(key)) {
    ctrl.invalidEnvVars.push(key);
  }
});
```

app/views/directives/add-secret-to-application.html
jeff-phillips-18: If I remove it then clicking on the label does not toggle the radio. It's hard to click on just the little radio button.

app/scripts/directives/addSecretToApplication.js
jeff-phillips-18: Yeah, I was going do do that but wasn't sure if we wanted to be able to list the invalid keys at some point. Thoughts?

app/views/directives/add-secret-to-application.html
spadgett: Really? Hm. It seems to work here, but maybe browser specific?

https://getbootstrap.com/docs/3.3/css/#checkboxes-and-radios

OK to leave if it's needed.

jeff-phillips-18: You're right, dunno what I was doing wrong before...

app/scripts/directives/addSecretToApplication.js
jeff-phillips-18: Still working on getting the validator to work correctly. 
      var keyValidator = new RegExp("[A-Za-z_][A-Za-z0-9_]*");

is allowing 'database-name' (as well as 'database=name')...

spadgett: Maybe

```js
var keyValidator = new RegExp("^[A-Za-z_][A-Za-z0-9_]+$");
```

jeff-phillips-18: I believe I've got it working with:
```javascript
      var keyValidator = new RegExp("^[A-Za-z_]{1}[A-Za-z0-9_]{0,}$");
```

spadgett: Isn't `*` OK instead of `{0,}`?

jeff-phillips-18: Yes, it is. Got carried away with some other attempts... ;)

jeff-phillips-18:
Believe I have addressed all of @spadgett's comments.

beanh66:
Thanks @jeff-phillips-18 the new visuals look great 👍

spadgett:
[merge][severity: bug]

spadgett:
[merge][severity: bug]

spadgett:
WebDriver flake

[merge][severity:bug]

spadgett:
[merge][severity:bug]

spadgett:
[merge][severity:bug]

PR #2046, opened 2017-09-11
Update notification drawer to use trackByID

benjaminapetersen:
Depends on release including [PR 180](https://github.com/openshift/origin-web-common/pull/180) from origin-web-common.

@spadgett @jeff-phillips-18

### DIFF
--- a/app/views/directives/notifications/notification-drawer-wrapper.html
+++ b/app/views/directives/notifications/notification-drawer-wrapper.html
@@ -7,7 +7,7 @@
   heading-include="{{$ctrl.headingInclude}}"
   notification-body-include="{{$ctrl.notificationBodyInclude}}"
   notification-groups="$ctrl.notificationGroups"
-  notification-track-field="uid"
+  notification-track-field="trackByID"
   on-close="$ctrl.onClose"
   on-clear-all="$ctrl.onClearAll"
   on-mark-all-read="$ctrl.onMarkAllRead"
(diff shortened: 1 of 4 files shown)


app/views/directives/notifications/notification-drawer-wrapper.html
```diff
@@ -7,7 +7,7 @@
   heading-include="{{$ctrl.headingInclude}}"
   notification-body-include="{{$ctrl.notificationBodyInclude}}"
   notification-groups="$ctrl.notificationGroups"
-  notification-track-field="uid"
+  notification-track-field="trackByID"
```
spadgett: Don't you need to set `trackByID` here for events?

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/notifications/notificationDrawerWrapper.js#L161

benjaminapetersen: Yup! looking at that now, I'll have an update in a second.

benjaminapetersen: I think mapping `event.metadata.uid` is fine.  not sure if it really needs the same format of ` _.uniqueId('notification-') + Date.now()`

benjaminapetersen: yeah has to be this way with `count` increments & things.

benjaminapetersen:
updated

spadgett:
[merge][severity: bug]

spadgett:
Flake #1685 (not the one we just fixed!)

[merge][severity: bug]

PR #2047, opened 2017-09-11
Add `serviceClass` url param to jump into service-catalog flow

david-martin:
https://trello.com/c/w8LiRRSg

If a `serviceClass` param is present, look up the serviceClass and
launch the service-catalog modal for that class.
The param can be the service id e.g. 'cakephp-mysql-persistent'
or the service display name e.g. CakePHP + MySQL (Persistent)

Fixes https://github.com/openshift/origin-web-catalog/issues/423

The reason for allowing the service id or the service display is to allow for 'not very deterministic' id's.
This is somewhat the case for service classes returned by the ansible-service-broker.
For example, an apb for keycloak would have an generated serviceclass name of `dh-myorg-keycloak-latest`,
where `dh` is the image stream, `myorg` is the org, `keycloak` is the apb name and `latest` is the apb image tag. Having to know all 4 of these details seems a little contrived, rather than having a simple id.

### DIFF
--- a/app/scripts/controllers/landingPage.js
+++ b/app/scripts/controllers/landingPage.js
@@ -111,6 +111,24 @@ angular.module('openshiftConsole')
     });
 
     function dataLoaded() {
+      // Check for a service class param to launch the catalog flow for
+      var paramClass = $location.search()['serviceClass'];
+      if (paramClass) {
+        // Search by class name e.g. cakephp-mysql-persistent
+        var paramItem = _.find($scope.catalogItems, {
+          resource: {
+            metadata: {
+              name: paramClass
+            }
+          }
+        });
+        // If a catalog item matches, lauch the catalog flow
+        if (paramItem) {
+          $scope.$broadcast('open-overlay-panel', paramItem);
+          return;
+        }
+      }
+
       if (!tourEnabled) {
         return;
       }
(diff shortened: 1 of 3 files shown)


david-martin:
@spadgett This is my first attempt at https://github.com/openshift/origin-web-catalog/issues/423.
I'm interested in your thoughts on the param value and what field to lookup the class on (see description)

app/scripts/controllers/landingPage.js
```diff
@@ -111,6 +111,30 @@ angular.module('openshiftConsole')
...
 
     function dataLoaded() {
+      // Check for a service class param to launch the catalog flow for
+      var paramClass = $location.search()['serviceClass'];
+      if (paramClass) {
+        // Lookup the class by class name e.g. cakephp-mysql-persistent
+        var paramItem = _.find($scope.catalogItems, {
+          name: paramClass
+        });
+        // Lookup the class by display name e.g. "CakePHP + MySQL (Persistent)"
```
spadgett: It looks like you might have the order backwards here, checking display name first above and then `metadata.name` here.

david-martin:
@spadgett Understood. I'm in agreement after thinking through it some more. I'll modify to search the kubernetes name only.

david-martin:
@spadgett Anything else I can do here? Is it enough to satisfy the card? https://trello.com/c/w8LiRRSg/1050-url-params-to-jump-into-service-catalog-flow
I'm not sure what exactly this part means in the context of the web console "So that I can see an unauthenticated view of applications"

spadgett:
> Anything else I can do here? Is it enough to satisfy the card? 

It should satisfy the card.

spadgett:
[merge]

PR #2052, opened 2017-09-12
Remove incorrect warning when deleting build config

spadgett:
Not all build history is necessarily deleted, so remove that message.
Builds without owner references will no be garbage collected.

Closes https://github.com/openshift/origin/issues/16177


spadgett:
[merge][severity: bug]

PR #2054, opened 2017-09-12
Bug 1490688: Do not disable filter input for importing a template from a project

jeff-phillips-18:
https://bugzilla.redhat.com/show_bug.cgi?id=1490688

The real issue in the bug is that the input should have been disabled to begin with since there is only 1 template available. The input is not yet available so it was not being disabled. No real need to disable the input at all. It was only disabled when there is only one item but that is kind of pointless since I don't really need to filter with 2 or 3 or 4...  (confirmed with @beanh66)


spadgett:
[merge][severity: blocker]

spadgett:
@jeff-phillips-18 @beanh66 Do you we want to make the same change on the landing page for consistency?

jeff-phillips-18:
@spadgett I would agree with that. I will open an issue for it.

PR #2058, opened 2017-09-12
Allow the user to select containers when adding a secret to an application

### LINKED ISSUE #2024: Allow the user to select containers when adding a secret to an application

jeff-phillips-18:
Fixes #2024

image

image

### DIFF
--- a/app/views/directives/add-secret-to-application.html
+++ b/app/views/directives/add-secret-to-application.html
@@ -70,6 +70,23 @@
             </div>
           </div>
         </div>
+        <legend ng-if-start="ctrl.application.spec.template.spec.containers.length > 1">Containers:</legend>
+        <div ng-if-end class="form-group container-options">
+          <div ng-if="ctrl.attachAllContainers">
+            The secret will be added to all containers. You can
+            <a href="" ng-click="ctrl.attachAllContainers = false">select specific containers</a>
+            instead.
+          </div>
+          <div ng-if="!ctrl.attachAllContainers" class="form-group">
+            <label class="sr-only required">Containers</label>
+            <select-containers
+              ng-model="ctrl.attachContainers"
+              pod-template="ctrl.application.spec.template"
+              ng-required="true"
+              help-text="Add the secret to the selected containers.">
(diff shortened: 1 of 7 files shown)


app/views/directives/add-secret-to-application.html
```diff
@@ -69,6 +69,23 @@
               </div>
             </div>
           </div>
+          <div ng-if="ctrl.addType === 'volume' && ctrl.application.spec.template.spec.containers.length > 1"  class="volume-options">
```
spadgett: We should go ahead do the same for environment variables as well I think

app/views/directives/add-secret-to-application.html
```diff
@@ -69,6 +69,23 @@
...
+            <h4>Containers</h4>
+            <div ng-if="ctrl.attachAllContainers">
+              The volume will be mounted into all containers. You can
+              <a href="" ng-click="ctrl.attachAllContainers = false">select specific containers</a>
+              instead.
+            </div>
+            <div ng-if="!ctrl.attachAllContainers" class="form-group">
```
spadgett: @jeff-phillips-18 Is there a max-height on the dialog in case there are a lot of containers?

jeff-phillips-18: nope, I suppose there ought to be. Thanks!

app/views/directives/add-secret-to-application.html
jeff-phillips-18: Ah, OK

jeff-phillips-18: Done

app/views/directives/add-secret-to-application.html
jeff-phillips-18: Set the max height on the list of containers.

jeff-phillips-18:
Updates to allow container selection for both Environment Variables and Volume

image

with scrolling containers:

image

spadgett:
[merge][severity: bug]

spadgett:
[merge][severity: bug]

Issue #2070, opened 2017-09-13
Long unbroken image-name strings break the monitoring page layout and cause horizontal scrollbar


sg00dwin:
<img width="1011" alt="screen shot 2017-09-13 at 4 18 34 pm" src="https://user-images.githubusercontent.com/1874151/30398965-3583a85e-98a0-11e7-92f0-9c386f9d23d6.png">

sg00dwin:
I'm working on a fix

PR #2071, opened 2017-09-13
Refactor markup and wrap icon and text, then set break long string names...

sg00dwin:
so they wrap.

This corrects this issue. Also noting, after speaking with @rhamilto this page uses an older/modified `.list-view-pf` which is different from what's used on the overview `.list-pf`. We should look into switching to `.list-pf`

Fixes https://github.com/openshift/origin-web-console/issues/2070
Cross browser tested

screen shot 2017-09-13 at 4 02 41 pm
screen shot 2017-09-13 at 4 05 38 pm
screen shot 2017-09-13 at 4 06 03 pm
screen shot 2017-09-13 at 4 07 43 pm

### DIFF
--- a/app/views/_image-names.html
+++ b/app/views/_image-names.html
@@ -1,5 +1,9 @@
-<span>{{podTemplate.spec.containers[0].image | imageStreamName}}</span>
-<span ng-repeat="id in imageIDs" title="{{id}}">
-  <span class="hash">{{id | stripSHAPrefix | limitTo: 7}}</span><span ng-if="!$last">,</span>
-</span>
-<span ng-if="podTemplate.spec.containers.length > 1"> and {{podTemplate.spec.containers.length - 1}} other image<span ng-if="podTemplate.spec.containers.length > 2">s</span></span>
+<div class="text-prepended-icon">
+  <span class="pficon pficon-image" aria-hidden="true"></span>
+  <span class="word-break-all">{{podTemplate.spec.containers[0].image | imageStreamName}}
+  <span ng-repeat="id in imageIDs" title="{{id}}">
+    <span class="hash nowrap">{{id | stripSHAPrefix | limitTo: 7}}</span><span ng-if="!$last">,</span>
+  </span>
+  <span class="nowrap" ng-if="podTemplate.spec.containers.length > 1"> and {{podTemplate.spec.containers.length - 1}} other image<span ng-if="podTemplate.spec.containers.length > 2">s</span></span>
+  </span>
+</div>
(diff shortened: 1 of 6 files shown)


app/views/_image-names.html
```diff
@@ -1,5 +1,7 @@
-<span>{{podTemplate.spec.containers[0].image | imageStreamName}}</span>
+<span class="pficon fa-fw pficon-image"></span>
```
spadgett: `aria-hidden="true"`

I realize this wasn't there before, but we should fix.

app/views/monitoring.html
```diff
@@ -186,8 +185,8 @@
                       </div>
                       <div class="list-view-pf-additional-info">
                         <div class="list-view-pf-additional-info-item">
-                          <span class="pficon fa-fw pficon-image"></span>
                           <image-names
+                              class="text-prepended-icon word-break"
```
spadgett: Is there any way we can move these classes into the `image-names` view, or do they need to be on this element to work with `pf-list`?

sg00dwin:
@spadgett PTAL

app/views/_image-names.html
```diff
@@ -1,5 +1,9 @@
-<span>{{podTemplate.spec.containers[0].image | imageStreamName}}</span>
-<span ng-repeat="id in imageIDs" title="{{id}}">
-  <span class="hash">{{id | stripSHAPrefix | limitTo: 7}}</span><span ng-if="!$last">,</span>
-</span>
-<span ng-if="podTemplate.spec.containers.length > 1"> and {{podTemplate.spec.containers.length - 1}} other image<span ng-if="podTemplate.spec.containers.length > 2">s</span></span>
+<div class="text-prepended-icon">
+  <span class="pficon pficon-image"></span>
```
spadgett: aria-hidden="true"

app/views/monitoring.html
```diff
@@ -383,8 +379,8 @@
                       </div>
                       <div class="list-view-pf-additional-info">
                         <div class="list-view-pf-additional-info-item">
-                          <span ng-if="build.spec.source.type || build.spec.revision.git.commit || build.spec.source.git.uri">
-                            <span class="fa fa-fw fa-code"></span>
+                          <div class="text-prepended-icon word-break" ng-if="build.spec.source.type || build.spec.revision.git.commit || build.spec.source.git.uri">
+                            <span class="fa fa-code"></span>
```
spadgett: aria-hidden="true"

sg00dwin: yep, i missed that

sg00dwin: done

spadgett:
[merge][severity: bug]

PR #2090, opened 2017-09-15
Add track by to overview service instance row

spadgett:
This fixes a bug where the binding dialog can disappear when the provisioned service status changes.


spadgett:
Thanks @jeff-phillips-18 

[merge][severity: blocker]

PR #2092, opened 2017-09-15
Show creation timestamp in binding delete dialog

spadgett:
Per @ncameronbritt 

This makes it a little easier to tell the bindings apart. Since pod presets are disabled, we show the secret names only now, and they will all have **very** similar names.

openshift web console 2017-09-15 15-32-24


spadgett:
[merge][severity: bug]

spadgett:
@ncameronbritt @jeff-phillips-18 I'm thinking we might change the step label from "Applications" to something else since pod presets are disabled

PR #2093, opened 2017-09-15
Don't show "Applications" as step label when pod presets are disabled

spadgett:
@ncameronbritt Opinion on this change? The step title used to be "Applications". Noticed this after working on #2092

openshift web console 2017-09-15 17-46-48

### DIFF
--- a/app/scripts/directives/unbindService.js
+++ b/app/scripts/directives/unbindService.js
@@ -23,6 +23,7 @@
     var ctrl = this;
     var validityWatcher;
     var context;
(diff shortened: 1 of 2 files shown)


ncameronbritt:
@spadgett I think that makes sense. Good catch. It must have just been an oversight that it said "applications" before.

spadgett:
> It must have just been an oversight that it said "applications" before.

We used to show the name of the applications the service was bound to. Now that pod presets are disabled, it's always a secret.

spadgett:
[merge][severity: bug]

PR #2094, opened 2017-09-16
"Add to Application" for config maps

spadgett:
Implement "Add to Application" on the config map page like we've already done for secrets. Reusing the component @jeff-phillips-18 added in #2021

openshift web console 2017-09-16 13-14-46


spadgett:
[test]

spadgett:
@jeff-phillips-18 PTAL

app/views/browse/config-map.html
```diff
@@ -58,7 +74,7 @@ <h1 class="contains-actions">
                     <truncate-long-text
                         content="value"
                         limit="1024"
-                        newlineLimit="20"
+                        newline-limit="20"
```
jeff-phillips-18: This same error occurs at https://github.com/openshift/origin-web-console/blob/master/app/views/browse/config-map.html#L61

Not part of this PR but might be worth fixing here.

spadgett: Thanks, looks like that might be the same line as here?

Did a quick search for this, and I don't see any other references. I know @rhamilto fixed one in another place.

spadgett:
[merge]

spadgett:
[merge]

PR #2096, opened 2017-09-18
Update addConfigVolume controller to use getPreferredVersion

benjaminapetersen:
Moving these along.
Depends on web console [PR 161](https://github.com/openshift/origin-web-common/pull/161)

Question still lingers, is `getPreferredVersion` the best name? I pause probably because `version` becomes overloaded.  This just means "get what we are consistenly using", even though the method actually returns a `{resource, group}` w/o the version.

@spadgett

### DIFF
--- a/app/scripts/controllers/addConfigVolume.js
+++ b/app/scripts/controllers/addConfigVolume.js
@@ -62,6 +62,9 @@ angular.module('openshiftConsole')
       subpage: 'Add Config Files'
     });
 
+    $scope.configMapVersion = APIService.getPreferredVersion('configmaps');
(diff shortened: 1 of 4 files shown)


spadgett:
> Question still lingers, is getPreferredVersion the best name?

I can't come up with much better unfortunately. We could use something really short like `APIService.rgv()`, but there's already a `toResourceGroupVersion`, which might be confusing.

Any suggestions?

benjaminapetersen:
Agree, have been thinking about it over weekend, nothing better comes to mind either.

benjaminapetersen:
rebased & added view

spadgett:
[merge]

PR #2097, opened 2017-09-18
Update attachPVC controller to use getPreferredVersion

benjaminapetersen:
Another

Depends on web console [PR 161](https://github.com/openshift/origin-web-common/pull/161)

@spadgett

### DIFF
--- a/app/scripts/controllers/attachPVC.js
+++ b/app/scripts/controllers/attachPVC.js
@@ -67,6 +67,10 @@ angular.module('openshiftConsole')
       subpage: 'Add Storage'
     });
 
+    $scope.pvcVersion = APIService.getPreferredVersion('persistentvolumeclaims');
+    var resourceQuotasVersion = APIService.getPreferredVersion('resourcequotas');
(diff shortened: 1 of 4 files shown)


benjaminapetersen:
Rebased & added view.

app/scripts/controllers/attachPVC.js
```diff
@@ -67,6 +67,10 @@ angular.module('openshiftConsole')
       subpage: 'Add Storage'
     });
 
+    $scope.pvcVersion = APIService.getPreferredVersion('persistentvolumeclaims');
+    $scope.quotaVersion = APIService.getPreferredVersion('resourcequotas');
+    $scope.appliedQuotaVersion = APIService.getPreferredVersion('appliedclusterresourcequotas');
```
spadgett: If we're going to shorten the name, I prefer `clusterQuotaVersion`. That makes it more obvious it's cross namespace.

spadgett: I'm good with `clusterQuotaVersion`

benjaminapetersen:
rebased & updated vars to full name.

PR #2098, opened 2017-09-18
Update build controller to use getPreferredVersion

benjaminapetersen:
Depends on web common [PR 161](https://github.com/openshift/origin-web-common/pull/161)

### DIFF
--- a/app/views/browse/build.html
+++ b/app/views/browse/build.html
@@ -10,7 +10,7 @@
             <!-- Primary Actions -->
             <button class="btn btn-default hidden-xs"
                     ng-click="cancelBuild()"
-                    ng-if="!build.metadata.deletionTimestamp && (build | isIncompleteBuild) && ('builds' | canI : 'update')">Cancel Build</button>
+                    ng-if="!build.metadata.deletionTimestamp && (build | isIncompleteBuild) && (buildsVersion | canI : 'update')">Cancel Build</button>
             <button class="btn btn-default hidden-xs"
                     ng-click="cloneBuild()"
                     ng-hide="build.metadata.deletionTimestamp || (build | isIncompleteBuild) || !('builds/clone' | canI : 'create') || (build | isBinaryBuild)"
@@ -25,7 +25,7 @@
                class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
                data-toggle="dropdown"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
             <ul class="dropdown-menu dropdown-menu-right actions action-button">
-              <li ng-if="'buildconfigs' | canI : 'update'">
+              <li ng-if="buildConfigsVersion | canI : 'update'">
                 <a ng-href="{{buildConfig | editResourceURL}}" role="button">
                   <span ng-if="!(buildConfig | isJenkinsPipelineStrategy)">
                     Edit Configuration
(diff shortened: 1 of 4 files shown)


spadgett:
@benjaminapetersen Let's update the views together with the controllers. There are a lot of `canI` checks in the views. We might need a filter for preferred version.

benjaminapetersen:
Looking at the views now:

```html 
<div class="pull-right dropdown" ng-hide="!('builds' | canIDoAny)">
<li ng-if="'buildconfigs' | canI : 'update'">
```
`canI` & `canIDoAny` filter is just expect strings, as does `AuthorizationService.canI` and `APIService.toResourceGroupVersion`.

spadgett:
> canI & canIDoAny filter is just expect strings, as does AuthorizationService.canI and APIService.toResourceGroupVersion.

They can take objects, too.

benjaminapetersen:
So `canI` works with an object, but `canIDoAny` does not:

```JavaScript
// canIDoAny will only take a string currently.
// I'm not sure it makes sense to update to take an object:
.filter('canIDoAny', function(canIFilter) { 
  var resourceRulesMap = { 
     'buildConfigs': [
      ],
      // etc, etc
  };
  return function(resource) {
    // resourceRulesMap[resource] works if resource is a string 
    // if we pass an object & just pluck out resource.resource, 
    // does it make sense to bother with the `| preferredVersion` filter? 
    return _.some(resourceRulesMap[resource], function(rule) {
      return _.some(rule.verbs, function(verb) {
        return canIFilter({resource: rule.resource, group: rule.group}, verb);
      });
    });
  };
})
```
I also did not handle sub-resources like `builds/clone`, or `builds/log` such as `('builds/clone' | canI : 'create')` in [API_PREFERRED_VERSIONS](https://github.com/openshift/origin-web-common/blob/master/src/constants/apiPreferredVersions.js#L4).

spadgett:
`canIDoAny` is for specific console pages and doesn't take in a kubernetes resource. It takes in a string that corresponds to a page. I wouldn't worry about that one.

benjaminapetersen:
Right. ok wanted to make sure we were on the same page. It didn't seem productive to alter that filter, though I may leave a comment if `build` = build page, not build resource.

spadgett:
It could be commented better. Honestly, I'd prefer just to add that logic to each controller rather than using a filter since it's specific to the one page. I think it's confusing as implemented. But I wouldn't change it right now.

In general, I prefer this approach:

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/overview/serviceInstanceRow.js#L75-L95

The `canIDoAny` filter itself can be updated to use the API groups, though.

benjaminapetersen:
Created an issue, will ignore for now, thx!

app/views/browse/build.html
```diff
@@ -10,10 +10,11 @@ <h1 class="contains-actions">
             <!-- Primary Actions -->
             <button class="btn btn-default hidden-xs"
                     ng-click="cancelBuild()"
-                    ng-if="!build.metadata.deletionTimestamp && (build | isIncompleteBuild) && ('builds' | canI : 'update')">Cancel Build</button>
+                    ng-if="!build.metadata.deletionTimestamp && (build | isIncompleteBuild) && ('builds' | preferredVersion | canI : 'update')">Cancel Build</button>
```
benjaminapetersen: I think I'm not a fan of using the filter unless necessary.  This creates a lot of template noise.  Going to update to use `$scope.buildVersion` instead.

benjaminapetersen:
@spadgett
- includes template updates.
- does not update `events` or `builds/log` in template, though [started a PR](https://github.com/openshift/origin-web-common/pull/193) here for missing versions from the const.

spadgett:
[merge]

spadgett:
@benjaminapetersen Missed this on review, but can you fix in a follow-on: We need to update the `canIDoAny` check for this page to use the same resource version

https://github.com/openshift/origin-web-console/blob/master/app/scripts/filters/canI.js#L11-L14

PR #1935, opened 2017-08-15
Add canI checks for route custom hosts

spadgett:
See https://github.com/openshift/origin/issues/15772

@smarterclayton FYI

PR #2101, opened 2017-09-18
File path input should always be readonly

spadgett:
The text input for uploading files should always be readonly since you can't type a file name. You have to use the browse button or drag and drop.

Regression introduced in #1935


spadgett:
[merge]

PR #2104, opened 2017-09-18
Fix case: "View All Projects"

spadgett:
Fix project menu case of the last item, "View all Projects" -> "View All Projects"


spadgett:
[merge]

PR #2106, opened 2017-09-18
Make close notification link accessible

spadgett:
The link was missing an `href` and could not be triggered using the keyboard.


spadgett:
[merge]

PR #2107, opened 2017-09-18
Let administrators disable "Copy Login Command"

spadgett:
* Add a `DISABLE_COPY_LOGIN_COMMAND` constant that prevents users from copying a login command with their token.
* Let users permanently dismiss the toast notification warning telling users not to share their token.
* Update "Log Out" to use headline case like other menu items.

@wgordon17 @serenamarie125


spadgett:
Follow on to #1788

spadgett:
[merge]

PR #2116, opened 2017-09-19
Quota Notifications

dtaylor113:
Push quota notifications to notification drawer:

quotas

[Trello](https://trello.com/c/ITpnCvsj/991-2-improve-quota-warnings-extensibility-usability)
[Design](https://github.com/openshift/openshift-origin-design/blob/master/web-console/5-components/quota-warnings.md)

### DIFF
--- a/app/scripts/constants.js
+++ b/app/scripts/constants.js
@@ -555,5 +555,10 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     //   href: 'http://example.com/',
     //   tooltip: 'Open Dashboard'
     // }
-  ]
+  ],
+  QUOTA_NOTIFICATION_MESSAGE: {
+    // Example quota messages to show in notification drawer
+    // "pods": "Upgrade to <a href='http://www.openshift.com'>OpenShift Pro</a> if you need additional resources.",
+    // "limits.memory": "Upgrade to <a href='http://www.openshift.com'>OpenShift Online Pro</a> if you need additional resources."
+  }
 });
--- a/app/scripts/directives/notifications/notificationDrawerWrapper.js
+++ b/app/scripts/directives/notifications/notificationDrawerWrapper.js
@@ -17,6 +17,7 @@
         'Constants',
         'DataService',
         'EventsService',
+        'NotificationsService',
         NotificationDrawerWrapper
       ]
     });
@@ -91,6 +92,12 @@
           group.totalUnread = unread(group.notifications).length;
           group.hasUnread = !!group.totalUnread;
           $rootScope.$emit('NotificationDrawerWrapper.onUnreadNotifications', group.totalUnread);
+        });
+      };
+
+      var removeNotificationFromGroup = function(notification) {
+        _.each(drawer.notificationGroups, function(group) {
+          _.remove(group.notifications, { uid: notification.uid, namespace: notification.namespace });
         });
       };
 
@@ -172,7 +179,7 @@
         var id = notification.id || _.uniqueId('notification_') + Date.now();
         notificationsMap[project] = notificationsMap[project] || {};
         notificationsMap[project][id] = {
-          actions: null,
+          actions: notification.actions,
           unread: !EventsService.isRead(id),
           // using uid to match API events and have one filed to pass
           // to EventsService for read/cleared, etc
@@ -183,6 +190,7 @@
           // but we sort based on lastTimestamp first.
           lastTimestamp: notification.timestamp,
           message: notification.message,
+          isHTML: notification.isHTML,
           details: notification.details,
           namespace: project,
           links: notification.links
@@ -261,7 +269,8 @@
           onLinkClick: function(link) {
             link.onClick();
             drawer.drawerHidden = true;
-          }
+          },
+          countUnreadNotifications: countUnreadNotifications
         }
       });
 
@@ -285,6 +294,18 @@
         // event from the counter to signal the drawer to open/close
         rootScopeWatches.push($rootScope.$on('NotificationDrawerWrapper.toggle', function() {
           drawer.drawerHidden = !drawer.drawerHidden;
+        }));
+
+        // event to signal the drawer to close
+        rootScopeWatches.push($rootScope.$on('NotificationDrawerWrapper.hide', function() {
+          drawer.drawerHidden = true;
+        }));
+
+        // event to signal the drawer to clear a notification
+        rootScopeWatches.push($rootScope.$on('NotificationDrawerWrapper.clear', function(event, notification) {
+          EventsService.markCleared(notification.uid);
+          removeNotificationFromGroup(notification);
+          drawer.countUnreadNotifications();
         }));
       };
 
--- a/app/scripts/services/quota.js
+++ b/app/scripts/services/quota.js
@@ -3,12 +3,21 @@
 angular.module("openshiftConsole")
   .factory("QuotaService", function(APIService,
                                     $filter,
+                                    $location,
+                                    $rootScope,
+                                    $routeParams,
                                     $q,
+                                    Constants,
                                     DataService,
-                                    Logger) {
+                                    EventsService,
+                                    Logger,
+                                    NotificationsService) {
 
     var isNil = $filter('isNil');
     var usageValue = $filter('usageValue');
+    var usageWithUnits = $filter('usageWithUnits');
+    var percent = $filter('percent');
+
     var isBestEffortPod = function(pod) {
       // To be best effort a pod must not have any containers that have non-zero requests or limits
       // Break out as soon as we find any pod with a non-zero request or limit
@@ -254,6 +263,99 @@ angular.module("openshiftConsole")
       });
     };
 
+    var COMPUTE_RESOURCE_QUOTAS = [
+      "cpu",
+      "requests.cpu",
+      "memory",
+      "requests.memory",
+      "limits.cpu",
+      "limits.memory"
+    ];
+
+    var getNotificaitonMessage = function(used, usedValue, hard, hardValue, quotaKey) {
+      // Note: This function returns HTML markup, not plain text
+
+      var msgPrefix = "Your project is " + (hardValue < usedValue ? 'over' : 'at') + " quota. ";
+      var msg;
+      if (_.includes(COMPUTE_RESOURCE_QUOTAS, quotaKey)) {
+        msg = msgPrefix + "It is using " + percent((usedValue/hardValue), 0) + " of " + usageWithUnits(hard, quotaKey) + " " + humanizeQuotaResource(quotaKey) + ".";
+      } else {
+        msg = msgPrefix + "It is using " + usedValue + " of " + hardValue + " " + humanizeQuotaResource(quotaKey) + ".";
+      }
+
+      msg = _.escape(msg);
+
+      if (Constants.QUOTA_NOTIFICATION_MESSAGE && Constants.QUOTA_NOTIFICATION_MESSAGE[quotaKey]) {
+        // QUOTA_NOTICIATION_MESSAGE can contain HTML and shouldn't be escaped.
+        msg += " " + Constants.QUOTA_NOTIFICATION_MESSAGE[quotaKey];
+      }
+
+      return msg;
+    };
+
+    // Return notifications if you are at quota or over any quota for any resource. Do *not*
+    // warn about quota for 'resourcequotas' or resources whose hard limit is
+    // 0, however.
+    var getQuotaNotifications = function(quotas, clusterQuotas, projectName) {
+      var notifications = [];
+
+      var notificationsForQuota = function(quota) {
+        var q = quota.status.total || quota.status;
+        _.each(q.hard, function(hard, quotaKey) {
+          var hardValue = usageValue(hard);
+          var used = _.get(q, ['used', quotaKey]);
+          var usedValue = usageValue(used);
+
+          // We always ignore quota warnings about being out of
+          // resourcequotas since end users cant do anything about it
+          if (quotaKey === 'resourcequotas' || !hardValue || !usedValue) {
+            return;
+          }
+
+          if(hardValue <= usedValue) {
+            notifications.push({
+              id: "quota-limit-reached-" + quotaKey,
+              namespace: projectName,
+              type: (hardValue < usedValue ? 'warning' : 'info'),
+              message: getNotificaitonMessage(used, usedValue, hard, hardValue, quotaKey),
+              isHTML: true,
+              skipToast: true,
+              showInDrawer: true,
+              actions: [
+                {
+                  name: 'View Quotas',
+                  title: 'View project quotas',
+                  onClick: function() {
+                    $location.url("/project/" + $routeParams.project + "/quota");
+                    $rootScope.$emit('NotificationDrawerWrapper.hide');
+                  }
+                },
+                {
+                  name: "Don't Show Me Again",
+                  title: 'Permenantly hide this notificaiton until quota limit changes',
+                  onClick: function(notification) {
+                    NotificationsService.permanentlyHideNotification(notification.uid, notification.namespace);
+                    $rootScope.$emit('NotificationDrawerWrapper.clear', notification);
+                  }
+                },
+                {
+                  name: "Clear",
+                  title: 'Clear this notificaiton',
+                  onClick: function(notification) {
+                    $rootScope.$emit('NotificationDrawerWrapper.clear', notification);
+                  }
+                }
+              ]
+            });
+          }
+        });
+      };
+      _.each(quotas, notificationsForQuota);
+      _.each(clusterQuotas, notificationsForQuota);
+
+      return notifications;
+    };
+
     // Warn if you are at quota or over any quota for any resource. Do *not*
     // warn about quota for 'resourcequotas' or resources whose hard limit is
     // 0, however.
@@ -324,6 +426,7 @@ angular.module("openshiftConsole")
       getLatestQuotaAlerts: getLatestQuotaAlerts,
       isAnyQuotaExceeded: isAnyQuotaExceeded,
       isAnyStorageQuotaExceeded: isAnyStorageQuotaExceeded,
-      willRequestExceedQuota: willRequestExceedQuota
+      willRequestExceedQuota: willRequestExceedQuota,
+      getQuotaNotifications: getQuotaNotifications
     };
   });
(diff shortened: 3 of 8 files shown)


app/scripts/services/resourceAlerts.js
```diff
@@ -70,36 +71,14 @@ angular.module("openshiftConsole")
...
-    var setGenericQuotaWarning = function(quotas, clusterQuotas, projectName, alerts) {
-      var isHidden = AlertMessageService.isAlertPermanentlyHidden("overview-quota-limit-reached", projectName);
-      if (!isHidden && QuotaService.isAnyQuotaExceeded(quotas, clusterQuotas)) {
-        if (alerts['quotaExceeded']) {
-          // Don't recreate the alert or it will reset the temporary hidden state
-          return;
+    var setQuotaNotifications = function(quotas, clusterQuotas, projectName) {
+      var notifications = QuotaService.getQuotaNotifications(quotas, clusterQuotas, projectName);
+      _.each(notifications, function(notification) {
+        // NotificationsService.restorePermanentlyHiddenNotification(notification.id, projectName);
```
dtaylor113: Oops...debug op...I'll remove before final merge

app/scripts/services/quota.js
```diff
@@ -254,6 +257,84 @@ angular.module("openshiftConsole")
...
+      "openshift.io/imagestreams"
+    ];
+
+    var getNotificaitonMessage = function(used, usedValue, hard, hardValue, quotaKey) {
+      var msgPrefix = "Your project is " + (hardValue < usedValue ? 'over' : 'at') + " quota. ";
+      var msg;
+      if (_.includes(OBJECT_COUNT_QUOTAS, quotaKey)) {
+        msg = msgPrefix + "It is using " + usedValue + " of " + hardValue + " " + humanizeQuotaResource(quotaKey, true) + ".";
+      } else {
+        msg = msgPrefix + "It is using " + (usedValue / hardValue * 100) + "% of " + usageWithUnits(hard, quotaKey) + " " + humanizeQuotaResource(quotaKey, true) + ".";
```
spadgett: We should round that percentage to avoid things like repeating decimals. I suggest using the `percent` filter:

https://github.com/openshift/origin-web-console/blob/c6ac85fec6ebff414312f658f467833ee09d3b8b/app/scripts/filters/util.js#L461-L470

app/scripts/services/quota.js
```diff
@@ -254,6 +257,84 @@ angular.module("openshiftConsole")
...
+
+          if(hardValue <= usedValue) {
+            notifications.push({
+              id: "quota-limit-reached-" + quotaKey,
+              namespace: projectName,
+              type: (hardValue < usedValue ? 'warning' : 'info'),
+              message: getNotificaitonMessage(used, usedValue, hard, hardValue, quotaKey),
+              skipToast: true,
+              showInDrawer: true,
+              actions: [
```
spadgett: Move the action on click handlers from the notification drawer to here. You can pass it as an `onClick` property in the action.

app/views/directives/notifications/notification-body.html
```diff
@@ -35,7 +36,7 @@
           href=""
           class="secondary-action"
           title="{{action.title}}"
-          ng-click="$ctrl.customScope.handleAction(notification, action)">
+          ng-click="$ctrl.customScope.handleAction(notification, action, $parent.$parent.$index, notificationGroup)">
```
spadgett: I'd change this to simply `action.onClick()` if you take my suggestion above.

app/views/directives/notifications/notification-body.html
```diff
@@ -71,7 +72,7 @@
         </span>
 
         <span ng-if="!(notification.event.involvedObject)">
-          {{notification.message}}
+          <span ng-bind-html="notification.message"></span>
```
spadgett: Can't assume that the message is HTML. We need a flag that says use HTML for the message. Otherwise plain text messages will not properly be escaped.

dtaylor113: Ok, should I check the message for HTML tags, or if I detect that a custom message is being used, is that enough to process it as HTML?

spadgett: No way to reliably detect if it has markup. I would document that the custom quota messages are markup. Then add another a flag like `notification.isHTML` that tells the drawer how to render it.

dtaylor113:
Hi @spadgett, I believe I have addressed all of your review issues.
Thanks,
Dave

app/scripts/constants.js
```diff
@@ -534,5 +534,10 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     //   href: 'http://example.com/',
     //   tooltip: 'Open Dashboard'
     // }
-  ]
+  ],
+  QUOTA_NOTIFICATION_MESSAGE: {
+    // Example quota messages to show in notification drawer
+    // "pods": "Upgrade to <a href='http://www.google.com'>OpenShift Pro</a> if you need additional resources.",
+    // "limits.memory": "Upgrade to <a href='http://www.google.com'>OpenShift Online Pro</a> if you need additional resources."
```
spadgett: nit: can we change the URLs to openshift.com?

app/scripts/filters/util.js
```diff
@@ -515,4 +515,13 @@ angular.module('openshiftConsole')
     return function(feature) {
       return _.get(Constants, ['ENABLE_TECH_PREVIEW_FEATURE', feature], false);
     };
+  })
+  .filter('hasHTML', function() {
```
spadgett: Let's not try to test for HTML tags because this can have false positives. I would remove this filter.

app/scripts/services/quota.js
```diff
@@ -254,6 +263,100 @@ angular.module("openshiftConsole")
...
+      var msg;
+      if (_.includes(COMPUTE_RESOURCE_QUOTAS, quotaKey)) {
+        msg = msgPrefix + "It is using " + percent((usedValue/hardValue), 0) + " of " + usageWithUnits(hard, quotaKey) + " " + humanizeQuotaResource(quotaKey, true) + ".";
+      } else {
+        msg = msgPrefix + "It is using " + usedValue + " of " + hardValue + " " + humanizeQuotaResource(quotaKey, true) + ".";
+      }
+
+      if (Constants.QUOTA_NOTIFICATION_MESSAGE && Constants.QUOTA_NOTIFICATION_MESSAGE[quotaKey]) {
+        msg += " " + Constants.QUOTA_NOTIFICATION_MESSAGE[quotaKey];
+      }
```
spadgett: Always generate an HTML message here and set `isHTML: true` below.

```js
msg = _.escape(msg);
if (Constants.QUOTA_NOTIFICATION_MESSAGE && Constants.QUOTA_NOTIFICATION_MESSAGE[quotaKey]) {
  msg += " " + Constants.QUOTA_NOTIFICATION_MESSAGE[quotaKey];
}

return msg;
```

app/scripts/services/quota.js
```diff
@@ -254,6 +263,100 @@ angular.module("openshiftConsole")
...
+
+          if(hardValue <= usedValue) {
+            var msg = getNotificaitonMessage(used, usedValue, hard, hardValue, quotaKey);
+            var containsHTML = hasHTML(msg);
+            notifications.push({
+              id: "quota-limit-reached-" + quotaKey,
+              namespace: projectName,
+              type: (hardValue < usedValue ? 'warning' : 'info'),
+              message: msg,
+              isHTML: containsHTML,
```
spadgett: `isHTML: true` with the suggested changes above.

app/scripts/services/quota.js
```diff
@@ -254,6 +263,100 @@ angular.module("openshiftConsole")
...
+                  }
+                },
+                {
+                  name: "Don't Show Me Again",
+                  title: 'Permenantly hide this notificaiton until quota limit changes',
+                  onClick: function(notification, action, index, group, drawer) {
+                    NotificationsService.permanentlyHideNotification(notification.uid, notification.namespace);
+                    EventsService.markCleared(notification.uid);
+                    group.notifications.splice(index, 1);
+                    drawer.customScope.countUnreadNotifications();
```
spadgett: I'd change these onClick handlers to `return true` if they should clear the notification and then move these two lines back to the drawer. The `onClick` handler shouldn't need to take any parameters (you have a reference to the notification already through the closure).

app/scripts/services/quota.js
```diff
@@ -254,6 +263,100 @@ angular.module("openshiftConsole")
...
+              isHTML: containsHTML,
+              skipToast: true,
+              showInDrawer: true,
+              actions: [
+                {
+                  name: 'View Quotas',
+                  title: 'View project quotas',
+                  onClick: function(notification, action, index, group, drawer) {
+                    $location.url("/project/" + $routeParams.project + "/quota");
+                    drawer.drawerHidden = true;
```
spadgett: Can we use something like

```js
$rootScope.$emit('NotificationDrawerWrapper.hide');
```

to close the drawer? We're already listening for a toggle message in the drawer, so it should not be too tough to add another. I'd rather not pass the `drawer` as a parameter.

app/scripts/services/quota.js
spadgett: Notification drawer wrapper can check the onClick return value and then clear the notification if it's true.

dtaylor113:
> The onClick handler shouldn't need to take any parameters (you have a reference to the notification already through the closure).

Hi @spadgett, sorry, you lost me here, can you please elaborate?  -thanks

app/scripts/services/quota.js
```diff
@@ -254,6 +263,100 @@ angular.module("openshiftConsole")
...
+                  }
+                },
+                {
+                  name: "Don't Show Me Again",
+                  title: 'Permenantly hide this notificaiton until quota limit changes',
+                  onClick: function(notification, action, index, group, drawer) {
+                    NotificationsService.permanentlyHideNotification(notification.uid, notification.namespace);
```
spadgett: I don't think`notification.uid` is right. Where is `uid` getting set?

It should be something like

```js
var notificationID = 'openshift/hide-quote-warning/' + projectName + '/' + quotaKey;
```

dtaylor113: `notification.uid` is being set [here](https://github.com/openshift/origin-web-console/pull/2116/files#diff-e13814006e3a1d3d1f4fa5f45b6ebc5dR179)

spadgett: OK, I see. We should be scoping to a project. So maybe just above at line 313:

```js
var notificationID = 'quota-warning/' + projectName + '/' + quotaKey';
```

Then we can use that value here and don't need a reference to the notification itself.

spadgett: By the way, I don't think the NotificationDrawerWrapper is setting a `uid` anymore.

app/scripts/services/quota.js
dtaylor113: `action.onClick` isn't being called from the notificationDrawerWrapper.js, it is being called from the included `notification-body.html`, [here](https://github.com/openshift/origin-web-console/pull/2116/files#diff-900a7b20a52c72356948916b917b0ebdR39).  
In `quota.js`, I could throw `$rootScope.$emit('NotificationDrawerWrapper.clear')` and also pass along the notification.id/uid.

spadgett: > action.onClick isn't being called from the notificationDrawerWrapper.js, it is being called from the included notification-body.html

Right, I understand. I'm proposing changing that to follow this basic pattern:

https://github.com/openshift/origin-web-console/blob/master/app/views/_alerts.html#L24

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/alerts.js#L34-L42

app/scripts/services/quota.js
dtaylor113: common's [NotificationService](https://github.com/openshift/origin-web-common/blob/master/src/ui-services/notificationsService.js#L64) already has an API for permanently hiding a notification.  It takes an id and a 'namespace', which in this case is 'projectName'.

spadgett: Sorry, I see. That's fine then.

I think the point that the notification doesn't need to be passed back to the `onClick` handlers stands, though. We already have `quotaKey` and `projectName`.

dtaylor113: > By the way, I don't think the NotificationDrawerWrapper is setting a uid anymore.

I still see it being set in Ben's PR [here](https://github.com/openshift/origin-web-console/pull/2001/files#diff-e13814006e3a1d3d1f4fa5f45b6ebc5dR180)

spadgett: You're right. I don't think that's needed in Ben's PR -- we shouldn't need 3 separate IDs -- but I'll leave well enough alone since it's in the merge queue.

dtaylor113:
Hi @spadgett, updated with latest rounds of changes.  I believe this is more inline with the separation of services.  Quotas and the NotificationDrawer now communicate via events and hand off responsibilities appropriately.

app/scripts/services/quota.js
benjaminapetersen: 3 IDs, wheeeee... 😄  
Yeah I agree, oops on my part, I prob should have dropped one after adding `trackByID`.

app/scripts/services/quota.js
```diff
@@ -254,6 +263,96 @@ angular.module("openshiftConsole")
...
+      "requests.memory",
+      "limits.cpu",
+      "limits.memory"
+    ];
+
+    var getNotificaitonMessage = function(used, usedValue, hard, hardValue, quotaKey) {
+      var msgPrefix = "Your project is " + (hardValue < usedValue ? 'over' : 'at') + " quota. ";
+      var msg;
+      if (_.includes(COMPUTE_RESOURCE_QUOTAS, quotaKey)) {
+        msg = msgPrefix + "It is using " + percent((usedValue/hardValue), 0) + " of " + usageWithUnits(hard, quotaKey) + " " + humanizeQuotaResource(quotaKey, true) + ".";
```
spadgett: Let's not pass `true` to `humanizeResourceQuota` here and just below. We typically don't capitalize resource kinds in the console in sentences, only headings.

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -94,6 +95,14 @@
         });
       };
 
+      var removeNotificationFromGroup = function(notification) {
+        _.each(drawer.notificationGroups, function(group) {
+          _.remove(group.notifications, function(grpNotification) {
+            return grpNotification.uid === notification.uid && grpNotification.namespace === notification.namespace;
+          });
```
spadgett: Nit: this could be

```js
_.remove(group.notifications, { uid: notification.uid, namespace: notification.namespace });
```

spadgett:
I noticed that the dropdown can get clipped if there's only one notification:

screen shot 2017-09-20 at 7 00 35 am

app/scripts/services/quota.js
```diff
@@ -254,6 +263,96 @@ angular.module("openshiftConsole")
...
+      } else {
+        msg = msgPrefix + "It is using " + usedValue + " of " + hardValue + " " + humanizeQuotaResource(quotaKey, true) + ".";
+      }
+
+      msg = _.escape(msg);
+
+      if (Constants.QUOTA_NOTIFICATION_MESSAGE && Constants.QUOTA_NOTIFICATION_MESSAGE[quotaKey]) {
+        msg += " " + Constants.QUOTA_NOTIFICATION_MESSAGE[quotaKey];
```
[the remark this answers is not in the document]
dtaylor113: The custom message could just be a string and not contain HTML, right?

dtaylor113:
> I noticed that the dropdown can get clipped if there's only one notification:

Hi, yes, filed a-pf issue: https://github.com/patternfly/angular-patternfly/issues/614

dtaylor113:
Hi @spadgett, addressed your latest comments.  Still have a question about escaping the custom message though.

app/scripts/services/quota.js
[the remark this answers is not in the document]
spadgett: Sorry if I wasn't clear. I'm saying to add a comment like

```js
// QUOTA_NOTICIATION_MESSAGE can contain HTML and shouldn't be escaped.
```

I don't want someone to add an `_.escape` call later thinking it was an oversight.

spadgett:
> Hi, yes, filed a-pf issue: patternfly/angular-patternfly#614

Thanks. I'm not 100% sure it's a a-pf bug since we're adding the dropdown.

cc @rhamilto

dtaylor113:
> Thanks. I'm not 100% sure it's a a-pf bug since we're adding the dropdown.

I see the clipping in our a-pf example with the notification drawer.

app/scripts/services/quota.js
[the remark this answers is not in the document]
dtaylor113: ah..ok..will do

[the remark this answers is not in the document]
dtaylor113: Added comment

app/views/directives/notifications/notification-body.html
```diff
@@ -5,6 +5,8 @@
   <a
     class="pull-right"
     href=""
+    ng-if="!notification.actions.length"
+    tabindex="0"
```
spadgett: Remove `tabindex`. It's not needed now that we have an `href`. Looks like this was accidentally added back on rebase

dtaylor113: Ok, removed `tabindex`. -thanks

spadgett:
[merge]

dtaylor113:
Filed these issues to cover remaining work:

[Quota Notifications: Resetting/Restoring "Don't Show Me Again" notifications](https://github.com/openshift/origin-web-console/issues/2122)
[Quota Notifications: Add kabob menu to Toasts](https://github.com/openshift/origin-web-console/issues/2123)
[Quota Notifications: Toast when user takes action which reaches quota](https://github.com/openshift/origin-web-console/issues/2124)

spadgett:
WebDriver flake

[merge]

app/scripts/constants.js
```diff
@@ -555,5 +555,10 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     //   href: 'http://example.com/',
     //   tooltip: 'Open Dashboard'
     // }
-  ]
+  ],
+  QUOTA_NOTIFICATION_MESSAGE: {
```
benjaminapetersen: Did we intend to leave this in as an empty object?

dtaylor113: I believe so.  Similar to the other two empty objects right before this one.

spadgett: Yeah, they're for custom messages. We generate a default message in code.

PR #2119, opened 2017-09-20
Update buildConfig controller to use getPreferredVersion

benjaminapetersen:
After doing a few of these, trying to decide if I really like the pattern:

```JavaScript 
    $scope.buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
    $scope.buildsVersion = APIService.getPreferredVersion('builds');
```
or if it would be better to do something like:

```JavaScript 
$scope.versions = {
  buildConfig: APIService.getPreferredVersion('buildconfigs'),
  builds: APIService.getPreferredVersion('builds')
}
```
Not a huge difference, but 

```html 
<li ng-if="buildConfigsVersion | canI : 'update'">
<li ng-if="versions.buildConfig | canI : 'update'">
```
might be more readable?  

@spadgett

### DIFF
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -44,6 +44,9 @@ angular.module('openshiftConsole')
       title: $routeParams.buildconfig
     });
 
+    $scope.buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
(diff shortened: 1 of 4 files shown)


spadgett:
[merge]

spadgett:
WebDriver flake

[merge]

PR #2120, opened 2017-09-20
Update builds controller to use getPreferredVersion

benjaminapetersen:
@spadgett

### DIFF
--- a/app/scripts/controllers/builds.js
+++ b/app/scripts/controllers/builds.js
@@ -8,7 +8,17 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('BuildsController', function ($routeParams, $scope, DataService, $filter, LabelFilter, Logger, $location, BuildsService, ProjectsService) {
+  .controller('BuildsController', function (
+    $filter,
+    $location,
+    $routeParams,
+    $scope,
+    APIService,
+    BuildsService,
+    DataService,
+    LabelFilter,
+    Logger,
+    ProjectsService) {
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/builds.js
```diff
@@ -8,7 +8,17 @@
...
+    $routeParams,
+    $scope,
+    APIService,
+    DataService,
+    $filter,
+    LabelFilter,
+    Logger,
+    $location,
+    BuildsService,
+    ProjectsService) {
```
spadgett: Let's sort the dependencies to keep the lists manageable

benjaminapetersen: sort done

benjaminapetersen:
Yeah, this one didn't need a view update as there are no `canI` checks in the matching view.

app/scripts/controllers/builds.js
```diff
@@ -20,6 +30,9 @@ angular.module('openshiftConsole')
 
     var buildConfigForBuild = $filter('buildConfigForBuild');
 
+    $scope.buildsVersion = APIService.getPreferredVersion('builds');
+    $scope.buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
```
spadgett: Let's not use `$scope` vars if these aren't used in the view

benjaminapetersen: sounds good.  kept that for consistency with other views, but agree, no point in blowing up `$scope`.

benjaminapetersen:
updated

spadgett:
[merge]

PR #2127, opened 2017-09-20
Changing overlay layout to a single column

rhamilto:
https://trello.com/c/6zsvyyYu
Requires https://github.com/openshift/origin-web-catalog/pull/413

### DIFF
--- a/app/views/directives/process-template-dialog/process-template-info.html
+++ b/app/views/directives/process-template-dialog/process-template-info.html
@@ -1,0 +1,20 @@
+<div class="order-service-details">
+  <div class="order-service-details-top">
+    <div class="service-icon">
+      <span class="icon {{$ctrl.iconClass}}" aria-hidden="true"></span>
+    </div>
+    <div class="service-title-area">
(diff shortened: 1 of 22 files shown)


app/views/directives/process-template-dialog/process-template-info.html
```diff
@@ -0,0 +1,20 @@
+<div class="order-service-details" ng-if="!$ctrl.selectStep.selected">
```
spadgett: I don't think the `ng-if` is needed here because you check `ng-if="step.selected"` above before including the view.

app/views/directives/process-template-dialog/process-template-info.html
```diff
@@ -0,0 +1,20 @@
...
+  <div class="order-service-details-top">
+    <div class="service-icon">
+      <span class="icon {{$ctrl.iconClass}}"></span>
```
spadgett: aria-hidden="true"

spadgett:
[merge]

PR #2133, opened 2017-09-21
Support EnvFrom in the Env Editors

cdcabrera:
https://trello.com/c/rOB7X4pj

Add dropdown support to give the ability to add bulk additions within the environment editors. [Design documentation located here](http://openshift.github.io/openshift-origin-design/web-console/3-project-details/deployment-details)

Notes:
- ~~Comment clean up still required~~
- ~~CSS/Styling~~
- ~~Some awkward behavior around validation and allowing a save when removing the last "envFrom" entry~~

### DIFF
--- a/app/scripts/directives/editEnvironmentFrom.js
+++ b/app/scripts/directives/editEnvironmentFrom.js
@@ -1,0 +1,191 @@
+"use strict";
+(function() {
+  angular.module("openshiftConsole").component('editEnvironmentFrom', {
+    controller: [
+      '$attrs',
+      '$filter',
+      'keyValueEditorUtils',
+      EditEnvironmentFrom
+    ],
+    bindings: {
+      addRowLink: '@',              // creates a link to "add row" and sets its text label
+      entries: '=',                 // an array of objects containing configmaps and secrets
+      envFromSelectorOptions: '<',  // dropdown selector options, an array of objects
+      selectorPlaceholder: '@'      // placeholder copy for dropdown selector
+    },
+    templateUrl: 'views/directives/edit-environment-from.html'
+  });
+
+  function EditEnvironmentFrom($attrs,
+                               $filter,
+                               utils) {
+    var ctrl = this;
+
+    var canI = $filter('canI');
+    var humanizeKind = $filter('humanizeKind');
+    var uniqueId = _.uniqueId();
+
+    ctrl.setFocusClass = 'edit-environment-from-set-focus-' + uniqueId;
+
+    var addEntry = function(entries, entry) {
+      entries && entries.push(entry || {});
+    };
+
+    ctrl.onAddRow = function() {
+      addEntry(ctrl.envFromEntries);
+      utils.setFocusOn('.'+ ctrl.setFocusClass);
+    };
+
+    ctrl.deleteEntry = function(start, deleteCount) {
+      if(ctrl.envFromEntries && !ctrl.envFromEntries.length) {
+        return;
+      }
+
+      ctrl.envFromEntries.splice(start, deleteCount);
+      if(!ctrl.envFromEntries.length && ctrl.addRowLink) {
+        addEntry(ctrl.envFromEntries);
+      }
+
+      ctrl.updateEntries(ctrl.envFromEntries);
+      ctrl.editEnvironmentFromForm.$setDirty();
+    };
+
+    ctrl.isEnvFromReadonly = function(entry) {
+      return ctrl.isReadonlyAny ||
+        entry.isReadonlyValue === true ||
+        ((entry.secretRef || entry.configMapRef) && !entry.selectedEnvFrom) ||
+        _.isEmpty(ctrl.envFromSelectorOptions);
+    };
+
+    ctrl.groupByKind = function(object) {
+      return humanizeKind(object.kind);
+    };
+
+    //ctrl.uniqueForValue = utils.uniqueForValue;
+    ctrl.dragControlListeners = {
+      accept: function (sourceItemHandleScope, destSortableScope) {
+        return sourceItemHandleScope.itemScope.sortableScope.$id === destSortableScope.$id;
+      },
+      orderChanged: function() {
+        ctrl.editEnvironmentFromForm.$setDirty();
+      }
+    };
+
+    ctrl.envFromObjectSelected = function(index, entry, selected) {
+      var newEnvFrom = {};
+
+      switch (selected.kind) {
+        case 'Secret':
+          newEnvFrom.secretRef = {
+            name: selected.metadata.name
+          };
+          delete ctrl.envFromEntries[index].configMapRef;
+          break;
+        case 'ConfigMap':
+          newEnvFrom.configMapRef = {
+            name: selected.metadata.name
+          };
+          delete ctrl.envFromEntries[index].secretRef;
+          break;
+      }
+
+      _.assign(ctrl.envFromEntries[index], newEnvFrom);
+      ctrl.updateEntries(ctrl.envFromEntries);
+    };
+
+    ctrl.updateEntries = function(entries) {
+      ctrl.entries = _.filter(entries, function (val) {
+        return val.secretRef || val.configMapRef;
+      });
+    };
+
+    var updateEnvFromEntries = function(entries) {
+      ctrl.envFromEntries = entries || [];
+
+      if(!ctrl.envFromEntries.length) {
+        addEntry(ctrl.envFromEntries);
+      }
+
+      _.each(ctrl.envFromEntries, function(entry) {
+        if(entry) {
+          if(entry.configMapRef && !canI('configmaps', 'get')) {
+            entry.isReadonlyValue = true;
+          }
+
+          if(entry.secretRef && !canI('secrets', 'get')) {
+            entry.isReadonlyValue = true;
+          }
+        }
+      });
+    };
+
+    var getReferenceValue = function(option) {
+      var referenceValue;
+
+      switch(option.kind) {
+        case 'ConfigMap':
+          referenceValue = _.find(ctrl.envFromEntries, {configMapRef: {name: option.metadata.name}});
+          break;
+        case 'Secret':
+          referenceValue = _.find(ctrl.envFromEntries, {secretRef: {name: option.metadata.name}});
+          break;
+      }
+
+      return referenceValue;
+    };
+
+    ctrl.checkEntries = function(option) {
+      return !!(getReferenceValue(option));
+    };
+
+    var findReferenceValueForEntries = function(entries, envFromSelectorOptions) {
+      ctrl.cannotAdd = (ctrl.isReadonlyAny || _.isEmpty(envFromSelectorOptions));
+
+      if(envFromSelectorOptions) {
+        _.each(envFromSelectorOptions, function(option) {
+          var referenceValue = getReferenceValue(option);
+
+          if (referenceValue) {
+            _.set(referenceValue, 'selectedEnvFrom', option);
+          }
+        });
+      }
+    };
+
+    ctrl.$onInit = function() {
+      updateEnvFromEntries(ctrl.entries);
+      findReferenceValueForEntries(ctrl.entries, ctrl.envFromSelectorOptions);
+
+      if('cannotDelete' in $attrs) {
+        ctrl.cannotDeleteAny = true;
+      }
+
+      if('cannotSort' in $attrs) {
+        ctrl.cannotSort = true;
+      }
+
+      if('isReadonly' in $attrs) {
+        ctrl.isReadonlyAny = true;
+      }
+
+      if('showHeader' in $attrs) {
+        ctrl.showHeader = true;
+      }
+
+      if(ctrl.envFromEntries && !ctrl.envFromEntries.length) {
+        addEntry(ctrl.envFromEntries);
+      }
+    };
+
+    ctrl.$onChanges = function(changes) {
+      if(changes.entries) {
+        updateEnvFromEntries(changes.entries.currentValue);
+      }
+
+      if(changes.envFromSelectorOptions) {
+        findReferenceValueForEntries(ctrl.envFromEntries, changes.envFromSelectorOptions.currentValue);
+      }
+    };
+  }
+})();
+
--- a/app/views/directives/edit-environment-from.html
+++ b/app/views/directives/edit-environment-from.html
@@ -1,0 +1,81 @@
+<ng-form name="$ctrl.editEnvironmentFromForm" novalidate>
+  <div
+    ng-if="$ctrl.showHeader"
+    class="environment-from-entry environment-from-editor-entry-header">
+    <div class="form-group environment-from-editor-header value-header">
+      <div class="input-group">
+        <span class="help-block">{{$ctrl.selectorPlaceholder}}</span>
+      </div>
+    </div>
+  </div>
+
+  <div ng-model="$ctrl.entries" class="environment-from-editor" as-sortable="$ctrl.dragControlListeners">
+    <div
+      class="environment-from-entry"
+      ng-class-odd="'odd'"
+      ng-class-even="'even'"
+      ng-repeat="entry in $ctrl.envFromEntries"
+      as-sortable-item>
+
+      <div class="form-group environment-from-input">
+        <div ng-if="$ctrl.isEnvFromReadonly(entry)" class="faux-input-group">
+          <div ng-if="!entry.configMapRef.name && !entry.secretRef.name">
+            No secrets or config maps have been added as Environment From.
+          </div>
+          <div ng-if="entry.configMapRef.name || entry.secretRef.name" class="faux-form-control readonly">
+            Use all keys and values from
+            <span ng-if="entry.configMapRef.name">config map {{entry.configMapRef.name}}</span>
+            <span ng-if="entry.secretRef.name">secret {{entry.secretRef.name}}</span>
+          </div>
+        </div>
+
+        <div ng-if="!$ctrl.isEnvFromReadonly(entry)">
+          <div class="ui-select">
+            <ui-select ng-model="entry.selectedEnvFrom"
+                       ng-required="entry.selectedEnvFrom"
+                       on-select="$ctrl.envFromObjectSelected($index, entry, $select.selected)"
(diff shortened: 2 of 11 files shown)


app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -0,0 +1,204 @@
...
+
+    ctrl.isReadonlySome = function(name) {
+      return _.includes(readOnlySome, name);
+    };
+
+    ctrl.cannotDeleteSome = function(name) {
+      return _.includes(cannotDeleteSome, name);
+    };
+
+    ctrl.isValueFromReadonly = function() {
```
spadgett: We should avoid `valueFrom` in names throughout this control. That refers specifically to choosing an individual key from a secret. It's going to get confusing if we use that in names for `envFrom`.

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -0,0 +1,204 @@
...
+      '$scope',
+      'keyValueEditorUtils',
+      EditEnvironmentFrom
+    ],
+    controllerAs: 'ctrl',
+    //require: '^form',
+    bindings: {
+      addRowLink: '@',
+      entries: '=',
+      valueFromSelectorOptions: '<',
```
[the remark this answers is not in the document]
cdcabrera: np, can change. Just reused the ```valueFromSelectorOptions``` since it helps generate th e same list... can change the name no problem

cdcabrera:
@openshift/team-ux-review need some input on the tooltip copy for the Environment From section...

[The original tooltip copy that shows up in the designs...](http://openshift.github.io/openshift-origin-design/web-console/3-project-details/deployment-details#environment-from)

> All keys from these config maps and secrets will be added as environment variables.

And a possible alternative...

> Environment From lets you add all key value pairs from a config map or secret.

cdcabrera:
@spadgett looks like updating the Environment Service requires an update to the associated unit test, or some additional checks within the service.

I could update the test to avoid this failure by placing an empty array for ```envFrom: []``` within the expected results?

spadgett:
> I could update the test to avoid this failure by placing an empty array for envFrom: [] within the expected results?

This is fine as long as we handle the case where `envFrom` is missing from the resource we pass in. It's an optional field.

cdcabrera:
@spadgett verified the "env" vs "envFrom" aspects of the unit tests. Behavior appears to be the same with the addition of an empty array vs the missing resource. Going to push the updated unit test change.

app/views/directives/edit-environment-from.html
```diff
@@ -0,0 +1,94 @@
+<ng-form name="forms.editEnvironmentFrom" novalidate>
+  <div
+    ng-if="$ctrl.showHeader"
+    class="key-value-editor-entry key-value-editor-entry-header">
+    <div class="form-group key-value-editor-header value-header">
+      <div class="input-group">
```
spadgett: Why wrap in a `form-group` and `input-group`?

app/views/directives/edit-environment-from.html
```diff
@@ -0,0 +1,94 @@
+<ng-form name="forms.editEnvironmentFrom" novalidate>
```
spadgett: Should this be `$ctrl.editEnvironmentFrom` here and below?

app/views/directives/edit-environment-from.html
```diff
@@ -0,0 +1,94 @@
...
+      as-sortable-item>
+
+      <div
+        class="form-group key-value-editor-input"
+        ng-class="{ 'has-error': (forms.editEnvironmentFrom[uniqueForValue(unique, $index)].$invalid && forms.editEnvironmentFrom[uniqueForValue(unique, $index)].$touched) }">
+        <div ng-if="$ctrl.isEnvFromReadonly(entry)" class="faux-input-group">
+          <div class="faux-form-control readonly">
+            Set to values in {{entry.selectedEnvFrom.kind | humanizeKind : true | lowercase}}
+            <span
+              ng-if="!('configmaps' | canI : 'get') || !('secrets' | canI : 'get')">
```
spadgett: This doesn't look right. If the value is a config map, we should be able to link to that even if I can't view secrets.

app/views/directives/edit-environment-from.html
```diff
@@ -0,0 +1,94 @@
...
+        </div>
+      </div>
+    </div>
+
+    <div class="key-value-editor-entry form-group" ng-if="(!$ctrl.cannotAdd)">
+      <a
+        href=""
+        class="add-row-link"
+        role="button"
+        aria-label="Add row"
```
spadgett: Probably don't want `aria-label` here so that the screen reader uses `addRowLink`

app/views/directives/edit-environment-variables.html
```diff
@@ -32,6 +34,21 @@
       add-row-with-selectors-link="Add Environment Variable Using a Config Map or Secret"
       show-header>
     </key-value-editor>
+
+    <h4>
+      Environment From
+      <span class="pficon pficon-info"
+            aria-hidden="true"
+            data-toggle="tooltip"
+            data-original-title="Environment From lets you add all key value pairs from a config map or secret."></span>
```
spadgett: "Environment From lets you add all key-value pairs from a config map or secret as environment variables."

spadgett:
We might rethink the "View Secret" link since it navigates away from the page, losing your edits.

cc @beanh66

spadgett:
I noticed the highlighted item is not very legible:

screen shot 2017-09-26 at 9 23 12 am

app/views/directives/edit-environment-variables.html
```diff
@@ -32,6 +34,21 @@
       add-row-with-selectors-link="Add Environment Variable Using a Config Map or Secret"
       show-header>
     </key-value-editor>
+
+    <h4>
+      Environment From
+      <span class="pficon pficon-info"
```
spadgett: We usually use a `?`(`pficon-help`) for these tooltips, for instance:

openshift web console 2017-09-26 09-25-53

@beanh66 Any objection to switching for consistency?

spadgett:
These labels should be updated as well in the design:

openshift web console 2017-09-26 09-30-50

spadgett:
Mobile styles are off:

openshift web console 2017-09-26 09-39-33

spadgett:
Need to fix experience when logged in as a project viewer. You can add a user with `view` role using the Resources -> Membership page.

openshift web console 2017-09-26 10-10-03

app/views/directives/edit-environment-from.html
cdcabrera: I can remove sure thing, just heads up... it's copied from the variables display

app/views/directives/edit-environment-variables.html
cdcabrera: Copy update, np

app/views/directives/edit-environment-from.html
spadgett: Let's fix it there, too.

app/views/directives/edit-environment-from.html
cdcabrera: Is that a vote to remove the ```ng-form```? 

I may be able to, but was under the impression we were trying to maintain the parent-child relationship with the parent form?

app/views/directives/edit-environment-from.html
cdcabrera: Part of the classes copied over from the variables component (heads up there's may be some clean up there too), also part of the CSS clean up we've been going through, lots of extras to remove.

app/views/directives/edit-environment-from.html
cdcabrera: Can adjust the logic, look at separating it out

app/views/directives/edit-environment-from.html
spadgett: You should be able to keep `ng-form` and use `$ctrl.editEnvironmentForm` I think. Is this sharing the same form object as the parent? I don't think that should be necessary.

app/views/directives/edit-environment-from.html
spadgett: OK to leave it's reusing the styles and needed

beanh66:
@spadgett agreed on the view secret link! I talked to @cdcabrera about having the link be called "View Details" or "View Variables" instead and clicking the link would pop a modal to display the secret or config map variables (rather than navigating away). I also mentioned that if we go this route, the link should be available right away after a selection is made. Previously we were only showing that link after the save button was clicked. Thoughts?

benjaminapetersen:
Seeing some premature collapse of the "x" to a new column @ tablet-ish size:

screen shot 2017-09-26 at 4 38 43 pm

app/views/directives/edit-environment-from.html
```diff
@@ -0,0 +1,94 @@
...
+    <div
+      class="key-value-editor-entry"
+      ng-class-odd="'odd'"
+      ng-class-even="'even'"
+      ng-repeat="entry in $ctrl.envFromEntries"
+      as-sortable-item>
+
+      <div
+        class="form-group key-value-editor-input"
+        ng-class="{ 'has-error': (forms.editEnvironmentFrom[uniqueForValue(unique, $index)].$invalid && forms.editEnvironmentFrom[uniqueForValue(unique, $index)].$touched) }">
```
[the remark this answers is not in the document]
benjaminapetersen: Also in `uniqueForValue(unique, $index)`, `unique` is `$scope.unique`, which is also missing here (would be `ctrl.unique`). 
https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/keyValueEditor.js#L202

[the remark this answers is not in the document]
benjaminapetersen: Btw, the `uniqueFor` helpers are to generate uniqueIds for the inputs, and match up the labels, `has-error` classes, etc. 

https://github.com/openshift/origin-web-console/blob/master/app/views/directives/key-value-editor.html#L37

The `ctrl.unique` is used in `key-value-editor` to ensure that you can have more than one editor on a page w/o collisions.  Basically `forms.editEnvironmentFrom[uniqueForValue(unique, $index)]` returns the matching `id` of one of the `<input id="???">` elements.

app/views/directives/edit-environment-from.html
```diff
@@ -0,0 +1,94 @@
...
+            class="pficon pficon-close delete-row as-sortable-item-delete"
+            role="button"
+            aria-label="Delete row"
+            ng-hide="$ctrl.cannotDeleteAny"
+            ng-click="$ctrl.deleteEntry($index, 1)"></a>
+          <a
+            ng-href="{{entry.selectedEnvFrom | navigateResourceURL}}"
+            class="pficon"
+            ng-show="entry.selectedEnvFrom"
+            ng-click="$ctrl.viewDetail(entry)">View {{entry.selectedEnvFrom.kind | humanizeKind : true}}</a>
```
benjaminapetersen: I think `$ctrl.viewDetail` doesn't exist?  the href works, tho.

app/views/directives/edit-environment-variables.html
```diff
@@ -32,6 +34,21 @@
...
+            aria-hidden="true"
+            data-toggle="tooltip"
+            data-original-title="Environment From lets you add all key-value pairs from a config map or secret as environment variables."></span>
+    </h4>
+    <edit-environment-from
+      entries="container.envFrom"
+      selector-placeholder="Secret/Config Map"
+      env-from-selector-options="$ctrl.valueFromObjects"
```
benjaminapetersen: See above about `env-from-selector-options`.

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -0,0 +1,192 @@
...
+  angular.module("openshiftConsole").component('editEnvironmentFrom', {
+    controller: [
+      '$attrs',
+      '$filter',
+      'keyValueEditorUtils',
+      EditEnvironmentFrom
+    ],
+    bindings: {
+      addRowLink: '@',
+      entries: '=',
```
benjaminapetersen: I'm not sure I'm crazy about `envFromSelectorOptions`...I'd at least like a comment about what it means.  Its config for something, but at a glance I'm not sure what (might be my ignorance re: EnvironmentFrom).

app/views/directives/edit-environment-from.html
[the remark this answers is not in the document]
cdcabrera: Agreed! Removing the forms reference from the earlier review and the lack of the uniqueForValue method seemed to build, looks like an adjustment for the readonly section needs to take place

app/views/directives/edit-environment-from.html
cdcabrera: Yeahhhhhhhh =), removing appears superfluous looks like it was from a previous iteration

app/scripts/directives/editEnvironmentFrom.js
cdcabrera: An annotation couldn't hurt, we'll add one in

cdcabrera:
@spadgett @benjaminapetersen there's still a missing method around ```edit-environment-from.html``` and ```uniqueForValue``` to help generate unique select Ids...

- the Read Only, in general, logic probably needs additional adjustment... 
- and the "has-error" class, which feels like it should be removed from the Read Only section
```js
ng-class="{ 'has-error': ($ctrl.editEnvironmentFromForm[uniqueForValue(unique, $index)].$invalid && $ctrl.editEnvironmentFromForm[uniqueForValue(unique, $index)].$touched) }">
```

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -0,0 +1,208 @@
...
+
+  function EditEnvironmentFrom($attrs,
+                               $filter,
+                               $routeParams,
+                               utils) {
+    var ctrl = this;
+
+    var canI = $filter('canI');
+    var humanizeKind = $filter('humanizeKind');
+    var namespace = $routeParams.project;
```
spadgett: Better to pass the project in than read it from route params. But why is it needed?

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -0,0 +1,208 @@
...
+          var entryName = _.get(entry.configMapRef, 'name') || _.get(entry.secretRef, 'name');
+
+          _.set(entry, 'apiObj', {
+            kind: entryKind,
+            metadata: {
+              name: entryName,
+              namespace: namespace
+            }
+          });
+        });
```
spadgett: I'd rather not set dummy values. Is there another way to handle this? Is this for handling secrets that have been deleted?

app/views/directives/edit-environment-from.html
```diff
@@ -0,0 +1,91 @@
...
+          as-sortable-item-handle></span>
+        <a
+          href=""
+          class="pficon pficon-close delete-row as-sortable-item-delete"
+          role="button"
+          aria-label="Delete row"
+          ng-hide="$ctrl.cannotDeleteAny"
+          ng-click="$ctrl.deleteEntry($index, 1)"></a>
+        <a
+          ng-if="!$ctrl.editEnvironmentFromForm.$dirty"
```
spadgett: We can't just check this form because the normal env var form above might have changed.

I suggest we handle "View Secret" as a follow on issue and remove it for now.

app/views/directives/edit-environment-variables.html
```diff
@@ -32,6 +34,21 @@
       add-row-with-selectors-link="Add Environment Variable Using a Config Map or Secret"
       show-header>
```
spadgett: Just above this line:

```html
add-row-link="Add Variable"
add-row-with-selectors-link="Add Variable from Config Map or Secret"
```

spadgett:
Let's fix the readonly state

openshift web console 2017-09-27 13-16-39

beanh66:
@spadgett Do we want to change the Environment From title?  @cdcabrera the field level help next to Environment From also doesn't appear to be using the PatternFly styling here, can you take a look? I liked your suggestion for the wording change regarding the tooltip though.

spadgett:
> @cdcabrera the field level help next to Environment From also doesn't appear to be using the PatternFly styling here, can you take a look?

@beanh66 It's using the style we use for help tooltips throughout the console. I'm not opposed to changing, but I'd like to change them everywhere to be consistent (and probably out of scope for this PR).

That said, the icon is larger here than it is in other places. For instance, this is the build config editor with a similar help tooltip:

screen shot 2017-09-27 at 2 09 59 pm

app/scripts/directives/editEnvironmentFrom.js
cdcabrera: For the comment chain, related to the below dummy values referenced in ```editEnvironmentFrom.js```

app/scripts/directives/editEnvironmentFrom.js
cdcabrera: For the comment chain, we're removing this as discussed

spadgett:
openshift web console 2017-09-27 14-44-34

app/views/directives/edit-environment-from.html
```diff
@@ -0,0 +1,81 @@
...
+
+      <div
+        class="form-group environment-from-input"
+        ng-class="{ 'has-error': ($ctrl.editEnvironmentFromForm[uniqueForValue(unique, $index)].$invalid && $ctrl.editEnvironmentFromForm[uniqueForValue(unique, $index)].$touched) }">
+        <div ng-if="$ctrl.isEnvFromReadonly(entry)" class="faux-input-group">
+          <div ng-if="!entry.configMapRef.name && !entry.secretRef.name">
+              No secrets or config maps have been added as Environment From.
+          </div>
+          <div ng-if="entry.configMapRef.name || entry.secretRef.name" class="faux-form-control readonly">
+              Set to values in {{entry.configMapRef.name || entry.secretRef.name}}
```
spadgett: Probably should say `Use all keys and values from <kind> <name>.`

spadgett:
```html
Use all keys and values from
<span ng-if="entry.configMapRef.name">config map {{entry.configMapRef.name}}</span>
<span ng-if="entry.secretRef.name">secret {{entry.secretRef.name}}</span>
```

spadgett:
[merge]

spadgett:
@cdcabrera 

spadgett:
[merge]

PR #2134, opened 2017-09-21
Set ui-select-dropdown z-index greater than navbar-fixed so that it overlays

sg00dwin:
Fixes bug https://bugzilla.redhat.com/show_bug.cgi?id=1479114

screen shot 2017-09-21 at 11 16 06 am


spadgett:
[merge]

PR #2135, opened 2017-09-21
Copy login command shouldn't be wrapped in quotes

### LINKED ISSUE #2131: Copy Login Command includes single quotes in copied command
Login go the web console
And click on your name -> 'Copy Login Command'

The copied command is surrounded by single quotes as in 
'oc login https://blah.blah'

I would expect it to not be surrounded by quotes so I can paste it directly in my shell

spadgett:
* Fix the copy login command menu item so that it doesn't include single quotes around the copied command
* HTML-escape special characters in login command
* Use NotificationsService.permanentlyHideNotification to simply copyLoginToClipboard directive

Fixes #2131

### DIFF
--- a/app/scripts/directives/util.js
+++ b/app/scripts/directives/util.js
@@ -103,12 +103,12 @@ angular.module('openshiftConsole')
       }
     };
   })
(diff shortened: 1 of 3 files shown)


spadgett:
[merge]

PR #2136, opened 2017-09-21
Show application kind when add to application dropdown is open

spadgett:
On the secret page, when clicking "Add to Application", show the kind
when the "Application" dropdown is open. Currently the kind is only
displayed after you pick an item.

Before:

openshift web console 2017-09-21 14-17-57

After:

openshift web console 2017-09-21 14-14-42

### DIFF
--- a/app/scripts/directives/addConfigToApplication.js
+++ b/app/scripts/directives/addConfigToApplication.js
@@ -24,6 +24,7 @@
 
   function AddConfigToApplication($filter, $scope, APIService, ApplicationsService, DataService, Navigate, NotificationsService, StorageService) {
     var ctrl = this;
+    var humanizeKind = $filter('humanizeKind');
 
     var getApplications = function() {
       var context = {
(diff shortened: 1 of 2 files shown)


spadgett:
@jeff-phillips-18 It looks like my fix in https://github.com/openshift/origin-web-common/pull/189 was bad and some RCs are showing up that shouldn't. Debugging that now.

spadgett:
> It looks like my fix in openshift/origin-web-common#189 was bad and some RCs are showing up that shouldn't. Debugging that now.

Fixed by https://github.com/openshift/origin-web-common/pull/197

spadgett:
@jeff-phillips-18 Actually I might switch this to use ui-select groups like we do for the env value from picker

spadgett:
OK changed to group by kind like other selects:

openshift web console 2017-09-22 13-37-38

spadgett:
@jeff-phillips-18 another look?

spadgett:
[merge]

PR #2140, opened 2017-09-22
List serviceclasses as a cluster-scoped resource

spadgett:
Do not pass a project context when listing serviceclasses since they are not scoped to a namespace.


spadgett:
@jeff-phillips-18 PTAL, I'm surprised this was working.

spadgett:
[merge]

PR #2146, opened 2017-09-22
Fix deployment empty state message when multiple ICT

spadgett:
Show the correct empty state message on the overview when a deployment config has multiple image change triggers.

openshift web console 2017-09-22 15-04-36


spadgett:
[merge]

PR #2150, opened 2017-09-22
Delete bindings when deleting a service instance

### LINKED ISSUE #2149: Warn when deleting a service instance that still has bindings
Right now we give no warning and show no message on the overview when a provisioned service is deleted when it has bindings. Deprovision will never complete until the bindings are also deleted. Also the dialog styles are not great and don't look like other dialogs.

cc @ncameronbritt

jeff-phillips-18:
Fixes #2149

### DIFF
--- a/app/scripts/services/serviceInstances.js
+++ b/app/scripts/services/serviceInstances.js
@@ -2,65 +2,117 @@
 
 angular.module("openshiftConsole")
   .factory("ServiceInstancesService", function($filter,
+                                               $q,
                                                $uibModal,
+                                               APIService,
+                                               BindingService,
+                                               CatalogService,
                                                DataService,
+                                               Logger,
                                                NotificationsService) {
+    var getBindingsIfNecessary = function(apiObject, bindings) {
+      if (angular.isDefined(bindings)) {
+        return $q.when(bindings);
+      }
 
-    function deprovision(apiObject) {
-      var getErrorDetails = $filter('getErrorDetails');
-      var modalScope = {
-        alerts: {
-          deprovision: {
-            type: 'error',
-            message: 'Service \'' + apiObject.spec.serviceClassName + '\' will be deleted and no longer available.'
-          }
-        },
-        detailsMarkup: 'Delete Service?',
-        okButtonText: 'Delete',
-        okButtonClass: 'btn-danger',
-        cancelButtonText: 'Cancel'
+      var context = {namespace: apiObject.metadata.namespace};
+      var resource = APIService.getPreferredVersion('serviceinstancecredentials');
+
+      return DataService.list(resource, context).then(function(serviceBindings) {
+        bindings = serviceBindings.by('metadata.name');
+        return BindingService.getBindingsForResource(bindings, apiObject);
+      });
+    };
+
+    var deprovisionInstance = function(apiObject) {
+      var context = {namespace: apiObject.metadata.namespace};
+      var resource = APIService.getPreferredVersion('serviceinstances');
+
+      NotificationsService.hideNotification("deprovision-service-error");
+
+      // TODO - remove once this is resolved https://github.com/kubernetes-incubator/service-catalog/issues/942
+      var opts = {
+        propagationPolicy: null
       };
 
-      // TODO: we probably have to handle bindings in here.
-      // either:
-      // - automatically remove the bindings
-      // - tell the user they must manually unbind before continue
-      return $uibModal.open({
+      return DataService.delete(resource, apiObject.metadata.name, context, opts).then(function() {
+        NotificationsService.addNotification({
+          type: "success",
+          message: "Provisioned service '" + apiObject.metadata.name + "' was marked for deletion."
+        });
+      }, function(err) {
+        NotificationsService.addNotification({
+          id: "deprovision-service-error",
+          type: "error",
+          message: "An error occurred while deleting provisioned service " + apiObject.metadata.name + ".",
+          details: $filter('getErrorDetails')(err)
+        });
+        Logger("An error occurred while deleting provisioned service " + apiObject.metadata.name + ".", err);
+      });
+    };
+
+    var deprovisionBindings = function(apiObject, bindings) {
+      if (!CatalogService.SERVICE_CATALOG_ENABLED) {
+        return;
+      }
+
+      var context = {namespace: apiObject.metadata.namespace};
+      var resource = APIService.getPreferredVersion('serviceinstancecredentials');
+      getBindingsIfNecessary(apiObject, bindings).then(function(serviceBindings) {
+        _.each(serviceBindings, function (binding) {
+          if (!binding.metadata.deletionTimestamp) {
+            return;
+          }
+
+          DataService.delete(resource, binding.metadata.name, context)
+            .then(function () {
+              NotificationsService.addNotification({
+                type: "success",
+                message: 'Binding ' + binding.metadata.name + "' was marked for deletion."
+              });
+            })
+            .catch(function (err) {
+              NotificationsService.addNotification({
+                type: "error",
+                message: 'Binding ' + binding.metadata.name + "' could not be deleted.",
+                details: $filter('getErrorDetails')(err)
+              });
+              Logger.error('Binding ' + binding.metadata.name + "' could not be deleted.", err);
+            });
+        });
+      });
+    };
+
+    var deprovision = function (apiObject, bindings) {
+      var modalInstance;
+
+      var modalScope = {
+        kind: apiObject.kind,
+        displayName: apiObject.metadata.name,
+        okButtonText: 'Delete',
+        okButtonClass: 'btn-danger',
+        cancelButtonText: 'Cancel',
+        delete: function() {
+          modalInstance.close('delete');
+        }
+      };
+
+      modalInstance =  $uibModal.open({
         animation: true,
-        templateUrl: 'views/modals/confirm.html',
+        templateUrl: 'views/modals/delete-resource.html',
         controller: 'ConfirmModalController',
         resolve: {
           modalConfig: function() {
             return modalScope;
           }
         }
-      })
-      .result.then(function() {
-        NotificationsService.hideNotification("deprovision-service-error");
-
-        return DataService.delete({
-            group: 'servicecatalog.k8s.io',
-            resource: 'serviceinstances'
-          },
-          apiObject.metadata.name,
-          { namespace: apiObject.metadata.namespace },
-          { propagationPolicy: null
-          }) // TODO - remove once this is resolved https://github.com/kubernetes-incubator/service-catalog/issues/942
-          .then(function() {
-            NotificationsService.addNotification({
-              type: "success",
-              message: "Successfully deleted provisioned service " + apiObject.metadata.name + "."
-            });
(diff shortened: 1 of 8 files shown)


jeff-phillips-18:
image

image

app/scripts/services/serviceInstances.js
```diff
@@ -3,64 +3,105 @@
...
+
+    var deprovision = function (apiObject) {
+      var context = {namespace: apiObject.metadata.namespace};
+      var resource = {
+        group: 'servicecatalog.k8s.io',
+        resource: 'serviceinstancecredentials'
+      };
+
+      if (CatalogService.SERVICE_CATALOG_ENABLED) {
+        DataService.list(resource, context, function (bindings) {
```
spadgett: I believe everywhere we call this, we already have a list of the bindings. I'd rather just pass those in if we have them.

app/scripts/services/serviceInstances.js
```diff
@@ -3,64 +3,105 @@
...
+              title: "Unable to Deprovision",
+              strongMessage: "This service cannot be deprovisioned because it has bindings.",
+              message: "You must delete all bindings from the service before you can deprovision it."
+            };
+          }
+        }
+      });
+    };
+
+    var deprovision = function (apiObject) {
```
spadgett: I realize this isn't new, but I'd rather call the parameter `serviceInstance` because the function is very specific to instances. `apiObject` sounds like any object.

app/views/modals/warning.html
```diff
@@ -0,0 +1,14 @@
+<div class="modal-resource-action">
+  <div class="modal-body">
+    <h1>
+      <span class="pficon pficon-warning-triangle-o"></span>
```
spadgett:
```html
<span class="sr-only">warning</span>
```

app/scripts/services/serviceInstances.js
```diff
@@ -3,64 +3,105 @@
...
+        var context = {namespace: apiObject.metadata.namespace};
+        var resource = {
+          group: 'servicecatalog.k8s.io',
+          resource: 'serviceinstances'
+        };
+
+        // TODO - remove once this is resolved https://github.com/kubernetes-incubator/service-catalog/issues/942
+        var opts = {
+          propagationPolicy: null
+        };
```
spadgett: Not related to this PR, but we should double-check that this isn't needed for bindings as well. Right now we're using `delete-link` for bindings and in some cases and not passing `propogationPolicy: null`

app/scripts/services/serviceInstances.js
```diff
@@ -3,64 +3,105 @@
...
+
+    var showUnableToDeprovisionDialog = function() {
+      return $uibModal.open({
+        animation: true,
+        templateUrl: 'views/modals/warning.html',
+        controller: 'WarningModalController',
+        resolve: {
+          modalConfig: function() {
+            return {
+              title: "Unable to Deprovision",
```
spadgett: We call the action "Delete" in the UI, so I'd be consistent in the messages we use (in the dialog and notifications).

ncameronbritt:
The original idea was to block the user from deleting until the bindings had been deleted, but I'm not sure we considered the possibility that deleting the bindings could take some time. 

I think it would be a better experience to allow users to delete the bindings from that dialog, rather than just telling them to go do it. The bindings would be then marked for deletion and the deprovision could go ahead as @spadgett suggested. Maybe something like this?
screen shot 2017-09-23 at 10 10 15 am

Related but probably outside of this scope, should we even be showing things that are marked for deletion in the UI? I know it doesn't work this way in practice, but deleting seems like an action that should be able to happen more or less instantaneously. Do users need to care about things they deleted but which are just hanging around waiting for the system to delete them?

ncameronbritt:
That "delete all bindings" checkbox should probably say "delete all
bindings for $service"

app/scripts/services/serviceInstances.js
spadgett: Disregard, looks like we are handling it: https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/deleteLink.js#L153

jeff-phillips-18:
@ncameronbritt How about the title?

jeff-phillips-18:
@ncameronbritt @spadgett  How about something like:

image

ncameronbritt:
@spadgett and I were just talking about this. I think the best experience is to simply delete the bindings for the user. Deleting the service will also delete any deployments created by the service, and there's no reason for the user to think that bindings needed to be treated in a special way, and a binding doesn't have any value without the service.

Maybe it makes sense to add a sentence like we have [here](https://github.com/openshift/openshift-origin-design/blob/master/web-console/3-project-details/binding-in-project.md#delete-service): "$service and its data will no longer be available to your application."

jeff-phillips-18:
Updated to automatically delete bindings when deleting a service instance (updated PR title as well).
Confirmation Dialog:
image

ncameronbritt:
Works for me. I would probably change "It cannot be undone" to "This cannot be undone." And in the sentence before it should be "*your* applications".

app/scripts/services/serviceInstances.js
jeff-phillips-18: The service instance page does not have the bindings. I can take them optionally and only retrieve if not given.

spadgett: > The service instance page does not have the bindings.

Hey @jeff-phillips-18 I'm OK with it in that case.

I want to look at the instance page anyway later. It looks like it's requesting service classes twice, and that's potentially a lot of data.

In general, I think it's better for the page controllers to get the data and pass them down to the components to avoid requesting things more than once.

jeff-phillips-18: I can take a look at that as part of this. Basically its the ResourceServiceBindings that gets the bindings for the service instance page (an all the 'application' pages. I can update that logic as well if you'd like.

spadgett: It might be better to wait for #2155 since I have some changes to instance page and resource service bindings component in flight :/

jeff-phillips-18:
>  I would probably change "It cannot be undone" to "This cannot be undone." 

This is in the delete utility code (been there for a while). I can change it, but it would change for most deletion confirmation dialogs. @ncameronbritt Please advise.

app/scripts/services/serviceInstances.js
jeff-phillips-18: Pushed changes to only get bindings when not passed. Holding off on updating all the controllers using the ResourceServiceBindings.

jeff-phillips-18:
@spadgett Also added checks to not show the "no bindable services" message when the object is either a service instance or it is marked for deletion. Removed the "Create Binding" link when the object is marked for deletion.

app/scripts/services/serviceInstances.js
```diff
@@ -2,65 +2,129 @@
...
+        };
+
+        DataService.list(resource, context, function (serviceBindings) {
+          bindings = serviceBindings.by('metadata.name');
+          bindings = BindingService.getBindingsForResource(bindings, apiObject);
+          deferred.resolve(bindings);
+        });
+      }
+
+      return deferred.promise;
```
spadgett: You can just return the promise from `DataService.list` instead of creating your own deferred.

```js
    return DataService.list(resource, context, function (serviceBindings) {
      bindings = serviceBindings.by('metadata.name');
      return BindingService.getBindingsForResource(bindings, apiObject);
    });
```

app/scripts/services/serviceInstances.js
```diff
@@ -2,65 +2,129 @@
...
+
+      var context = {namespace: apiObject.metadata.namespace};
+      var resource = {
+        group: 'servicecatalog.k8s.io',
+        resource: 'serviceinstancecredentials'
+      };
+
+      getBindingsIfNecessary(apiObject, bindings).then(function(serviceBindings) {
+        _.each(serviceBindings, function (binding) {
+          if (!binding.metadata.deletionTimestamp) {
```
spadgett: This is a style nit, but I personally find early returns easier to reason about.

```js
if (binding.metadata.deletionTimestamp) {
  // The binding is marked for deletion.
  return;
}

// <code to delete the binding>
```

ncameronbritt:
@jeff-phillips-18 I don't feel that strongly about "this" vs. "it", so if it's going to change a bunch of other stuff, I'd say leave it as is for now, and we can address the language change taking into account all of the different contexts in which it occurs later, if we feel like we need to.

app/scripts/services/serviceInstances.js
jeff-phillips-18: @spadgett What comes back from the promise is not the bindings manipulated by the callback function. Maybe I'm doing something else wrong?

app/scripts/services/serviceInstances.js
```diff
@@ -2,65 +2,129 @@
...
-            type: 'error',
-            message: 'Service \'' + apiObject.spec.serviceClassName + '\' will be deleted and no longer available.'
+      if (angular.isDefined(bindings)) {
+        return $q.when(bindings);
+      } else {
+        var context = {namespace: apiObject.metadata.namespace};
+        var resource = {
+          group: 'servicecatalog.k8s.io',
+          resource: 'serviceinstancecredentials'
+        };
```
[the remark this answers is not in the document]
jeff-phillips-18: fixed

app/scripts/services/serviceInstances.js
```diff
@@ -2,65 +2,129 @@
...
+
+      return deferred.promise;
+    };
+
+    var deprovisionInstance = function(apiObject) {
+      var context = {namespace: apiObject.metadata.namespace};
+      var resource = {
+        group: 'servicecatalog.k8s.io',
+        resource: 'serviceinstances'
+      };
```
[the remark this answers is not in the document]
jeff-phillips-18: fixed

app/scripts/services/serviceInstances.js
```diff
@@ -2,65 +2,129 @@
...
+    var deprovisionBindings = function(apiObject, bindings) {
+      if (!CatalogService.SERVICE_CATALOG_ENABLED) {
+        return;
+      }
+
+      var context = {namespace: apiObject.metadata.namespace};
+      var resource = {
+        group: 'servicecatalog.k8s.io',
+        resource: 'serviceinstancecredentials'
+      };
```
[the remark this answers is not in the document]
jeff-phillips-18: fixed

app/scripts/services/serviceInstances.js
jeff-phillips-18: fixed

app/scripts/services/serviceInstances.js
spadgett: Sorry my JS was wrong:

```js
   return DataService.list(resource, context).then(function(serviceBindings) {
      bindings = serviceBindings.by('metadata.name');
      return BindingService.getBindingsForResource(bindings, apiObject);
    });
```

jeff-phillips-18: Ah yes, that makes sense. Sorry, I should have seen that.

jeff-phillips-18: Fixed. Thanks @spadgett

spadgett:
[merge]

PR #2152, opened 2017-09-23
Don't show parameters step when bind has no parameters

### LINKED ISSUE #2151: Parameters step visible in Create Binding dialog when no parameters
openshift web console 2017-09-23 10-06-16

spadgett:
* Remove unnecessary `pf-wizard-no-back` class since it's handled by the wizard component itself now
* Don't hide button in `process-template-dialog` since there's now an extra info step

Fixes #2151


spadgett:
WebDriver flake

[merge]

spadgett:
flake #1684

PR #2154, opened 2017-09-25
Fix binding ready logic

spadgett:
The binding ready logic was flipped in the service-binding component, so the View Secret link was showing up at the wrong time.

openshift web console 2017-09-25 07-51-38

### DIFF
--- a/app/views/directives/_service-binding.html
+++ b/app/views/directives/_service-binding.html
@@ -46,7 +46,7 @@
       project-name="{{$ctrl.binding.metadata.namespace}}"
       stay-on-current-page="true">
     </delete-link>
-    <a ng-if="('secrets' | canI : 'get') && !($ctrl.binding | isBindingFailed) && !(binding | isBindingReady)"
+    <a ng-if="('secrets' | canI : 'get') && ($ctrl.binding | isBindingReady)"
        ng-href="{{$ctrl.binding.spec.secretName | navigateResourceURL : 'Secret' : $ctrl.namespace}}">
       View Secret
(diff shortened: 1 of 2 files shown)


app/views/directives/_service-binding.html
```diff
@@ -46,7 +46,7 @@
       project-name="{{$ctrl.binding.metadata.namespace}}"
       stay-on-current-page="true">
     </delete-link>
-    <a ng-if="('secrets' | canI : 'get') && !($ctrl.binding | isBindingFailed) && !(binding | isBindingReady)"
+    <a ng-if="('secrets' | canI : 'get') && ($ctrl.binding | isBindingReady)"
```
spadgett: @jeff-phillips-18 The failed check isn't necessary because the binding is never ready when the `Failed` condition is true. (Checked with @pmorie on this a few days ago.)

spadgett: Note that the logic before was `!(binding | isBindingReady)`

jeff-phillips-18: Thanks @spadgett

spadgett:
[merge]

spadgett:
@ncameronbritt Is there any reason we don't just make the secret name itself a link instead of having a separate "View Secret" link? I think this might be a hold-over from designs when we had pod presets.

ncameronbritt:
That's a holdover from pod presets. Not sure there's any reason not to have
the secret name be the link, and remove the "view secret" action as long as
pod presets are out.

PR #2156, opened 2017-09-25
Update create controller to use getPreferredVersion

benjaminapetersen:
No view changes needed.

@spadgett


spadgett:
@benjaminapetersen I suggest just skipping this page since it's now hidden by default and will be removed later on.

benjaminapetersen:
Works for me, I can close this out.

PR #2158, opened 2017-09-25
Update createConfigMap controller to use getPreferredVersion

benjaminapetersen:
another w/o matching view updates

@spadgett


spadgett:
[merge]

PR #2159, opened 2017-09-25
Update CreateFromUrl controller to use getPreferredVersion

benjaminapetersen:
no `canI` in view

@spadgett


spadgett:
@benjaminapetersen Suggest we defer this one since we don't test these often in day-to-day development. I know we have automated tests, but I'd rather leave it for 3.8 just in case.

benjaminapetersen:
@spadgett works for me.

benjaminapetersen:
Updated.
Watching the build on this one as we have a good # of tests covering.

PR #2162, opened 2017-09-25
Update deployment controller to get getPreferredVersion

benjaminapetersen:
DEPENDS on [web-common PR 201](https://github.com/openshift/origin-web-common/pull/201) to fix `autoscaling` -> `horizontalpodautoscalers`. and probably [web-common 202](https://github.com/openshift/origin-web-common/pull/202) to bump the version.

@spadgett

### DIFF
--- a/app/scripts/controllers/deployment.js
+++ b/app/scripts/controllers/deployment.js
@@ -11,6 +11,7 @@ angular.module('openshiftConsole')
               function ($scope,
                         $filter,
                         $routeParams,
+                        APIService,
                         DataService,
                         DeploymentsService,
                         HPAService,
@@ -42,10 +43,18 @@ angular.module('openshiftConsole')
         title: $routeParams.deployment
       }
     ];
+    var buildsVersion = APIService.getPreferredVersion('builds');
+    var replicaSetsVersion = APIService.getPreferredVersion('replicasets');
+    var limitRangesVersion = APIService.getPreferredVersion('limitranges');
+    var imageStreamsVersion = APIService.getPreferredVersion('imagestreams');
+    $scope.deploymentsVersion = APIService.getPreferredVersion('deployments');
+    $scope.eventsVersion = APIService.getPreferredVersion('events');
+    $scope.horizontalPodAutoscalersVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
+
     $scope.healthCheckURL = Navigate.healthCheckURL($routeParams.project,
                                                     "Deployment",
                                                     $routeParams.deployment,
-                                                    "apps");
+                                                    $scope.deploymentsVersion.group);
     var watches = [];
 
     ProjectsService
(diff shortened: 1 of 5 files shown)


spadgett:
@benjaminapetersen Still WIP or OK to merge?

app/views/browse/deployment.html
```diff
@@ -15,39 +15,39 @@ <h1 class="contains-actions">
                class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
                data-toggle="dropdown"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
             <ul class="dropdown-menu dropdown-menu-right actions action-button">
-              <li ng-if="!deployment.spec.paused && !updatingPausedState && ({ group: 'apps', resource: 'deployments' } | canI : 'update')">
+              <li ng-if="!deployment.spec.paused && !updatingPausedState && (deploymentsVersion | canI : 'update')">
```
spadgett: We need to update the `canIDoAny` check above to use the same resource versions

benjaminapetersen: got it

benjaminapetersen:
@spadgett think its good to roll now.

app/scripts/controllers/deployment.js
```diff
@@ -42,6 +43,14 @@ angular.module('openshiftConsole')
...
+    var replicaSetsVersion = APIService.getPreferredVersion('replicasets');
+    var limitRangesVersion = APIService.getPreferredVersion('limitranges');
+    var imageStreamsVersion = APIService.getPreferredVersion('imagestreams');
+    $scope.deploymentsVersion = APIService.getPreferredVersion('deployments');
+    $scope.eventsVersion = APIService.getPreferredVersion('events');
+    $scope.horizontalPodAutoscalersVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
+
     $scope.healthCheckURL = Navigate.healthCheckURL($routeParams.project,
                                                     "Deployment",
                                                     $routeParams.deployment,
```
spadgett: Need to add group to the health check URL.

benjaminapetersen: I think I missed something:

```JavaScript 
// group is 4th param:
healthCheckURL: function(projectName, kind, name, group) {
```
Looks like we are sending it `apps`?  Generated url: `https://localhost:9000/dev-console/project/foo/edit/health-checks?kind=Deployment&name=deployment-example&group=apps`

spadgett: Yeah, I think you can add `$scope.deploymentsVersion.group` as the last param, then it will always be the same as everything else.

I see we were passing `apps` before, which is right, but it's better to use the value from preferred version.

benjaminapetersen: ah, gotcha. will update.

benjaminapetersen:
Updated

spadgett:
[merge]

spadgett:
[merge]

benjaminapetersen:
```bash
Verifying that checked in built files under dist match the source...
```
updating

benjaminapetersen:
```
Verifying that checked in built files under dist match the source...
diff --git a/dist/scripts/scripts.js b/dist/scripts/scripts.js
```
hmm. updating again.

benjaminapetersen:
Can't generate a difference in `/dist`, rebased & reinstalled deps.

[test] again

benjaminapetersen:
rebased

spadgett:
[merge]

PR #2166, opened 2017-09-25
Update DeploymentsController to use getPreferredVersion

benjaminapetersen:
No `canI` matching view changes.

@spadgett


spadgett:
[merge]

PR #2068, opened 2017-09-13
Update unit test dependencies

benjaminapetersen:
https://trello.com/c/w3vdwj85

Updates `karma` related dependencies for unit testing.

screen shot 2017-09-13 at 3 12 37 pm

Looking into why a few tests now fail, will update again shortly (did we always have this pretty coverage tool?)(is a lot of red reported by that pretty coverage tool 😄 ). 

to fix:
- [x] one of the updates, probably jasmine, is more strictly comparing objects than previously, resulting in some tests failing.
- [x] the coverage table is being output twice since I'm running the tests in Firefox & Chrome concurrently by default.  This is a little annoying.
  
@spadgett

PR #2169, opened 2017-09-26
Bug 1495460 - Fix build config source namespace

spadgett:
Correctly set the build config source image namespace when adding a builder using advanced options.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1495460


spadgett:
@jeff-phillips-18 After looking at this more, I think it's better to undo the `ApplicationGenerator` change in #2068 and update the test. `input.namespace` should always be defined. Do you mind taking another look?

spadgett:
Thanks @jeff-phillips-18 

[merge]

Issue #2114, opened 2017-09-19
canIDoAny filter is confusing


benjaminapetersen:
Since the `canIDoAny` filter looks a lot like the `canI` filter, it is confusing to use.  `canI` takes a kubernetes resource, but `canIDoAny` takes a symbol representing the current page.  Since the string `'builds'` fulfills the requirement either way, it is a bit misleading.

@spadgett suggested a different approach in [this comment](https://github.com/openshift/origin-web-console/pull/2098#issuecomment-330544026).

Creating this as a tech debt issue to deal with later.

PR #2173, opened 2017-09-26
Update canIDoAny filter & canIDoAny on builds page

benjaminapetersen:
- canIDoAny will now accept an object or a string, other PRs will depend on this small tweak.

Follow-on to #2098

@spadgett

### DIFF
--- a/app/scripts/filters/canI.js
+++ b/app/scripts/filters/canI.js
@@ -2,15 +2,19 @@
 
 angular
   .module('openshiftConsole')
-  .filter('canIDoAny', function(canIFilter) {
-    var resourceRulesMap = {
+  .filter('canIDoAny', function(APIService, canIFilter) {
+    // Top level keys are representing pages.
+    // Within each page may one or more resources.
+    // canIDoAny returns truthy if the user passes the canIFilter
+    // for at least one of the resources and verbs listed in the page.
+    var pageRulesMap = {
       'buildConfigs': [
         {group: '', resource: 'buildconfigs',             verbs: ['delete', 'update']},
         {group: '', resource: 'buildconfigs/instantiate', verbs: ['create']}
       ],
       'builds': [
-        {group: '', resource: 'builds/clone', verbs: ['create']},
-        {group: '', resource: 'builds',       verbs: ['delete', 'update']}
+        _.assign({}, APIService.getPreferredVersion('builds/clone'), {verbs: ['create']}),
+        _.assign({}, APIService.getPreferredVersion('builds'), {verbs: ['delete', 'update']})
       ],
       'configmaps': [
         {group: '', resource: 'configmaps', verbs: ['update', 'delete']}
@@ -58,12 +62,15 @@ angular
       'projects': [
         {group: '', resource: 'projects', verbs: ['delete', 'update']}
       ],
+      // FIXME: inconsistent case (camel)
       'statefulsets': [
         {group: 'apps', resource: 'statefulsets', verbs: ['update', 'delete']}
       ]
     };
-    return function(resource) {
-      return _.some(resourceRulesMap[resource], function(rule) {
+
+    // the primary key is the page name, NOT the actual resource, though they appear to match.
+    return function(pageName) {
+      return _.some(pageRulesMap[pageName], function(rule) {
         return _.some(rule.verbs, function(verb) {
           return canIFilter({resource: rule.resource, group: rule.group}, verb);
         });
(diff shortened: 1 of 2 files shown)


app/views/browse/build.html
```diff
@@ -6,7 +6,7 @@
       <div ng-if="!loaded" class="mar-top-xl">Loading...</div>
       <div ng-if="build">
         <h1 class="contains-actions">
-          <div class="pull-right dropdown" ng-hide="!('builds' | canIDoAny)">
+          <div class="pull-right dropdown" ng-hide="!(buildsVersion | canIDoAny)">
```
spadgett: This should still be `'builds'`. It's not referring to an API version. It's a string identifier for the console page.

app/scripts/filters/canI.js
```diff
@@ -9,8 +9,8 @@ angular
         {group: '', resource: 'buildconfigs/instantiate', verbs: ['create']}
       ],
       'builds': [
-        {group: '', resource: 'builds/clone', verbs: ['create']},
-        {group: '', resource: 'builds',       verbs: ['delete', 'update']}
+        {group: 'build.openshift.io', resource: 'builds/clone', verbs: ['create']},
+        {group: 'build.openshift.io', resource: 'builds',       verbs: ['delete', 'update']}
```
spadgett: Use `APIService.getPreferredVersion` here so it always matches what we use in the view.

app/views/browse/build.html
spadgett: It has to match a key in the `canIDoAny` map

benjaminapetersen: yeah i though this was supposed to stay a string per our previous/first convo about `canIDoAny`. Hmm. Will revert.

app/scripts/filters/canI.js
benjaminapetersen: So `APIService.getPreferredVersion('builds')` -> `{group: 'build.openshift.io',     resource: 'builds' }`.  Do we want to take that & add the `verbs` property?

Example:
```JavaScript 
// update to this?
'builds': [
  _.extend(APIService.getPreferredVersion('builds/clone'), {verbs: ['create']});
  _.extend(APIService.getPreferredVersion('builds'), {verbs: ['delete', 'update']});
],
```

spadgett: Yes, except `_.extend` will mutate the preferred version object. Maybe put `verbs` first to avoid that problem.

```js
 _.extend({ verbs: ['create'], APIService.getPreferredVersion('builds/clone'));
```

benjaminapetersen:
Updated w/some comments... I'm open to revisiting this on the second part of the trello card.

app/scripts/filters/canI.js
```diff
@@ -2,15 +2,22 @@
...
+    var pageRulesMap = {
       'buildConfigs': [
         {group: '', resource: 'buildconfigs',             verbs: ['delete', 'update']},
         {group: '', resource: 'buildconfigs/instantiate', verbs: ['create']}
       ],
       'builds': [
-        {group: '', resource: 'builds/clone', verbs: ['create']},
-        {group: '', resource: 'builds',       verbs: ['delete', 'update']}
+        _.extend({}, APIService.getPreferredVersion('builds/clone'), {verbs: ['create']}),
+        _.extend({}, APIService.getPreferredVersion('builds'), {verbs: ['delete', 'update']})
```
spadgett: We should use `_.assign` instead of `_.extend`. I think we only want own properties.

app/scripts/filters/canI.js
```diff
@@ -2,15 +2,22 @@
 
 angular
   .module('openshiftConsole')
-  .filter('canIDoAny', function(canIFilter) {
-    var resourceRulesMap = {
+  .filter('canIDoAny', function(APIService, canIFilter) {
+    // Top level keys are representing pages.
+    // Within each page may one or more resources.
+    // canIDoAny returns truthy if the user passes the canIFilter
+    // for at least one of the resources listed in the page.
```
spadgett: nit: for at least one of the resources and verbs

app/scripts/filters/canI.js
```diff
@@ -2,15 +2,22 @@
...
+    // TODO: this is confusing, we should come back & fix
```
spadgett: I think having an issue is enough. Would rather not clutter the code with TODO comments for things we're already tracking in issues (#2114).

benjaminapetersen: agree, will remove

benjaminapetersen:
Updated

spadgett:
[merge]

spadgett:
@benjaminapetersen Diff in vendor.js

benjaminapetersen:
rebase/rebuilt

spadgett:
[merge]

PR #2174, opened 2017-09-26
Overview instance row improvements

spadgett:
* Show documentation URL
* Show support URL
* Add a delete action to failed deployment alerts
* Truncate long text in alert messages

openshift web console 2017-09-26 14-06-39

@rhamilto @ncameronbritt

### DIFF
--- a/app/scripts/directives/overview/serviceInstanceRow.js
+++ b/app/scripts/directives/overview/serviceInstanceRow.js
@@ -32,11 +32,6 @@
 
     var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
 
-    var getDescription = function() {
-      var serviceClassName = row.apiObject.spec.serviceClassName;
-      return _.get(row, ['state','serviceClasses', serviceClassName, 'description']);
-    };
-
     var updateInstanceStatus = function() {
       if (_.get(row.apiObject, 'metadata.deletionTimestamp')) {
         row.instanceStatus = 'deleted';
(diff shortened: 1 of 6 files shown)


ncameronbritt:
Looks good. I wonder if we shouldn't think about changing "Learn More" to "View Documentation" so that it's clearer where that link goes.

ncameronbritt:
I think in the wizard link too. "Learn More" is vague.

spadgett:
> I think in the wizard link too. "Learn More" is vague.

OK, makes sense. I'll tackle the catalog change in a separate PR since it's another repo anyway.

ncameronbritt:
Not suggesting we tackle this now, but we use "Learn More" to link to openshift documentation throughout the console. 

But here I think there's value in making the distinction between documentation and support clear.

spadgett:
@ncameronbritt I agree. It's only in one place in the catalog, so I just made the change there, too.

spadgett:
Updated the link text and switched to the `learn-more-inline` console class for consistency. It does change the font size. @ncameronbritt Look OK or should we change back to normal font size here?

openshift web console 2017-09-26 15-04-53

ncameronbritt:
@spadgett I know I already approved, but I think I like the larger font size for those links. I feel like they serve a slightly different function in this context as the other "learn more" links that we have in the console.

spadgett:
OK, back to normal font size with some small fixes for mobile styles.

spadgett:
[merge]

Issue #2045, opened 2017-09-11
Provide equivalent of 'oc set volume --overwrite'.


GrahamDumpleton:
If you attempt to mount a volume in a container of a pod, and try to name the volume definition the same as an existing volume definition, the web console will give you an error and not allow it.

![image](https://user-images.githubusercontent.com/507637/30256912-80931d9e-96f1-11e7-99bb-b48228f5bfac.png)

The problem is that the ability to reuse the existing volume definition, when it is the same underlying persistent volume claim, is necessary if you are attempting to share a volume between two different containers within the same pod.

When using the command line you can override any such error by using the ``oc set volume --overwrite`` option. A similar ability is needed in the web console.

There are possibly two choices. If the error comes up, show a checkbox to allow you to overwrite it. Or, maybe not even show the error if the volume definition exists but is for the same persistent volume claim. That is, make it intelligent enough to know that it can reuse the existing volume definition which has already been created for the same persistent volume claim.

Note that you cannot just create a second volume definition for the same underlying persistent volume as it seems having multiple volume definitions with different names, but for the same underlying persistent volume claim, against the one deployment configuration, cause deployment to hang.

spadgett:
Reopening because there is an issue with the current fix.

spadgett:
@GrahamDumpleton Making sure I understand the scenario: You want to be able to add the same PVC to different containers in the same pod? Do they need different mount paths? Otherwise you should be able to add it to all containers at once.

GrahamDumpleton:
They needed to be added at different mount points in each container and with different sub path from within the volume as it was not intended they use the same data, but different parts of the shared volume.

GrahamDumpleton:
As example:

```
                    "spec": {
                        "volumes": [
                            {
                                "name": "data",
                                "persistentVolumeClaim": {
                                    "claimName": "${APPLICATION_NAME}-data"
                                }
                            }
                        ],
                        "containers": [
                            {
                                "name": "wordpress",
                                ...
                                "volumeMounts": [
                                    {
                                        "name": "data",
                                        "mountPath": "/opt/app-root/src",
                                        "subPath": "wordpress-data"
                                    }
                                ]
                            },
                            {
                                "name": "mysql",
                                ...
                                "volumeMounts": [
                                    {
                                        "name": "data",
                                        "mountPath": "/var/lib/mysql/data",
                                        "subPath": "mysql-database"
                                    }
                                ]
                            }
```

[...]

PR #2176, opened 2017-09-26
add "overwrite" option to attachPVC view

juanvallejo:
Depends on: https://github.com/openshift/origin-web-common/pull/203
Fixes https://github.com/openshift/origin-web-console/issues/2045

Work in progress. Just wanted to begin gathering feedback.
Will post images.

cc @spadgett @benjaminapetersen

### DIFF
--- a/app/scripts/controllers/attachPVC.js
+++ b/app/scripts/controllers/attachPVC.js
@@ -118,6 +118,63 @@ angular.module('openshiftConsole')
         $scope.$watchGroup(['attach.resource', 'attach.allContainers'], updateMountPaths);
         $scope.$watch('attach.containers', updateMountPaths, true);
 
+        var checkVolumeMountPath = function(newVolumeMount, container) {
+          var duplicateMount = _.find(container.volumeMounts, function(mount) {
+            return mount.mountPath === newVolumeMount && mount.name !== newVolumeMount.name;
+          });
+
+          // if a new volumeMount matches an existing mountPath,
+          // fail if their names differ. This can happen when the
+          // "overwrite" option is specified.
+          if (duplicateMount) {
+              displayError('The volume mount "' + duplicateMount.mountPath + '" with name "' + duplicateMount.name +'" already exists for container "' + container.name + '"');
+              return false;
+          }
+
+          return true;
+        };
+
+        var replaceExistingVolumeMount = function(newVolumeMount, container) {
+          // if the volume mount we are trying to add already exists,
+          // replace the existing mount with the newly created one.
+          // This can happen when the "overwrite" option is specified.
+          var index = _.findIndex(container.volumeMounts, { name: newVolumeMount.name });
+          if (index === -1) {
+            return false;
+          }
+
+          container.volumeMounts[index] = newVolumeMount;
+          return true;
+        };
+
+        var setVolumeMount = function(podTemplate, name, mountPath, subPath, readOnly) {
+          var success = true;
+          _.each(podTemplate.spec.containers, function(container) {
+            if (!isContainerSelected(container)) {
+              return;
+            }
+
+            var newVolumeMount =
+              StorageService.createVolumeMount(name, mountPath, subPath, readOnly);
+            if (!container.volumeMounts) {
+              container.volumeMounts = [];
+            }
+
+            if (!checkVolumeMountPath(newVolumeMount, container)) {
+              success = false;
+              return false;
+            }
+
+            if (replaceExistingVolumeMount(newVolumeMount, container)) {
+              return false;
+            }
+
+            container.volumeMounts.push(newVolumeMount);
+          });
+
+          return success;
+        };
+
         // load resources required to show the page (list of pvcs and deployment or deployment config)
         var load = function() {
           DataService.get(resourceGroupVersion, $routeParams.name, context).then(
@@ -175,16 +232,10 @@ angular.module('openshiftConsole')
             var readOnly = $scope.attach.readOnly;
             if (mountPath) {
               // for each container in the pod spec, add the new volume mount
-              angular.forEach(podTemplate.spec.containers, function(container) {
-                if (isContainerSelected(container)) {
-                  var newVolumeMount =
-                    StorageService.createVolumeMount(name, mountPath, subPath, readOnly);
-                  if (!container.volumeMounts) {
-                    container.volumeMounts = [];
-                  }
-                  container.volumeMounts.push(newVolumeMount);
-                }
-              });
+              if(!setVolumeMount(podTemplate, name, mountPath, subPath, readOnly)) {
+                $scope.disableInputs = false;
+                return;
+              }
             }
 
             // add the new volume to the pod template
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/attachPVC.js
```diff
@@ -118,6 +118,51 @@ angular.module('openshiftConsole')
         $scope.$watchGroup(['attach.resource', 'attach.allContainers'], updateMountPaths);
         $scope.$watch('attach.containers', updateMountPaths, true);
 
+        var setVolumeMount = function(podTemplate, name, mountPath, subPath, readOnly) {
+          for (var i = 0; i < podTemplate.spec.containers.length; i++) {
```
juanvallejo: Will switch to an `angular.forEach`.

app/scripts/controllers/attachPVC.js
```diff
@@ -118,6 +118,51 @@ angular.module('openshiftConsole')
...
+              continue;
+            }
+
+            var newVolumeMount =
+              StorageService.createVolumeMount(name, mountPath, subPath, readOnly);
+            if (!containers[i].volumeMounts) {
+              containers[i].volumeMounts = [];
+            }
+
+            for (var idx = 0; idx < containers[i].volumeMounts.length; idx++) {
```
juanvallejo: Will switch to `angular.forEach`

app/scripts/controllers/attachPVC.js
```diff
@@ -118,6 +118,51 @@ angular.module('openshiftConsole')
...
+              // if a new volumeMount matches an existing mountPath,
+              // fail if their names differ. This can happen when the
+              // "overwrite" option is specified.
+              if (m.mountPath === newVolumeMount.mountPath && m.name !== newVolumeMount.name) {
+                displayError('The volume mount "' + m.mountPath + '" with name "' + m.name +'" already exists for container "' + containers[i].name + '"');
+                return false;
+              }
+            }
+
+            for (var mountIdx = 0; mountIdx < containers[i].volumeMounts.length; mountIdx++) {
```
juanvallejo: Will also switch to `angular.forEach`

app/scripts/controllers/attachPVC.js
spadgett: We usually use Lodash `_.each`

https://lodash.com/docs/4.17.4#forEach

app/views/attach-pvc.html
```diff
@@ -211,7 +225,7 @@ <h2 class="text-center">No persistent volume claims.</h2>
                   <button type="submit"
                       class="btn btn-primary btn-lg"
                       ng-click="attachPVC()"
-                      ng-disabled="attachPVCForm.$invalid || disableInputs || !attachPVC">Add</button>
+                      ng-disabled="(attachPVCForm.$invalid) || disableInputs || !attachPVC">Add</button>
```
spadgett: nit: shouldn't need the parens

app/scripts/controllers/attachPVC.js
```diff
@@ -175,24 +220,32 @@ angular.module('openshiftConsole')
...
+
+            // if the newly created volume already exists, only
+            // fail if the "overwrite" option was not set
+            var volumeExists = false;
+            for (var i = 0; i < podTemplate.spec.volumes.length; i++) {
+              if (podTemplate.spec.volumes[i].name === newVolume.name) {
+                volumeExists = true;
+                break;
+              }
+            }
```
spadgett: You should use `_.some` to make this a one-liner:

```js
var volumeExists = _.some(podTemplate.spec.volumes, { name: newVolume.name });
```

https://lodash.com/docs/4.17.4#some

app/scripts/controllers/attachPVC.js
benjaminapetersen: fwiw, `var` isn't block scoped (inside if statements), switching to `_.each(function() { })` will fix.  It wouldn't error, but its a bit sketch. yay js 😄

app/views/attach-pvc.html
```diff
@@ -83,6 +83,7 @@ <h2 class="text-center">No persistent volume claims.</h2>
                       ng-model="attach.mountPath"
                       ng-pattern="/^\/.*$/"
                       osc-unique="existingMountPaths"
+                      osc-unique-disabled="attach.overwrite"
```
benjaminapetersen: +1

juanvallejo:
@spadgett thanks for the feedback, review comments addressed

juanvallejo:
@benjaminapetersen thanks, PR re-built and rebased

juanvallejo:
re[test]

benjaminapetersen:
```bash
```
😄 
try:
```bash 
hack/clean-deps.sh
hack/install-deps.sh
grunt build 
// commit this stuff
```

juanvallejo:
@benjaminapetersen thanks! that seems to have done the trick

benjaminapetersen:
Yup, now you just have the same e2e test errors that all of our other PRs are struggling with. Woohoo!

benjaminapetersen:
[test] since [Jenkins file changed](https://github.com/openshift/aos-cd-jobs/pull/709)

juanvallejo:
@spadgett friendly ping

juanvallejo:
@spadgett per our conversation, went ahead and broke `setVolumeMount` into smaller functions https://github.com/openshift/origin-web-console/pull/2176/commits/f4f32b0ad38d27d2db309bffbc036e2d535485f8

app/views/attach-pvc.html
```diff
@@ -174,6 +176,18 @@ <h2 class="text-center">No persistent volume claims.</h2>
                 <div class="form-group">
                   <div class="checkbox">
                     <label>
+                      <input type="checkbox" ng-model="attach.overwrite" aria-describedby="overwrite-help">
+                      Overwrite
+                    </label>
+                    <div id="overwrite-help" class="help-block">
+                      Overwrite the volume mount config (if it already exists).
```
spadgett: I'd just say "Overwrite a volume mount if it already exists."

app/scripts/controllers/attachPVC.js
```diff
@@ -175,24 +238,27 @@ angular.module('openshiftConsole')
...
-            podTemplate.spec.volumes.push(newVolume);
+
+            // if the newly created volume already exists, only
+            // fail if the "overwrite" option was not set
+            var volumeExists = _.some(podTemplate.spec.volumes, { name: newVolume.name });
+
+
+            if (!$scope.attach.overwrite || ($scope.attach.overwrite && !volumeExists)) {
+              podTemplate.spec.volumes.push(newVolume);
+            }
```
spadgett: Is it correct to continue if the volume exists and overwrite is not set?

app/scripts/controllers/attachPVC.js
```diff
@@ -175,24 +238,27 @@ angular.module('openshiftConsole')
...
-                if (isContainerSelected(container)) {
-                  var newVolumeMount =
-                    StorageService.createVolumeMount(name, mountPath, subPath, readOnly);
-                  if (!container.volumeMounts) {
-                    container.volumeMounts = [];
-                  }
-                  container.volumeMounts.push(newVolumeMount);
-                }
-              });
+              // angular.forEach(podTemplate.spec.containers, function(container) {
```
spadgett: Remove commented out code

app/scripts/controllers/attachPVC.js
juanvallejo: > Is it correct to continue if the volume exists and overwrite is not set?

Ah, no pushing a volume containing a duplicate name field would cause an error, however I think I was relying on the client-side checks that we currently have to stop a user from submitting the form if a duplicate name or mount path was found.

Will go ahead and update this to just be 
```js
if (!volumeExists) {
  podTemplate.spec.volumes.push(newVolume);
}
```

PR #2184, opened 2017-09-28
Making next steps/results messaging consistent

rhamilto:
To Do:

- [ ] Update styling of wizard contents once comps are finalized per @ncameronbritt and @jennyhaines
- [x] Implement changes to remove mega alert and show simple bulleted list like [create-from-builder-results.html](https://github.com/openshift/origin-web-catalog/blob/48cc8ad4258a937ff7eb52c2d5cfc8890749e7da/src/components/create-from-builder/create-from-builder-results.html#L34-L55) for select from project (wizard) and import (next steps and wizard)
  screen shot 2017-09-28 at 3 17 19 pm
- [ ] Display webhook info when appropriate inside wizard results (currently only shows on standalone next steps)
- [ ] Show proper type of imported item in import wizard results (non-templates are generically called "YAML / JSON")
- [x] Remove import toast notifications (currently needed for standalone import next steps as it redirects to overview)
- [ ] Add appropriate actions when failure occurs

Screenshots:

screen shot 2017-09-28 at 2 44 17 pm
screen shot 2017-09-28 at 3 09 26 pm
screen shot 2017-09-28 at 3 16 47 pm
screen shot 2017-09-28 at 3 17 19 pm
screen shot 2017-09-28 at 12 06 05 pm
screen shot 2017-09-28 at 12 10 22 pm
screen shot 2017-09-28 at 12 11 35 pm
screen shot 2017-09-29 at 2 33 57 pm
screen shot 2017-09-29 at 2 36 59 pm
screen shot 2017-09-29 at 2 53 40 pm


spadgett:
> Display webhook info when appropriate inside wizard results (currently only shows on standalone next steps)

I think the problem here is that we're just not setting the webhook when not using advanced options. This is arguably just a bug.

spadgett:
> Implement changes to remove mega alert and show simple bulleted list like create-from-builder-results.html for select from project (wizard) and import (next steps and wizard)

I wonder if we shouldn't just change the styling of the tasks component and remove it from the overview.

ncameronbritt:
Since we have the different webhook options (bitbucket, gitlab), is that something we would want to let users set up as part of this flow, and then show the results on that last step? I'm not sure how common it is that a user would want to set those up.

@jhaines was working on restyling the tasks component, but that's on hold because I was wanting to figure out what a user should do after a partial successes/failure, and what's important for the user to know to take that next step.

rhamilto:
@spadgett, I think this is ready for review/merge with the understanding we'll tackle the remaining to-dos as follow-ons.

spadgett:
> Since we have the different webhook options (bitbucket, gitlab), is that something we would want to let users set up as part of this flow, and then show the results on that last step? I'm not sure how common it is that a user would want to set those up.

We should probably be consistent in what we set up by default in the dialog vs advanced. There is a checkbox for it in advanced options.

If you are using GitHub, Bitbucket, etc., you probably want them. But if you're not using GitHub and we set up a GitHub webhook for you, that's not helpful. I don't know if we can reliably detect from the Git URL which ones to create.

app/views/directives/next-steps.html
```diff
@@ -39,62 +87,33 @@ <h1 ng-if="!$ctrl.pendingTasks(tasks()).length && $ctrl.erroredTasks(tasks()).le
...
-    <pre class="code prettyprint">oc login {{$ctrl.loginBaseUrl}}
-oc project {{$ctrl.projectName}}
-oc status</pre>
-
-    <p>For more information about the command line tools, check the <a target="_blank" href="{{'cli' | helpLink}}">CLI Reference</a> and <a target="_blank" href="{{'basic_cli_operations' | helpLink}}">Basic CLI Operations</a>.</p>
-  </div>
-
-  <div ng-if="$ctrl.createdBuildConfig" class="col-md-6">
-    <h2>Making code changes</h2>
+<div ng-controller="TasksController" ng-if="$ctrl.createdBuildConfig">
```
spadgett: Remove `ng-controller` here

rhamilto:
Comment addressed.

rhamilto:
> We should probably be consistent in what we set up by default in the dialog vs advanced. There is a checkbox for it in advanced options.

And that checkbox is enabled by default.  So if you use the wizard, you get no webhook.  But if you used advanced with the default, you do.  Seems a bit confusing?

ncameronbritt:
Agree with your comments above re webhooks @spadgett. If it's useful, maybe down the line we can look at adding a step to the wizard or including it in the builder flow some other way.

spadgett:
[merge]

Issue #2179, opened 2017-09-27
 Black background doesn't extend to bottom of page for chromeless loading message.


spadgett:
The area below the message is white instead of black while the log is loading.

![screen shot 2017-09-27 at 2 22 26 pm](https://user-images.githubusercontent.com/1167259/30930340-7351c9b4-a38f-11e7-92ff-1e86c12dbee0.png)

PR #2185, opened 2017-09-28
Make .chromeless background full height while logs are loading.

sg00dwin:
Remove old unused ellipsis class

Fixes https://github.com/openshift/origin-web-console/issues/2179

chromeless-log

### DIFF
--- a/app/styles/_log.less
+++ b/app/styles/_log.less
@@ -1,6 +1,7 @@
 .chromeless {
   background-color:  @log-bg-color;
   color: #d1d1d1;
+  height: 100%; // extends dark background to full height while logs are loading
   .log-loading-msg {
     .empty-state-message();
     .h2();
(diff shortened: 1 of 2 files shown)


sg00dwin:
@spadgett

app/styles/_log.less
```diff
@@ -165,17 +166,6 @@
     }
   }
   &.log-fixed-height {
-    .ellipsis-loader {
```
spadgett: `log-fixed-height` is used on the monitoring page. Have you confirmed this doesn't break anything there?

sg00dwin: It doesn't. This should have been removed when`.ellipsis-loader` replaced by `.ellipsis-pulser`

spadgett:
[merge]

Issue #1590, opened 2017-05-25
Autoscaler's "CPU Limit Target" in UI does not match what the CLI returns


alikhajeh1:
To reproduce this issue:
1. From the Deployment UI > Actions > Add autoscaler, then set "CPU Limit Target" to 72. Click on save and confirm that the UI shows 72%:
<img width="237" alt="screen shot 2017-05-25 at 6 15 34 pm" src="https://cloud.githubusercontent.com/assets/638577/26452771/6857b558-4176-11e7-8089-82658e47cc5b.png">

2. Using the CLI, run `oc get hpa` and notice it doesn't match the UI:
```
NAME               REFERENCE                           TARGET    CURRENT   MINPODS   MAXPODS   AGE
php-laravel-todo   DeploymentConfig/php-laravel-todo   600%      2%        1         3         14m
```

`oc describe hpa/php-laravel-todo` also shows the issue:
```
Name:				php-laravel-todo
Namespace:			examples
Labels:				app=php-laravel-todo
Annotations:			<none>
CreationTimestamp:		Thu, 25 May 2017 18:00:53 +0430
Reference:			DeploymentConfig/php-laravel-todo
Target CPU utilization:		600% <<<<<<<< not 72%
Current CPU utilization:	4%
Min replicas:			1
Max replicas:			3
No events.
```

The min and max replica counts match.

```
$ oc version
oc v1.5.0+031cbe4
kubernetes v1.5.2+43a9be4
features: Basic-Auth

Server XXX
openshift v1.5.0
kubernetes v1.5.2+43a9be4
```

Tested with Chrome and FireFox.

spadgett:
@alikhajeh1 When overcommit is configured for the cluster, the web console shows the target as a percentage of CPU **limit**, whereas the CLI is showing it as a percentage of CPU **request**. See also

https://github.com/openshift/origin/issues/13907#issuecomment-297714714

Unfortunately this is confusing if you use both the CLI and web console. @jwforres I think we should look at the user experience here.

jwforres:
Closing this one since it is a duplicate of https://github.com/openshift/origin/issues/13907

PR #2187, opened 2017-09-29
Bug 1496687 - Consistent binding names

spadgett:
* Show binding name rather than secret name in collapsed overview row to be consistent with what we show when expanded.
* Remove "Secret" heading above binding name to avoid confusion. We shouldn't label a ServiceBinding as a Secret since they're different resources.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1496687

openshift web console 2017-09-29 07-54-40

openshift web console 2017-09-29 07-55-31

cc @ncameronbritt

### DIFF
--- a/app/views/directives/_service-binding.html
+++ b/app/views/directives/_service-binding.html
@@ -3,9 +3,6 @@
     <div ng-class="{'col-sm-5 col-md-6': $ctrl.isOverview,
                     'col-sm-8 col-md-6 col-lg-8': !$ctrl.isOverview}">
       <h3>
-        <div class="component-label">
-          Secret
-        </div>
         {{$ctrl.binding.metadata.name}}
         <span ng-if="$ctrl.refApiObject.kind !== 'ServiceInstance'">
           <small ng-if="$ctrl.serviceClass">
--- a/app/views/overview/_service-instance-row.html
+++ b/app/views/overview/_service-instance-row.html
@@ -48,7 +48,7 @@
               {{application.metadata.name}}
             </span>
             <span ng-if="!application">
-              {{firstBinding.spec.secretName}}
+              {{firstBinding.metadata.name}}
             </span>
             <span ng-if="row.bindings.length > 1">
               and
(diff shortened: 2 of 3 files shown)


ncameronbritt:
@spadgett We should definitely address the inconsistency btw collapsed and expanded, but I want to think through which option would be more helpful for the user to see. I think the most confusing case we came up with was when you had multiple bindings and wanted to make sure you were deleting the correct one. Isn't it the secret that ends up getting tied to the application, not the binding? Can the user even do anything with the binding object? Also worth thinking about how things will be when pod presets are in.

spadgett:
> I think the most confusing case we came up with was when you had multiple bindings and wanted to make sure you were deleting the correct one.

To me, the most confusing thing is to show a list of bindings, and it's not binding names that we show, but secret names. If you run `oc get servicebindings` on the CLI, you'll see a totally different list.

Some users also don't have authority to view secrets, but will be able view bindings.

> Can the user even do anything with the binding object? 

You can check status and delete bindings.  You can see what parameters were used for the binding. At some point, you might be able to update binding parameters.

If the user wants to do anything with a binding on the CLI, they'll need to know the name. 

> Also worth thinking about how things will be when pod presets are in.

True. Right now (in 3.6 tech preview), I believe we should the application name with the binding name in smaller, muted text.

FWIW, in the past, I've regretted it every time we've decided to pretend one thing is really another. Deployments vs replication controllers is one example. HPA target CPU is another example. We've gotten a lot of bugs like #1590 because what we show in the web console is different than the CLI. I feel like in the end it makes things more confusing and does the user a disservice.

spadgett:
Also: you can always get to the secret from the binding, but you can't get to the binding from the secret.

ncameronbritt:
I really thought that in 3.6 tech preview we were showing the secret or the application. I'm pretty sure that's how we designed the Provisioned Service details page too. I feel like either the binding object didn't exist when we were first working on these designs, or if it did, we just weren't aware of it.

spadgett:
Got OK from @ncameronbritt on this after talking in person (thanks @ncameronbritt)

[merge]

spadgett:
[merge]

PR #2188, opened 2017-09-29
Fixes bug where required field indicator was misaligned in deploy ima…

rhamilto:
…ge overlay

Before:

screen shot 2017-09-29 at 9 32 52 am

After:

screen shot 2017-09-29 at 9 32 24 am


spadgett:
[merge]

PR #2194, opened 2017-09-29
Don't let users accidentally and permanently dismiss router hostname message

spadgett:
Knowing the router hostname is vital for setting up a CNAME for routes
with custom hosts. Remove the "Don't Show Me Again" link to prevent
users from accidentally dismissing the message without being able to get
it back.

See https://github.com/openshift/origin-web-console/pull/1194#issuecomment-333160295

@Download FYI, we're also looking at the other feedback you've given.


spadgett:
[merge]

Download:
Wow guys you are on top of it! Much appreciated!

PR #2198, opened 2017-10-02
EnvFrom Issue Updates

### LINKED ISSUE #2182: EnvFrom in the Env Editors Interface Issues
EnvFrom
- [x] The drag and drop handle associated with updating the sequence order of EnvFrom values doesn't remain hidden when there is a single entry
- [x] Missing link for viewing Config Maps and Secret details. It would be helpful to see these details at the time of adding these bulk additions.
- [x] Minor copy revisit
- [x] Minor CSS color highlight and contrast issue on disabled selected values. The color needs adjustment, either the copy or the background... or both.
- Prefixes for EnvFrom values, see issue #2362 

EnvFrom & Variables 
- [x] Random refresh issue where a background process appears to run and reset some of the drop-downs causing them to collapse even though the user has them open
  - This refresh goes towards a larger issue around how watches on the parent view are providing updates, which also happens to affect the setting of individual environment values.
- [x] On saving, dependent on background processes, there appears to be a random flash of a hidden form. This also appears to happen when saving environment variables.

cdcabrera:
EnvFrom issue fixes for CSS, copy, and UX. Updates parts of #2182 

- Drag & Drop for single entries
- Copy
- CSS highlighting on disabled options

### DIFF
--- a/app/scripts/directives/editEnvironmentFrom.js
+++ b/app/scripts/directives/editEnvironmentFrom.js
@@ -134,7 +134,11 @@
       return referenceValue;
     };
 
-    ctrl.checkEntries = function(option) {
+    ctrl.checkEntries = function(option, entrySelectedEnvFrom) {
+      if(option === entrySelectedEnvFrom) {
+        return false;
+      }
+
       return !!(getReferenceValue(option));
     };
 
--- a/app/views/directives/edit-environment-variables.html
+++ b/app/views/directives/edit-environment-variables.html
@@ -46,7 +46,7 @@
       entries="container.envFrom"
       selector-placeholder="Secret/Config Map"
       env-from-selector-options="$ctrl.valueFromObjects"
-      add-row-link="Add ALL Values from a Resource"
+      add-row-link="Add ALL Values from Secret or Config Map"
       show-header>
     </edit-environment-from>
   </div>
--- a/app/views/directives/edit-environment-from.html
+++ b/app/views/directives/edit-environment-from.html
@@ -42,7 +42,7 @@
                 </span>
               </ui-select-match>
               <ui-select-choices
-                ui-disable-choice="$ctrl.checkEntries(source)"
+                ui-disable-choice="$ctrl.checkEntries(source, entry.selectedEnvFrom)"
                 repeat="source in $ctrl.envFromSelectorOptions | filter : { metadata: { name: $select.search } } track by (source | uid)"
                 group-by="$ctrl.groupByKind">
                 <span ng-bind-html="source.metadata.name | highlight : $select.search"></span>
@@ -54,7 +54,7 @@
 
       <div ng-if="!$ctrl.isReadonlyAny && !entry.isReadonlyValue" class="environment-from-editor-button">
         <span
-          ng-if="!$ctrl.cannotSort"
+          ng-if="!$ctrl.cannotSort && $ctrl.entries.length > 1"
           class="fa fa-bars sort-row"
           role="button"
           aria-label="Move row"
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -9158,8 +9158,8 @@ name: e.metadata.name
 }
 return t;
 };
-a.checkEntries = function(e) {
-return !!l(e);
+a.checkEntries = function(e, t) {
+return e !== t && !!l(e);
 };
 var u = function(e, t) {
 a.cannotAdd = a.isReadonlyAny || _.isEmpty(t), t && _.each(t, function(e) {
--- a/dist/scripts/templates.js
+++ b/dist/scripts/templates.js
@@ -6674,7 +6674,7 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
     "<small class=\"text-muted\">&ndash; {{$select.selected.kind | humanizeKind : true}}</small>\n" +
     "</span>\n" +
     "</ui-select-match>\n" +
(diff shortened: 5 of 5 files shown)


app/views/directives/edit-environment-variables.html
```diff
@@ -46,7 +46,7 @@ <h4 class="section-label">
       entries="container.envFrom"
       selector-placeholder="Secret/Config Map"
       env-from-selector-options="$ctrl.valueFromObjects"
-      add-row-link="Add ALL Values from a Resource"
+      add-row-link="Add ALL Variables from Secret or Config Map"
```
spadgett: I think "Add ALL Values..." is more accurate since secrets and config maps don't have "variables."

@beanh66 Are you OK with that change?

beanh66: @spadgett That works for me! So it would be "Add ALL Values from Secret or Config Map" correct?

spadgett: > So it would be "Add ALL Values from Secret or Config Map" correct?

@beanh66 Yup! That's what I was thinking

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -134,7 +134,11 @@
       return referenceValue;
     };
 
-    ctrl.checkEntries = function(option) {
+    ctrl.checkEntries = function(option, entrySelectedEnvFrom) {
+      if(_.isEqual(option, entrySelectedEnvFrom)) {
```
spadgett: Does `option === entrySelectedEnvFrom` work or is a copy?

app/views/directives/edit-environment-variables.html
```diff
@@ -46,7 +46,7 @@ <h4 class="section-label">
       entries="container.envFrom"
       selector-placeholder="Secret/Config Map"
       env-from-selector-options="$ctrl.valueFromObjects"
-      add-row-link="Add ALL Values from a Resource"
+      add-row-link="Add ALL Values from Secret or Config Map"
```
benjaminapetersen: +1 for clarity here.

app/scripts/directives/editEnvironmentFrom.js
cdcabrera: So it might... but the data within "entrySelectedEnvFrom" is "moved" over so kind of a copy ... I had thought about placing a check for ```option.kind === entrySelectedEnvFrom.kind``` to avoid firing the ```_.isEqual``` every time

spadgett: You could check the UIDs, something like (untested)

```js
if (entrySelectedEnvFrom.metadata.uid === option.metadata.uid) {
  return false;
}
```

assuming the neither is ever null.

spadgett: Secrets and config maps can contains megabytes of data, so would be good to avoid a deep comparison if possible.

cdcabrera: That works, and understood on the object complexity... Went ahead and checked the ```===``` comparison, it does appear to function as expected, they evaluate as equal. Was a little unsure since one of the parameters is being pulled through a nested ```ng-repeat```

spadgett:
@beanh66 Opinion on these proposed changes? I feel like we're repeating ourselves with some of the labels and headings.

Before:

openshift web console 2017-10-02 11-32-32

After:

openshift web console 2017-10-02 11-33-55

benjaminapetersen:
Hmm. True, that does feel a touch overkill...

spadgett:
I'm going to go ahead and merge this since it's fixes several problems. Thanks @cdcabrera 

@beanh66 @ncameronbritt I still wouldn't mind your thoughts on my screenshots just above, though: https://github.com/openshift/origin-web-console/pull/2198#issuecomment-333572355

[merge]

spadgett:
Looks like a webdriver flake

[merge]

beanh66:
@cdcabrera @spadgett see my comments below: 

- It might be best to have the two dropdowns say "Select a resource" for simplicity. 
- I am fine with removing the top "Variables" label as it doesn't add much. 
- In terms of the links, whether you are adding a single or multiple values form a config map/secret, the language should be consistent. If you are changing the last link to be "Add ALL **Values** from Secret or Config Map" then I think the one above should also be "Add **Value** from Secret or Config Map." I'm fine with the word Value or Variable here but I do think they should match.
- I also noticed the links say "Config Map or Secret" but in the actual dropdown the secrets are listed first so we should be consistent. Either change the order in the dropdown or change the link to "Secret or Config Map"

Thoughts?

spadgett:
> In terms of the links, whether you are adding a single or multiple values form a config map/secret, the language should be consistent.

Yeah, I see your point. I prefer "Add Value from Secret or Config Map"

It also has the nice property of matching the field you see (`valueFrom`) when editing YAML.

> I also noticed the links say "Config Map or Secret" but in the actual dropdown the secrets are listed first so we should be consistent. Either change the order in the dropdown or change the link to "Secret or Config Map"

Yeah, we should fix the group ordering. It might be a random order now.

spadgett:
[merge]

spadgett:
[test]

spadgett:
Opened flake #2200

spadgett:
[merge]

spadgett:
[test]

spadgett:
[merge]

spadgett:
[test]

PR #2205, opened 2017-10-03
Remove "tasks" message from overview

spadgett:
The wizard results step and next step pages show the tasks result. It's not needed on the overview and results in multiple different alert styles.

I removed one of these in #2184, but I missed the other:

https://github.com/openshift/origin-web-console/pull/2184/files#diff-361579fdd045e14f6dc3e106d5feb425L5

### DIFF
--- a/app/views/overview.html
+++ b/app/views/overview.html
@@ -101,7 +101,6 @@
       </div>
       <div class="middle-content">
         <div class="container-fluid">
-          <tasks></tasks>
           <alerts alerts="overview.state.alerts"></alerts>
           <div ng-if="overview.everythingFiltered && overview.viewBy !== 'pipeline'">
             <div class="empty-state-message text-center h2">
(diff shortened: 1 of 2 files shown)


rhamilto:
Do we want to go ahead and remove the tasks directive since it's no longer being used?  https://github.com/openshift/origin-web-console/blob/master/app/views/directives/next-steps.html#L54-L83 is duplication of the task directive's markup (i've not done a line-by-line comparison to see if there are differences).

rhamilto:
> Do we want to go ahead and remove the tasks directive since it's no longer being used? https://github.com/openshift/origin-web-console/blob/master/app/views/directives/next-steps.html#L54-L83 is duplication of some of the task directive's markup (they differ a bit).

I've removed the tasks directive in https://github.com/openshift/origin-web-console/pull/2209

spadgett:
Closing in favor of #2209

Issue #2200, opened 2017-10-02
Test Flake: A Jasmine spec timed out. Resetting the WebDriver Control Flow.


spadgett:
```
Running "protractor:default" (protractor) task
Starting selenium standalone server...
[launcher] Running 1 instances of WebDriver
Selenium standalone server started at http://172.18.14.245:57776/wd/hub
Spec started
visiting url: create-project
visiting url: project/console-test-project-1506967814825/create
visiting url: project/console-test-project-1506967814825/browse/images
A Jasmine spec timed out. Resetting the WebDriver Control Flow.
The last active task was: 
Protractor.get(https://localhost:9000/dev-console/project/console-test-project-1506967814825/browse/images) - reset url
```

spadgett:
https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_origin_web_console/305

spadgett:
Screenshot is just a loading screen:

![user adds a template to a project after creating a new project using the import yaml tab should process and create the objects in the template 6](https://user-images.githubusercontent.com/1167259/31092902-eb819f80-a77d-11e7-84a4-c4cccde95d92.png)

cc @benjaminapetersen

benjaminapetersen:
Yeah I think in all this `e2e` stuff the dependencies aren't nailed down as well as we would like.  Selenium or webdriver versions seems to suddenly change and things get funny. :/

spadgett:
> Selenium or webdriver versions seems to suddenly change and things get funny.

We probably should commit package-lock.json to avoid this stuff.

benjaminapetersen:
Re: #2083

spadgett:
Might be unrelated, but Firefox recently changed from 

52.3.0-2.el7_4 -> 52.4.0-1.el7_4

spadgett:
https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_origin_web_console/304/
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_web_console/244/

It's failing on `Protractor.get`, but not every time. I wonder if we just need to bump the timeout.

spadgett:
Timeout is only 3s. I can see the page taking longer than that to load. Since the tests take 20-30 minutes to run in Jenkins, we might want to bump this to something much higher.

> Error: Wait timed out after 3050ms

benjaminapetersen:
I think its in here:

```bash
Spec started
visiting url: create-project
[...]

PR #2083, opened 2017-09-15
[WIP] Update e2e test dependency updates & get them running on Mac

benjaminapetersen:
Based off of #2068, though not really dependent.

Gets Protractor running again on Mac:
- via Chrome
- Firefox will not run due to cert signing issues (should still be fine in travis since we define an older firefox)

Currently at least one test is broken, so they do not pass locally, but they do run.  Working on fixing these up. 

@spadgett

PR #2210, opened 2017-10-03
Temporarily disable some integration tests to unblock queue

spadgett:
Run the integration tests, but don't fail if only the integration tests fail. See #2200

Merge will still fail on dist differences or unit test failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,5 +44,7 @@ build: install
 test: build
 	hack/verify-dist.sh
 	hack/test-headless.sh test $(GRUNT_FLAGS)
-	hack/test-headless.sh test-integration $(GRUNT_FLAGS)
+	# Run the integration tests, but don't fail on test failures.
+	# Temporary until we are able to upgrade protractor.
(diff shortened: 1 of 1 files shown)


spadgett:
[test]

spadgett:
[test]

spadgett:
@jeff-phillips-18 @benjaminapetersen Any objection to this? Our merge queue has been broken for several days now. We haven't been able to find a simple fix.

I think the right path forward is to update Protractor in #2083, which we need to do anyway.

Makefile
```diff
@@ -44,5 +44,5 @@ build: install
 test: build
 	hack/verify-dist.sh
 	hack/test-headless.sh test $(GRUNT_FLAGS)
-	hack/test-headless.sh test-integration $(GRUNT_FLAGS)
+	-hack/test-headless.sh test-integration $(GRUNT_FLAGS)
```
benjaminapetersen: only thought, comment this?  the `-` runs it but wont fail?

spadgett:
[merge]

PR #2216, opened 2017-10-05
Bug 1498897 - Don't break if quota status is not set

spadgett:
Fixes a runtime error in the console checking resource quota when the quota status has not been set.

https://bugzilla.redhat.com/show_bug.cgi?id=1498897

cc @jupierce


spadgett:
I'm still testing, but should be ready for code review.

jupierce:
@spadgett We will need it in stage branch too. Thanks.

spadgett:
@jupierce Sure thing, stage PR coming shortly.

[merge]

PR #2222, opened 2017-10-05
Correctly hide/show quota notifications per project

dtaylor113:
Addresses issues when switching between projects and viewing and hiding/clearing quota notifications.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1495455#c4

### DIFF
--- a/app/scripts/directives/notifications/notificationDrawerWrapper.js
+++ b/app/scripts/directives/notifications/notificationDrawerWrapper.js
@@ -99,6 +99,7 @@
         _.each(drawer.notificationGroups, function(group) {
           _.remove(group.notifications, { uid: notification.uid, namespace: notification.namespace });
         });
+        delete notificationsMap[$routeParams.project][notification.uid];
       };
 
       var formatAPIEvents = function(apiEvents) {
@@ -172,11 +173,13 @@
       };
 
       var notificationWatchCallback = function(event, notification) {
-        if(!notification.showInDrawer) {
+        var project = notification.namespace || $routeParams.project;
+        var id = notification.id ? project + "/" + notification.id : _.uniqueId('notification_') + Date.now();
+
+        if(!notification.showInDrawer || EventsService.isCleared(id)) {
           return;
         }
-        var project = notification.namespace || $routeParams.project;
-        var id = notification.id || _.uniqueId('notification_') + Date.now();
+
         notificationsMap[project] = notificationsMap[project] || {};
         notificationsMap[project][id] = {
           actions: notification.actions,
--- a/app/scripts/services/quota.js
+++ b/app/scripts/services/quota.js
@@ -317,7 +317,7 @@ angular.module("openshiftConsole")
 
           if(hardValue <= usedValue) {
             notifications.push({
-              id: "quota-limit-reached-" + quotaKey,
+              id: projectName + "/quota-limit-reached-" + quotaKey,
               namespace: projectName,
               type: (hardValue < usedValue ? 'warning' : 'info'),
               message: getNotificaitonMessage(used, usedValue, hard, hardValue, quotaKey),
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -3598,7 +3598,7 @@ var t = e.status.total || e.status;
 _.each(t.hard, function(e, s) {
 var c = m(e), l = _.get(t, [ "used", s ]), d = m(l);
 "resourcequotas" !== s && c && d && c <= d && i.push({
-id: "quota-limit-reached-" + s,
+id: o + "/quota-limit-reached-" + s,
 namespace: o,
 type: c < d ? "warning" : "info",
 message: I(0, d, e, c, s),
@@ -14418,7 +14418,7 @@ _.remove(t.notifications, {
 uid: e.uid,
 namespace: e.namespace
 });
-});
+}), delete h[r.project][e.uid];
 }, P = function(e) {
 return _.map(e, function(e) {
 return {
@@ -14456,9 +14456,8 @@ l && l(), l = null;
 }, $ = function(e) {
 g[r.project] = P(R(e.by("metadata.name"))), T();
 }, B = function(e, t) {
-if (t.showInDrawer) {
-var n = t.namespace || r.project, a = t.id || _.uniqueId("notification_") + Date.now();
-h[n] = h[n] || {}, h[n][a] = {
+var n = t.namespace || r.project, a = t.id ? n + "/" + t.id : _.uniqueId("notification_") + Date.now();
+t.showInDrawer && !c.isCleared(a) && (h[n] = h[n] || {}, h[n][a] = {
 actions: t.actions,
 unread: !c.isRead(a),
 trackByID: t.trackByID,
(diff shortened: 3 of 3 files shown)


app/scripts/services/resourceAlerts.js
```diff
@@ -74,7 +74,8 @@ angular.module("openshiftConsole")
     var setQuotaNotifications = function(quotas, clusterQuotas, projectName) {
       var notifications = QuotaService.getQuotaNotifications(quotas, clusterQuotas, projectName);
       _.each(notifications, function(notification) {
-        if(!NotificationsService.isNotificationPermanentlyHidden(notification)) {
+        // Only add a quota notification if it wasn't specifically set to be hidden by user
+        if(!NotificationsService.isNotificationHidden(notification)) {
```
spadgett: I don't think this should be necessary since the drawer tracks what was cleared in session storage. I think the drawer should know whether to show or hide the notification already.

dtaylor113: > My worry is that this doesn't fix the problem for other notifications that don't include the project name in the ID.

That's why I was working in `namespace` into the common NotificationsService :smile: 
> 
> @dtaylor113 If you prepend the project name to the uid the notification drawer assigns, does it just work?
> 
> https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/notifications/notificationDrawerWrapper.js#L187
> 
> This might be a one line fix without needing the common changes?

I'll have to test after removing the `isNotificationHidden(..)` call in resourceAlerts.js.   The drawer currently uses EventsService.isCleared() for filtering the APIEvents; I need to add the `EventsService.isCleared()` call to the notifications callback.

dtaylor113:
Until `namespace` is built into `NotificationsService.isNotificationVisible()`, I'll need to add `projectName` to the quota notification id.   Setting the `uid` happens after `NotificationsService.addNotification(...)`

dtaylor113:
> My worry is that this doesn't fix the problem for other notifications that don't include the project name in the ID.

I had to add projectName into the notification id in order for the NotificationsService to process it correctly.  I added a conditional for the `uid` to include the project name if it isn't already in the `id`.

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -184,7 +185,7 @@
           // using uid to match API events and have one filed to pass
           // to EventsService for read/cleared, etc
           trackByID: notification.trackByID,
-          uid: id,
+          uid: _.includes(id, project) ? id : project + "/" + id,
```
spadgett: I think we should always prepend project rather than trying to check if it's in `id`. There's no harm and it makes things simpler.

spadgett:
Thanks @dtaylor113. I understand your point now about the memory leak in NotificationsService. I'm good with this change (with one small comment), and I think we should tackle the other problem separately.

app/scripts/directives/notifications/notificationDrawerWrapper.js
dtaylor113: Ok, fixed -thanks

benjaminapetersen: Agree on keeping it simpler. If the `id` is being created on `line 180`, can we keep all the `id` manipulation logic there?  Manipulating it in a couple spots might open doors to bugs...

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -99,6 +99,7 @@
         _.each(drawer.notificationGroups, function(group) {
           _.remove(group.notifications, { uid: notification.uid, namespace: notification.namespace });
         });
+        delete notificationsMap[$routeParams.project][notification.uid];
```
spadgett: @benjaminapetersen Is this a bug in remove? It seems like the `_.each` above should take care of this

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -184,7 +185,7 @@
           // using uid to match API events and have one filed to pass
           // to EventsService for read/cleared, etc
           trackByID: notification.trackByID,
```
benjaminapetersen: Would it make sense to keep all the `id` generation/manipulation together on line `180`? I'd tend to think it should always be the same value if possible.

benjaminapetersen: Might not be relevant, reviewing again, we have `id`, `trackById`, and `uid`.  Bleh.

app/scripts/directives/notifications/notificationDrawerWrapper.js
dtaylor113: The notificationsMap is different from the group.notifications.  notificationsMap actually generates the group.notifications when switching projects, so it needs to be cleared in the map as well

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -172,7 +173,7 @@
       };
 
       var notificationWatchCallback = function(event, notification) {
-        if(!notification.showInDrawer) {
+        if(!notification.showInDrawer || EventsService.isCleared(notification.id)) {
```
spadgett: Shouldn't we be using `notification.uid` for cleared? We seem to use that in other places.

dtaylor113: The notification in this callback doesn't have the `uid`; the `uid` is set if it gets past this check.

benjaminapetersen: Yeah, that relates to my comment above.  Should always be consistent in the drawer.  Think I settled on `uid` originally because of `apiEvents` first.

spadgett: But I don't think we're not checking the right ID.... If we're clearing `uid` but checking `id`, isn't this check just wrong?

app/scripts/directives/notifications/notificationDrawerWrapper.js
dtaylor113: Either way, I addressed this particular comment. -thanks

app/scripts/directives/notifications/notificationDrawerWrapper.js
dtaylor113: Ok, addressed this logic issue.

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -172,11 +173,13 @@
       };
 
       var notificationWatchCallback = function(event, notification) {
-        if(!notification.showInDrawer) {
+        var project = notification.namespace || $routeParams.project;
+        var id = notification.id ? project + "/" + notification.id : _.uniqueId('notification_') + Date.now();
+
+        if(!notification.showInDrawer || EventsService.isCleared(id)) {
```
spadgett: Thanks @dtaylor113, this looks better

app/scripts/directives/notifications/notificationDrawerWrapper.js
spadgett: I understand, thanks.

spadgett:
[merge]

app/scripts/directives/notifications/notificationDrawerWrapper.js
benjaminapetersen: I would think so as well, looking.

dtaylor113:
Follow up issue https://github.com/openshift/origin-web-common/issues/220

PR #2225, opened 2017-10-06
Bug 1498897 - Add additional quota status check

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1498897


spadgett:
Disregard, this PR is not needed since we have an earlier check for status before the method is called.

PR #2230, opened 2017-10-07
Fix bug deleting bindings for service instance

spadgett:
The logic checking if the binding was already deleted was flipped.


spadgett:
[merge]

PR #2155, opened 2017-09-25
Adopt service catalog API changes

spadgett:
https://trello.com/c/y1lbFrbp

Bumps origin-web-common 0.0.63 -> 0.0.64
Bumps origin-web-catalog 0.0.52 -> 0.0.53

Requires https://github.com/openshift/origin-web-catalog/pull/448
Requires https://github.com/openshift/origin-web-common/pull/199

@jeff-phillips-18 FYI

PR #2233, opened 2017-10-09
Fix preferred version for service bindings

spadgett:
@jeff-phillips-18 FYI, missed this in #2155. PTAL


spadgett:
@jeff-phillips-18 Thanks. I found a few places in views that I need to change anyway for the SerivceBinding resource rename. I have them changed locally.

spadgett:
[merge]

Issue #2227, opened 2017-10-06
Should not be able to delete an already deleted service


spadgett:
We should disable this menu item if the service is already deleted. Note that service instances can potentially stay in deleted state for a while until they can be fully deprovisioned

![openshift web console 2017-10-06 12-42-28](https://user-images.githubusercontent.com/1167259/31288667-e3720a28-aa93-11e7-9525-76d5fb5802d2.png)

PR #2234, opened 2017-10-09
Disable service-instance delete link when already schedualed for deletion

jhadvig:
@spadgett just FYI ... will properly test once the web-catalog is fixed 

Closes https://github.com/openshift/origin-web-console/issues/2227

### DIFF
--- a/app/scripts/controllers/serviceInstance.js
+++ b/app/scripts/controllers/serviceInstance.js
@@ -21,6 +21,9 @@ angular.module('openshiftConsole')
     ];
 
     $scope.deprovision = function() {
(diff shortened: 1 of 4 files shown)


spadgett:
LGTM

jhadvig:
.... adding dist

spadgett:
Hey @jhadvig I wasn't sure if you had been able to verify the fix or if we're still waiting? Let me know.

It does need rebase now.

jhadvig:
@jeff-phillips-18 @spadgett I've tested the change and it works as expected. if the service-instance is deleted then the `Delete` link is disabled.

Screen:
disable

spadgett:
Flake #1685

/retest

PR #2237, opened 2017-10-09
Service instance details configuration and edit

### LINKED ISSUE #2278: SVG images clipping in IE11
I noticed PHP and MySQL, but there may be others.  Make sure to double check post-minified versions, too (e.g., `grunt serve:dist`).

screen shot 2017-10-17 at 10 21 02 am
screen shot 2017-10-17 at 10 20 49 am
screen shot 2017-10-17 at 10 20 40 am

jeff-phillips-18:
Depends on https://github.com/openshift/origin-web-catalog/pull/474

Fixes #2278 with catalog version bump

Bumped origin-web-catalog to 0.0.56

### DIFF
--- a/app/scripts/controllers/serviceInstance.js
+++ b/app/scripts/controllers/serviceInstance.js
@@ -6,13 +6,19 @@ angular.module('openshiftConsole')
                                                      $routeParams,
                                                      APIService,
                                                      BindingService,
+                                                     AuthorizationService,
+                                                     Catalog,
                                                      DataService,
+                                                     Logger,
                                                      ProjectsService,
+                                                     SecretsService,
                                                      ServiceInstancesService) {
     $scope.alerts = {};
     $scope.projectName = $routeParams.project;
     $scope.serviceInstance = null;
     $scope.serviceClass = null;
+    $scope.serviceClasses = null;
+    $scope.editDialogShown = false;
 
     $scope.breadcrumbs = [
       {
@@ -28,9 +34,26 @@ angular.module('openshiftConsole')
       ServiceInstancesService.deprovision($scope.serviceInstance, $scope.bindings);
     };
 
+    $scope.showEditDialog = function() {
+      $scope.editDialogShown = true;
+    };
+
+    $scope.showParameterValues = false;
+
+    $scope.toggleShowParameterValues = function() {
+      $scope.showParameterValues = !$scope.showParameterValues;
+    };
+
+    $scope.closeEditDialog = function() {
+      $scope.editDialogShown = false;
+    };
+
     var watches = [];
+    var secretWatchers = [];
+    var serviceClassPromise;
 
     var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
+    var serviceInstanceReady = $filter('isServiceInstanceReady');
 
     // API Versions
     var serviceBindingsVersion = APIService.getPreferredVersion('servicebindings');
@@ -42,28 +65,76 @@ angular.module('openshiftConsole')
       });
     };
 
-    var serviceClassPromise;
-    var updateServiceClass = function() {
-      // If we've previously loaded the service class or a request is in flight, don't do anything.
-      if ($scope.serviceClass || serviceClassPromise) {
+    var updateParameterData = function() {
+      if (!$scope.serviceInstance || !$scope.parameterSchema) {
         return;
       }
 
-      serviceClassPromise = ServiceInstancesService.fetchServiceClassForInstance($scope.serviceInstance).then(function(serviceClass) {
-        $scope.serviceClass = serviceClass;
-        $scope.displayName = serviceInstanceDisplayName($scope.serviceInstance, serviceClass);
-        updateBreadcrumbs();
-        serviceClassPromise = null;
+      DataService.unwatchAll(secretWatchers);
+      secretWatchers = [];
+
+      $scope.parameterData = {};
+      _.each(_.keys(_.get($scope.parameterSchema, 'properties')), function(key) {
+        $scope.parameterData[key] = $scope.parameterSchema.properties[key].default;
       });
+
+      $scope.parameterData = angular.extend($scope.parameterData, _.get($scope.serviceInstance, 'spec.parameters', {}));
+
+      if (AuthorizationService.canI('secrets', 'get', $scope.projectName)) {
+        _.each(_.get($scope.serviceInstance, 'spec.parametersFrom'), function (parametersSource) {
+          secretWatchers.push(DataService.watchObject("secrets", _.get(parametersSource, 'secretKeyRef.name'), $scope.projectContext, function (secret) {
+            try {
+              _.extend($scope.parameterData, JSON.parse(SecretsService.decodeSecretData(secret.data)[parametersSource.secretKeyRef.key]));
+            } catch (e) {
+              Logger.warn('Unable to load parameters from secret ' + _.get(parametersSource, 'secretKeyRef.name'), e);
+            }
+          }));
+        });
+      }
     };
 
-    var updatePlan = function() {
-      if (ServiceInstancesService.isCurrentPlan($scope.serviceInstance, $scope.plan)) {
+    var updateEditable = function() {
+      if (!$scope.plan || !$scope.serviceClass || !$scope.serviceInstance) {
         return;
       }
 
-      ServiceInstancesService.fetchServicePlanForInstance($scope.serviceInstance).then(function(plan) {
-        $scope.plan = plan;
+      var updateSchema = _.get($scope.plan, 'spec.instanceUpdateParameterSchema');
+      var planUpdatable = (_.size(_.get(updateSchema, 'properties')) > 0) || (_.get($scope.serviceClass, 'spec.planUpdatable') && (_.size($scope.servicePlans) > 1));
+
+      $scope.editAvailable = planUpdatable && serviceInstanceReady($scope.serviceInstance) && !_.get($scope.serviceInstance, 'metadata.deletionTimestamp');
+    };
+
+    var updateParameterSchema = function() {
+      $scope.parameterFormDefinition = angular.copy(_.get($scope.plan, 'spec.externalMetadata.schemas.service_instance.update.openshift_form_definition'));
+      $scope.parameterSchema = _.get($scope.plan, 'spec.instanceCreateParameterSchema');
+    };
+
+    var updateServiceClass = function() {
+      // If we've previously loaded the service class or a request is in flight, don't do anything.
+      if (!$scope.serviceInstance || $scope.serviceClass || serviceClassPromise) {
+        return;
+      }
+
+      serviceClassPromise = ServiceInstancesService.fetchServiceClassForInstance($scope.serviceInstance).then(function (serviceClass) {
+        $scope.serviceClass = serviceClass;
+        $scope.displayName = serviceInstanceDisplayName($scope.serviceInstance, $scope.serviceClass);
+
+        updateBreadcrumbs();
+        serviceClassPromise = null;
+
+        Catalog.getServicePlans().then(function (plans) {
+          plans = plans.by('metadata.name');
+
+          var plansByServiceClassName = Catalog.groupPlansByServiceClassName(plans);
+          $scope.servicePlans = plansByServiceClassName[$scope.serviceClass.metadata.name];
+
+          var servicePlanName = _.get($scope.serviceInstance, 'spec.clusterServicePlanRef.name');
+          $scope.plan = plans[servicePlanName];
+
+          updateParameterSchema();
+          updateParameterData();
+          updateEditable();
+        });
       });
     };
 
@@ -79,7 +150,8 @@ angular.module('openshiftConsole')
       }
 
       updateServiceClass();
-      updatePlan();
+      updateParameterData();
+      updateEditable();
     };
 
     ProjectsService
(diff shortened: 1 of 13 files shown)


app/scripts/controllers/serviceInstance.js
```diff
@@ -24,11 +28,28 @@ angular.module('openshiftConsole')
       ServiceInstancesService.deprovision($scope.serviceInstance);
     };
 
+    $scope.showEditDialog = function() {
+      $scope.editDialogShown = true;
+    };
+
+    $scope.showConfigValues = false;
+
+    $scope.toggleConfigValues = function() {
```
spadgett: nit: I have a small preference for calling them parameter values in the code since it's what they're called in the API.

app/scripts/controllers/serviceInstance.js
```diff
@@ -37,25 +58,68 @@ angular.module('openshiftConsole')
...
-        updateBreadcrumbs();
+      $scope.parameterData = {};
+      _.each(_.keys(_.get($scope.parameterSchema, 'properties')), function(key) {
+        $scope.parameterData[key] = $scope.parameterSchema.properties[key].default;
+      });
+
+      $scope.parameterData = angular.extend($scope.parameterData, _.get($scope.serviceInstance, 'spec.parameters', {}));
+
+      _.each(_.get($scope.serviceInstance, 'spec.parametersFrom'), function(parametersSource) {
+        secretWatcher = DataService.watchObject("secrets", _.get(parametersSource, 'secretKeyRef.name'), $scope.projectContext, function (secret) {
```
spadgett: If there's more than one secret, you'll lose the previous `secretWatcher` reference and leak watches

app/scripts/controllers/serviceInstance.js
```diff
@@ -37,25 +58,68 @@ angular.module('openshiftConsole')
...
+          } catch (e) {
+          }
+        });
       });
     };
 
-    var updatePlan = function() {
-      if (ServiceInstancesService.isCurrentPlan($scope.serviceInstance, $scope.plan)) {
+    var updateParameterSchema = function() {
+      $scope.parameterFormDefinition = angular.copy(_.get($scope.plan, 'spec.externalMetadata.schemas.service_instance.create.openshift_form_definition'));
```
spadgett: For update I think this should be

```
spec.externalMetadata.schemas.service_instance.update.openshift_form_definition
```

app/scripts/controllers/serviceInstance.js
```diff
@@ -37,25 +58,68 @@ angular.module('openshiftConsole')
...
+
+      $scope.parameterSchema = angular.copy(_.get($scope.plan, 'spec.instanceCreateParameterSchema'));
```
spadgett:
```
spec.instanceUpdateParameterSchema
```

app/scripts/controllers/serviceInstance.js
```diff
@@ -37,25 +58,68 @@ angular.module('openshiftConsole')
...
       }
 
-      ServiceInstancesService.fetchServicePlanForInstance($scope.serviceInstance).then(function(plan) {
-        $scope.plan = plan;
+      var serviceClassName = _.get($scope.serviceInstance.spec, 'serviceClassRef.name');
+      $scope.serviceClass = _.get($scope.serviceClasses, [serviceClassName]);
+      $scope.displayName = serviceInstanceDisplayName($scope.serviceInstance, $scope.serviceClass);
+      updateBreadcrumbs();
+
+      Catalog.getServicePlans().then(function(plans) {
```
spadgett: We should be able to only request plans for a specific service class, but OK for now. I worry down the road that this might be a lot of data, though.

app/scripts/controllers/serviceInstance.js
```diff
@@ -94,8 +159,24 @@ angular.module('openshiftConsole')
             };
           });
 
-        $scope.$on('$destroy', function(){
-          DataService.unwatchAll(watches);
+        DataService.list(serviceClassesVersion, {}, function(serviceClasses) {
```
spadgett: We should be able to get just the one service class for this instance. (You might just need to rebase to pick up my changes for this.)

app/scripts/directives/serviceBinding.js
```diff
@@ -17,22 +20,62 @@
...
+          } catch (e) {
+          }
+        });
+      });
+    };
+
+    var updateParameterSchema = function() {
+      var resource = APIService.getPreferredVersion('clusterserviceplans');
+      DataService.get(resource, _.get(ctrl.serviceInstance, 'spec.servicePlanRef.name'), context).then(function(servicePlan) {
+        // TODO: Is there a form definition for this?
```
spadgett: > Is there a form definition for this?

```
spec.externalMetadata.schemas.service_binding.create.openshift_form_definition
```

app/scripts/directives/serviceBinding.js
```diff
@@ -17,22 +20,62 @@
...
+        ctrl.bindParameterSchema = angular.copy(_.get(servicePlan, 'spec.serviceInstanceCredentialCreateParameterSchema'));
```
spadgett: Will eventually be renamed to `serviceBindingCreateParameterSchema`

app/scripts/directives/serviceBinding.js
```diff
@@ -17,22 +20,62 @@
...
+
+    var context = {
+      namespace: ctrl.namespace
+    };
+
+    var updateParameterData = function() {
+      ctrl.parameterData = angular.copy(_.get(ctrl.binding, 'spec.parameters', {}));
+
+      _.each(_.get(ctrl.binding, 'spec.parametersFrom'), function(parametersSource) {
+        DataService.get("secrets", _.get(parametersSource, 'secretKeyRef.name'), context).then(function(secret) {
```
spadgett: Need to check if this user can get secrets. Project viewers can't, but they will be able to see the instance.

app/views/browse/service-instance.html
```diff
@@ -93,6 +134,17 @@ <h1 class="contains-actions">
             </div><!-- /col-* -->
           </div>
         </div>
+        <overlay-panel show-panel="editDialogShown" handle-close="closeEditDialog">
+          <update-service
+            project="project"
+            display-name="serviceInstance | serviceInstanceDisplayName:serviceClasses"
```
spadgett: You'll need to rebase. This now takes in one service class instead of a map.

app/views/directives/_service-binding.html
```diff
@@ -48,4 +48,18 @@
       View Secret
     </a>
   </div>
+  <div class="service-binding-parameters" ng-if="$ctrl.bindParameterSchema.properties && (true || !($ctrl.binding | isBindingFailed) && ($ctrl.binding | isBindingReady))">
```
spadgett: Is there a reason not to show parameter for failed bindings? It might help the user debug.

app/views/directives/_service-binding.html
```diff
@@ -48,4 +48,18 @@
...
+    <span class="parameters-heading">PARAMETERS</span>
```
spadgett:
```html
<span class="parameters-heading">Parameters</span>
```

and use `text-transform` to make it uppercase in CSS

app/views/directives/_service-binding.html
```diff
@@ -48,4 +48,18 @@
...
+    <a href="" ng-click="$ctrl.toggleShowParameterValues()">{{$ctrl.showParameterValues ? 'Hide Values' : 'Reveal Values'}}</a>
```
spadgett:
```
role="button"
```

app/views/directives/_service-binding.html
```diff
@@ -48,4 +48,18 @@
...
+    <a ng-if="!$ctrl.binding.metadata.deletionTimestamp" href="" ng-click="$ctrl.editParameters()">Edit</a>
```
spadgett:
```
role="button"
```

jeff-phillips-18:
Believe I have addressed @spadgett comments.

Have not been able to test with multiple plans and/or editable plans.

spadgett:
@jmontleon FYI, this is the plan and parameter update PR. We saw some problems (in our code) when working against your ASB changes that we're debugging now.

jeff-phillips-18:
Fixed updating the plan editability while being updated. No known issues atm, but have not tested updating the service instance's parameters with an instanceUpdateParameterSchema.

app/scripts/controllers/serviceInstance.js
```diff
@@ -42,28 +65,76 @@ angular.module('openshiftConsole')
...
+      secretWatchers = [];
+
+      $scope.parameterData = {};
+      _.each(_.keys(_.get($scope.parameterSchema, 'properties')), function(key) {
+        $scope.parameterData[key] = $scope.parameterSchema.properties[key].default;
+      });
+
+      $scope.parameterData = angular.extend($scope.parameterData, _.get($scope.serviceInstance, 'spec.parameters', {}));
+
+      if (AuthorizationService.canI('secrets', 'get', $scope.projectName)) {
```
spadgett: This means that you will not be able to see that there's a parameter at all when it comes from a secret and you can't get the secret, correct?

We could possibly fill in the redacted values using `status.externalProperties.parameters` when users can't list secrets (without letting them reveal). I'm OK handling this as a follow on, however.

spadgett:
Checkboxes look a little odd to me in read only mode:

openshift web console 2017-10-18 10-24-20

jeff-phillips-18:
The checkboxes issue is fixed in the catalog-parameters component

app/scripts/controllers/serviceInstance.js
jeff-phillips-18: @spadgett  What should we do when the user does reveal and we have redacted values for some of the parameters? Continue to show ***** for those parameters?

spadgett: Maybe remove the "reveal" action?

spadgett: I think we're close enough here it should be a follow-on anyway.

app/styles/_core.less
```diff
@@ -105,6 +105,32 @@
     border-bottom: 0;
     padding-bottom: 0;
   }
+
+  .config-parameters-form {
```
rhamilto: Nit:  alpha of this rule and the rules nested within

app/views/browse/service-instance.html
```diff
@@ -67,26 +81,67 @@ <h1 class="contains-actions">
                           <dd ng-if-end>
                             {{serviceInstance | serviceInstanceConditionMessage}}
                           </dd>
-                          <dt ng-if-start="serviceClass.spec.description || serviceClass.spec.externalMetadata.longDescription">Description:</dt>
-                          <dd ng-if-end>
+                        </dl>
+                        <div class="hidden-lg">
+                          <h3 ng-if-start="serviceClass.spec.description || serviceClass.spec.externalMetadata.longDescription">Description</h3>
+                          <dl ng-if-end class="dl-horizontal left">
```
rhamilto: A dl should only contain [dt, dd, or div/script/template elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl).  So either don't use a dl or change the p elements to divs and add a margin using a helper class (e.g., mar-bottom-md)?  I think I prefer the former as that maintains the semantic value of the p.

app/views/browse/service-instance.html
```diff
@@ -67,26 +81,67 @@ <h1 class="contains-actions">
...
+                              read-only="true">
+                            </catalog-parameters>
+                          </form>
+                        </div>
                       </div>
-                      <div class="col-lg-6 mar-bottom-xl">
+                      <div class="col-lg-6">
+                        <div class="hidden-xs hidden-sm hidden-md">
+                          <h3 ng-if-start="serviceClass.spec.description || serviceClass.spec.externalMetadata.longDescription">Description</h3>
+                          <dl ng-if-end class="dl-horizontal left">
```
rhamilto: Same dl issue as before.

app/views/browse/service-instance.html
rhamilto: No need for the dl.  Just remove it and render the p as individual elements.

app/styles/_core.less
rhamilto: Generally ignore the non-alpha characters as I think most folks don't think about 'em when they're scanning the code.  The overall goal is to make the Less more human-readable.  You would not believe the number of times I've seen rules declared multiple times or even the same declaration twice within the same rule.  Either way this rule should come first in this block.  :)

jeff-phillips-18:
Addressed @rhamilto comments. Removed WIP status 🙌  🤘

app/scripts/controllers/serviceInstance.js
```diff
@@ -42,28 +65,76 @@ angular.module('openshiftConsole')
...
         updateBreadcrumbs();
         serviceClassPromise = null;
-      });
-    };
 
-    var updatePlan = function() {
-      if (ServiceInstancesService.isCurrentPlan($scope.serviceInstance, $scope.plan)) {
-        return;
-      }
+        Catalog.getServicePlans().then(function (plans) {
```
spadgett: When we fix the removed plans, let's use a field filter on list to get just the plans for this service class and filter removed plans client side (follow on)

app/scripts/directives/serviceBinding.js
```diff
@@ -20,23 +23,62 @@
...
+      namespace: ctrl.namespace
+    };
+
+    var updateParameterData = function() {
+      ctrl.parameterData = angular.copy(_.get(ctrl.binding, 'spec.parameters', {}));
+      if (AuthorizationService.canI('secrets', 'get', ctrl.namespace)) {
+        _.each(_.get(ctrl.binding, 'spec.parametersFrom'), function (parametersSource) {
+          DataService.get('secrets', _.get(parametersSource, 'secretKeyRef.name'), context).then(function (secret) {
+            try {
+              _.extend(ctrl.parameterData, SecretsService.decodeSecretData(secret.data).parameters);
```
spadgett: This is not always `parameters`. It is the `secretKeyRef.key`

spadgett: Surprised you don't need a `JSON.parse` here

spadgett:
Looks like a dist mismatch

app/scripts/directives/serviceBinding.js
jeff-phillips-18: Yes. Thanks, I do need a JSON.parse here.

PR #2238, opened 2017-10-09
Bug 1505782 - Environment From Fix Drag & Order Display

cdcabrera:
Fix for making sure the drag and drop/order handles are hidden within a read only display. Related to #2182 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1505782

### DIFF
--- a/app/scripts/directives/editEnvironmentFrom.js
+++ b/app/scripts/directives/editEnvironmentFrom.js
@@ -11,7 +11,8 @@
       addRowLink: '@',              // creates a link to "add row" and sets its text label
       entries: '=',                 // an array of objects containing configmaps and secrets
       envFromSelectorOptions: '<',  // dropdown selector options, an array of objects
-      selectorPlaceholder: '@'      // placeholder copy for dropdown selector
+      selectorPlaceholder: '@',     // placeholder copy for dropdown selector
+      isReadonly: '<?'              // display as read only values
     },
     templateUrl: 'views/directives/edit-environment-from.html'
   });
@@ -20,7 +21,6 @@
                                $filter,
                                utils) {
     var ctrl = this;
-
     var canI = $filter('canI');
     var humanizeKind = $filter('humanizeKind');
     var uniqueId = _.uniqueId();
@@ -51,6 +51,7 @@
       }
 
       ctrl.envFromEntries.splice(start, deleteCount);
+
       if(!ctrl.envFromEntries.length && ctrl.addRowLink) {
         addEntry(ctrl.envFromEntries);
       }
@@ -59,11 +60,18 @@
       ctrl.editEnvironmentFromForm.$setDirty();
     };
 
+    ctrl.hasOptions = function() {
+      return !_.isEmpty(ctrl.envFromSelectorOptions);
+    };
+
+    ctrl.hasEntries = function() {
+      return _.some(ctrl.entries, function(entry) {
+        return _.get(entry, 'configMapRef.name') || _.get(entry, 'secretRef.name');
+      });
+    };
+
     ctrl.isEnvFromReadonly = function(entry) {
-      return ctrl.isReadonlyAny ||
-        entry.isReadonlyValue === true ||
-        ((entry.secretRef || entry.configMapRef) && !entry.selectedEnvFrom) ||
-        _.isEmpty(ctrl.envFromSelectorOptions);
+      return ctrl.isReadonly === true || entry && entry.isReadonly === true;
     };
 
     ctrl.groupByKind = function(object) {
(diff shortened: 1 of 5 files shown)


app/views/directives/edit-environment-from.html
```diff
@@ -52,7 +52,7 @@
         </div>
       </div>
 
-      <div ng-if="!$ctrl.isReadonlyAny && !entry.isReadonlyValue" class="environment-from-editor-button">
+      <div ng-if="$ctrl.isEnvFromSelectable()" class="environment-from-editor-button">
```
cdcabrera: Going to look at place the ```isReadonlyAny``` and ```isReadonlyValue``` back in...

cdcabrera: Separated out the options length check and encased both ```isReadonlyAny``` and ```isReadonlyValue``` within the ```isEnvFromReadonly``` method

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -50,11 +50,12 @@
...
+    ctrl.isEnvFromSelectable = function() {
+      return !_.isEmpty(ctrl.envFromSelectorOptions);
+    };
+
     ctrl.isEnvFromReadonly = function(entry) {
-      return ctrl.isReadonlyAny ||
-        entry.isReadonlyValue === true ||
-        ((entry.secretRef || entry.configMapRef) && !entry.selectedEnvFrom) ||
-        _.isEmpty(ctrl.envFromSelectorOptions);
+      return ctrl.isReadonlyAny || entry.isReadonlyValue === true || ((entry.secretRef || entry.configMapRef) && !entry.selectedEnvFrom);
```
spadgett: We could simplify this a lot by just saying the whole thing is either read-only or it isn't. We currently don't have a need to make individual values read-only.

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -50,11 +50,12 @@
       ctrl.editEnvironmentFromForm.$setDirty();
     };
 
+    ctrl.isEnvFromSelectable = function() {
```
spadgett: I suggest `hasOptions`

app/scripts/directives/editEnvironmentFrom.js
cdcabrera: That works, removing the "isReadonlyAny" and tweaking some of the related logic

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -50,11 +50,12 @@
...
+    ctrl.hasOptions = function() {
+      return !_.isEmpty(ctrl.envFromSelectorOptions);
+    };
+
     ctrl.isEnvFromReadonly = function(entry) {
-      return ctrl.isReadonlyAny ||
-        entry.isReadonlyValue === true ||
-        ((entry.secretRef || entry.configMapRef) && !entry.selectedEnvFrom) ||
-        _.isEmpty(ctrl.envFromSelectorOptions);
+      return entry.isReadonlyValue === true || ((entry.secretRef || entry.configMapRef) && !entry.selectedEnvFrom);
```
spadgett: I was proposing getting rid of `isReadonlyValue` and keeping only `isReadonlyAny` :)

I don't think we have a need for only certain values being read only. Either the entire list is read only or it isn't

cdcabrera: Hahahaha whoops, I'll go the other direction

cdcabrera: After trying to pick this apart looks like the piece around Secrets or Config maps and duplicate entries falls into a strange area. Specifically around when a user edits the Yaml and purposefully enters duplicates. 

With the current version these duplicates become read only because they lack the additional property around "selectedEnvFrom" due to the way lodash and find is being used... this is unintended but has kinda given us an inadvertent solution to the duplicate issue.

spadgett: I'm not sure I understand why we made duplicates read only, or missing secrets for that matter. Why prevent the  user from editing those to fix the problem?

I think it should be a warning if it's there twice, but we shouldn't block it since it's not a validation error.

cdcabrera: Concur on the blocking at all, that shouldn't happen, and doesn't... kinda. The current implementation doesn't block a user from deleting or rearranging an item, just creating an additional duplicate/triplicate/quad/etc entry through the interface. 

Can and will look at using lodash filter instead of find to handle the checks, this may correct the problem, keep you posted

cdcabrera: @spadgett looks like the aiming towards ```_.filter``` achieved the desired result, no more entries being read only if they're duplicates. Leveraged the same concept to apply a temporary placeholder ```duplicate``` class around the individual UI-Selects

@spadgett  & @benjaminapetersen 

On the Readonly state as an all or nothing piece... can push them up, just need a confirmation for the conditions on my local, which so far encompass... 
- If options aren't populated, everything is Read Only
- If either the Secrets OR Config Maps "Can I" checks fail, everything is Read Only
- And finally, removed the block (similar to below) for binding attribute pass-through for Read Only 
  ```js
    if('isReadonly' in $attrs) {
      ctrl.isReadonlyAny = true;
    }
  ```

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -99,8 +100,23 @@
       });
     };
 
-    var updateEnvFromEntries = function(entries) {
-      ctrl.envFromEntries = entries || [];
+    var getReferenceValues = function(option) {
```
benjaminapetersen: You might be able to nix the switch statements & eliminate some duplication with:

```JavaScript
var getReferenceValues = function(option) {
      var kindRef = _.camelCase(option.kind) + 'Ref';
      return _.filter(ctrl.envFromEntries, [kindRef, 'name', option.metadata.name]);      
    };
```
(untested)

benjaminapetersen: Prob similar could be done to trim down [lines 76-91](https://github.com/openshift/origin-web-console/pull/2238/files#diff-a09b8976dc42ea9f1ff076ac68341ba9R78)

cdcabrera: @benjaminapetersen looks like coincidentally wiped some of the duplication out. Good catch on the switch

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -50,18 +51,22 @@
       ctrl.editEnvironmentFromForm.$setDirty();
     };
 
+    ctrl.hasOptions = function() {
+      return !_.isEmpty(ctrl.envFromSelectorOptions);
+    };
+
+    ctrl.hasEntries = function() {
+      return angular.toJson(ctrl.entries) !== '[{}]' && ctrl.entries && ctrl.entries.length >= 1;
```
spadgett:
```js
return _.some(ctrl.entries, function(entry) {
   return _.get(entry, 'configMapRef.name') || _.get(entry, 'secretRef.name');
});
```

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -109,56 +119,28 @@
...
-      switch(option.kind) {
-        case 'ConfigMap':
-          referenceValue = _.find(ctrl.envFromEntries, {configMapRef: {name: option.metadata.name}});
-          break;
-        case 'Secret':
-          referenceValue = _.find(ctrl.envFromEntries, {secretRef: {name: option.metadata.name}});
-          break;
-      }
+      _.each(ctrl.envFromSelectorOptions, function(option) {
+        _.each(getReferenceValues(option), function(val, i) {
```
spadgett: You should create two maps of options `secretsByName` and `configMapsByName` when options change. Then you can index directly into the map using `configMapsByName[option.configMapRef.name]`, etc.

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -107,66 +115,55 @@
...
+      var entriesMap = {};
+
+      ctrl.envFromEntries = ctrl.entries || [];
 
       if(!ctrl.envFromEntries.length) {
         addEntry(ctrl.envFromEntries);
       }
+      _.each(ctrl.envFromSelectorOptions, function(option) {
+        optionsByName[option.metadata.name] = option;
+      });
```
spadgett:
```js
var secretsByName = {};
var configMapsByName = {};
_.each(ctrl.envFromSelectorOptions, function(option) {
  switch(option.kind) {
    case 'Secret':
      secretsByName[option.metadata.name] = option;
      break;
    // ...
  }
});
```

sg00dwin: Select with `.has-warning` and `.help-block`
screen shot 2017-10-20 at 3 37 30 pm

spadgett:
@cdcabrera The form is showing up as read only for me when it should be editable:

openshift web console 2017-10-24 14-08-07

spadgett:
I can no longer reproduce the readonly problem.

spadgett:
/lgtm
/hold

Needs #2367 to go in first or merge will fail

Issue #2226, opened 2017-10-06
Should not be able to add the same secret or config map to an app twice


spadgett:
It's doesn't make sense to add the same secret twice.

cc @cdcabrera

jhadvig:
@spadgett do you mean the in the DC's Environment Variables section ? Or BC's Build Secrets section ?

jhadvig:
@jeff-phillips-18 mind if I take this one ?

jeff-phillips-18:
go for it @jhadvig

spadgett:
@jhadvig I mean on the secret or config map page. There's an "Add to Application" button.

jhadvig:
@spadgett what if the application contains two containers, where one of them already contains the secret/configMap... shouldn't we show the application if at least one of its containers doesn't contain the secret/configMap ?

spadgett:
> shouldn't we show the application if at least one of its containers doesn't contain the secret/configMap ?

Good question. I think so.

PR #2249, opened 2017-10-11
Should not be able to add the same secret or config map to an application twice

jhadvig:
The `Add to Application` modal will only show applications which DC's won't contain the secret/configMap that user wants to add. In case one of the containers won't contain the secret/configMap the application will be shown in the dropbox, but the ref will be added only to the containers that don't contain the it.

Unfortunately if, in case of multiple container DC, user chooses to specify a specific container, the checkbox of the container that already contains the ref wont be pre-selected, cause that could mislead the user that by unchecking the checkbox he could remove the ref from that specific container. Other option would be to pre-select the checkbox and also disable it, but that would require add the disable logic to the `select-container` directive. Wasn't sure if we need that ATM...

@jeff-phillips-18 @spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/2226

### DIFF
--- a/app/scripts/directives/addConfigToApplication.js
+++ b/app/scripts/directives/addConfigToApplication.js
@@ -26,6 +26,20 @@
     var ctrl = this;
     var humanizeKind = $filter('humanizeKind');
 
+    var conatinerHasRef = function(container) {
+      var addRefName = ctrl.apiObject.metadata.name;
+      if (ctrl.apiObject.kind === "ConfigMap") {
+        return _.some(container.envFrom, {configMapRef: {name: addRefName}});
+      } else {
+        return _.some(container.envFrom, {secretRef: {name: addRefName}});
+      }
+    };
+
+    ctrl.checkApplicationContainersRefs = function(application) {
+      var containers = _.get(application, 'spec.template.spec.containers');
+      ctrl.canAddRefToApplication = !_.every(containers, conatinerHasRef);
+    };
+
     var getApplications = function() {
       var context = {
         namespace: ctrl.project.metadata.name
@@ -91,7 +105,7 @@
 
         // For each container, add the new volume mount.
         _.each(podTemplate.spec.containers, function(container) {
-          if (isContainerSelected(container)) {
+          if (isContainerSelected(container) && !conatinerHasRef(container)) {
             container.envFrom = container.envFrom || [];
             container.envFrom.push(newEnvFrom);
           }
--- a/app/views/directives/add-config-to-application.html
+++ b/app/views/directives/add-config-to-application.html
@@ -6,9 +6,9 @@
     <form name="addToApplicationForm" novalidate>
       <fieldset ng-disabled="disableInputs">
         <legend>Add this {{ctrl.apiObject.kind | humanizeKind}} to application:</legend>
-        <div class="form-group">
+        <div class="form-group"  ng-class="{'has-error' : ctrl.addType === 'env' && ctrl.application && !ctrl.canAddRefToApplication}">
           <div class="application-select">
-            <ui-select id="application" ng-model="ctrl.application" required="true" ng-disabled="ctrl.disableInputs">
+            <ui-select id="application" ng-model="ctrl.application" on-select="ctrl.checkApplicationContainersRefs($item)" required="true" ng-disabled="ctrl.disableInputs">
               <ui-select-match placeholder="{{ctrl.applications.length ? 'Select an application' : 'There are no applications in this project'}}">
               <span>
                 {{$select.selected.metadata.name}}
@@ -23,6 +23,11 @@
             </ui-select>
           </div>
         </div>
+        <div class="has-error" ng-if="ctrl.addType === 'env' && ctrl.application && !ctrl.canAddRefToApplication">
+          <span class="help-block">
+            The {{ctrl.apiObject.kind | humanizeKind}} has already been added to this application.
+          </span>
+        </div>
         <legend>Add {{ctrl.apiObject.kind | humanizeKind}} as:</legend>
         <div class="form-group">
           <div class="radio">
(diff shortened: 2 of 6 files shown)


app/scripts/directives/addConfigToApplication.js
```diff
@@ -26,12 +26,35 @@
     var ctrl = this;
     var humanizeKind = $filter('humanizeKind');
 
+    var checkContainerForRef = function(container) {
```
benjaminapetersen: Tested, I think we might need an additional message option:

screen shot 2017-10-11 at 11 31 37 am

It says "There are no applications in this project", but that isn't true in this case.  I've just already added this secret to both of the apps in my project.

jhadvig: @benjaminapetersen what about "There is not any application with containers to add secret to"

spadgett: We might want to show all apps in the list, but show an error with a disabled save button when the user tries to pick one that already has the secret. Otherwise it might seem like a bug that their app isn't showing up, and it's not obvious why.

jhadvig:
@benjaminapetersen as suggested by @spadgett I've reworked the logic so the selectbox will list all the applications but we will throw an warning and disable the **Save** button if the ref is already present in the application containers.

Screen:
twice

PTAL

benjaminapetersen:
Nice, I'm good with that!

app/scripts/directives/addConfigToApplication.js
```diff
@@ -25,6 +25,26 @@
   function AddConfigToApplication($filter, $scope, APIService, ApplicationsService, DataService, Navigate, NotificationsService, StorageService) {
     var ctrl = this;
     var humanizeKind = $filter('humanizeKind');
+    ctrl.canAddRefToApplication = false;
+
+    var checkContainerForRef = function(container) {
+      var addRefName = ctrl.apiObject.metadata.name;
+      if (ctrl.apiObject.kind === "ConfigMap") {
+        return _.find(container.envFrom, {configMapRef: {name: addRefName}});
```
spadgett: You can use `_.some` instead of `_.find` if you are just checking if it exists.

app/scripts/directives/addConfigToApplication.js
```diff
@@ -25,6 +25,26 @@
...
+      } else {
+        return _.find(container.envFrom, {secretRef: {name: addRefName}});
```
spadgett: _.some

app/scripts/directives/addConfigToApplication.js
```diff
@@ -25,6 +25,26 @@
   function AddConfigToApplication($filter, $scope, APIService, ApplicationsService, DataService, Navigate, NotificationsService, StorageService) {
     var ctrl = this;
     var humanizeKind = $filter('humanizeKind');
+    ctrl.canAddRefToApplication = false;
+
+    var checkContainerForRef = function(container) {
```
spadgett: Suggest `conatinerHasRef`

app/scripts/directives/addConfigToApplication.js
```diff
@@ -25,6 +25,26 @@
...
+    };
+
+    ctrl.checkApplicationContainersRefs = function(application) {
+      ctrl.canAddRefToApplication = false;
+      _.each(_.get(application, 'spec.template.spec.containers'), function(container) {
+        if (!checkContainerForRef(container)) {
+          ctrl.canAddRefToApplication = true;
+          return false;
+        }
+      });
```
spadgett: I think this could be shortened to.

```js
var containers = _.get(application, 'spec.template.spec.containers');
ctrl.canAddRefToApplication = !_.every(containers, checkContainerForRef);
```

app/views/directives/add-config-to-application.html
```diff
@@ -5,10 +5,14 @@
   <div class="dialog-body">
     <form name="addToApplicationForm" novalidate>
       <fieldset ng-disabled="disableInputs">
+        <div class="alert alert-warning" ng-if="ctrl.addType === 'env' && ctrl.application && !ctrl.canAddRefToApplication">
+          <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
+          <strong>Containers of the selected application already contain this {{ctrl.apiObject.kind | humanizeKind}}.</strong>
```
spadgett: Maybe just

```html
<strong>The {{ctrl.apiObject.kind | humanizeKind}} has already been added to this application.</strong>
```

spadgett: Suggest using the help-block / has-error style (instead of an alert) and moving it below the select.

jhadvig:
@spadgett I've update the PR based on your comments :)
PTAL

spadgett:
Looks good. A couple small comments:

openshift web console 2017-10-16 08-41-56

1. There should be more margin between the error message and the "Add secret as:" label below.
2. The select should be highlighted red as well. If you have a `has-error` class on the form group, the entire select is highlighted. See https://github.com/openshift/origin-web-catalog/blob/master/src/components/select-project/selectProject.html#L2
3. We should probably use a consistent style for the "Some keys for..." warning as well. It looks odd to have two error styles. I'd suggest switching the env warning to the `help-block` / `has-warning` style.

jhadvig:
@spadgett comments addressed

screen:
twice4

PTAL

app/views/directives/add-config-to-application.html
```diff
@@ -108,7 +111,7 @@
                   class="btn btn-primary"
                   ng-class="{'dialog-btn': isDialog}"
                   ng-click="ctrl.addToApplication()"
-                  ng-disabled="ctrl.addType === 'volume' && addToApplicationForm.$invalid || !ctrl.application"
+                  ng-disabled="(ctrl.addType === 'volume' && addToApplicationForm.$invalid) || (ctrl.addType === 'env' && !ctrl.canAddRefToApplication)"
```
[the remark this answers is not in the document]
jeff-phillips-18: I think you still need to ensure an application is selected either way.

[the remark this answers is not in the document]
jhadvig: @jeff-phillips-18 I've update the mount path input with the check if the application was selected as @spadgett suggested. This will force user to pick the application in order to make the form valid. Added separate commit so the changes are obvious.
PTAL

spadgett:
@jhadvig Looks like a dist mismatch

jhadvig:
@spadgett will rebase and regenerate dist

PR #2252, opened 2017-10-11
Remove nameGenerator service

benjaminapetersen:
We no longer use this service:

```
bpeterse at dhcp129-24 in ~/github/bap-origin-web-console on master [$]
$ grep -ri 'NameGenerator' app/
app//index.html:        <script src="scripts/services/nameGenerator.js"></script>
app/scripts/services/nameGenerator.js:  .service("NameGenerator", function(){

bpeterse at dhcp129-24 in ~/github/bap-origin-web-console on master [$]
$ grep -ri suggestFromSourceUrl app/
app/scripts/services/nameGenerator.js:      suggestFromSourceUrl: function(sourceUrl){
```


spadgett:
I think it would be to actually use it like we do for the catalog repo.

https://github.com/openshift/origin-web-catalog/blob/master/src/components/create-from-builder/create-from-builder.controller.ts#L141-L157

spadgett:
I have been meaning to come back and do that so we're consistent.

benjaminapetersen:
Not sure I follow ~ meaning we do want to keep it?

PR #2256, opened 2017-10-12
Bug 1501015 - Don't load service class more than once

### LINKED ISSUE #2255: Provisioned service name added twice to breadcrumbs
openshift web console 2017-10-11 17-38-27

spadgett:
Don't try to load the service class a second time if a request is
already in flight on the service instance page. Fixes a problem where
the breadcrumb with the service class name appears more than once.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1501015
Fixes #2255


spadgett:
[merge]

PR #2262, opened 2017-10-13
Update PHP and MySQL icons

spadgett:
Updated icons from @jennyhaines

openshift web console 2017-10-13 09-09-27

@jeff-phillips-18


jennyhaines:
Looks good @spadgett !

PR #2263, opened 2017-10-13
Revert "add "overwrite" option to attachPVC view"

spadgett:
Reverts openshift/origin-web-console#2176

The change has introduced some bugs. Reverting for now. I've reopened the original issue.


spadgett:
/assign @jeff-phillips-18 

PTAL, rolling back one fix because we found some bugs.

PR #2266, opened 2017-10-13
Special case some humanized reasons codes

spadgett:
* BackOff -> Back-off (instead of "Back off")
* OAuth -> OAuth (instead of "O auth")

Used in the events page, notification drawer, and for pod status, which
shows failed container reason codes.

@ncameronbritt Fixing your "back off" bug


spadgett:
openshift web console 2017-10-13 12-47-15

spadgett:
@jeff-phillips-18 Can I get the lgtm comment? :)

PR #2268, opened 2017-10-13
Add junit reporter for unit tests

benjaminapetersen:
@spadgett @stevekuznetsov 

Adds the junit reporter for our unit tests.  Will work on adding this for e2e next (seems like there is some tinkering that needs to be done).

Related to https://github.com/openshift/aos-cd-jobs/pull/730.
- generates the directory & files expected for the above PR to be meaningful.


benjaminapetersen:
test stuck?

/test

PR #2277, opened 2017-10-17
[WIP] e2e video reporter

benjaminapetersen:
Depends on #2083 

- Adds `protractor-video-reporter`.
- Consolidates all the reports under `test/reports` across unit & e2e tests

The video reporter plugin is a little quirky to configure. For example, locally I cannot get it to record my test browser yet, it instead records arbitrary browser windows.  That said, running in `travis`, it may be pretty helpful.  We shouldn't need a video recording of a local running e2e suite anyway, we can just watch the test directly.

@spadgett @jwforres


benjaminapetersen:
@stevekuznetsov thoughts on this?  How often do we clean up old tests?  If we had a video/animated gif in our e2e reports, concerns about file size?

stevekuznetsov:
How much data are we talking here

benjaminapetersen:
Only 7-10 gigs.  Per spec test.  Is fine, totally fine :smile:
Dunno yet, I'm experimenting with what this generates... video vs animated gif, etc.

stevekuznetsov:
OK, depending on the codec i would really not expect the video to take a lot of space, the UI is fairly static frame-to-frame. But if it's more than like 50MB per test we should be careful

Issue #1219, opened 2017-02-03
On membership, in edit mode, on the edge between tablet & larger screens, there is a potential layout wobble


benjaminapetersen:
With long or multiple roles, between ~765px and ~820px screen width there is the potential for this:
![screen shot 2017-02-03 at 11 35 03 am](https://cloud.githubusercontent.com/assets/280512/22599425/eaa6eaa2-ea04-11e6-9e83-c329be379250.png)

spadgett:
Honestly I wonder if Bootstrap grid or even a table isn't better than flex for layouts like this.

benjaminapetersen:
Agree, might be worth a restructure after all.

benjaminapetersen:
Ideas:
- Could drop the 3rd column if we had a `+` action in the roles list that would open a dialog
- Could also eliminate edit/non-edit mode with the above `+` actions
- Or could list the role chips under the name, which would leave plenty of space (2 column again)

sg00dwin:
There are various width/states, both edit and non-edit mode has content that is breaking the layout.

Adding examples:
<img width="322" alt="screen shot 2017-06-27 at 12 28 59 pm" src="https://user-images.githubusercontent.com/1874151/27600907-6dbb305e-5b3b-11e7-8ad5-85473666c8cd.png">
<img width="628" alt="screen shot 2017-06-27 at 12 28 11 pm" src="https://user-images.githubusercontent.com/1874151/27600903-6bcb199e-5b3b-11e7-89e3-24cd476234c2.png">

<img width="635" alt="screen shot 2017-06-27 at 1 17 14 pm" src="https://user-images.githubusercontent.com/1874151/27600914-7273497e-5b3b-11e7-8b2b-fbdce3f385ae.png">
<img width="852" alt="screen shot 2017-06-27 at 12 31 03 pm" src="https://user-images.githubusercontent.com/1874151/27600915-72784b7c-5b3b-11e7-8f6b-d36353c5038e.png">

sg00dwin:
Posting some exploratory wireframes showing members and roles using PF list-view expand rows to address scalability and be consistent with other views.

View by member
![membership by user01](https://user-images.githubusercontent.com/1874151/28336438-ef7c05e2-6bcf-11e7-9211-8741547d0390.png)

View by member expanded
![membership by user expanded01](https://user-images.githubusercontent.com/1874151/28336439-ef7ea14e-6bcf-11e7-961e-d3b58e70e9b3.png)

[...]

PR #2283, opened 2017-10-17
Membership updates to correct broken layouts when multiple roles assigned

sg00dwin:
The current layout breaks are occurring between 768-1200px when there are 3+ roles assigned. And are especially severe in edit mode . 

To fix these, roles are stacked and allowed to flex-wrap. And in edit mode from 768-991px, the select field is positioned at the bottom of the stack.  Also more space is given to the `col-name` column when the vertical nav is collapsed so that long user names don't wrap.

Before and after screens below.

Fixes https://github.com/openshift/origin-web-console/issues/1219

**Before**

**After**
6b

_____

**Before**

**After**
3b

_____

**Before**

**After**
7b

_____

**Before**

**After**
5b

### DIFF
--- a/app/styles/_membership.less
+++ b/app/styles/_membership.less
@@ -1,70 +1,100 @@
 .membership {
-  // TODO: make generic?
-  h1 {
-    .learn-more-inline {
-      white-space: nowrap;
-      margin-right: 10px;
-      margin-left: 0px;
-      display: inline-block;
+  .action-set {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    width: 100%;
+    @media(min-width: @screen-sm-min) {
+      flex-wrap: nowrap;
+    }
+    @media(min-width: @screen-md-min) {
+      flex-direction: row;
     }
   }
   .content-pane {
     width: 100%;
-    .col-add-role {
-      padding: 10px 0;
-      min-width: 150px;
-    }
     .add-role-to {
       margin-left: -1px;
     }
-    .item-row,
+    .col-add-role {
+      width: 100%;
+      @media(min-width: @screen-xs-min) {
+        min-width: 245px;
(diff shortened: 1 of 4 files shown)


app/styles/_membership.less
```diff
@@ -8,12 +8,31 @@
...
     }
+    .col-roles {
+      display: flex;
+      flex-wrap: wrap;
+      padding-bottom: 5px;
+      @media(min-width: @screen-md-min) {
+        padding: 0;
+      }
+    }
     .add-role-to {
```
rhamilto: Nit:  alpha

app/styles/_membership.less
```diff
@@ -25,27 +44,28 @@
...
+      padding: 10px 10px 10px 5px;
     }
-    .col-heading {
-      margin-bottom: 20px;
-      .col-add-role {
-        h3 {
-          display: none;
-        }
-      }
+    .col-heading .col-add-role h3 {
```
rhamilto: Nit: alpha

app/styles/_membership.less
```diff
@@ -25,27 +44,28 @@
...
+      display: none;
     }
     .col-name {
+      @media(min-width: @screen-xs-min) {
+        max-width: 30%;
+        padding: 0 5px 0 0;
+        input {
```
rhamilto: Nit: alpha

sg00dwin:
@rhamilto ready for second look. 
overall code cleanup and rebased

spadgett:
@sg00dwin should be ready, except needs rebase

sg00dwin:
@spadgett rebased and ready

spadgett:

spadgett:
I cherry-picked this commit and rebuilt the dist in #2344

Issue #2293, opened 2017-10-17
Click on a read notification just leaves focus on the notification


jwforres:
Right now if you click on an already read notification it looks like this (see blue focus outlines):

![screen shot 2017-10-17 at 4 51 59 pm](https://user-images.githubusercontent.com/1467629/31688868-afa60d2c-b35b-11e7-822a-187691a9e571.png)

I question whether we should even still have a click handler on a read notification, clicking it does absolutely nothing.

jwforres:
@openshift/team-ux-review

beanh66:
@jeff-phillips-18 I don't think we have this issue in PatternFly, but curious if you have seen it? Typically if a row is already marked as read, we would not see the blue outline. @jwforres I agree it is unnecessary.

jwforres:
Yeah @spadgett thought this was probably a bug in our special handling of click events in our impl of the drawer.  Once its been marked as read I think we should remove the click handler and click target (no longer show the hand when hovering).

spadgett:
It's specific to OpenShift notifications:

https://github.com/openshift/origin-web-console/blob/master/app/views/directives/notifications/notification-body.html#L1-L4

cc @benjaminapetersen

benjaminapetersen:
We can remove that handler pretty easily:

```
ng-click="notification.unread && $ctrl.customScope.markRead(notification)">
```
There is no longer a click, but that doesn't affect the styling.  Working on a solution.

benjaminapetersen:
Looking @ this with @sg00dwin.  So the patternfly [demo](http://www.patternfly.org/angular-patternfly/#/api/patternfly.notification.component:pfNotificationDrawer) appears to be tinkering with the focus. If I manually trigger the "focus" attrib in the web inspector, I can achieve the same effect:

Angular Patternfly:
![screen shot 2017-10-18 at 11 50 05 am](https://user-images.githubusercontent.com/280512/31728629-cfbf8788-b3fa-11e7-8fa7-4c538a532de9.png)

Patternfly: 
![screen shot 2017-10-18 at 11 53 30 am](https://user-images.githubusercontent.com/280512/31728697-fe7bbfce-b3fa-11e7-8bb0-a78d72278d08.png)

So there is something happening in the demo CSS that is not inherited in the actual component that is causing the focus to be hidden or clipped.

benjaminapetersen:
The cause is our `tabindex="0"`:

```
<div
  class="drawer-pf-notification-inner"
[...]

PR #2299, opened 2017-10-18
Fix issue 2293: click on read notification focus bug

benjaminapetersen:
Fix issue #2293 

You can still click the notification to clear it, but now there is a `sr-only` button for tabbing.  The gif below shows the tabbing experience.  There is clearly a hidden element in the flow.

@jwforres @spadgett

### DIFF
--- a/app/views/directives/notifications/notification-body.html
+++ b/app/views/directives/notifications/notification-body.html
@@ -1,7 +1,7 @@
 <div
   class="drawer-pf-notification-inner"
-  tabindex="0"
-  ng-click="$ctrl.customScope.markRead(notification)">
+  ng-class="{ 'is-clickable': notification.unread }"
+  ng-click="notification.unread && $ctrl.customScope.markRead(notification)">
   <button
     class="btn btn-link pull-right drawer-pf-notification-close"
     type="button"
@@ -88,7 +88,14 @@
             <span ng-if="!$last" class="toast-action-divider">|</span>
           </span>
         </span>
-
+        <span class="sr-only">Message Unread. </span>
+        <a
+          ng-if="notification.unread"
+          href=""
+          class="sr-only sr-only-focusable"
+          ng-click="$ctrl.customScope.markRead(notification)">
+          <span>Mark Read</span>
+        </a>
       </div>
       <div
         ng-if="notification.event.count > 1"
--- a/dist/scripts/templates.js
+++ b/dist/scripts/templates.js
@@ -7860,7 +7860,7 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
 
 
   $templateCache.put('views/directives/notifications/notification-body.html',
-    "<div class=\"drawer-pf-notification-inner\" tabindex=\"0\" ng-click=\"$ctrl.customScope.markRead(notification)\">\n" +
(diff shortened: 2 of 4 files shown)


jwforres:
I'm ok with this, I doubt a typical user would tab through that part of the UI

app/views/directives/notifications/notification-body.html
```diff
@@ -10,6 +9,11 @@
     <span class="sr-only">Clear notification</span>
     <span aria-hidden="true" class="pficon pficon-close"></span>
   </button>
+  <button
```
jwforres: I would only expect this button to exist if the notification is unread

jwforres:
Are we doing anything from an accessibility perspective to notify the user that the event is unread?  Bolding the text but not providing sr-only text somewhere isn't accessible.

dist/scripts/templates.js
```diff
@@ -7785,11 +7785,14 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
 
 
   $templateCache.put('views/directives/notifications/notification-body.html',
-    "<div class=\"drawer-pf-notification-inner\" tabindex=\"0\" ng-click=\"$ctrl.customScope.markRead(notification)\">\n" +
+    "<div class=\"drawer-pf-notification-inner\" tabindex=\"0\" ng-click=\"notification.unread && $ctrl.customScope.markRead(notification)\">\n" +
```
spadgett: dist looks wrong

app/views/directives/notifications/notification-body.html
spadgett: Maybe move it down after the notification message, too?

benjaminapetersen: Will do.

benjaminapetersen:
I can do something like this to also indicate to a screen reader that the message is unread:

```
<a
       ng-if="notification.unread"
       href=""
       class="sr-only sr-only-focusable"
       ng-click="$ctrl.customScope.markRead(notification)">
       <span class="sr-only">Message Unread.</span><!-- this never becomes visible -->
       <span>Mark Read?</span>
</a>
```

We support a list of links following the message (though in practice I've not seen any with more than one link), so adding the "mark read" message here seems to make sense to me:



Though @spadgett you mentioned margins.  Think this could be better?

spadgett:
Your screenshot looks good to me, although I'd leave off the `?`

jwforres:
Agreed, no ?

app/views/directives/notifications/notification-body.html
```diff
@@ -88,7 +87,14 @@
             <span ng-if="!$last" class="toast-action-divider">|</span>
           </span>
         </span>
-
+        <a
+          ng-if="notification.unread"
+          href=""
+          class="sr-only sr-only-focusable"
+          ng-click="$ctrl.customScope.markRead(notification)">
+          <span class="sr-only">Message Unread. </span>
```
spadgett: nit: I'd pull this `span` outside of the `a` since it's not part of the action.

benjaminapetersen:
Dropping the `?`, also adjusting the pointer icon, will be good to remove `[WIP]` shortly.

benjaminapetersen:
rebased

/test

PR #2300, opened 2017-10-18
Fix bug with very long expanded toast notification text

### LINKED ISSUE #2284: Alert error 'see all' long message extends off screen and can not be scrolled.
This can occur even at full screen. 

A larger question is should messages beyond a certain length be displayed in a different view? Open in a new window? Formatted?

screen shot 2017-10-17 at 1 41 40 pm

screen shot 2017-10-17 at 1 42 02 pm

rhamilto:
Fixes #2284 

IE
screen shot 2017-10-18 at 1 44 52 pm

Chrome/mac
screen shot 2017-10-18 at 2 13 09 pm

### DIFF
--- a/app/styles/_toast.less
+++ b/app/styles/_toast.less
@@ -1,0 +1,11 @@
+//
+// Toast notifications
+// --------------------------------------------------
+
+// so that expanded text isn't unbounded
+// fixes https://github.com/openshift/origin-web-console/issues/2284
+.toast-notification-details {
+  margin-right: -15px;
+  max-height: 300px;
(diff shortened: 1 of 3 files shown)


spadgett:
LGTM, but it should this be in https://github.com/openshift/origin-web-common/blob/master/src/styles/_notifications.less ?

rhamilto:
> LGTM, but it should this be in https://github.com/openshift/origin-web-common/blob/master/src/styles/_notifications.less ?

Yes.  Thanks for catching that.  I forgot to look in common.  Will open a PR there.

PR #2305, opened 2017-10-18
Mount path is not required for envFrom

### LINKED ISSUE #2303: Save button not enabled in "Add to Application" dialog
openshift web console 2017-10-18 14-56-09

spadgett:
In the add to application dialog, don't require a mount path when adding environment variables.

Fixes #2303 

cc @cfchase


spadgett:
@jeff-phillips-18 can you "lgtm"? :)

Issue #2228, opened 2017-10-06
Setting wizard title adds a tooltip to the entire dialog


spadgett:
Since the `title` attribute is a standard HTML attribute used for tooltips, setting `title` on the `pf-wizard` component causes the browser to show a tooltip when you hover the mouse over any part of the dialog. This is especially confusing when you hover over a "Cancel" button and it shows a tooltip like "Create Binding."

One example: https://github.com/openshift/origin-web-console/blob/master/app/views/directives/bind-service.html#L3

But it's now a problem in all of our dialogs. This might need to be fixed upstream in angular-patternfly.

https://github.com/patternfly/angular-patternfly/blob/master/src/wizard/wizard.component.js#L314

cc @jeff-phillips-18 @dtaylor113 @rhamilto

spadgett:
I see now there's an angular-patternfly issue: https://github.com/patternfly/angular-patternfly/issues/639

spadgett:
Closing as this is now tracked by https://bugzilla.redhat.com/show_bug.cgi?id=1504001

spadgett:
This has been fixed.

PR #2306, opened 2017-10-19
Bug 1504001 - Fix tooltip over entire dialog

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1504001
See also #2228
See also https://github.com/openshift/origin-web-catalog/pull/514


jeff-phillips-18:
This requires an angular-patternfly update to version 4.11.4 or above

spadgett:
> This requires an angular-patternfly update to version 4.11.4 or above

Yes, the update is in bower.json

Issue #2294, opened 2017-10-17
Left nav in mini-mode is missing alt text for icons


jwforres:
When you click the hamburger to collapse the left nav, the text is set to display:none which makes it invisible to a screen reader.  There is no alternative text for the icons in this case.

spadgett:
This should probably be addressed in patternfly. They make `.list-group-item-value` hidden, but maybe it should be `.sr-only()` instead.

PR #2309, opened 2017-10-19
Update vertical nav items adding title attribute and aria-hidden true.

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/2294

### DIFF
--- a/app/views/_sidebar.html
+++ b/app/views/_sidebar.html
@@ -24,13 +24,13 @@
         ng-mouseleave="onMouseLeave(primaryItem)"
         class="list-group-item">
         <a ng-if="primaryItem.href" ng-href="{{navURL(primaryItem.href)}}" ng-click="itemClicked(primaryItem)">
-          <span class="{{primaryItem.iconClass}}"></span> <span class="list-group-item-value">{{primaryItem.label}}</span>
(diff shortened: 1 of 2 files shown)


app/views/_sidebar.html
```diff
@@ -24,13 +24,13 @@
         ng-mouseleave="onMouseLeave(primaryItem)"
         class="list-group-item">
         <a ng-if="primaryItem.href" ng-href="{{navURL(primaryItem.href)}}" ng-click="itemClicked(primaryItem)">
-          <span class="{{primaryItem.iconClass}}"></span> <span class="list-group-item-value">{{primaryItem.label}}</span>
+          <span alt="{{primaryItem.label}}" aria-hidden="true" class="{{primaryItem.iconClass}}"></span> <span class="list-group-item-value">{{primaryItem.label}}</span>
```
spadgett: Maybe (untested)

```html
<span title="{{primaryItem.label}}" class="{{primaryItem.iconClass}}" aria-hidden="true"></span>
<span class="list-group-item-value">{{primaryItem.label}}</span>
<span ng-if="nav.collapsed && !isMobile" class="sr-only">{{primaryItem.label}}</span>
```

But I think it should be fixed in Patternfly.

spadgett:
Replaced by #2359

PR #2315, opened 2017-10-19
Allow unlimited websockets on Edge - also fix leaking websockets on monitoring page

jwforres:


app/scripts/controllers/monitoring.js
```diff
@@ -29,7 +29,8 @@ angular.module('openshiftConsole')
     $scope.renderOptions.showEventsSidebar = true;
     $scope.renderOptions.collapseEventsSidebar = localStorage.getItem('monitoring.eventsidebar.collapsed') === 'true';
 
-
+    var limitWatches = $filter('isIE')();
```
spadgett: Wonder if we shouldn't set `LIMIT_WATCHES` as a constant once in app.js. Then if we ever need to change it again, we're always consistent.

jwforres:
This half fixes https://bugzilla.redhat.com/show_bug.cgi?id=1379641

PR #2318, opened 2017-10-19
Fix phantom cleared notifications

### LINKED ISSUE #2302: Marking all notifications can resurrect a cleared notification
Steps to reproduce:

- Open the notification drawer with a set of notifications
- Clear via the "x" a notification 
- Click "mark all read" at the bottom of the drawer 

All notifications currently in view are marked read, notifications previously cleared will suddenly become visible & will not be marked read.

benjaminapetersen:
Fix #2302: Marking all notifications read can resurrect cleared notifications.


spadgett:
Closing since the same commit is rebased in #2346

PR #2322, opened 2017-10-20
Bug 1504511 - Remember template values when clicking back

spadgett:
When editing template parameter values, remember the values when the
user clicks back and then forward again. This will still correctly clear
the values if the user switches templates in the "Select from Project"
dialog.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1504511

@jeff-phillips-18 @jwforres


spadgett:
Sounds like they're OK from the bug:

> Ordering image/serviceclass does not have the issue

I'll double check

spadgett:
The other dialogs remember parameters.

Issue #2323, opened 2017-10-20
Overview bind parameters look odd


spadgett:
The bind parameters are displayed on the overview now. I think this might be unintentional, but it's odd to show bind parameters and not instance parameters. I'd vote just to remove them.

If we keep them, the spacing looks wrong.

![screen shot 2017-10-20 at 8 51 17 am](https://user-images.githubusercontent.com/1167259/31821680-ef40e1cc-b573-11e7-889d-9f5668e478c5.png)

@jwforres @jeff-phillips-18 FYI

spadgett:
Similar spacing problems on the provisioned service page:

![screen shot 2017-10-20 at 8 58 44 am](https://user-images.githubusercontent.com/1167259/31821968-e98c0652-b574-11e7-8c99-2e5effdf809f.png)

jeff-phillips-18:
Fixed by #2326

PR #2326, opened 2017-10-20
Do not show binding parameters on the overview page

jeff-phillips-18:
See #2323

### DIFF
--- a/app/scripts/directives/serviceInstanceBindings.js
+++ b/app/scripts/directives/serviceInstanceBindings.js
@@ -9,6 +9,7 @@ angular.module('openshiftConsole').component('serviceInstanceBindings', {
   ],
   controllerAs: '$ctrl',
   bindings: {
+    isOverview: '<?',
     showHeader: '<?',
     project: '<',
     bindings: '<',
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -200,28 +200,22 @@ code.command {
   margin-left: 20px;
 }
 .service-binding-parameters {
-  margin-left: 20px;
-  > a {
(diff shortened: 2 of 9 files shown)


app/scripts/directives/serviceInstanceBindings.js
```diff
@@ -9,6 +9,7 @@ angular.module('openshiftConsole').component('serviceInstanceBindings', {
   ],
   controllerAs: '$ctrl',
   bindings: {
+    isOverview: '<?',
```
spadgett: Maybe call this `showParameters` and flip the logic?

spadgett:
@jeff-phillips-18 verified the fix works

spadgett:
But we should still look at the margins:

openshift web console 2017-10-20 11-10-44

app/scripts/directives/serviceInstanceBindings.js
jeff-phillips-18: We can but isOverview is already used in the sub-component  (serviceBinding) so this is really just a pass thru to that.

spadgett: I'm OK with it then

spadgett:
Going to leave #2323 open to track the other display issue

spadgett:
@jeff-phillips-18 dist mismatch

jeff-phillips-18:
Rebased to fix dist issues.

spadgett:
still some dist problems

jeff-phillips-18:
Added visual changes for binding parameters as well. See discussion in https://github.com/openshift/origin-web-catalog/pull/518

image

PR #2327, opened 2017-10-20
Fix bug where key in key-value-table didn't align top

### LINKED ISSUE #2320: Config map keys should be top-aligned in details table
The keys are vertically center aligned in the table cell. When the config map has a large value that's expanded, the key is pushed off the screen. Also the width of the key column is sometimes small, causing the key name to wrap when the table could give it more width.

screen shot 2017-10-19 at 5 51 08 pm

screen shot 2017-10-19 at 5 51 12 pm

openshift web console 2017-10-19 17-53-42

rhamilto:
Fixes #2320 

screen shot 2017-10-20 at 11 49 34 am
screen shot 2017-10-20 at 11 49 25 am
screen shot 2017-10-20 at 11 49 17 am

### DIFF
--- a/app/styles/_tables.less
+++ b/app/styles/_tables.less
@@ -84,21 +84,20 @@
   }
 }
 
-.key-value-table > tbody > tr > td {
-  &.key {
-    padding-right: 10px;
-    vertical-align: top;
-  }
-  &.value {
-    width: 100%;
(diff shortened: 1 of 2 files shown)


spadgett:
webdriver flake

/retest

Issue #2332, opened 2017-10-21
Select from Project sort not obvious


spadgett:
I believe the projects are sorted by most recent, but it's not obvious since we don't add the grouping to the ui-select. It looks unsorted.

![openshift web console 2017-10-21 07-41-42](https://user-images.githubusercontent.com/1167259/31851433-97bb091c-b633-11e7-8e25-38ea7619b5d0.png)

PR #2336, opened 2017-10-23
Add grouping to Select from Project

jhadvig:
@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/2332

### DIFF
--- a/app/scripts/directives/processTemplateDialog.js
+++ b/app/scripts/directives/processTemplateDialog.js
@@ -208,6 +208,15 @@
       }));
     };
 
+    // TODO: Update the select-project component in the origin-web-catalog to optionally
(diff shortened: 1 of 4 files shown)


app/scripts/directives/processTemplateDialog.js
```diff
@@ -208,6 +208,14 @@
       }));
     };
 
+    // TODO: Make this as a reusable component since we are using similar method also in origin-web-catalog
```
spadgett: I think the TODO is really to reuse the select-project component. We'd need to update the component to optionally disable creating new projects, however. Do you mind updating the comment?

jhadvig:
@spadgett comment addressed

spadgett:
WebDriver flake #1684

/retest

PR #2339, opened 2017-10-23
Align hover and expanded list-group-item bg colors with list-pf

rhamilto:
PatternFly deprecated list-view, so the hover and expanded bg colors
are still the old standard.  This fix remedies that until we can update
the monitoring page to use list-pf (I will open a separate tech debt
issue for that).
screen shot 2017-10-23 at 9 40 09 am
screen shot 2017-10-23 at 9 40 15 am

### DIFF
--- a/app/styles/_list-view.less
+++ b/app/styles/_list-view.less
@@ -56,8 +56,13 @@
   h2 + & {
     margin-top: 20px;
   }
-  .list-group-item:first-child {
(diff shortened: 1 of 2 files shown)


spadgett:
Hm, Jenkins is complaining about the dist, but it passed Travis. @rhamilto can you try rebasing and rebuilding?

PR #2259, opened 2017-10-12
Replacing empty tables with empty state message

rhamilto:
Fixes #109
Replaces #497

HACK DAY!

Before:
127 0 0 1-8443-console-project-my-project-dc-browse-deployments
127 0 0 1-8443-console-project-my-project-dc-browse-deployments 1

After:
localhost-9000-dev-console-project-my-project-empty-quota
localhost-9000-dev-console-project-my-project-delete-browse-config-maps-labelfilter 7b 22matchexpressions 22- 5b 7b 22key 22- 22foo 22 22operator 22- 22exists 22 22values 22- 5b 5d 7d 5d 7d

PR #2341, opened 2017-10-23
Update storage filter empty state to new standard

rhamilto:
I overlooked the filtered empty state message for storage in https://github.com/openshift/origin-web-console/pull/2259.  This fixes that.

screen shot 2017-10-23 at 11 05 38 am


spadgett:
WebDriver flake #1684

/retest

PR #2344, opened 2017-10-23
Membership updates to correct broken layouts when multiple roles assigned

spadgett:
Overall code cleanup
Fixes https://github.com/openshift/origin-web-console/issues/1219

@sg00dwin is the commit author.

### DIFF
--- a/app/styles/_membership.less
+++ b/app/styles/_membership.less
@@ -1,70 +1,100 @@
 .membership {
-  // TODO: make generic?
-  h1 {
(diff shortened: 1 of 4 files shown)


spadgett:
Replaces #2283

Same as #2283 with dist rebuilt

spadgett:
Adding lgtm label since #2283 is already approved

sg00dwin:
Thanks @spadgett

PR #2346, opened 2017-10-23
Fix issue #2302: marking all notifications read can resurrect cleared notifications

spadgett:
This is the same as #2318, but rebased on current master.

@benjaminapetersen I've left you as the commit author


spadgett:
Adding lgtm label since #2346 is already approved

PR #2353, opened 2017-10-23
Center-align "Back to..." link on attach pvc page

### LINKED ISSUE #2352: Storage empty state back link should be centered on dc add storage page?
current:
screen shot 2017-10-23 at 2 34 47 pm

proposed:
screen shot 2017-10-23 at 2 49 49 pm

rhamilto:
Fixes #2352 

screen shot 2017-10-23 at 4 29 17 pm

### DIFF
--- a/app/views/attach-pvc.html
+++ b/app/views/attach-pvc.html
@@ -6,15 +6,15 @@
           <breadcrumbs breadcrumbs="breadcrumbs"></breadcrumbs>
           <div ng-show="!pvcs || !attach.resource">Loading...</div>
 
-          <div ng-show="pvcs && !pvcs.length && attach.resource" class="empty-state-message empty-state-full-page">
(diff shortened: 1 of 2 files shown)


app/views/attach-pvc.html
```diff
@@ -23,7 +23,7 @@ <h2 class="text-center">No persistent volume claims.</h2>
               To claim storage from a persistent volume, refer to the documentation on <a target="_blank" ng-href="{{'persistent_volumes' | helpLink}}">using persistent volumes</a>.
             </p>
```
spadgett: Should we center this paragraph too for consistency with other empty state messages?

rhamilto:
Moved .text-center to .empty-state-message so all nodes are centered.  Good catch, @spadgett.

PR #2355, opened 2017-10-23
Handle displaying parameters when secrets are not available.

jeff-phillips-18:
Bump catalog version to 0.0.58


app/scripts/controllers/serviceInstance.js
```diff
@@ -73,18 +73,40 @@ angular.module('openshiftConsole')
...
 
+      $scope.allowParametersReveal = AuthorizationService.canI('secrets', 'get', $scope.projectName);
       $scope.parameterData = {};
-      _.each(_.keys(_.get($scope.parameterSchema, 'properties')), function(key) {
-        $scope.parameterData[key] = $scope.parameterSchema.properties[key].default;
+      $scope.opaqueParameterKeys = [];
+
+      if (!$scope.allowParametersReveal) {
+        _.each(_.keys(_.get($scope.parameterSchema, 'properties')), function (key) {
+          $scope.parameterData[key] = '*****';
```
spadgett: I think we should be able to just fill in the parameter data from the status parameters below and just remove this `if` block. Are you seeing parameters in the schema, but not in status?

app/scripts/controllers/serviceInstance.js
```diff
@@ -73,18 +73,40 @@ angular.module('openshiftConsole')
...
+          $scope.opaqueParameterKeys.push(key);
+        }
       });
 
-      $scope.parameterData = angular.extend($scope.parameterData, _.get($scope.serviceInstance, 'spec.parameters', {}));
+      // Fill in the secret values if they are available to the user
+      if ($scope.allowParametersReveal) {
+        // Set default values for those items that are not specifically set, secrets will override
+        _.each(_.keys(_.get($scope.parameterSchema, 'properties')), function (key) {
+          $scope.parameterData[key] = $scope.parameterSchema.properties[key].default;
```
spadgett: I think the defaults are only for prefiling the create form. A missing value isn't necessarily treated as a default.

app/scripts/controllers/serviceInstance.js
```diff
@@ -73,18 +73,40 @@ angular.module('openshiftConsole')
...
+        });
 
-      if (AuthorizationService.canI('secrets', 'get', $scope.projectName)) {
+        // Get the data from each secret
         _.each(_.get($scope.serviceInstance, 'spec.parametersFrom'), function (parametersSource) {
           secretWatchers.push(DataService.watchObject("secrets", _.get(parametersSource, 'secretKeyRef.name'), $scope.projectContext, function (secret) {
             try {
-              _.extend($scope.parameterData, JSON.parse(SecretsService.decodeSecretData(secret.data)[parametersSource.secretKeyRef.key]));
+              var secretData = JSON.parse(SecretsService.decodeSecretData(secret.data)[parametersSource.secretKeyRef.key]);
+              // TODO: ONly include fields from the secret that are part of the schema
```
spadgett: typo: "Only"

app/scripts/controllers/serviceInstance.js
spadgett: I guess there are parameters that haven't been set. But since users can already see that in the instance YAML under status, I don't think we need to obfuscate it here.

spadgett: Disregard my comment here. Thinking about this more, I think this behavior is better.

app/scripts/controllers/serviceInstance.js
jeff-phillips-18: Fixed.

spadgett:
Needs #2367 to fix the dist problem

PR #2358, opened 2017-10-23
Making add config files empty state formatting consistent with add volume

rhamilto:
Companion to https://github.com/openshift/origin-web-console/pull/2353

This one is unlikely to ever be seen given there are always secrets, but I figured I'd make it consistent anyhow...

screen shot 2017-10-23 at 4 41 27 pm


spadgett:
> This one is unlikely to ever be seen given there are always secrets

It's possible the user can't list secrets, although by default editors can.

PR #2360, opened 2017-10-23
Use correct HTTP status from failed DataService.list

### LINKED ISSUE #2354: Deploy image dialog should not show an error when it can't list secrets
Currently, you see an error toast when the deploy image dialog can't list secrets. It should suppress this error since they're not needed and only there as options for the env value from options.

spadgett:
Bumps origin-web-common to 0.0.69

Requires https://github.com/openshift/origin-web-common/pull/236
Fixes #2354

### DIFF
--- a/app/scripts/controllers/addConfigVolume.js
+++ b/app/scripts/controllers/addConfigVolume.js
@@ -143,7 +143,7 @@ angular.module('openshiftConsole')
         DataService.list($scope.configMapVersion, context, null, { errorNotification: false }).then(function(configMapData) {
           $scope.configMaps = orderByDisplayName(configMapData.by("metadata.name"));
         }, function(e) {
-          if (e.code === 403) {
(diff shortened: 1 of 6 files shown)


jwforres:
so all the list error handlers were broken?

are there list calls in catalog that we would need to fix?

spadgett:
> so all the list error handlers were broken?

Yeah :/

> are there list calls in catalog that we would need to fix?

I swept all 3 repos looking at list calls. The ones in the catalog repo weren't doing anything with the failure args (catalog service)

jwforres:
lgtm, just needs the common update it depends on

spadgett:
Updated with dist

@jwforres The logos changes in the dist are because we cut a logos release and we were using a `^` in bower.json. @rhamilto is removing the `^` in #2367

spadgett:
Rebased

@jwforres another look?

Issue #2335, opened 2017-10-23
"Select from Project" should not have an "Add to Project" dropdown when inside project


spadgett:
When already inside a project, we should always add to the current project.

![screen shot 2017-10-23 at 8 20 59 am](https://user-images.githubusercontent.com/1167259/31888752-42cbb57a-b7cb-11e7-8ef6-e4e9be5e3637.png)

cc @jeff-phillips-18

PR #2361, opened 2017-10-24
Preselect current project in 'Select from Project'

jhadvig:
Preselecting the current project in the `Selection` step will remove the selectbox from the `Configuration` step.

7fee4eaa-e28f-44fd-y677-2e1bf20d1eeb

@spadgett PTAL

Closes https://github.com/openshift/origin-web-console/issues/2335

### DIFF
--- a/app/scripts/directives/processTemplateDialog.js
+++ b/app/scripts/directives/processTemplateDialog.js
@@ -85,6 +85,10 @@
     ctrl.$onInit = function() {
       ctrl.loginBaseUrl = DataService.openshiftAPIBaseUrl();
       ctrl.preSelectedProject = ctrl.selectedProject = ctrl.project;
+      if (ctrl.project) {
+        ctrl.templateProject = ctrl.project;
+        ctrl.templateProjectChange();
+      }
       listProjects();
 
       ctrl.projectEmptyState = {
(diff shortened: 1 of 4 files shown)


app/scripts/directives/processTemplateDialog.js
```diff
@@ -184,7 +185,7 @@
     };
 
     ctrl.templateProjectChange = function () {
-      ctrl.templateProjectName = _.get(ctrl.templateProject, 'metadata.name');
+      ctrl.templateProjectName = _.get(ctrl.selectedProject, 'metadata.name');
```
spadgett: I kind of like the original `templateProject` name better. `selectedProject` sounds like the project you are instantiating the template into.

jhadvig:
@spadgett comment addressed :)

app/scripts/directives/processTemplateDialog.js
```diff
@@ -137,7 +138,7 @@
     };
 
     $scope.$on('templateInstantiated', function(event, message) {
-      ctrl.selectedProject = message.project;
+      ctrl.templateProject = message.project;
```
spadgett: @jhadvig There are two different projects. We need to make sure we're not getting them mixed up:

- `templateProject` -- the source project the template is from
- `selectedProject` -- the target project the template is instantiated into

You can selected a project to instantiate into on the landing page, but current project is always used when showing the dialog inside the project.

I think this is probably a two line fix: calling `ctrl.templateProjectChange()` on init and setting `currentProject` as the project in the view.

jhadvig: @spadgett we need also the change the `process-template-select` so it uses the `ctrl.selectedProject` not to use the `ctrl.templateProject`, cause we are not even initiating that var

jhadvig:
@spadgett updated the PR so it covers all the cases, in the landing page and also in projects.
PTAL

spadgett:
@jhadvig dist mismatch :/

spadgett:
You might need to rebase, install assets, and rebuild?

spadgett:
/lgtm
/hold

Need to wait for #2367 or merge will fail

PR #2370, opened 2017-10-25
Change cursor to a pointer when hover over read notification

jhadvig:
@jwforres PTAL

Closes https://github.com/openshift/origin-web-console/issues/2293

### DIFF
--- a/app/styles/_notifications.less
+++ b/app/styles/_notifications.less
@@ -39,9 +39,9 @@ notification-drawer-wrapper {
   }
   // the whole block is clickable, need to set pointer on all of these
   // for the correct visual
-  .drawer-pf-notification-inner.is-clickable,
-  .is-clickable .drawer-pf-notification-info,
-  .is-clickable .drawer-pf-notification-message {
+  .drawer-pf-notification-inner,
+  .drawer-pf-notification-inner .drawer-pf-notification-info,
+  .drawer-pf-notification-inner .drawer-pf-notification-message {
     cursor: pointer;
   }
   .drawer-pf-notification {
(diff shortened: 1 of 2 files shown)


spadgett:
I think the fix is to use the default (arrow) cursor when read. Is that what this does? It looks like it makes it a pointer (hand) cursor always.

benjaminapetersen:
So on Master I see what @spadgett mentioned, which is to use the default arrow once it has been read:



This PR will always use the pointer:



@jwforres can you clarify what you mentioned on [this comment](https://github.com/openshift/origin-web-console/issues/2293#issuecomment-338790856)?  Do we just want to eliminate the hover effect entirely?

app/styles/_notifications.less
```diff
@@ -39,9 +39,9 @@ notification-drawer-wrapper {
   }
   // the whole block is clickable, need to set pointer on all of these
   // for the correct visual
```
benjaminapetersen: I'm not sure this is quite right, might be that we just want to eliminate the hover (bg blue) as well as return the cursor.

spadgett:
If we switch to the default cursor, I think we'd want to remove the blue hover style as well? But I might have misunderstood @jwforres' comment.

I personally don't have a strong preference here.

jwforres:
weird, i thought i was on master when i checked it...

i would at least see what it looks like without the hover color

jhadvig:
So from [this comment](https://github.com/openshift/origin-web-console/issues/2293#issuecomment-338790856) I was in a impression we want to have to pointer cursor on unread and also read notifications.

jhadvig:
Thinking about it now Im not really sure it would make sense to have the pointer on read notification. If we align on this it would be better to merge @benjaminapetersen [PR](https://github.com/openshift/origin-web-console/pull/2383/files)
If align on the pointer for both cases the I would rather keep the changed background on the hover, just to to have a similar behavior of  read and unread notifications.

jhadvig:
Closing this PR in favor of https://github.com/openshift/origin-web-console/pull/2383
@benjaminapetersen fyi :)

benjaminapetersen:
Cool, think mine is gonna get closed as well. Yay discussions!

Issue #2175, opened 2017-09-26
Marked for deletion "in a few seconds"


spadgett:
This should say "just now" or "a few seconds ago." (I might have a client clock skew, but we should never show a future value.)

![screen shot 2017-09-26 at 4 28 32 pm](https://user-images.githubusercontent.com/1167259/30882696-d2ecf9f6-a2d7-11e7-9649-1a25e8eab6c3.png)

jhadvig:
in favour for the "few seconds ago", since term "just now" is pretty relative in this terms

jwforres:
So we already started normalizing all negative durations to 0 since we knew they were caused by clock skew.  I would be tempted to do the same with the relative timestamp helpers since nothing we work with today has future dates.

The exception will be CronJobs that are scheduled in the future.  We could do something like adding a flag to relative timestamp that allows future dates and you have to opt in to that.

jwforres:
So actually we do not have a relative timestamp in any of the other deletion messages.  We should remove this timestamp entirely for consistency. That is the least risky fix at this point.

PR #2372, opened 2017-10-25
Remove timestamp from Service Instance Row deletion alert

jhadvig:
@spadgett removed the timestamp as suggested... PTAL

Closes https://github.com/openshift/origin-web-console/issues/2175

### DIFF
--- a/app/views/overview/_service-instance-row.html
+++ b/app/views/overview/_service-instance-row.html
@@ -101,8 +101,7 @@
               <div class="alert word-break alert-warning">
                 <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
                 <span class="sr-only">warning</span>
-                <span class="strong">The service was marked for deletion</span>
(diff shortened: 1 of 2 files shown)


app/views/overview/_service-instance-row.html
```diff
@@ -101,8 +101,7 @@
               <div class="alert word-break alert-warning">
                 <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
                 <span class="sr-only">warning</span>
-                <span class="strong">The service was marked for deletion</span>
-                <span class="strong" am-time-ago="row.apiObject.metadata.deletionTimestamp"></span>.
+                <span class="strong">The service was marked for deletion</span>.
```
spadgett: nit: Any reason not to put the period inside the span so it has the strong style like the rest of the sentence?

jhadvig: no reason at all... I've realized it after hitting the Open PR btn... fixing atm

PR #2376, opened 2017-10-25
Update getting started resources for enterprise catalog

### LINKED ISSUE #2282: Getting started links need to be updated for OCP
The documentation link may be pointing to the wrong doc. We should decide which of the other links we want to be there.

The documentation link needs to be pulled from the openshift constants if its not already so that it correctly points to customized doc links, such as for Online.

spadgett:
Closes #2282

screen shot 2017-10-25 at 1 38 27 pm

@sspeiche 
/assign @jwforres

### DIFF
--- a/app/scripts/constants.js
+++ b/app/scripts/constants.js
@@ -64,6 +64,33 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
   CLI: {
     "Latest Release":          "https://access.redhat.com/downloads/content/290"
   },
+
+  // Change the minishift link to the CDK for enterprise.
+  CATALOG_HELP_RESOURCES: {
+    links: [
+      {
+        title: 'Documentation',
+        help: ''
+      },
+      {
+        title: 'Interactive Learning Portal',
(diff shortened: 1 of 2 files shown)


app/scripts/constants.js
```diff
@@ -64,6 +64,33 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
...
+      {
+        title: 'Documentation',
+        help: ''
+      },
+      {
+        title: 'Interactive Learning Portal',
+        href: 'https://learn.openshift.com'
+      },
+      {
+        title: 'Red Hat Container Development Kit',
```
jwforres: feels a bit wordy compared to the other links, can we drop the "Red Hat" from it?  @sspeiche ?

spadgett: > feels a bit wordy compared to the other links, can we drop the "Red Hat" from it?

nudge @sspeiche :)

spadgett:
@jwforres updated

spadgett:
openshift web console 2017-10-30 14-09-07

PR #2377, opened 2017-10-25
Allow EnvFrom Prefix

### LINKED ISSUE #2362: Need to handle envFrom prefix
Our environment from implementation does not handle the `prefix`. We need to make sure we're not dropping values with different prefixes minimally, but we should also let users see and change the prefix values.

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L1698-L1700

cdcabrera:
Opened envFrom to allow prefix input.

- [x] Minor display logic updates
- [x] Design/UX modifications

Related to #2182 and closes #2362 

screen shot 2017-10-25 at 5 06 37 pm

### DIFF
--- a/app/scripts/directives/editEnvironmentFrom.js
+++ b/app/scripts/directives/editEnvironmentFrom.js
@@ -8,11 +8,9 @@
       EditEnvironmentFrom
     ],
     bindings: {
-      addRowLink: '@',              // creates a link to "add row" and sets its text label
-      entries: '=',                 // an array of objects containing configmaps and secrets
-      envFromSelectorOptions: '<',  // dropdown selector options, an array of objects
-      selectorPlaceholder: '@',     // placeholder copy for dropdown selector
-      isReadonly: '<?'              // display as read only values
+      entries: '=',                         // an array of objects containing configmaps and secrets
+      envFromSelectorOptions: '<',          // dropdown selector options, an array of objects
+      isReadonly: '<?'                      // display as read only values
     },
     templateUrl: 'views/directives/edit-environment-from.html'
   });
@@ -52,7 +50,7 @@
 
       ctrl.envFromEntries.splice(start, deleteCount);
 
-      if(!ctrl.envFromEntries.length && ctrl.addRowLink) {
+      if(!ctrl.envFromEntries.length) {
         addEntry(ctrl.envFromEntries);
       }
 
@@ -105,6 +103,10 @@
           break;
       }
 
+      if (entry.prefix) {
+        newEnvFrom.prefix = entry.prefix;
+      }
+
       _.assign(ctrl.envFromEntries[index], newEnvFrom);
       ctrl.updateEntries(ctrl.envFromEntries);
     };
@@ -121,7 +123,7 @@
 
       ctrl.envFromEntries = ctrl.entries || [];
 
-      if(!ctrl.envFromEntries.length) {
+      if (!ctrl.envFromEntries.length) {
         addEntry(ctrl.envFromEntries);
       }
 
--- a/app/styles/_kve.less
+++ b/app/styles/_kve.less
@@ -115,18 +115,20 @@
     width: (@as-sortable-item-button-width * 2);
   }
 
-  .key-value-editor-entry,
   .environment-from-entry {
     display: table;
     margin-bottom: 15px;
-    padding-right: (@as-sortable-item-button-width * 2);
+    padding-right: 0;
+    @media(max-width: @screen-xxs-max) {
+      padding-right: (@as-sortable-item-button-width * 2);
+    }
     position: relative;
     table-layout: fixed;
     width: 100%;
-    @media(min-width: @screen-md-min)  {
+    @media(min-width: @screen-xs-min)  {
       .environment-from-editor-button {
         float: left;
-        padding-right: 5px;
+        padding: 0 5px;
         position: relative;
         width: auto;
       }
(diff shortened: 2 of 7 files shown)


app/styles/_kve.less
```diff
@@ -170,7 +170,8 @@
 }
 
 .key-value-editor .key-value-editor-input,
-.key-value-editor-header {
+.key-value-editor-header,
+.environment-from-editor-header {
```
cdcabrera: @sg00dwin made adjustment for the column widths see PR screenshot. Still looking the column containing the prefix field is a few pixels off

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -87,6 +87,11 @@
       }
     };
 
+    ctrl.removePrefix = function (entry) {
```
spadgett: Should no longer be needed.

app/views/directives/edit-environment-from.html
```diff
@@ -4,10 +4,35 @@
...
+        {{$ctrl.selectorPlaceholder}}
+      </div>
+    </div>
+    <div class="form-group environment-from-editor-header value-header">
+      <div class="input-group">
+        Prefix
+        <small class="pficon pficon-help"
+               aria-hidden="true"
+               data-toggle="tooltip"
+               data-original-title="Optional prefix added to each environment variable name for this resource."></small>
```
spadgett: Maybe remove "for this resource." since it's a column header with many resources.

I know I'm picking nits on my own suggested wording :)

app/views/directives/edit-environment-from.html
```diff
@@ -23,8 +48,13 @@
...
             Use all keys and values from
-            <span ng-if="entry.configMapRef.name">config map {{entry.configMapRef.name}}</span>
-            <span ng-if="entry.secretRef.name">secret {{entry.secretRef.name}}</span>
+            <span ng-if="entry.prefix">prefixed</span>
+            <span ng-if="entry.configMapRef.name">config map
+              <span ng-if="entry.prefix">"{{entry.prefix}}"</span>
+              {{entry.configMapRef.name}}</span>
+            <span ng-if="entry.secretRef.name">secret
+              <span ng-if="entry.prefix">"{{entry.prefix}}"</span>
+              {{entry.secretRef.name}}</span>
```
spadgett: Maybe...?

```
Use all keys and values from config map my-config-map, adding name prefix "my-prefix-"
```

@jwforres is better at wording stuff then me, though.

app/views/directives/edit-environment-from.html
```diff
@@ -50,27 +80,40 @@
...
-          ng-click="$ctrl.viewOverlayPanel(entry.selectedEnvFrom)">View Details</a>
+      <div class="form-group environment-from-input">
+        <div class="environment-from-input" ng-if="!$ctrl.isEnvFromReadonly(entry) && $ctrl.hasOptions()">
+          <label for="envfrom-prefix-{{$index}}" class="sr-only">Prefix</label>
+          <input type="text"
+                 class="form-control"
+                 placeholder="Add prefix"
+                 id="envfrom-prefix-{{$index}}"
+                 name="envfrom-prefix-{{$index}}"
+                 ng-model="entry.prefix"/>
```
spadgett: We usually don't include the self-closing `/`

app/views/directives/edit-environment-from.html
```diff
@@ -50,27 +80,40 @@
...
-      </div>
-      <div class="environment-from-view-details">
-        <a
-          href=""
-          ng-if="entry.selectedEnvFrom"
-          ng-click="$ctrl.viewOverlayPanel(entry.selectedEnvFrom)">View Details</a>
+      <div class="form-group environment-from-input">
+        <div class="environment-from-input" ng-if="!$ctrl.isEnvFromReadonly(entry) && $ctrl.hasOptions()">
+          <label for="envfrom-prefix-{{$index}}" class="sr-only">Prefix</label>
+          <input type="text"
```
spadgett: This should have a pattern with a corresponding validation error. (I'll need to look up the correct regex for prefix.)

spadgett:
@sg00dwin Can you help with the suggested style changes in #2362? (OK as a separate PR.)

spadgett:
Looks like prefix has to match the env name regex

https://github.com/kubernetes/kubernetes/blob/5f030f4568822e791054d5c4c3aff7f9f5c85151/pkg/api/validation/validation.go#L1808-L1812

We're using `[A-Za-z_][A-Za-z0-9_]*`

app/scripts/directives/editEnvironmentFrom.js
cdcabrera: Agreed, on the clean up list

app/views/directives/edit-environment-from.html
cdcabrera: That works, yeah send it over

app/views/directives/edit-environment-from.html
cdcabrera: Old habits, can remove

app/views/directives/edit-environment-from.html
jwforres: agree, drop the "for this resource"

app/views/directives/edit-environment-from.html
jwforres: tough to stuff that into one sentence, would probably read better as two.

"Use all keys and values from config map my-config-map. Names will be prefixed with "my-prefix".

cdcabrera:
@beanh66 @jwforres @spadgett 

Prefix and the modal display, is this something we should try to work in?

screen shot 2017-10-26 at 11 58 53 pm

spadgett:
@cdcabrera Are you asking if we should show the prefix in the modal? I would just show the names as they are defined in the secret.

cdcabrera:
@spadgett indeed was wondering if the prefix needed to be displayed in some fashion within the modal itself

cdcabrera:
@spadgett per request removed the extra component bindings. Should now only have the main ones 
- entries
- envFromSelectorOptions
- isReadonly

Also as noted the "ng-pattern" copied over from key/values component misses out on special characters tacked onto the end, corrected it. There's still some funkiness around empty strings that happens on both Key/Values and EnvFrom

spadgett:
I feel like in read-only mode we should use the same two column display instead of trying to write things out as sentences. Also I don't think we need to use break-all, which breaks in the middle of words when not needed. We might look at this as a follow on, however, to get the blocker fixed.

openshift web console - private browsing 2017-10-30 09-24-34

PR #2382, opened 2017-10-25
Only transition vertical nav at mobile on show

rhamilto:
As a follow on to https://github.com/openshift/origin-web-console/pull/2381, it feels better to me if the vertical nav at mobile only transitions on show.  This way the vertical nav goes away immediately on clicking a nav item rather than lingering around as it transitions off the screen.


spadgett:
Agreed. Thanks @rhamilto 

/lgtm

PR #2383, opened 2017-10-25
Remove notification hover background if already read

### LINKED ISSUE #2293: Click on a read notification just leaves focus on the notification

benjaminapetersen:
Yet another fix #2293  😄 

Eliminates the background color for notifications already marked read.

@jwforres @spadgett

### DIFF
--- a/app/styles/_notifications.less
+++ b/app/styles/_notifications.less
@@ -37,6 +37,13 @@ notification-drawer-wrapper {
       display: block !important;
     }
   }
+  // override .drawer-pf-notification bg when not clickable
+  .drawer-pf-notification-inner {
+    background-color: #fff;
+  }
+  .drawer-pf-notification-inner.is-clickable:hover {
+    background-color: #def3ff;
+  }
   // the whole block is clickable, need to set pointer on all of these
   // for the correct visual
   .drawer-pf-notification-inner.is-clickable,
--- a/dist/styles/main.css
+++ b/dist/styles/main.css
@@ -1,8 +1,7 @@
 div.code,pre,textarea{overflow:auto}
 body,figure{margin:0}
 .text-left,caption,th{text-align:left}
-.navbar-fixed-bottom .navbar-collapse,.navbar-fixed-top .navbar-collapse,.pre-scrollable{max-height:340px}
-.c3 svg,html{-webkit-tap-highlight-color:transparent}
+.list-view-pf-top-align .list-view-pf-actions,.list-view-pf-top-align .list-view-pf-checkbox{align-self:flex-start}
 .fa,.font-icon,.glyphicon{-moz-osx-font-smoothing:grayscale}
 @font-face{font-family:"Open Sans";font-style:normal;font-weight:300;src:url(../styles/fonts/OpenSans-Light-webfont.eot);src:local("Open Sans Light"),local("OpenSans-Light"),url(../styles/fonts/OpenSans-Light-webfont.eot?#iefix) format("embedded-opentype"),url(../styles/fonts/OpenSans-Light-webfont.woff2) format("woff2"),url(../styles/fonts/OpenSans-Light-webfont.woff) format("woff"),url(../styles/fonts/OpenSans-Light-webfont.ttf) format("truetype"),url(../styles/fonts/OpenSans-Light-webfont.svg#OpenSans) format("svg")}
 @font-face{font-family:"Open Sans";font-style:normal;font-weight:400;src:url(../styles/fonts/OpenSans-Regular-webfont.eot);src:local("Open Sans"),local("OpenSans"),url(../styles/fonts/OpenSans-Regular-webfont.eot?#iefix) format("embedded-opentype"),url(../styles/fonts/OpenSans-Regular-webfont.woff2) format("woff2"),url(../styles/fonts/OpenSans-Regular-webfont.woff) format("woff"),url(../styles/fonts/OpenSans-Regular-webfont.ttf) format("truetype"),url(../styles/fonts/OpenSans-Regular-webfont.svg#OpenSans) format("svg")}
@@ -73,6 +72,7 @@ h2,h3{page-break-after:avoid}
 .img-thumbnail,body{background-color:#fff}
 @font-face{font-family:'Glyphicons Halflings';src:url(../styles/fonts/glyphicons-halflings-regular.eot);src:url(../styles/fonts/glyphicons-halflings-regular.eot?#iefix) format('embedded-opentype'),url(../styles/fonts/glyphicons-halflings-regular.woff2) format('woff2'),url(../styles/fonts/glyphicons-halflings-regular.woff) format('woff'),url(../styles/fonts/glyphicons-halflings-regular.ttf) format('truetype'),url(../styles/fonts/glyphicons-halflings-regular.svg#glyphicons_halflingsregular) format('svg')}
 .glyphicon{position:relative;top:1px;display:inline-block;font-family:'Glyphicons Halflings';font-style:normal;font-weight:400;line-height:1;-webkit-font-smoothing:antialiased}
+.popover,.tooltip,body{font-family:"Open Sans",Helvetica,Arial,sans-serif}
 .glyphicon-asterisk:before{content:"\002a"}
 .glyphicon-plus:before{content:"\002b"}
 .glyphicon-eur:before,.glyphicon-euro:before{content:"\20ac"}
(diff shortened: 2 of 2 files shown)


benjaminapetersen:
@jhadvig I compete with #2370 lol.

app/styles/_notifications.less
```diff
@@ -37,6 +37,13 @@ notification-drawer-wrapper {
       display: block !important;
     }
   }
+  // override .drawer-pf-notification bg when not clickable
+  .drawer-pf-notification-inner {
+    background-color: #fff;
+  }
+  .drawer-pf-notification-inner.is-clickable:hover {
+    background-color: #def3ff;
```
jwforres: Do we have a variable for this

jwforres:
got the +1 from ux, @benjaminapetersen it needs to be rebased

jwforres:
/assign @rhamilto 

since its css changes

beanh66:
@jwforres @ncameronbritt @jhadvig may be worth more discussion here. Looking through PF, we consistently use the hover on list views and tables views regardless of clickability because it helps with readability. At first glance, I was thinking removing the blue from "read" notifications made sense, but it might be a larger consistency question. The notification drawer is really just a mini list view.

app/styles/_notifications.less
rhamilto: I would probably write this like so:

```
.drawer-pf-notification-inner {
  // override .drawer-pf-notification bg when not clickable
  background-color: @color-pf-white;
  &.is-clickable:hover {
     background-color: @color-pf-blue-50;
  }
}
```

But it's not clear to me that these should have a hover state at all.  [The design docs](https://www.patternfly.org/pattern-library/communication/notification-drawer/#/design) don't specify when to use a hover, but the demo pages utilize an [optional class that turns the hover state off](https://github.com/patternfly/patternfly/blob/master/src/less/notifications-drawer.less#L186-L188), which leads me to believe that's the intended behavior (no hover).

rhamilto: And I just realized that the core demos use the optional class, but the angular demos do not.  PatternFly bug?

rhamilto: It may be advantageous to do:

```
  .drawer-pf-notification {
    padding: 0;
    &:hover {
      background-color: transparent;
    }
    &.ng-leave {
      transition: 0.25s linear all;
      opacity: 1;
      &.ng-leave-active {
        opacity: 0;
      }
    }
    &.unread:hover {
      background-color: @color-pf-blue-50;
    }
  }
```

Because this unsets the [default hover color](https://github.com/patternfly/patternfly/blob/master/src/less/notifications-drawer.less#L162) rather than masking it with another background color **and** clearly defines the intentions of the hover (only occurs on unread items).

serenamarie125:
@jwforres I am concerned deviating from PatternFly again.  I don't fully understand the issue and why we need to deviate.

spadgett:
angular-patternfly has the hover style, though. So there is already inconsistency between angular-patternfly and core.

jwforres:
@serenamarie125 we are trying to figure out what is the right thing and initially thought the issue was with our implementation, hence the PR.  At this point we need to understand what the correct behavior is from a usability perspective and then get it fixed in the appropriate place.

serenamarie125:
The PF documentation and the Angular PF implementation match, but the core implementation does not match the documentation :( 
¯\_(ツ)_/¯
Saw the issue raised by @rhamilto in PF, thank you for adding this!

benjaminapetersen:
Shall I assume we can close this one out then & inherit behavior from PF?

spadgett:
> Shall I assume we can close this one out then & inherit behavior from PF?

It's still not clear to me when to use hand vs arrow cursors. Since the Patternfly implementation is not accessible (and thus always uses the arrow), it's hard to know what the right behavior is.

1. It's strange to me to always use the same hover style, but change the cursor styles for different rows.
2. It's also strange to me not to use the hand cursor when the row is clickable.
3. It's also strange to me to use the hand cursor for read rows that don't do anything when clicked.

I guess I don't see a great option. I have a slight preference for (3) just for consistency and have the click on a read row be a no-op. But I don't think @jwforres likes that behavior.

@serenamarie125 Opinion? The design doesn't specify what cursor to use from what I can tell.

benjaminapetersen:
Yeah I tend to think this is still the most consistent behavior:



My strongest opinion is probably just to get the accessibility right, I can waffle on some of the presentation.

serenamarie125:
@spadgett @benjaminapetersen 

As I mentioned before, I do not think we should remove the hover state, that's inconsistent with the intent (PF design & ang implementation).  Can you confirm if you agree with this?

I agree that the hand cursor should only be shown if the item is unread OR when the user is hovering over the "X" or an embedded link

Re: @benjaminapetersen comment "Shall I assume we can close this one out then & inherit behavior from PF?"  the problem here is that core has a bug ... so if you are already over riding the PF behavior for other things, continue so the experience is better

beanh66:
@serenamarie125 I agree with this approach and I can create the bugs in PF

spadgett:
@serenamarie125 @beanh66 -- @benjaminapetersen captured a gif of what that looks like if you want to see:

https://github.com/openshift/origin-web-console/pull/2370#issuecomment-339375174

serenamarie125:
@spadgett i like this behavior ( the first which looks like master ).  I can't see if the mouse changes to the hand cursor when hovering over a link & "X" though, can we confirm that this would occur on items which are already read?

jwforres:
Yes the links/buttons get the hand cursor.

benjaminapetersen:
@serenamarie125 

    @spadgett @benjaminapetersen

    As I mentioned before, I do not think we should remove the hover state, that's inconsistent with the intent (PF design & ang implementation). Can you confirm if you agree with this?

I think if core has a bug/is incorrect, its worth investigating the best possible solution, even if the ang implementation matches the docs.  I think the gif above is the most correct visual for following reasons:

- Blue hover typically indicate clickable & is a stronger visual signal than cursor change. 
- A non-blue hover (soft gray color?) could be used for visual clarification.  For example, if the user is hovering over a wide table with many rows, it makes sense to highlight the hovered row.  It will improve the user's ability to correlate related data. Change to a pointer & bg color blue if the row is also clickable.

I'm wondering if our discussion is just missing distinction between these two use cases.

I can def update the PR to whatever everyone thinks is best!

serenamarie125:
@benjaminapetersen I'm not sure what else to say :)  I think that UX has given our feedback.  We think the hover should remain, as it is in the PF documentation.

serenamarie125:
I guess I would add one more thing ... rather than changing the behavior of OpenShift at this late date, why not let UX bring this back up from a design perspective to a broader audience so that it's researched and updated in PatternFly, so that things are consistent between products.  IMO, minimizing these deviations is a win in the long run for consistency through our portfolio and a better user experience for our customers.

benjaminapetersen:
Works for me! So leave the hover blue, as long as the cursor changes.  I can update.

serenamarie125:
@benjaminapetersen would you mind sending me an email with your concerns about the current approach of hover and gray versus blue?  I also vaguely remember that there was a request to make all hover treatments gray, and that may have come from OpenShift, but unsure.  I'm happy to get an effort around trying to understand your concerns, and see if we can research/solve if necessary at a higher level.

spadgett:
@serenamarie125 I feel like any hover background color on list rows / table rows makes it seem like you can click to interact. In particular, blue usually means you can click on something in Patternfly, but I think the problem is also there with a gray hover.

Not proposing we change anything in OpenShift 3.7, but I feel like it might be worth considering for Patternfly.

[1] https://getbootstrap.com/docs/3.3/css/#tables-striped

serenamarie125:
Thanks @spadgett and @benjaminapetersen by no means am i trying to ignore your input, just trying  to make sure that we understand the issues and come up with a solution that can be used by all.  I will bring up with PatternFly, and then report back with the plan.

ncameronbritt:
I think there are a several potential research questions here for PF: 
- Does the hover state suggest to users that the row/item is interactive when its not? Is that confusing?
- Does the hover state improve readability for lists, or affect usability in some other positive way?
- If the answer to both is yes, does one outweigh the other?

benjaminapetersen:
Closing as we also closed the original issue.

PR #2385, opened 2017-10-26
Wait for service classes and plans to load

### LINKED ISSUE #2384: Provisioned service ordering on overview is not stable
I see different ordering of provisioned services different times I visit the overview. I believe I've seen two rows flip while I was on the page as well.

@jwforres FYI

spadgett:
Wait for all service classes and plans to load on the overview before
checking bindability and sorting service instances.

Fixes #2384

Requires https://github.com/openshift/origin-web-common/pull/239

origin-web-common 0.0.70 -> 0.0.71

https://github.com/openshift/origin-web-common/releases/tag/v0.0.71

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -3,6 +3,7 @@
 angular.module('openshiftConsole').controller('OverviewController', [
   '$scope',
   '$filter',
+  '$q',
   '$routeParams',
   'AlertMessageService',
   'APIService',
@@ -24,6 +25,7 @@ angular.module('openshiftConsole').controller('OverviewController', [
   'OwnerReferencesService',
   'PodsService',
   'ProjectsService',
+  'PromiseUtils',
   'ResourceAlertsService',
   'RoutesService',
   'ServiceInstancesService',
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/overview.js
```diff
@@ -1336,46 +1338,69 @@ function OverviewController($scope,
...
 
-          fetchServiceClass(instance);
-          fetchServicePlan(instance);
+          promises.push(fetchServiceClass(instance));
+          promises.push(fetchServicePlan(instance));
+        });
+
+        // Wait for all promises to complete before trying to sort the
+        // instances and check bindability.
+        $q.all(promises).then(function() {
```
jwforres: my concern with doing this is that as soon as one of them fails then it will reject the q.all and this will never fire.  i dont think thats really what we want.

spadgett:
@jwforres Updated with the `$q.all` change, but requires openshift/origin-web-common#239

spadgett:
/hold cancel

Updated common release

spadgett:
Rebase only

spadgett:
I'm not sure why it's saying Jenkins failed. The build is still running.

Also I don't see Travis listed? Hm.

/retest

PR #2387, opened 2017-10-26
Update template service broker flag name

spadgett:
The broker is no longer tech preview.

/assign @jwforres 
cc @bparees


spadgett:
I have the cluster up change ready if we agree on the flag name.

bparees:
name looks ok to me.

PR #2391, opened 2017-10-27
Improve verify dist reporting

benjaminapetersen:
- Generate junit style XML file reporting success/fail
- CI expects an XML file for a better reporting experience.
- Include hack/lib/utils for helper to improve bash scripting

@jwforres @spadgett 
@stevekuznetsov

### DIFF
--- a/hack/verify-dist.sh
+++ b/hack/verify-dist.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
 set -e
+# show everything the shell is doing in case exit 1 isn't called on a fail
+# this is very noisy
+# set -o xtrace
 
-RED='\033[0;31m'
-NOCOLOR='\033[0m'
-GREEN='\033[0;32m'
+source "$(dirname "${BASH_SOURCE}")/lib/init.sh"
 
 # We don't need grunt to be installed globally for the system,  so
 # we can amend our path to look into the local node_modules for the
@@ -12,13 +13,75 @@ GREEN='\033[0;32m'
 repo_root="$( dirname "${BASH_SOURCE}" )/.."
 export PATH="${PATH}:${repo_root}/node_modules/grunt-cli/bin"
 
+# vars to generate a fake junit test output file
+TIMESTAMP=$(date +"%x %r %Z")
+SUITE_NAME="$BASH_SOURCE/$FILE"
+TEST_NAME="User should generate /dist by running grunt build"
+TESTS_COUNT=1
+SKIPPED_COUNT=0
+DISABLED=0
+FAILURE_COUNT=0
+ERROR_COUNT=0
+PASSED=false
+POTENTIAL_FAILED_TEST=""
+
+# run the build to generate dist
+os::log::info "Rebuilding dist..."
 grunt build
 
-echo "Verifying that checked in built files under dist match the source..."
-if [[ $(git status -s -u dist*) ]]; then
-    echo -e "${RED}Built /dist does not match what is committed, run 'grunt build' and include the results in your commit.${NOCOLOR}"
-    git diff --exit-code dist*
-    exit 1
+# validate generated against what is committed
+os::log::info "Verifying that checked in built files under dist match the source..."
+GIT_STATUS="$( git status --short --untracked-files -- dist/ )"
+GIT_DIFF="$( git diff )"
+
+# -n = not null
+if [[ -n "$GIT_STATUS" ]]; then
+  FAILURE_COUNT=1
+  FAILURE_MESSAGE="Built /dist does not match what is committed, run grunt build and include the results in your commit."
+  ERROR_COUNT=1
+  POTENTIAL_FAILED_TEST="<failure type=\"DIST_FAILURE\" message=\"$FAILURE_MESSAGE\"><![CDATA[ $GIT_DIFF ]]></failure>"
+  PASSED=false
 else
-    echo -e "${GREEN}Verified. Rebuilt /dist matches what has been committed.${NOCOLOR}"
+  PASSED=true
+fi
+
+# generate a dummy junit style xml file for better CI output
+DEST_OUTPUT_PATH="${repo_root}/test/junit/"
+DEST_OUTPUT_FILE="${DEST_OUTPUT_PATH}/verify-dist.xml"
+
+mkdir -p "${DEST_OUTPUT_PATH}"
+
+# print the results
+cat <<END_OF_XML >"${DEST_OUTPUT_FILE}"
+<?xml version="1.0" encoding="UTF-8" ?>
+<testsuites>
+  <properties>
+    <property name="skuznets" value="is amazing"/>
+  </properties>
+  <testsuite
+    name="$SUITE_NAME"
+    timestamp="$TIMESTAMP"
+    errors="$ERROR_COUNT"
+    tests="$TESTS_COUNT"
+    skipped="$SKIPPED_COUNT"
+    disabled="$DISABLED"
+    failures="$FAILURE_COUNT"
+    time="1">
+    <testcase
+      classname="$SUITE_NAME"
+      name="$TEST_NAME"
+      time="1">
+      $POTENTIAL_FAILED_TEST
+    </testcase>
+  </testsuite>
+</testsuites>
+END_OF_XML
+
+# finally, print results message & write the file
+if [[ "${PASSED}" == "true" ]]; then
+  os::log::success "Verified. Rebuilt /dist matches what has been committed."
+else
+  os::log::fatal "Built /dist does not match what is committed, run 'grunt build' and include the results in your commit."
+  git diff --exit-code dist*
+  exit 1
 fi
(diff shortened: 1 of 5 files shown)


benjaminapetersen:
Hmm. I break server.  Oops.

screen shot 2017-10-27 at 4 01 49 pm

/test again?

hack/verify-dist.sh
```diff
@@ -5,20 +5,80 @@ set -e
 RED='\033[0;31m'
 NOCOLOR='\033[0m'
 GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
```
stevekuznetsov: Are you intending to use the `os::log::*` functions that you pulled in with this PR?

hack/verify-dist.sh
```diff
@@ -5,20 +5,80 @@ set -e
...
+PASSED=false
+POTENTIAL_FAILED_TEST=""
+
+# run the build to generate dist
 grunt build
 
-echo "Verifying that checked in built files under dist match the source..."
+# validate generated against what is committed
+echo -e "${YELLOW}Verifying that checked in built files under dist match the source...${NOCOLOR}"
 if [[ $(git status -s -u dist*) ]]; then
```
stevekuznetsov: This is using an implicit check right now on the output of the subshell. Is the intent to update the dist files and if there are any changes, fail? Mabe use:

```sh
if [[ -n "$( git status --short --untracked-files -- dist/ )" ]]; then
```

Just to explicitly check the output for empty string

hack/verify-dist.sh
```diff
@@ -5,20 +5,80 @@ set -e
...
+  PASSED=false
+else
+  PASSED=true
+fi
+
+# generate a dummy junit style xml file for better CI output
+DEST_OUTPUT_PATH="${repo_root}/test/junit/"
+DEST_OUTPUT_FILE="${DEST_OUTPUT_PATH}/verify-dist.xml"
+
+if [ ! -d DEST_OUTPUT_PATH ]; then
```
stevekuznetsov: Always prefer `[[` over `[`

stevekuznetsov: Also right now this is checking if the directory `"DEST_OUTPUT_PATH"` exists, not the var:

```sh
if [[ ! -d "${DEST_OUTPUT_PATH}" ]]; then
```

hack/verify-dist.sh
```diff
@@ -5,20 +5,80 @@ set -e
...
+  mkdir -p $DEST_OUTPUT_PATH
```
stevekuznetsov: Always quote and brace variable expansions:

```sh
mkdir -p "${DEST_OUTPUT_PATH}"
```

stevekuznetsov: Also, `mkdir -p` will not fail if the dir already exists, so you can get rid of the `if` statement here and just run `mkdir -p`

hack/verify-dist.sh
```diff
@@ -5,20 +5,80 @@ set -e
...
+fi
+
+XML_JUNIT_OUTPUT=$(cat <<END_OF_XML
```
stevekuznetsov: No need to save this data into a var, just use:


```sh
cat <<END_OF_XML >"${DEST_OUTPUT_FILE}"
<?xml version="1.0" encoding="UTF-8" ?>
<testsuites>
...
END_OF_XML
```

hack/verify-dist.sh
```diff
@@ -5,20 +5,80 @@ set -e
...
+      name="$TEST_NAME"
+      time="1">
+      $POTENTIAL_FAILED_TEST
+    </testcase>
+  </testsuite>
+</testsuites>
+END_OF_XML)
+
+# finally, print results message & write the file
+if [ ! $PASSED ]; then
```
stevekuznetsov: Prefer `[[` over `[`

stevekuznetsov: Don't use implicit string comparison, so if you want to check that `"${PASSED}"` is set, use:

```sh
if [[ -n "${PASSED}" ]]; then
```

Or, since you know the value that it will hold,

```sh
if [[ "${PASSED}" == "false" ]];
```

hack/verify-dist.sh
benjaminapetersen: yup. updated.

benjaminapetersen:
bash is weird. but updates made. lets see if this next test fails (which is desirable).

benjaminapetersen:
@jwforres @spadgett this is ready for review.  failed page looks like this now:

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_origin-web-console/2391/test_pull_request_origin_web_console/712/

Thoughts?  
(I still have a deliberately broken `/dist` here to force the fail)

benjaminapetersen:
New report: 

screen shot 2017-10-30 at 4 21 42 pm

jwforres:
is there a way for us to get the actual diff reported? or is that more output than we want here

benjaminapetersen:
This page can only show output in junit xml format.  So it would probably be a bit gnarly to try to get the output dumped here.  You can still click the build log link & see the output, which includes the diff.  

I'd be willing to give this a swing, and if its insufficient do a follow-on and try to capture the diff output on this page somehow.  I'm sure @stevekuznetsov would appreciate me adding a bit more crazy bash.

benjaminapetersen:
With the screenshot added, I'm going to remove the incorrect dist.

stevekuznetsov:
You can capture the output of the `diff` and shove it into the XML output

hack/verify-dist.sh
```diff
@@ -1,24 +1,83 @@
...
 
-echo "Verifying that checked in built files under dist match the source..."
-if [[ $(git status -s -u dist*) ]]; then
-    echo -e "${RED}Built /dist does not match what is committed, run 'grunt build' and include the results in your commit.${NOCOLOR}"
-    git diff --exit-code dist*
-    exit 1
+# validate generated against what is committed
+os::log::info "Verifying that checked in built files under dist match the source..."
+
+if [[ -n "$( git status --short --untracked-files -- dist/ )" ]]; then
```
stevekuznetsov: Could capture diff here

hack/verify-dist.sh
```diff
@@ -1,24 +1,83 @@
...
+# if [[ $(git status -s -u dist*) ]]; then
+  FAILURE_COUNT=1
+  ERROR_COUNT=1
+  POTENTIAL_FAILED_TEST='<failure type="DIST_FAILURE">Built /dist does not match what is committed, run `grunt build` and include the results in your commit.</failure>'
```
stevekuznetsov: and put the output into this block

hack/verify-dist.sh
```diff
@@ -1,24 +1,83 @@
...
+  </testsuite>
+</testsuites>
+END_OF_XML
+
+# finally, print results message & write the file
+if [[ "${PASSED}" == "true" ]]; then
+  os::log::success "Verified. Rebuilt /dist matches what has been committed."
 else
-    echo -e "${GREEN}Verified. Rebuilt /dist matches what has been committed.${NOCOLOR}"
+  os::log::error "Built /dist does not match what is committed, run 'grunt build' and include the results in your commit."
```
stevekuznetsov: `error` + `exit` = `fatal`

```sh
os::log::fatal "Built /dist does not match what is committed, run 'grunt build' and include the results in your commit.
$( git diff --exit-code dist* )"
```

benjaminapetersen:
Progress.  This isn't showing the diff'd lines, but it is showing which files failed:

screen shot 2017-11-02 at 2 48 12 pm 1

benjaminapetersen:
/test

jenkins, plz

benjaminapetersen:
We now have a dist in the xml:

screen shot 2017-11-03 at 9 27 04 am

hack/verify-dist.sh
```diff
@@ -1,24 +1,87 @@
...
+os::log::info "Verifying that checked in built files under dist match the source..."
+GIT_STATUS="$( git status --short --untracked-files -- dist/ )"
+GIT_DIFF="$( git diff )"
+
+# -n = not null
+if [[ -n "$GIT_STATUS" ]]; then
+  FAILURE_COUNT=1
+  FAILURE_MESSAGE="Built /dist does not match what is committed, run grunt build and include the results in your commit."
+  ERROR_COUNT=1
+  POTENTIAL_FAILED_TEST="<failure type=\"DIST_FAILURE\" message=\"$FAILURE_MESSAGE\"><![CDATA[ $GIT_DIFF ]]></failure>"
```
benjaminapetersen: @stevekuznetsov so long as this `CDATA` block isn't objectionable, I think I'm set on this PR & can stop bothering you about bash for a bit.  :stuck_out_tongue:

stevekuznetsov:
/hold
@jwforres should final approve

benjaminapetersen:
@jwforres @spadgett good to go.

benjaminapetersen:
looking forward to this improving the /dist mismatch.

PR #2392, opened 2017-10-27
Check 'auth' field when displaying .dockercfg config

jhadvig:
While working on https://bugzilla.redhat.com/show_bug.cgi?id=1506998 I've noticed that we are not checking for `auth` field in the old **.dockercfg** format, which looks like:
```
{
        "registryaddress": {
                "auth": "auth string",
                "email": "email@domain.com"
        }
}
```

The `registryaddress` could also contain `username` and `password` fields.

@spadgett PTAL

Will add the dist after review

### DIFF
--- a/app/scripts/services/secrets.js
+++ b/app/scripts/services/secrets.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module("openshiftConsole")
-  .factory("SecretsService", function(){
+  .factory("SecretsService", function($filter, Logger, NotificationsService){
 
     var groupSecretsByType = function(secrets) {
       var secretsByType = {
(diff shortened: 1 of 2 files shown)


app/scripts/services/secrets.js
```diff
@@ -28,17 +28,29 @@ angular.module("openshiftConsole")
       return secretsByType;
     };
 
+    var setDataParams = function(decodedSecretData, serverName, data) {
```
spadgett: If this is specific to dockercfg, we should have that somewhere in the function name.

app/scripts/services/secrets.js
```diff
@@ -28,17 +28,29 @@ angular.module("openshiftConsole")
...
+        usernamePassword = window.atob(data.auth).split(":");
+      } else if (data.username && data.password) {
+        usernamePassword = [data.username, data.password];
+      }
+
+      decodedSecretData.auths[serverName] = {
+        username: usernamePassword[0],
+        password: usernamePassword[1],
+        email: data.email
+      };
```
spadgett: I would let the caller set the value in the auths object and just let this function return the data. Then the function takes fewer arguments and doesn't need to know anything about the structure of the `decodedSecretData` object passed to it.

(untested)

```js
var getDockercfgServerParams = function(serverData) {
  var params = _.pick(serverData, ['email', 'username', 'password']);
  if (serverData.auth) {
    try {
      // Decode Base64-encoded username:password string.
      _.spread(window.atob(data.auth).split(":"), function(username, password) {
        params.username = username;
        params.password = password;
      });
    } catch(e) {}
  }

  return params;
};
```

then below in the `decodeDockercfg` function

```js
decodedSecretData.auths[serverName] = getDockercfgServerParams(data);
```

https://lodash.com/docs/4.17.4#pick
https://lodash.com/docs/4.17.4#spread

app/scripts/services/secrets.js
```diff
@@ -28,17 +28,29 @@ angular.module("openshiftConsole")
       return secretsByType;
     };
 
+    var setDataParams = function(decodedSecretData, serverName, data) {
+      var usernamePassword = [];
+      if (data.auth) {
+        usernamePassword = window.atob(data.auth).split(":");
```
spadgett: We should catch exceptions to avoid a runtime error if it's not properly encoded.

https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/atob

app/scripts/services/secrets.js
jhadvig: This method is not intended to be specific for `dockercfg` or `config.json`. It's purpose is to assemble an object that will contain server address, username, password.

jhadvig:
@spadgett I've updated the PR based on your comments. 
I know that you suggested to rename the method for getting the server params to `getDockercfgServerParams` but that method could be Docker config format agnostic (should cover both `.dockercfg` and `docker/config.json`).
Also I've refactored both `decodeDockercfg` and `decodeDockerconfigjson` method into single `decodeDockerConfig` method since both of them have been doing almost the same operations, there have been just slight differences which should  be covered with a simple `if`.
Also added decoding error handling, here is the example with error notification:
auth_err


Tested the change with single and multi server `.dockercfg` and `docker/config.json` and works as expected.

PTAL

spadgett:
/kind bug

@jhadvig You have a dist mismatch

spadgett:
Hm.

```
pm ERR! phantomjs-prebuilt@2.1.15 install: `node install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the phantomjs-prebuilt@2.1.15 install script 'node install.js'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the phantomjs-prebuilt package,
npm ERR! not with npm itself.
```

/retest

PR #2400, opened 2017-10-30
Follow-on updates from jennyhaines

sg00dwin:
Requested changes
https://github.com/openshift/origin-web-console/issues/2362#issuecomment-339710253 and https://github.com/openshift/origin-web-console/issues/2182#issuecomment-340546643

screen shot 2017-10-30 at 3 49 34 pm

screen shot 2017-10-30 at 4 08 01 pm

### DIFF
--- a/app/styles/_kve.less
+++ b/app/styles/_kve.less
@@ -125,11 +125,13 @@
     position: relative;
     table-layout: fixed;
     width: 100%;
-    @media(min-width: @screen-xs-min)  {
-      .environment-from-editor-button {
+    .environment-from-editor-button {
+      top: 26px; // position to align with config-map ui-select when stacked
+      @media(min-width: @screen-xs-min)  {
         float: left;
         padding: 0 5px;
         position: relative;
+        top: 0; // position when not stacked
         width: auto;
       }
     }
@@ -180,6 +182,25 @@
       @media(min-width: @screen-md-min) {
         padding-top: 0;
         padding-left: 5px;
+      }
+    }
+  }
+}
+
+.environment-from-editor-header {
+  &.config-map-header,
+  &.prefix-header {
+    @media(max-width: @screen-xxs-max) {
+      display: none;
+      width: 100%;
+    }
+    &.config-map-header-mobile,
+    &.prefix-header-mobile {
+      display: block;
+      margin-bottom: 0;
+      margin-top: 5px;
+      @media(min-width: @screen-xs-min) {
+        display: none;
       }
     }
   }
--- a/app/styles/_tooltip.less
+++ b/app/styles/_tooltip.less
@@ -2,6 +2,11 @@
 // Tooltips
 // --------------------------------------------------
 
+// TODO Go thru and apply consistantly across pages
+.tooltip-default-icon {
+  cursor: help;
+  font-size: @font-size-base - 1;
+}
 .tooltip-inner {
   .word-break(); // so that long, unbroken strings don't overflow
 }
--- a/app/views/directives/edit-environment-from.html
+++ b/app/views/directives/edit-environment-from.html
@@ -2,20 +2,17 @@
   <div
     ng-if="$ctrl.showHeader"
     class="environment-from-entry environment-from-editor-entry-header">
-    <div class="form-group environment-from-editor-header value-header">
-      <div class="input-group">
+    <div class="environment-from-editor-header config-map-header">
         Config Map/Secret
-      </div>
     </div>
-    <div class="form-group environment-from-editor-header value-header">
-      <div class="input-group"
-           ng-if="!$ctrl.isEnvFromReadonly() && $ctrl.hasOptions()">
+    <div
+      class="environment-from-editor-header prefix-header"
+      ng-if="!$ctrl.isEnvFromReadonly() && $ctrl.hasOptions()">
         Prefix
-        <small class="pficon pficon-help"
+        <small class="pficon pficon-help tooltip-default-icon"
                aria-hidden="true"
                data-toggle="tooltip"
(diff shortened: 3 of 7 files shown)


app/views/directives/key-value-editor.html
```diff
@@ -4,14 +4,10 @@
     ng-if="showHeader"
     class="key-value-editor-entry key-value-editor-entry-header">
     <div class="form-group key-value-editor-header key-header">
```
spadgett: Do we still need `form-group`?

app/views/directives/key-value-editor.html
```diff
@@ -4,14 +4,10 @@
...
-      <div class="input-group">
-        <span class="help-block">{{keyPlaceholder}}</span>
-      </div>
+      {{keyPlaceholder}}
     </div>
     <div class="form-group key-value-editor-header value-header">
```
spadgett: Do we still need `form-group`?

app/styles/_kve.less
```diff
@@ -192,6 +192,12 @@
   margin-bottom: 0;
   padding-right: 5px;
   width: 50%;
+  &.prefix-header {
+    display: none;
```
spadgett: If the header is totally hidden, I'm not sure how you know what that field does on mobile.

app/views/directives/key-value-editor.html
sg00dwin: well, it's setting the margin-bottom but that could be set using the specific class *-header

If so we'd need to do the same for the KVE ones above to be consistent

app/styles/_kve.less
sg00dwin: well when empty it does display `placeholder="Add prefix"` :)
I'll append it to the single header and show at mobile.

screen shot 2017-10-30 at 5 24 10 pm

sg00dwin:
@spadgett updated 

screen shot 2017-10-30 at 5 24 10 pm

spadgett:
Hey @sg00dwin I'd suggest keeping the current layout for desktop, but stacking the labels above at mobile. Something like

```
Config Map / Secret
[my-config-map]

Prefix
_____________

Config Map / Secret
[my-secret]

Prefix
MY_PREFIX_
```

sg00dwin:
@spadgett Is this the structure you suggest?

env-from sp

spadgett:
Yeah, that's what I was thinking. Seeing it, I'm good with either approach, though.

@beanh66 Opinion?

sg00dwin:
@spadgett PTAL

screen shot 2017-10-31 at 13 53 51

spadgett:
openshift web console 2017-10-31 16-44-09

spadgett:
Nevermind on the prefix width: I see it's to give room for the view details link

app/styles/_tooltip.less
```diff
@@ -2,6 +2,10 @@
 // Tooltips
 // --------------------------------------------------
 
+.tooltip-default-icon {
+  font-size: @font-size-base - 1;
+  vertical-align: top;
```
spadgett: I like creating a common style, but currently most other tooltip icons aren't top-aligned. For instance,

openshift web console 2017-10-31 16-59-39

app/views/directives/edit-environment-from.html
```diff
@@ -2,20 +2,17 @@
...
     </div>
-    <div class="form-group environment-from-editor-header value-header">
-      <div class="input-group"
-           ng-if="!$ctrl.isEnvFromReadonly() && $ctrl.hasOptions()">
+    <div
+      class="environment-from-editor-header prefix-header"
+      ng-if="!$ctrl.isEnvFromReadonly() && $ctrl.hasOptions()">
         Prefix
-        <small class="pficon pficon-help"
+        <small class="pficon pficon-help tooltip-default-icon"
```
spadgett: We should add a style to switch the cursor to the help icon on hover like other tooltips.

app/views/directives/edit-environment-from.html
```diff
@@ -62,7 +64,16 @@
         </div>
       </div>
 
-      <div class="form-group environment-from-input prefix">
+      <div class="environment-from-input prefix">
+        <div
```
spadgett: Ideally this is a `<label for="...">` for accessibility, but that's a bit tricky since you have to generate IDs. I'm OK leaving that for later.

app/views/directives/edit-environment-variables.html
```diff
@@ -1,7 +1,7 @@
 <form ng-if="$ctrl.apiObject" name="$ctrl.form" class="mar-bottom-xl">
   <confirm-on-exit ng-if="$ctrl.canIUpdate && !$ctrl.ngReadonly" dirty="$ctrl.form.$dirty"></confirm-on-exit>
   <div ng-repeat="container in $ctrl.containers">
-    <h3>Container {{container.name}}</h3>
+    <h2>Container {{container.name}}</h2>
```
spadgett: I know this change was suggested by @jennyhaines, but I'm a little concerned that we're not making it on other browse pages / other tabs. All of the other headers are `h3` on the browse pages.

We might hold off on this and look at it in the large context of all the pages later. Any objections @jennyhaines ?

If so, we'd probably keep the "Environment From" header an `h4` (without the `section-label` class) keep the hierarchy.

app/styles/_tooltip.less
sg00dwin: Yeah, thought about that. Maybe add a todo to sweep through and common class.

sg00dwin:
@spadgett updated with changes PTAL

jennyhaines:
@spadgett @sg00dwin - understandable about the text size change. I think that works for now so the text styles map across the other tab content.

PR #2404, opened 2017-10-31
Bug 1507753 - Make config page and environment tab actions consistent

jhadvig:
When editing DC environment variables the action links are following: 
current-env

But the DC's Environment tab actions links are:
old-env

To be consistent (eg BC editor and BC Environment tab)  the action links wording in the DC env tab should be:
new-env
 
@spadgett PTAL

### DIFF
--- a/app/views/create/fromimage.html
+++ b/app/views/create/fromimage.html
@@ -283,8 +283,8 @@
                                 value-from-selector-options="valueFromObjects"
                                 key-validator="[a-zA-Z_][a-zA-Z0-9_]*"
                                 key-validator-error-tooltip="A valid environment variable name is an alphanumeric (a-z and 0-9) string beginning with a letter that may contain underscores."
-                                add-row-link="Add Environment Variable"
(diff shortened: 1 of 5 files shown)


spadgett:
Let's change these instances to use the wording from `edit-environment-variables`

```
app/views/directives/deploy-image.html
171:            add-row-with-selectors-link="Add Environment Variable Using a Config Map or Secret"></key-value-editor>

app/views/edit/deployment-config.html
75:                          add-row-with-selectors-link="Add Environment Variable Using a Config Map or Secret"></key-value-editor>
360:                      add-row-with-selectors-link="Add Environment Variable Using a Config Map or Secret"></key-value-editor>

app/views/directives/edit-lifecycle-hook.html
61:            add-row-with-selectors-link="Add Environment Variable Using a Config Map or Secret"

app/views/create/fromimage.html
287:                                add-row-with-selectors-link="Add Environment Variable Using a Config Map or Secret"></key-value-editor>
```

jhadvig:
@spadgett comments addressed

spadgett:
Thanks @jhadvig 

/kind bug
/lgtm

PR #2406, opened 2017-10-31
Fix bugzilla 15077030 rolebindings for serviceaccounts

benjaminapetersen:
Fix bugzilla 15077030 where deleting a rolebinding for a serviceaccount can delete additional rolebindings for serviceaccounts from another namespace.

Hopefully will follow-up with some tests soon.  Time elapse is def showing some subtle bugs here.

### DIFF
--- a/test/spec/services/membership/roleBindingsSpec.js
+++ b/test/spec/services/membership/roleBindingsSpec.js
@@ -88,10 +88,13 @@ describe('RoleBindingsService', function() {
     it('should update the rolebinding by removing the subject from the subject list', function() {
       var bindings = [{
         roleRef: { name: 'admin'},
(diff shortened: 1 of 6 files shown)


benjaminapetersen:
yup, updating test

spadgett:
Fix looks good to me pending test changes.

/kind bug

test/spec/services/membership/roleBindingsSpec.js
```diff
@@ -102,13 +105,61 @@ describe('RoleBindingsService', function() {
       $rootScope.$digest();
     });
```
benjaminapetersen: Added new test to cover this use case here.

benjaminapetersen:
@spadgett tests updated & new test added.

PR #2409, opened 2017-10-31
Fix bugzilla 1507822 - Update plan info on service instance update

jeff-phillips-18:
Fixes bugzilla [1507822](https://bugzilla.redhat.com/show_bug.cgi?id=1507822) where the service plan information is not updated automatically when the service instance is edited.

### DIFF
--- a/app/scripts/controllers/serviceInstance.js
+++ b/app/scripts/controllers/serviceInstance.js
@@ -51,6 +51,7 @@ angular.module('openshiftConsole')
     var watches = [];
     var secretWatchers = [];
     var serviceClassPromise;
+    var servicePlansPromise;
 
     var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
     var serviceInstanceReady = $filter('isServiceInstanceReady');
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/serviceInstance.js
```diff
@@ -128,33 +128,39 @@ angular.module('openshiftConsole')
       updateParameterData();
     };
 
-    var updateServiceClass = function() {
-      // If we've previously loaded the service class or a request is in flight, don't do anything.
-      if (!$scope.serviceInstance || $scope.serviceClass || serviceClassPromise) {
-        return;
-      }
+    var updateServicePlan = function() {
+      Catalog.getServicePlansForServiceClass($scope.serviceClass).then(function (plans) {
```
spadgett: This re-requests plans for the service class on every watch update, correct? It would be better to only request them once on page load and remember them. Watches can trigger often due to status updates on the instance when it's being provisioned.

jeff-phillips-18: Updated.

app/scripts/controllers/serviceInstance.js
```diff
@@ -128,33 +129,52 @@ angular.module('openshiftConsole')
...
     };
 
-    var updateServiceClass = function() {
-      // If we've previously loaded the service class or a request is in flight, don't do anything.
-      if (!$scope.serviceInstance || $scope.serviceClass || serviceClassPromise) {
+    var updateServicePlan = function() {
+      var servicePlanName = _.get($scope.serviceInstance, 'spec.clusterServicePlanRef.name');
+      $scope.plan = _.find($scope.servicePlans, function(plan) {
+        return _.get(plan, 'metadata.name') ===  servicePlanName;
+      });
```
spadgett: You can use the shorthand here:

```js
$scope.plan = _.find($scope.servicePlans, { metadata: { name: servicePlanName } });
```

app/scripts/controllers/serviceInstance.js
```diff
@@ -128,33 +129,52 @@ angular.module('openshiftConsole')
...
-
-        Catalog.getServicePlansForServiceClass($scope.serviceClass).then(function (plans) {
-          plans = plans.by('metadata.name');
-
+      if (!$scope.servicePlans) {
+        servicePlansPromise = Catalog.getServicePlansForServiceClass($scope.serviceClass).then(function (plans) {
           var servicePlanName = _.get($scope.serviceInstance, 'spec.clusterServicePlanRef.name');
-          $scope.servicePlans = _.reject(plans, function(plan) {
+          $scope.servicePlans = _.reject(plans.by('metadata.name'), function(plan) {
             return _.get(plan, 'status.removedFromBrokerCatalog') && (plan.metadata.name !== servicePlanName);
```
spadgett: I just realized if you change the instance plan twice without leaving the page, you'll probably be able to select a removed plan again if it was initially selected. Nothing related to this PR, but we might want to test and open an issue.

app/scripts/controllers/serviceInstance.js
jeff-phillips-18: Thanks, updated.

app/scripts/controllers/serviceInstance.js
jeff-phillips-18: Yeah, I thought of the same thing but really really really edge case. Can't think of a good fix except to re-request the plans on a change.

spadgett: Check if current instance plan != $scope.plan on instance watch updates and re-filter the existing $scope.servicePlans list?

jeff-phillips-18: True, I was actually thinking of plans that were removed after the original fetch.

PR #2412, opened 2017-10-31
Fix missing RolloutCancelled event in notification drawer

benjaminapetersen:
Fix bugzilla #1490720 ([link](https://bugzilla.redhat.com/show_bug.cgi?id=1490720))

Adds this event to the notification drawer:

screen shot 2017-10-31 at 4 05 47 pm


app/scripts/constants.js
```diff
@@ -189,6 +189,7 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     BuildConfigInstantiateFailed: true,
     // Deployment
     DeploymentCancelled: true,
+    RolloutCancelled: true,
```
spadgett: nit: better to keep it alphabetical by type so it's easier to find things

app/scripts/constants.js
```diff
@@ -189,6 +189,7 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     BuildConfigInstantiateFailed: true,
     // Deployment
     DeploymentCancelled: true,
```
spadgett: Does this reason exist if RolloutCancelled is for cancelled deployments

app/scripts/constants.js
spadgett: If this is for deployment configs, better to move it under that section

spadgett:
I don't think `DeploymentCancelled` is a real event reason. I'd like to remove that as part of this PR.

app/scripts/constants.js
benjaminapetersen: My doc indicates that it did in my [first review](https://docs.google.com/document/d/11SYwEtnq0S6PT9FPXZuyktCdVsdoYfU_9lDghkTf02A/edit) of all the events, but I don't see a `c.recorder.Eventf` with `grep -ri 'DeploymentCancellationFailed' pkg` so I'm pretty sure its not real.   I'll remove this.

spadgett: Might have changed from 3.6 -> 3.7

app/scripts/constants.js
benjaminapetersen: will do.

benjaminapetersen:
Updated.

spadgett:
```diff
Built /dist does not match what is committed, run 'grunt build' and include the results in your commit.
diff --git a/dist/scripts/scripts.js b/dist/scripts/scripts.js
index 354721d..2ae32d6 100644
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -723,11 +723,10 @@ BuildCompleted: !0,
 BuildFailed: !0,
 BuildStarted: !0,
 BuildConfigInstantiateFailed: !0,
-DeploymentCancelled: !0,
-RolloutCancelled: !0,
 Failed: !0,
 DeploymentCreated: !0,
 DeploymentCreationFailed: !0,
+RolloutCancelled: !0,
 FailedRescale: !0,
 SuccessfulRescale: !0,
 BackOff: !0,
```

benjaminapetersen:
/test

`/dist` should match.

spadgett:
@benjaminapetersen I can look at the commit and see the dist is wrong :)

benjaminapetersen:
:neutral_face: more :coffee: :coffee: :coffee:

PR #2414, opened 2017-11-01
add k8s compliant validation for label

mjudeikis:
If I have storageclass like this: 
```
kind: StorageClass
apiVersion: storage.k8s.io/v1beta1
metadata:
  name: fast
provisioner: kubernetes.io/mystorageprvider
parameters:
  pool: default
  description: Kubernetes volume
  fsType: ext4
  adminSecretNamespace: default
  adminSecretName: mystorageprvider-secret
```

I want to use UI and create storage claim and provide label selector. As per K8S docs labels can have . and / as keys (https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)

But our validation does not allow this from UI:
image

We want this to be valid:
image

### DIFF
--- a/app/views/directives/osc-persistent-volume-claim.html
+++ b/app/views/directives/osc-persistent-volume-claim.html
@@ -221,8 +221,8 @@
           entries="claim.selectedLabels"
           key-placeholder="label"
           value-placeholder="value"
-          key-validator="[a-zA-Z][a-zA-Z0-9_-]*"
-          key-validator-error-tooltip="A valid label name is an alphanumeric (a-z and 0-9) string beginning with a letter that may contain underscores and dashes."
+          key-validator="[a-zA-Z][a-zA-Z0-9_.-\/]*"
+          key-validator-error-tooltip="A valid label name is an alphanumeric (a-z and 0-9) string beginning with a letter that may contain underscores, dots and dashes."
           add-row-link="Add Label"></key-value-editor>
       </fieldset>
     </div>
--- a/dist/scripts/templates.js
+++ b/dist/scripts/templates.js
@@ -8323,7 +8323,7 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
     "<a ng-href=\"{{'selector_label' | helpLink}}\" target=\"_blank\">Learn More&nbsp;<i class=\"fa fa-external-link\" aria-hidden=\"true\"></i></a>\n" +
     "</div>\n" +
     "</div>\n" +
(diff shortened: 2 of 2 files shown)


app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -221,8 +221,8 @@
           entries="claim.selectedLabels"
           key-placeholder="label"
           value-placeholder="value"
-          key-validator="[a-zA-Z][a-zA-Z0-9_-]*"
-          key-validator-error-tooltip="A valid label name is an alphanumeric (a-z and 0-9) string beginning with a letter that may contain underscores and dashes."
+          key-validator="[a-zA-Z][a-zA-Z0-9_.-\/]*"
```
spadgett: It looks like the label validation is more complicated than this, and the original regex on this page was incorrectly copied from an env variables editor. See

https://github.com/spadgett/origin-web-console/blob/57ae92d906931b0f60b6fa73833c9dc800f26fca/app/scripts/directives/labels.js#L38-L41

Can you try to use the `label-editor` directive here instead? Here's an example:

https://github.com/spadgett/origin-web-console/blob/57ae92d906931b0f60b6fa73833c9dc800f26fca/app/views/create-route.html#L21-L26

mjudeikis: But this one still does not do "/" - like kubernetes.io/test ?

mjudeikis: Do we want to update label regexp too?

spadgett: It should split on `/` before testing the regex if you look further down.

https://github.com/spadgett/origin-web-console/blob/57ae92d906931b0f60b6fa73833c9dc800f26fca/app/scripts/directives/labels.js#L75

mjudeikis: k, let me check what I can do :)

mjudeikis: k, having some trouble to use `label-editor` directive...  there is some styles clashing, which makes directive not visible (my assumption)... my UI skills are rusty :/

mjudeikis: when it comes to directives and angularjs - I have to say - pass :/

spadgett: @mjudeikis I hope you don't mind: I went ahead and opened #2423 to use the label-editor

mjudeikis: all good :) Thanks

PR #2415, opened 2017-11-01
Show proper type of imported item in import wizard results

jhadvig:
Show the kind thats actually being created and not default "YAML / JSON"
propertype

Although not sure if in case of List we dont wanna print all the created resources from the List itself, but on the other hand we are not doing it with Templates.

@spadgett @rhamilto thoughts ?

### DIFF
--- a/app/scripts/directives/fromFile.js
+++ b/app/scripts/directives/fromFile.js
@@ -302,7 +302,8 @@ angular.module("openshiftConsole")
           }
           else if ($scope.isDialog) {
             $scope.$emit('fileImportedFromYAMLOrJSON', {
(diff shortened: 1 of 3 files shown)


spadgett:
@jhadvig You and I are working on the same issues. I have some changes locally for this :/

I think it should say something like "Stateful set my-stateful-set created." I realize the "completed" wording is not new.

spadgett:
Maybe just "Resources created successfully." for a list?

I know it might be hard to get these exact messages in `next-steps` given the way the directive works.

spadgett:
Since this needed to get in today for 3.7, I opened a separate PR (#2426). Closing since #2426 has merged.

PR #2418, opened 2017-11-01
Adjust spacing around 'Events' heading in sidebar on monitoring page

rhamilto:
In #2389 I adjusted the spacing around the h1 so borders aligned, but I overlooked the border underneath "Events" in the monitoring page sidebar.  This fixes that, and does so in a way that the border is consistent across browsers (it wasn't before--only Chrome had a correctly aligned border).

Before:
screen shot 2017-11-01 at 10 46 26 am

After:
screen shot 2017-11-01 at 10 46 01 am

### DIFF
--- a/app/styles/_sidebar-right.less
+++ b/app/styles/_sidebar-right.less
@@ -139,10 +139,13 @@
     display: flex;
     justify-content: space-between;
     // So that sidebar bottom border aligns with main content header bottom border.
-    margin-top: 17.5px;
+    margin-top: 0;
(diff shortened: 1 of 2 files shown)


serenamarie125:
@rhamilto 
Should Events & View Details be bottom aligned?  Looks like View Details is a pixel or two higher

rhamilto:
> Should Events & View Details be bottom aligned? Looks like View Details is a pixel or two higher

Probably.  Lemme see what's going on there.  :)

PR #2423, opened 2017-11-01
Use label-editor for PVC labels

spadgett:
Fixes a problem where the regex used for validating label names is
wrong. Names with a `/` were incorrectly blocked.

See discussion in #2414

openshift web console 2017-11-01 12-12-24

openshift web console 2017-11-01 12-19-48

/kind bug
/assign @jwforres 
cc @mjudeikis


jwforres:
The Learn More feels out of place as a block, try the inline one instead.

spadgett:
@jwforres Updated to use a checkbox instead:

openshift web console 2017-11-01 12-45-38

mjudeikis:
@rusenask :) FIY . Done.

PR #2430, opened 2017-11-01
[WIP] Set ui-select dropdown to append-to-body with z-index in order to overlay outside of overflow scroll area and prevent Bug 1479114 where options are not selectable. Bump origin-web-common version 0.0.73

sg00dwin:
fixes [bug 1479114 ](https://bugzilla.redhat.com/show_bug.cgi?id=1479114)

### DIFF
--- a/app/views/directives/istag-select.html
+++ b/app/views/directives/istag-select.html
@@ -3,7 +3,7 @@
     <div class="row">
       <div class="form-group col-sm-4">
         <label class="sr-only">Namespace</label>
-        <ui-select ng-required="selectRequired" ng-model="istag.namespace" ng-disabled="selectDisabled" ng-change="istag.imageStream = null; istag.tagObject = null;">
+        <ui-select ng-required="selectRequired" ng-model="istag.namespace" ng-disabled="selectDisabled" ng-change="istag.imageStream = null; istag.tagObject = null;" append-to-body="true">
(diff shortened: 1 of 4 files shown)


app/views/directives/istag-select.html
```diff
@@ -5,7 +5,7 @@
         <label class="sr-only">Namespace</label>
         <ui-select ng-required="selectRequired" ng-model="istag.namespace" ng-disabled="selectDisabled" ng-change="istag.imageStream = null; istag.tagObject = null;">
           <ui-select-match placeholder="Namespace">{{$select.selected}}</ui-select-match>
-          <ui-select-choices repeat="namespace in (namespaces | filter : $select.search)">
+          <ui-select-choices repeat="namespace in (namespaces | filter : $select.search)" position='down'>
```
spadgett: nit: we generally use double quotes `"` for HTML attributes

sg00dwin:
@spadgett  updated ptal

spadgett:
Judging from the last screenshot in the bug, this now only happens if the browser window is very short. I'm not too worried about it for 3.7.

Would it make sense to change the z-index so that the popup lays over the top wizard steps?

If we go the position route, it would be nice to have an optional position attribute on the directive so that we can let ui-select calculate the best position when it makes sense. For instance, on the edit build config page, it's better to let it pop up when needed:

openshift web console 2017-11-01 15-53-09

PR #2208, opened 2017-10-03
Add SVG icons

spadgett:
Icons from @jennyhaines

Requires https://github.com/openshift/origin-web-catalog/pull/465
Requires https://github.com/openshift/origin-web-common/pull/213

screen shot 2017-10-03 at 3 03 29 pm

Before and After:

screen shot 2017-10-04 at 9 42 10 am

@serenamarie125 @sspeiche @rhamilto

PR #2432, opened 2017-11-01
Fix bug where custom img icons don't appear in catalog cards

rhamilto:
Introduced with #2208

screen shot 2017-11-01 at 4 49 26 pm


spadgett:
WebDriver flake

/retest

PR #2436, opened 2017-11-02
Add watcher to secrets browse page

jhadvig:
This is a follow up of https://github.com/openshift/origin-web-console/pull/2421

@spadgett PTAL

### DIFF
--- a/app/scripts/controllers/secret.js
+++ b/app/scripts/controllers/secret.js
@@ -8,7 +8,12 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('SecretController', function ($routeParams, $filter, $scope, DataService, ProjectsService, SecretsService) {
+  .controller('SecretController', function ($routeParams,
+                                            $filter,
+                                            $scope,
+                                            DataService,
+                                            ProjectsService,
+                                            SecretsService) {
     $scope.projectName = $routeParams.project;
     $scope.secretName = $routeParams.secret;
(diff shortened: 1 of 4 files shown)


jhadvig:
cc'ing @jwforres

app/scripts/controllers/secret.js
```diff
@@ -45,13 +65,21 @@ angular.module('openshiftConsole')
...
+            $scope.alerts["load"] = {
+              type: "error",
+              message: "The secret details could not be loaded.",
+              details: $filter('getErrorDetails')(e)
+            };
           });
+
+        $scope.$on('$destroy', function(){
+          DataService.unwatchAll(watches);
+        });
```
jwforres: you should pass a final opts parameter to the get with `{errorNotification: false}` so that you don't get both a toast notification and the inline error

spadgett: Looks like it's already set in the call to `DataService.get`

jhadvig: @jwforres not really sure what you mean

jwforres: nevermind @jhadvig i see it being passed above now, ignore this

jhadvig:
@jwforres comments addressed :)
PTAL

spadgett:

jhadvig:
@spadgett updated. Sorry missed updated scripts.js when was rebasing. Should be ok now

PR #2439, opened 2017-11-02
Don't wrap "Clear Filter" text

spadgett:
Before:

iphone x - ios 11 0 2017-11-02 13-04-13

After:

iphone x - ios 11 0 2017-11-02 13-08-36

@jwforres @rhamilto

### DIFF
--- a/app/views/browse/config-maps.html
+++ b/app/views/browse/config-maps.html
@@ -38,7 +38,7 @@
                 </p>
               </div>
               <div ng-if="filterWithZeroResults">
-                <h2>The filter is hiding all config maps. <a href="" ng-click="clearFilter()">Clear Filter</a></h2>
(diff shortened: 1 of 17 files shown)


jwforres:
something is wrong with the line height, those two lines of text are too close together

spadgett:
> something is wrong with the line height, those two lines of text are too close together

This is using the normal `h2` style. Seems to be the `line-height: 1.1` rule on headers coming from Bootstrap.

jwforres:
We adjust the bootstrap header sizes in our variables, maybe we screwed up the line heights

rhamilto:
We have many instances of "Clear Filter" now (browse pages).  We should probably change them all?

spadgett:
@jwforres another look since I added the class to other instances?

PR #2445, opened 2017-11-02
Fix broken route links

sg00dwin:
Noticed a couple of broken links to docs for routes.


spadgett:
/kind bug
/lgtm
/hold

Holding for @jwforres approval, but this is probably a blocker.

jwforres:
yup
/hold cancel

but @sg00dwin can you do a pass on the rest of the links and make sure they are still good

jwforres:
@ahardin-rh there are at least two cases here where doc links are broken, these should have at least had redirects put in place

sg00dwin:
@jwforres All other links referenced in https://github.com/openshift/origin-web-console/blob/master/app/scripts/constants.js are good

PR #2448, opened 2017-11-03
Bug 1509142 - Should not display the 'Reveal Secret' link when secret is without 'data' field

jhadvig:
@spadgett PTAL

### DIFF
--- a/app/views/browse/secret.html
+++ b/app/views/browse/secret.html
@@ -19,6 +19,7 @@
             <button ng-if="project.metadata.name | canIAddToProject"
                     type="button"
                     class="btn btn-default hidden-xs"
+                    ng-disabled="!secret.data"
                     ng-click="addToApplication()">
               Add to Application
             </button>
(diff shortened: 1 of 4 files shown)


app/views/browse/secret.html
```diff
@@ -90,6 +90,9 @@ <h2 class="mar-top-none">
                 </div>
               </div>
             </dl>
+            <p ng-if="!secret.data" class="mar-bottom-xl">
+              There are no data in this secret.
+            </p>
```
spadgett: This is grammatically correct, but reads a little funny to me. Maybe use an empty state message?

```html
<div ng-if="!secret.data" class="empty-state-message text-center">
  <h2>No data.</h2>
  <p>This secret has no data.</p>
</div>
```

jhadvig:
@spadgett so you are right regarding the secret without data field in the`Environment From`... there is nothing to select in the value input.

bitmap

Do we want to disable the value input in that case, or just set the placeholder to just say something like "No key" ?

spadgett:
> Do we want to disable the value input in that case, or just set the placeholder to just say something like "No key" ?

As long as it's not broken, we might just leave it for 3.7 and come back to it in 3.8.

cc @jwforres

jwforres:
yeah if its not throwing a JS error i wouldnt touch that code in 3.7

jhadvig:
@spadgett disabling the btn when no `data`
no-add

PR #2451, opened 2017-11-03
Wizard fixes for iPhone

### LINKED ISSUE #2450: Extremely difficult to scroll inside wizard on iPhone
Sometimes the page behind the wizard scrolls instead of the wizard content itself. When this happens, it's difficult to figure out how to make the content itself scroll.

When you finally are able to scroll the content, it feels sticky / sluggish.

spadgett:
* Prevent page scrolling under the wizard dialog, making it difficult to scroll the wizard content itself
* Don't add custom scrollbar styles on iOS, which cause layout problems on an iPhone
* Reduce wizard max-height on iPhone to prevent the buttons being hidden on iPhone X

Closes #2450 

@rhamilto @jwforres

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -7,38 +7,46 @@
   display: none !important;
 }
 
-// Prevent zooming when focusing input fields in mobile Safari. Try to only
-// target mobile devices to avoid larger-than-wanted fonts elsewhere.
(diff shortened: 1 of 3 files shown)


spadgett:
/kind bug

Largely whitespace changes.

spadgett:
/hold

typo in one rule

PR #2462, opened 2017-11-06
[WIP] Append notification drawer kebab menu to body to fix overflow clipping

### LINKED ISSUE #2449: Kebab menu clipped when last item in drawer
screen shot 2017-11-03 at 12 26 38 pm

cc @jwforres

rhamilto:
Fixes #2449 

This fix comes with some concessions:

1. The styling of the dropdown no longer includes the caret/arrow and the alignment to the kebab now differs since the rules that provide the styling no longer apply now that the dropdown is appended to body. [1]
1.  The dropdown no longer scrolls with the notification drawer contents. [2]

Upgrading UI Bootstrap to v1.0.0+ should allow us to fix both issues as those versions include a uib-dropdown setting to append the dropdown to an arbitrary element (dropdown-append-to) in addition to the ability to add an optional class to the open dropdown (appendToOpenClass).

Upgrading a dependency is too risky at this point, so it seems like our options are to:

1.  Not merge this PR and let the bug remain unfixed
1.  Merge the PR with the understanding it comes with the above concessions.

[1]
screen shot 2017-11-06 at 9 17 16 am
screen shot 2017-11-06 at 9 17 25 am

[2]
screen shot 2017-11-06 at 9 24 02 am

### DIFF
--- a/app/styles/_notifications.less
+++ b/app/styles/_notifications.less
@@ -2,6 +2,11 @@
 // Notification drawer
 // -----------------------------------------------
 
+// so that .uib-dropdown-menu inside drawer-pf render above the drawer pf
+body > .uib-dropdown-menu {
+  z-index: @zindex-navbar-fixed - 2; // two less than navbar, one less than project-bar, same as drawer-pf
+}
+
 notification-drawer-wrapper {
   .blank-state-pf-title {
     font-size: @font-size-h4;
--- a/app/views/directives/notifications/notification-body.html
+++ b/app/views/directives/notifications/notification-body.html
@@ -12,6 +12,7 @@
   </button>
   <div
     uib-dropdown
+    dropdown-append-to-body
     class="dropdown pull-right dropdown-kebab-pf"
     ng-if="notification.actions.length">
     <button
@@ -25,7 +26,7 @@
       <span class="fa fa-ellipsis-v"></span>
     </button>
     <ul
-      class="dropdown-menu dropdown-menu-right"
+      class="uib-dropdown-menu dropdown-menu dropdown-menu-right"
       aria-labelledby="dropdownKebabRight">
       <li
         ng-repeat="action in notification.actions"
--- a/dist/scripts/templates.js
+++ b/dist/scripts/templates.js
@@ -7894,11 +7894,11 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
     "<span class=\"sr-only\">Clear notification</span>\n" +
     "<span aria-hidden=\"true\" class=\"pficon pficon-close\"></span>\n" +
     "</button>\n" +
-    "<div uib-dropdown class=\"dropdown pull-right dropdown-kebab-pf\" ng-if=\"notification.actions.length\">\n" +
+    "<div uib-dropdown dropdown-append-to-body class=\"dropdown pull-right dropdown-kebab-pf\" ng-if=\"notification.actions.length\">\n" +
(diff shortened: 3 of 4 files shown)


dtaylor113:
> The dropdown no longer scrolls with the notification drawer contents

IMO this is kinda a showstopper.  It looks strange to have a kabob menu floating unattached.  Easy for users to lose track on which notification the kabob menu is for.

rhamilto:
@jeff-phillips-18 pointed out the upstream does not have this issue as the scrollable area expands to accommodate the dropdown.  So this bug is specific to the console because of a [markup addition](https://github.com/openshift/origin-web-console/blob/master/app/views/directives/notifications/notification-body.html#L1-L4).  @benjaminapetersen, do you remember why you needed the markup addition?

spadgett:
@rhamilto That element's there for the "mark as read" click handler, but we can probably change it if it's causing the menu to be cut off.

benjaminapetersen:
The drawer is a tad difficult to work with as the separation of concerns is not clean.  You can kind of customize the markup.... but it expects a lot of your structure &&  some of it is up in the `notification-drawer` itself and untouchable.  

The upstream does `mark as read` only if you click the bold title text of the notification itself, not if you click the background of the notification.  However, the whole notification turns blue if you hover over it, implying that it is clickable.  (The upstream never gives you a pointer indicator,  which is also a bit confusing & related to the topic of one of our spinoff discussions, see http://www.patternfly.org/angular-patternfly/#/api/patternfly.notification.component:pfNotificationDrawer).

I'm fine with changes.  If we discussion moving things to upstream, I may like to see the elimination of my `wrapper` in favor of putting the actual `notification` div in control of the user (in `notification-body.html`).  That is prob outside this bug fix, tho.

jeff-phillips-18:
@benjaminapetersen Where does the mark read happen upstream? The only mark read that occurs is due to adding a Mark all Read button with a mark all read callback.

serenamarie125:
Just to confirm, you are not changing the hover behavior with this PR, are you @benjaminapetersen ?

jeff-phillips-18:
The drawer does not expect any particular structure. It provides some structure in order to help lead the application into the patternfly design pattern for the notification drawer. The notification body can contain whatever the application sees fit. Unclear what you mean by:

> putting the actual notification div in control of the user (in notification-body.html).

spadgett:
@serenamarie125 This PR does not change the hover. It's only to fix the menu getting cut off.

spadgett:
Best I can tell you can already scroll the notification when the kebab is open. So maybe that is "good enough" for 3.7? At least you can get to the actions.

rhamilto:
> So this bug is specific to the console because of a markup addition.

That's incorrect.  The console implementation scrolls the same way upstream does.

spadgett:
Not ideal, but at least you can get to the other items.

screencast 2017-11-06 14-36-39

benjaminapetersen:
Surprising amount of detail here :)

So my original impl of notifications didn't have kebabs, everything we added just had the "x" icon & mostly based off of api events.  Looking into this some more now.

@serenamarie125 I am not changing behavior ( this is @rhamilto's PR ).

@jeff-phillips-18 quick gif of me clicking around on http://www.patternfly.org/angular-patternfly/#/api/patternfly.notification.component:pfNotificationDrawer shows that the "mark read" is a bit tricky to pin down (every pause is me clicking):



so it was a little surprising when the `mark read` fired & when it did not.   I had to add the `markRead` handler to the extra node because https://github.com/patternfly/angular-patternfly/blob/master/src/notification/notification-drawer.html#L24 the actual node is up in the directive & not something I could access via `notification-body.html`.

PR #2464, opened 2017-11-06
Create dockercodnfigjson as default secret

jhadvig:
When creating secret we should start defaulting to the new file type and format which is `docker/config.json`
```
{
        "auths": {
                "registryaddress": {
                        "auth": "auth string",
                        "email": "email@domain.com"
                }
        }
}
```

So when selecting an `Image registry Credentials` Authentication Type on the create-secret page, it shall create secret of `kubernetes/dockerconfigjson` type, that is in the mentioned format.

`Image registry Credentials` Authentication Type will create a `kubernetes/dockercfg` type of secret with the old format, unless user provides a config in a new format that contains `auths` as a root field in the object, in that case `kubernetes/dockerconfigjson` type of a secret is created.

@jwforres @spadgett PTAL

### DIFF
--- a/app/views/directives/create-secret.html
+++ b/app/views/directives/create-secret.html
@@ -214,7 +214,7 @@
         </div>
       </div>
 
-      <div ng-if="newSecret.authType === 'kubernetes.io/dockercfg'">
+      <div ng-if="newSecret.authType === 'kubernetes.io/dockerconfigjson'">
         <div class="form-group" ng-class="{ 'has-error' : secretForm.dockerServer.$invalid && secretForm.dockerServer.$touched }">
           <label for="docker-server" class="required">Image Registry Server Address</label>
           <div>
@@ -298,7 +298,7 @@
         </div>
       </div>
 
-      <div ng-if="newSecret.authType === 'kubernetes.io/dockerconfigjson'">
+      <div ng-if="newSecret.authType === 'kubernetes.io/dockercfg'">
         <div class="form-group" id="docker-config">
           <label for="dockerConfig" class="required">Configuration File</label>
           <osc-file-input
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -49,11 +49,14 @@ angular.module("openshiftConsole")
             label: "Image Secret",
             authTypes: [
               {
-                id: "kubernetes.io/dockercfg",
+                id: "kubernetes.io/dockerconfigjson",
                 label: "Image Registry Credentials"
               },
+              // User can paste content of his docker configuration file, which could be either in old
+              // '.dockercfg' format, or new '.docker/config.json' format. Because of that the secret type
+              // is determined upon creating the secret, where we scan for the type of format.
               {
-                id: "kubernetes.io/dockerconfigjson",
+                id: "kubernetes.io/dockercfg",
                 label: "Configuration File"
               }
             ]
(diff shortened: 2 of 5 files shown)


app/views/directives/create-secret.html
```diff
@@ -313,6 +313,7 @@
       </div>
     </div>
   </div>
+  {{newSecret.authType}}
```
jwforres: did this get left in accidentally?

jwforres:
My concern with just flipping what the labels point to is that the user won't realize the behavior has changed. Maybe the Configuration File option should now say Legacy Configuration File Format or something

@spadgett thoughts?

jwforres:
Also can the docker config contain more than just the auth info? By changing this we would be taking away the option people were using to provide a whole config file in the new format, since the Image Registry Credentials option only lets you specify certain fields.

app/views/directives/create-secret.html
jhadvig: it did.. sorry, will remove and update

jwforres:
@miminar are we breaking anything if we continue defaulting to the old format in 3.7, I'm concerned with these changes this late in the release

cc @pweil-

jhadvig:
Like I'm completely fine with dont merge it and deal with the issue after the 3.7

miminar:
@jwforres the change shouldn't break anything, but waiting with merge for 3.8 seems reasonable.

Adding the upstream PR for reference: moby/moby#12009

jwforres:
Ok lets hold then for 3.8 and we talk in more detail about the UX details around the change

/hold

xingxingxia:
Subscribing it and will verify it when available. Thx

jhadvig:
@openshift/team-ux-review

xingxingxia:
Hi, will this go to 3.9 web console? 3.9 CLI already fixed in https://bugzilla.redhat.com/show_bug.cgi?id=1531511

jhadvig:
@spadgett rebased && updated. PTAL

spadgett:
@jhadvig It's strange to me we're keeping the same two labels, but changing the types of secrets they create.

Before:

Image Registry Credentials -> kubernetes.io/dockercfg
Configuration File -> kubernetes.io/dockerconfigjson

After:

Image Registry Credentials -> kubernetes.io/dockerconfigjson
Configuration File -> kubernetes.io/dockercfg

Were the labels incorrect before?

jhadvig:
@spadgett no they were not incorrect, but the secret type was diferent when the image registry secret was created via CLI and UI. The PR was open to align those two processes so they create secrets of the same type

jhadvig:
Not really sure if CLI was creating the old format in the past but its creating the `kubernetes.io/dockerconfigjson` secret type.
The thing is that when user picked `Image Registry Credentials`, as a Authentication File, we create the secret in the old format, not like the CLI which create the new one. If user picks `Configuration File` he can paste content of his docker config file, no matter in which format it is, cause if it's in the old format we create secret with `kubernetes.io/dockercfg` typ. Otherwise secret with `kubernetes.io/dockerconfigjson` type is created.

xingxingxia:
> Does the CLI support both formats as well or did it switch to always creating the new format?

In 3.9, commands always create new format, as confirmed in [bug 1531511 comment 13](https://bugzilla.redhat.com/show_bug.cgi?id=1531511#c13), no matter command `oc create secret docker-registry` or the depreciated `oc secrets new-dockercfg`
> Not really sure if CLI was creating the old format in the past

3.7 was. Including latest 3.7.z of v3.7.42

spadgett:
@jhadvig Looks like a dist mismatch

PR #2466, opened 2017-11-07
Allow editing service instances again if previous update failed

spadgett:
Service catalog now allows users to edit a service instance if the previous
update failed. We were incorrectly blocking it in the UI since we check for
readiness instead of failure.

/kind bug
/assign @jwforres @jeff-phillips-18

### DIFF
--- a/app/scripts/controllers/serviceInstance.js
+++ b/app/scripts/controllers/serviceInstance.js
@@ -54,7 +54,7 @@ angular.module('openshiftConsole')
     var servicePlansPromise;
 
     var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
-    var serviceInstanceReady = $filter('isServiceInstanceReady');
+    var isServiceInstanceFailed = $filter('isServiceInstanceFailed');
 
     // API Versions
     var serviceBindingsVersion = APIService.getPreferredVersion('servicebindings');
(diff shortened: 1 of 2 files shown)


spadgett:
We might change the edit dialog to re-enable the back button on a failed update, but I think that change is probably too big for 3.7.

app/scripts/controllers/serviceInstance.js
```diff
@@ -120,7 +120,14 @@ angular.module('openshiftConsole')
       var updateSchema = _.get($scope.plan, 'spec.instanceUpdateParameterSchema');
       var planUpdatable = (_.size(_.get(updateSchema, 'properties')) > 0) || (_.get($scope.serviceClass, 'spec.planUpdatable') && (_.size($scope.servicePlans) > 1));
 
-      $scope.editAvailable = planUpdatable && serviceInstanceReady($scope.serviceInstance) && !_.get($scope.serviceInstance, 'metadata.deletionTimestamp');
+      $scope.editAvailable =
+        (updateSchema || planUpdatable) &&
```
spadgett: @jeff-phillips-18 I think there was a bug here before. It should be editable sometimes even if the plan is not updatable, correct?

spadgett:
/hold

I have a bug

PR #2472, opened 2017-11-07
Fix project list view styles per UXD feedback

TheRealJon:
# Description
Separated project list view item project name, description, and additional info fields into three columns and styled to match Patternfly "list-view" pattern more closely.

[Trello Card](https://trello.com/c/zHWEtHgj) 
[UXD Feedback Document](https://docs.google.com/a/redhat.com/document/d/1oowFKLNc19GuUPW3rxeM_JMe9ZfeAvdxC0vN1hgWUkU/edit?usp=sharing)

# Screen Shots 
Ignore the weird right margins on the extra small and small layouts, that's an issue with my screen capture extension.

## Extra Small
extra-small

## Small
small

## Medium
medium

## Large
large

### DIFF
--- a/app/styles/_projects.less
+++ b/app/styles/_projects.less
@@ -12,11 +12,8 @@
   display: inline-block;
 }
 
-.project-additional-info.list-view-pf-additional-info {
+.project-additional-content.list-pf-additional-content {
   display: block;
-  @media (min-width: @screen-md-min) {
-    width: 55%;
-  }
 }
 
 .project-date, // from origin-web-catalog
@@ -26,44 +23,42 @@
   }
 }
 
-.project-description.list-group-item-text {
-  width: 100%;
+.project-description.list-pf-description {
+  color: @color-pf-black-800;
   .highlighted-content,
   .truncated-content {
     display: block;
     white-space: pre-line;
     .word-break();
   }
+  .truncation-block {
+    margin: 0;
+  }
 }
 
 .project-info {
-  &.list-group-item {
+  &.list-pf-item {
     border-bottom: 0;
     border-color: #e0e0e0;
     padding: 10px 20px;
   }
-  .list-view-pf-description {
-    width: 100%;
-  }
-  .list-view-pf-main-info {
-    display: block;
-  }
-
 }
 
-.project-name-item.list-group-item-heading {
-  @media (min-width: @screen-md-min) {
-    white-space: normal;
-    width: 45%;
+.project-name-item.list-pf-title {
+  @media (max-width: @screen-sm-min) {
+    margin-bottom: 18px;
   }
-  .h1 {
+  .h3 {
     align-items: center;
     display: flex;
     margin-bottom: 0;
     margin-top: 0;
+    @media (max-width: @screen-sm-min) {
+      margin-bottom: 12px;
+    }
     [data-toggle='tooltip'] {
       cursor: help;
-      margin: 0 0 -5px 7px;
+      margin-left: 7px;
     }
     a {
       display: block;
@@ -72,9 +67,11 @@
       text-overflow: ellipsis;
     }
   }
-  p {
-    margin-bottom: 0;
-    margin-top: 5px;
+  small {
+    color: @color-pf-black-600;
+    font-size: @font-size-base;
+    font-weight: normal;
+    margin: 0;
   }
 }
 
@@ -100,7 +97,6 @@
       margin-top: 0;
       order: 2;
     }
-
     .origin-modal-popup {
       @media (min-width: @screen-sm-min) {
         width: 370px;
--- a/app/views/projects.html
+++ b/app/views/projects.html
@@ -94,64 +94,69 @@
                 The current filter is hiding all projects.
                 <a href="" ng-click="search.text = ''" role="button" class="nowrap">Clear Filter</a>
               </div>
-              <div class="list-group list-view-pf projects-list">
-                <div ng-repeat="project in projects | limitTo: limitListTo track by (project | uid)" class="list-group-item project-info tile-click">
-                  <div class="list-view-pf-main-info">
-                    <div class="list-view-pf-description project-names">
-                      <div class="list-group-item-heading project-name-item">
-                        <h2 class="h1">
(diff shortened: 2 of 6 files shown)


app/views/projects.html
```diff
@@ -97,27 +97,29 @@ <h2 class="text-center">Loading...</h2>
...
-                          <span ng-if="project.status.phase != 'Active'"
-                            data-toggle="tooltip"
-                            title="This project has been marked for deletion."
-                            class="pficon pficon-warning-triangle-o"></span>
-                        </h2>
-                        <small>
+                    <div class="list-view-pf-body">
+                      <div class="list-view-pf-description">
+                        <div class="list-group-item-heading project-name-item">
+                          <span class="h3">
```
rhamilto: span is semantically neutral, so this change actually removes semantic meaning.  It's better as an h2.

app/views/projects.html
```diff
@@ -97,27 +97,29 @@ <h2 class="text-center">Loading...</h2>
               <div class="list-group list-view-pf projects-list">
```
rhamilto: .list-view-pf has been deprecated in PatternFly.  I wonder if we should re-evaluate its use here?  Additionally, it's adding a lot of extraneous markup that we really don't need given the simplicity of the layout.

spadgett:
I know this is what is in the design doc, but I'm not sure the 3 column layout works as well with longer descriptions (medium screenshot above).

@openshift/team-ux-review opinions?

beanh66:
@spadgett in the medium view, the middle column looks the most narrow to me, which I think is part of the issue. I would rather see that a bit wider since the right column doesn't need much space. @TheRealJon can we give that a try?

beanh66:
@spadgett I see what you mean. I know we wanted an interim solution that moves us closer to the longer term design, but I'm chatting with the other designers on this again as well to figure out what makes sense for now!

jwforres:
As @rhamilto mentioned the deprecated list-view-pf is part of the problem here. It's pretty restrictive in how it handles the layout. Switching to the new pf-list (i think is the name?) will give us full control over the layout in the rows.

TheRealJon:
Do we want to try to transition over to the pf-list pattern instead of the pf-list-view? Or should I continue down the current path?

jeff-phillips-18:
I'd vote for list-pf if its not too unwieldy  (it shouldn't be).

spadgett:
Having the description below "created by" is also consistent with the landing page project summary.

openshift web console 2017-11-09 08-09-04

beanh66:
@spadgett I agree with you and the fixes @jennyhaines provided should change the order so at smaller screen sizes the created by is listed above the description.

TheRealJon:
@beanh66 and I just had a conversation about this task to answer a few questions that I had:

**Is it important that the `created by` column be on the far right in larger viewports?**
I was concerned that in order to make Jenny's style updates, we would not exactly match the original design recommendations that Colleen provided. The list-pf pattern doesn't provide the needed structure/styling to have the third `created by` column responsively move underneath the project title. To solve this, the `description` and `created by` columns would need to be switched, which doesn't match her original document. We concluded that the responsive behavior is the primary concern and that switching the columns would be an acceptable solution.

**If we switch these columns, are we concerned that the list-pf-item child class names do not semantically match their content?**
Another concern that arose was that the above solution causes a little bit of dissonance between class names and element content. To implement this solution, the `created by` information would be  in an element with class `list-pf-description` and the project description would be in an element with class `list-pf-additional-content`. This was a minor concern of mine, but it led to the next question.

**If we want to use the `list-pf-additional-content` section of the `list-pf-item` in the future, what will we do with the project description?**
Colleen mentioned that there might be a future need to display project metrics in the list view. When and if the need arises, the above implementation would need to be revisited and adjusted so that the `list-pf-additional-content` element could be freed up for project metrics.

We finally came to the conclusion that we should keep the list item structure similar to the current implementation. The `created by` info will remain below the title and the project description will be in the `list-pf-description` element. The text sizes and colors will be adjusted to more closely match the project information displayed on the home page. This way the additional content section can remain available for project metrics in the future.

beanh66:
FYI @jennyhaines, we will still be making the changes to follow your styling suggestions! @TheRealJon let me know if I'm off on that!

TheRealJon:
Yeah, I should have mentioned that too. I am including @jennyhaines' suggestions in my updates as well.

TheRealJon:
Here are screenshots for the latest changes:

**Large**
large

**Medium**
medium

**Small**
small

**Extra Small**
extra-small

app/styles/_projects.less
```diff
@@ -72,9 +67,11 @@
       text-overflow: ellipsis;
     }
   }
-  p {
-    margin-bottom: 0;
-    margin-top: 5px;
+  small {
+    margin: 0;
```
rhamilto: Nit:  [alphabetization of declarations by property](http://codeguide.patternfly.org/#css-declaration-order)

app/views/projects.html
```diff
@@ -94,64 +94,69 @@ <h2 class="text-center">Loading...</h2>
...
+                        </div>
                       </div>
-                      <div class="list-view-pf-additional-info project-additional-info">
-                        <span class="list-group-item-text project-description">
-                          <truncate-long-text ng-if="!keywords.length" content="project | description" limit="265" newline-limit="10" use-word-boundary="true"></truncate-long-text>
-                          <span class="highlighted-content" ng-if="keywords.length" ng-bind-html="project | description | truncate : 1000 | highlightKeywords : keywords"></span>
-                        </span>
+                      <div class="list-pf-actions" ng-if="project.status.phase == 'Active'">
+                        <div uib-dropdown class="dropdown pull-right dropdown-kebab-pf">
+                          <button uib-dropdown-toggle class="btn btn-link dropdown-toggle"><i class="fa fa-ellipsis-v" aria-hidden="true"></i><span class="sr-only">Actions</span></button>
```
spadgett: 👍  Thanks, this is better as a button for accessibility.

spadgett:
Looks like @beanh66 and @jennyhaines have signed off and codes changes LGTM

@TheRealJon feel free to take WIP off the PR title if you are ready to merge

/lgtm

PR #2473, opened 2017-11-07
Correct ProjectNameTaken error handling in various add-to-projects wizards

### LINKED ISSUE #1955: Project 'nameTaken' not being set properly in controllers
HTML files for process-template, from-file, and deploy-image all have: 

```<select-project ... name-taken="$ctrl.projectNameTaken"></select-project>```

However, the corresponding controllers do not set the variable ```projectNameTaken```, which should be set from a failure from the server after attempting to create a project which already has the name.  

Some comments from @spadgett:

"Normal users can't see all projects in the cluster, so you don't have a complete list of project names (used by ```osc-unique```). We still need to handle errors from the server creating the project because the name is already in use by someone else."

https://github.com/openshift/origin-web-console/search?utf8=%E2%9C%93&q=name-taken&type=

This is also true for ```order-service-controller``` in origin-web-catalog.

### PR DESCRIPTION
**Pull request by dtaylor113**
Updated deployImage, processTemplate, and fromFile wizards.

Fixes #1955 

image

### DIFF
--- a/app/scripts/directives/deployImage.js
+++ b/app/scripts/directives/deployImage.js
@@ -26,12 +26,16 @@ angular.module("openshiftConsole")
         // Must be initialized the controller. The link function is too late.
         $scope.forms = {};
         $scope.noProjectsCantCreate = false;
-      },
-      link: function($scope) {
+
         $scope.input = {
           selectedProject: $scope.project
         };
 
+        $scope.$watch('input.selectedProject.metadata.name', function() {
+          $scope.projectNameTaken = false;
+        });
+      },
+      link: function($scope) {
         // Pick from an image stream tag or Docker image name.
         $scope.mode = "istag"; // "istag" or "dockerImage"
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: app/scripts/directives/fromFile.js [hunk 1]
```diff
@@ -33,10 +33,16 @@ angular.module("openshiftConsole")
           $scope.noProjectsCantCreate = true;
         });
 
+        $scope.projectNameTaken = false;
```
spadgett: Shouldn't need to initialize here since the `$watch` below is called immediately and will set it to false.

app/scripts/directives/deployImage.js
```diff
@@ -378,13 +382,17 @@ angular.module("openshiftConsole")
...
-                details: getErrorDetails(e)
-              });
               $scope.disableInputs = false;
+              if (e.data.reason === 'AlreadyExists') {
+                $scope.projectNameTaken = true;
+              } else {
+                NotificationsService.addNotification({
+                  id: "deploy-image-create-project-error",
+                  type: "error",
+                  message: "An error occurred creating project",
```
spadgett: We make the message a complete sentence with punctuation when possible.

```
"An error occurred creating project."
```

app/scripts/directives/fromFile.js
```diff
@@ -181,14 +187,17 @@ angular.module("openshiftConsole")
...
-              message: "An error occurred creating project",
-              details: getErrorDetails(e)
-            });
+            if (e.data.reason === 'AlreadyExists') {
+              $scope.projectNameTaken = true;
+            } else {
+              NotificationsService.addNotification({
+                id: "import-create-project-error",
+                type: "error",
+                message: "An error occurred creating project",
```
spadgett: "An error occurred creating project."

app/scripts/directives/deployImage.js
dtaylor113: Fixed

app/scripts/directives/fromFile.js
dtaylor113: Fixed

PR #2476, opened 2017-11-08
Separate out defaultIfBlank filter

benjaminapetersen:
Reviewing some test coverage & noticed we have a standalone spec for `defaultIfBlank`.  Breaking the filter out so the file structure matches.


spadgett:
It looks like this filter is unused as well?

benjaminapetersen:
Yay finding stale code 😄

PR #2477, opened 2017-11-08
Add unit test for underscore filter, split out into separate file

benjaminapetersen:
Improving our test coverage.
Sticking with the pattern of splitting out filters into their own file to match the structure in the unit test directory.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -394,6 +394,7 @@
         <script src="scripts/filters/resources.js"></script>
         <script src="scripts/filters/canI.js"></script>
         <script src="scripts/filters/util.js"></script>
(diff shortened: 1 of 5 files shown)


spadgett:
+1 to improving test coverage :)

It looks like this filter isn't being used, though. Should we just delete it?

benjaminapetersen:
Saw that as well.  Im fine with just nixing it.

spadgett:
> Im fine with just nixing it.

I think it's better just to remove it.

benjaminapetersen:
Was going to update, but I think someone already removed.  Closing.

Issue #2422, opened 2017-11-01
Missing filters in the secrets page


jhadvig:
Noticed that we are missing input field for label filtering on the secrets page. We need to add those.

@spadgett fyi

PR #2479, opened 2017-11-08
Adding label filter to the secrets page

jhadvig:
Screens:

@spadgett @jwforres PTAL

Closes https://github.com/openshift/origin-web-console/issues/2422

### DIFF
--- a/app/scripts/controllers/secrets.js
+++ b/app/scripts/controllers/secrets.js
@@ -8,9 +8,13 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('SecretsController', function ($routeParams, $scope, DataService, ProjectsService) {
+  .controller('SecretsController', function ($routeParams, $scope, DataService, LabelFilter, ProjectsService) {
     $scope.projectName = $routeParams.project;
-    $scope.secretsByType = {};
+    $scope.labelSuggestions = {};
+    $scope.clearFilter = function() {
+      LabelFilter.clear();
+    };
+
     var watches = [];
 
     ProjectsService
@@ -19,9 +23,25 @@ angular.module('openshiftConsole')
         $scope.project = project;
         $scope.context = context;
         watches.push(DataService.watch("secrets", context, function(secrets) {
-          $scope.secrets = _.sortBy(secrets.by("metadata.name"), ["type", "metadata.name"]);
+          $scope.unfilteredSecrets = _.sortBy(secrets.by("metadata.name"), ["type", "metadata.name"]);
           $scope.secretsLoaded = true;
+          LabelFilter.addLabelSuggestionsFromResources($scope.unfilteredSecrets, $scope.labelSuggestions);
+          LabelFilter.setLabelSuggestions($scope.labelSuggestions);
+          $scope.secrets = LabelFilter.getLabelSelector().select($scope.unfilteredSecrets);
+          updateFilterMessage();
         }));
+
+        function updateFilterMessage() {
+          $scope.filterWithZeroResults = !LabelFilter.getLabelSelector().isEmpty() && _.isEmpty($scope.secrets) && !_.isEmpty($scope.unfilteredSecrets);
+        }
+
+        LabelFilter.onActiveFiltersChanged(function(labelSelector) {
+          // trigger a digest loop
+          $scope.$evalAsync(function() {
+            $scope.secrets = labelSelector.select($scope.unfilteredSecrets);
+            updateFilterMessage();
+          });
+        });
 
         $scope.$on('$destroy', function(){
           DataService.unwatchAll(watches);
--- a/app/views/secrets.html
+++ b/app/views/secrets.html
@@ -14,28 +14,39 @@
           </span>
         </h1>
       </div>
+      <div ng-if="unfilteredSecrets" class="data-toolbar">
+        <div class="data-toolbar-filter">
+          <project-filter></project-filter>
+        </div>
+      </div>
     </div>
   </div><!-- /middle-header-->
   <div class="middle-content">
     <div class="container-fluid">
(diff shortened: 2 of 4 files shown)


app/views/secrets.html
```diff
@@ -14,28 +14,39 @@
           </span>
         </h1>
       </div>
+      <div ng-if="(secrets | hashSize) > 0 || filterWithZeroResults" class="data-toolbar">
```
spadgett: I think we could just check `unfilteredSecrets` size here. Note that we've started using `| size` instead of `| hashSize`.

Does this work?

```html
<div ng-if="unfilteredSecrets | size" class="data-toolbar">
```

app/scripts/controllers/secrets.js
```diff
@@ -19,10 +23,26 @@ angular.module('openshiftConsole')
...
           $scope.secretsLoaded = true;
+          $scope.unfilteredSecrets = secrets.by("metadata.name");
+          LabelFilter.addLabelSuggestionsFromResources($scope.unfilteredSecrets, $scope.labelSuggestions);
+          LabelFilter.setLabelSuggestions($scope.labelSuggestions);
+          $scope.secrets = LabelFilter.getLabelSelector().select($scope.unfilteredSecrets);
+          updateFilterMessage();
         }));
 
+        function updateFilterMessage() {
+          $scope.filterWithZeroResults = !LabelFilter.getLabelSelector().isEmpty() && _.isEmpty($scope.secrets)  && !_.isEmpty($scope.unfilteredSecrets);
```
spadgett: nit: extra space before the last `&&`

app/views/secrets.html
```diff
@@ -14,28 +14,39 @@
...
-          <div ng-if="!secrets.length">
-            <div class="empty-state-message text-center">
-              <h2>No secrets.</h2>
-              <p>
-                No secrets have been added to project {{projectName}}.
-              </p>
-              <p ng-if="project && ('secrets' | canI : 'create') && !secrets.length">
-                <a ng-href="project/{{project.metadata.name}}/create-secret" class="btn btn-primary">Create Secret</a>
-              </p>
+          <div ng-if="(secrets | hashSize) == 0">
```
spadgett:
```html
<div ng-if="(secrets | size) === 0">
```

app/views/secrets.html
```diff
@@ -14,28 +14,39 @@
...
+            <p ng-if="!secretsLoaded">
+              Loading...
+            </p>
+            <div ng-if="secretsLoaded" class="empty-state-message text-center">
+              <div ng-if="!filterWithZeroResults">
+                <h2>No secrets.</h2>
+                <p>No secrets have been added to project {{projectName}}.</p>
+                <p ng-if="project && ('secrets' | canI : 'create') && !filterWithZeroResults">
```
spadgett: You can remove `!filterWithZeroResults` from this `ng-if` since it's already checked in the `div` above.

app/views/secrets.html
```diff
@@ -14,28 +14,39 @@
...
+                  <a ng-href="project/{{project.metadata.name}}/create-secret" class="btn btn-primary btn-lg">Create Secret</a>
+                </p>
+              </div>
+              <div ng-if="filterWithZeroResults">
+                <h2>The filter is hiding all secrets. <a href="" ng-click="clearFilter()" role="button" class="nowrap">Clear Filter</a></h2>
+              </div>
             </div>
           </div>
-          <table ng-if="secrets.length" class="table table-bordered table-mobile secrets-table table-layout-fixed">
+          <table ng-if="(secrets | hashSize) > 0" class="table table-bordered table-mobile secrets-table table-layout-fixed">
```
spadgett:
```html
ng-if="secrets | size"
```

app/views/secrets.html
jhadvig: It does :+1:

jhadvig:
@spadgett comments addressed

app/scripts/controllers/secrets.js
```diff
@@ -19,10 +23,26 @@ angular.module('openshiftConsole')
         $scope.project = project;
         $scope.context = context;
         watches.push(DataService.watch("secrets", context, function(secrets) {
-          $scope.secrets = _.sortBy(secrets.by("metadata.name"), ["type", "metadata.name"]);
```
spadgett: It looks like we've lost the sort?

jhadvig: That should be done by the `$scope.unfilteredSecrets = secrets.by("metadata.name");` on the line 27. No ?

spadgett: I don't see the call to `_.sortBy` there?

jhadvig: @spadgett updated :)

app/views/secrets.html
```diff
@@ -14,28 +14,39 @@
           </span>
         </h1>
       </div>
+      <div ng-if="(unfilteredSecrets | size) || filterWithZeroResults" class="data-toolbar">
```
spadgett: I don't think we need the filter with zero results check?

```html
<div ng-if="unfilteredSecrets | size" class="data-toolbar">
```

app/views/secrets.html
```diff
@@ -14,28 +14,39 @@
...
-          <div ng-if="!secrets.length">
-            <div class="empty-state-message text-center">
-              <h2>No secrets.</h2>
-              <p>
-                No secrets have been added to project {{projectName}}.
-              </p>
-              <p ng-if="project && ('secrets' | canI : 'create') && !secrets.length">
-                <a ng-href="project/{{project.metadata.name}}/create-secret" class="btn btn-primary">Create Secret</a>
-              </p>
+          <div ng-if="(secrets | size) == 0">
```
spadgett: Prefer `===`

```html
<div ng-if="(secrets | size) === 0">
```

jhadvig:
@spadgett updated :)
PTAL

PR #2481, opened 2017-11-08
Bug 1510786 - Error info should be more precise when create app in project with view role

jhadvig:
Instead of trying to send a request that will fail, we shouldn't show the "Create" button, if he doesn't have the permissions.

create

Not really sure if we shouldn't hide both `Create | Cancel` butwould rather keep the `Cancel` btn for the navigation.

@spadgett PTAL

### DIFF
--- a/app/scripts/controllers/newfromtemplate.js
+++ b/app/scripts/controllers/newfromtemplate.js
@@ -14,6 +14,7 @@ angular.module('openshiftConsole')
                        $parse,
                        $routeParams,
                        $scope,
+                       AuthorizationService,
                        CachedTemplateService,
                        DataService,
                        Navigate,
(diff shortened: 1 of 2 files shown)


spadgett:
@jhadvig I think it would be better to just show an error message when the page loads saying you don't have authority to instantiate templates.

It's possible the template fails for other reasons and we've lost the real error as well, so I think even hiding the button doesn't totally fix the problem.

spadgett:
> I think it would be better to just show an error message when the page loads saying you don't have authority to instantiate templates.

By this I mean a full page error with none of the fields visible

jhadvig:
@spadgett @jwforres comments addressed. Now there is a full page error:

err

PR #2484, opened 2017-11-08
Remember drawer expanded state across sessions

spadgett:
/assign @jwforres 
@benjaminapetersen

### DIFF
--- a/app/scripts/directives/notifications/notificationDrawerWrapper.js
+++ b/app/scripts/directives/notifications/notificationDrawerWrapper.js
@@ -11,9 +11,10 @@
         '$filter',
         '$interval',
         '$location',
-        '$timeout',
-        '$routeParams',
         '$rootScope',
+        '$routeParams',
+        '$scope',
+        '$timeout',
         'Constants',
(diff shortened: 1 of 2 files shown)


spadgett:
Fixing pet peeves :)

spadgett:
This is consistent with how we remember the nav expanded state.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/nav.js#L203-L219

app/scripts/directives/notifications/notificationDrawerWrapper.js
```diff
@@ -246,7 +249,7 @@
       angular.extend(drawer, {
         drawerHidden: true,
         allowExpand: true,
```
benjaminapetersen: Ah so you would prefer it would remember this state

spadgett: The idea is if you prefer it open to see descriptions, you don't have to click the little expand button every time. And we tend to remember other collapsed / expanded states like overview list rows and nav.

PR #2488, opened 2017-11-08
Allow setting environment variable prefix when adding secret to application

### LINKED ISSUE #2475: Allow environment variable prefix to be defined when adding secret to application.
When using service broker with service catalog under 3.7, to add the secret from the _ServiceBinding_, the _Add to Application_ screen should provide the ability to supply an environment variable prefix for names of environment variables created.

image

In other words, should provide the same functionality as:

```
oc set env --from secret/xxx --prefix POSTGRESQL_
```

Without this, you can't use web console to do it where a clash would occur as no way to override the names using a prefix.

This was found using 3.7.0-rc.0 using builtin in PostgreSQL template via template service broker.

jeff-phillips-18:
Fixes #2475 

image

image

image

### DIFF
--- a/app/views/directives/add-config-to-application.html
+++ b/app/views/directives/add-config-to-application.html
@@ -8,7 +8,7 @@
         <legend>Add this {{ctrl.apiObject.kind | humanizeKind}} to application:</legend>
         <div class="form-group"  ng-class="{'has-error' : ctrl.addType === 'env' && ctrl.application && !ctrl.canAddRefToApplication}">
           <div class="application-select">
-            <ui-select id="application" ng-model="ctrl.application" on-select="ctrl.checkApplicationContainersRefs($item)" required="true" ng-disabled="ctrl.disableInputs">
+            <ui-select autofocus id="application" ng-model="ctrl.application" on-select="ctrl.checkApplicationContainersRefs($item)" required="true" ng-disabled="ctrl.disableInputs">
               <ui-select-match placeholder="{{ctrl.applications.length ? 'Select an application' : 'There are no applications in this project'}}">
               <span>
                 {{$select.selected.metadata.name}}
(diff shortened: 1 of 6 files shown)


jeff-phillips-18:
@openshift/team-ux-review

app/views/directives/add-config-to-application.html
```diff
@@ -31,22 +31,42 @@
         <legend>Add {{ctrl.apiObject.kind | humanizeKind}} as:</legend>
         <div class="form-group">
           <div class="radio">
-            <div class="add-choice">
-              <label>
-                <input type="radio" ng-model="ctrl.addType" value="env" ng-disabled="ctrl.disableInputs">
-                Environment variables
-              </label>
+            <label class="add-choice" for="env">
+              <input id="env" type="radio" ng-model="ctrl.addType" value="env" ng-disabled="ctrl.disableInputs">
```
spadgett: Instead of having the `ctrl.disableInputs` check on every input, we could wrap the form in  a `<fieldset>` and add the check once there.

app/views/directives/add-config-to-application.html
```diff
@@ -31,22 +31,42 @@
...
+              Environment variables
+            </label>
+            <div class="choice-options">
+              <div class="form-horizontal">
+                <span class="control-label">Prefix</span>
```
spadgett: Should we be using `class="form-group"` and are the `choice-options` styles still necessary if we do?

app/views/directives/add-config-to-application.html
```diff
@@ -31,22 +31,42 @@
...
+                <span class="control-input" ng-class="{'has-error': addToApplicationForm.envPrefix.$error.pattern && addToApplicationForm.envPrefix.$touched}">
+                  <input class="form-control"
+                         name="envPrefix"
+                         id="envPrefix"
+                         placeholder="Optionally, specify a prefix for the variables"
+                         type="text"
+                         ng-pattern="/^[A-Za-z_][A-Za-z0-9_]*$/"
+                         aria-describedby="env-prefix-help"
```
spadgett: I don't see an element with this id

spadgett:
@jeff-phillips-18 We have a `take-focus` directive, but I don't know if it works with `ui-select`

jeff-phillips-18:
Updated the layout a bit keep the two sections more consistent:

http://g.recordit.co/FkVZCsN7uR.gif

jeff-phillips-18:
@openshift/team-ux-review

beanh66:
@jeff-phillips-18 Looks great! The only thing I noticed is the mount path doesn't have an * to show it is required when volume is selected.

jeff-phillips-18:
Added the required field marker.

beanh66:
LGTM!

jeff-phillips-18:
I believe all issues have been addressed.

spadgett:
I know the field level help is a Patternfly pattern, but we don't use it anywhere else in the console. I'm hesitant to introduce it in just one dialog when it's inconsistent with all of our other forms.

The placeholder text more of less repeats the label and help text. I'd rather just leave the placeholders off in that case.

I understand why the different indentation on the warning and error messages is there and that it's intentional, but it looks strange to the eye. I wonder how the env var name warning looks above the prefix field?

serenamarie125:
@spadgett although I appreciate your thoughts on consistency, I think that introducing a new element which is improving the experience makes sense.

A couple of other points - many OpenShift customers are using other RH products which are using this patterns & the field level help pattern isn't specific to PatternFly, I've seen it around a lot & it's pretty intuitive to use.

jeff-phillips-18:
@spadgett 
image

serenamarie125:
Removing the placeholder text is fine.  The only thing that it currently spells out is that the prefix is optional.  Mount path has an asterisk, so it's clear that it's required, but when Env variables is selected, the only field shown is optional, so I think it's nice to have

spadgett:
> Removing the placeholder text is fine. The only thing is, that it currently spells out is that the prefix is optional. Mount path has an asterisk, so it's clear that it's required. When Env variables is selected, the only field shown is optional, so I think it's nice to have

This is one reason why I'm hesitant to switch to tooltips. The tooltip states that it's optional, but it's not visible unless you hover. So we need to repeat the tooltip in a placeholder.

serenamarie125:
@spadgett, I think adding (optional) as a placeholder text would be fine.  The field level help string was not a 100% match of the placeholder text.

FYI field level help is on click not hover.

jeff-phillips-18:
Based on above:

image

image

image

image

image

image

beanh66:
Looks good @jeff-phillips-18

spadgett:
@serenamarie125 I'm concerned that introducing inconsistencies is going to make the user experience worse, not better.

>  I think adding (optional) as a placeholder text would be fine

Does this mean we should add "(optional)" in other places? I think this will make other forms look really cluttered where most fields are optional. But if we don't, then this field is inconsistent.

We generally use placeholders with example values when we use them in the console.

> FYI field level help is on click not hover.

For better or worse, every other tooltip in the console is hover. So while click is consistent with Patternfly, we're introducing an inconsistency with the rest of the console.

beanh66:
@spadgett I don't think showing "optional" is needed in other cases because usually forms have some fields that are required (marked with *) and some that are then clearly optional (not marked with *). 

Where there is only a single field shown (this particular case) and that field is optional, IMO we should clearly state it to the user. I would also say if there is a case where ALL fields are optional, we should clearly state that to the user since the * would not be present.

jeff-phillips-18:
@spadgett where do we stand with this? Is there something I need to change?

spadgett:
Flake #1685

/retest

PR #2492, opened 2017-11-09
Correctly handle Enter keypress in deploy image dialog

### LINKED ISSUE #2490: Pressing enter in Name field of deploy with Image Name, should not re-query image details.
When using the service catalog in 3.7-rc.0, if you select 'Image Name', and enter the image reference, eg., 'openshiftkatacoda/blog-django-py' and press enter, or select on the magnifying glass, it will query Docker Hub and pull down image details. That much is good.

The 'Name' field for the deployment will be set the name of the image, eg., 'blog-django-py'. If you edit the 'Name' field to change the name for the resources created and press enter in the 'Name' field, it will requery the details for the image from Docker Hub, throwing away the value of the 'Name' field you just entered.

Pressing enter in the 'Name' field should not do that. It should just accept the change.

image

cc @jorgemoralespou

spadgett:
Fixes #2490

/kind bug
/assign @jwforres

### DIFF
--- a/app/views/directives/deploy-image-dialog.html
+++ b/app/views/directives/deploy-image-dialog.html
@@ -22,7 +22,7 @@
     <div class="wizard-pf-main-inner-shadow-covers">
       <div class="order-service-config">
         <div class="wizard-pf-main-form-contents">
-          <form name="$ctrl.deployForm">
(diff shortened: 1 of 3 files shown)


jwforres:
what was the decision on this? were we going to try the other approach with the aria-hidden submit button within the form element?

Issue #1227, opened 2017-02-08
Unclear why Save is disabled on forms when validation is failing within hidden sections


jwforres:
If the user shows advanced options and changes something that fill fail form validation, then hides the advanced sections, the Save button will be disabled but they will not know why.

jwforres:
Follow up for https://bugzilla.redhat.com/show_bug.cgi?id=1419887

PR #2494, opened 2017-11-09
Show existing git settings by default in BC editor

### LINKED ISSUE #1910: When entering edit build config page, if I have a branch ref set for my source it should be visible
Currently the branch ref field is always hidden when entering edit build config, even if the value is set.  If a user has set that field we should always show it, otherwise a user doesn't know that it is set.

spadgett:
If git reference, context dir, or source secret are set for a build config,
show them by default when opening the build config editor. Don't make the user
click "advanced options" to see that values are already set.

Fixes #1910 

/kind bug
/assign @jwforres


spadgett:
I wonder if we shouldn't get away from the "advanced options" links in general. They seem to have some usability problems like this bug and #1227

PR #2495, opened 2017-11-09
Bug 1506998 - DC's environment tab not showing image secrets properly

jhadvig:
Image secrets of `kubernetes/dockercfg` and `kubernetes/dockerconfigjson` havent been shown properly on the DC's environment tab.

Screen:
dc_page 1

After fix:
dc_page

@spadgett PTAL

### DIFF
--- a/app/scripts/directives/editEnvironmentFrom.js
+++ b/app/scripts/directives/editEnvironmentFrom.js
@@ -24,13 +24,23 @@
     var canI = $filter('canI');
     var humanizeKind = $filter('humanizeKind');
     var uniqueId = _.uniqueId();
+    var keyValidator = /^[A-Za-z_][A-Za-z0-9_]*$/;
 
     ctrl.setFocusClass = 'edit-environment-from-set-focus-' + uniqueId;
+
+    ctrl.isEnvVarInvalid = function(keyName) {
+      return !keyValidator.test(keyName);
+    };
+
+    ctrl.hasInvalidEnvVar = function(secret) {
+      return _.some(secret, function(value, key) {
+        return ctrl.isEnvVarInvalid(key);
+      });
+    };
 
     ctrl.viewOverlayPanel = function(entry) {
       ctrl.decodedData = entry.data;
       ctrl.overlayPaneEntryDetails = entry;
-
       if (entry.kind === 'Secret') {
         ctrl.decodedData = SecretsService.decodeSecretData(entry.data);
       }
--- a/app/styles/_kve.less
+++ b/app/styles/_kve.less
@@ -156,6 +156,9 @@
         float: left;
         width: 100%;
       }
+      .has-warning {
+        display: inline-block;
+      }
     }
     .environment-from-view-details {
       float: left;
--- a/app/views/directives/edit-environment-from.html
+++ b/app/views/directives/edit-environment-from.html
@@ -61,6 +61,9 @@
               </ui-select-choices>
             </ui-select>
           </div>
+          <div class="has-warning" ng-if="$ctrl.hasInvalidEnvVar(entry.selectedEnvFrom.data)">
+            <div class="help-block">Some of the keys for {{entry.selectedEnvFrom.kind | humanizeKind}} <strong>{{entry.selectedEnvFrom.metadata.name}}</strong> are not valid environment variable names and will not be added.</div>
+          </div>
         </div>
       </div>
 
@@ -149,8 +152,11 @@
       <div ng-if="$ctrl.overlayPaneEntryDetails.data | size" class="table-responsive scroll-shadows-horizontal">
         <table class="table table-bordered table-bordered-columns config-map-table key-value-table">
           <tbody>
-            <tr ng-repeat="(prop, value) in $ctrl.decodedData">
-              <td class="key">{{prop}}</td>
+            <tr ng-repeat="(key, value) in $ctrl.decodedData">
+              <td class="key">
+                {{key}}
+                <span ng-if="$ctrl.isEnvVarInvalid(key)" class="pficon pficon-warning-triangle-o tooltip-default-icon" data-toggle="popover" data-trigger="hover" dynamic-content="{{key}} is not a valid environment variable name and will not be added."></span>
+              </td>
               <td class="value">
                 <truncate-long-text
                   ng-if="$ctrl.overlayPaneEntryDetails.kind === 'ConfigMap'"
@@ -168,7 +174,7 @@
                     newline-limit="2"
                     expandable="true">
                   </truncate-long-text>
-                  <div ng-if="decodedData.$$nonprintable[prop]" class="help-block">
+                  <div ng-if="decodedData.$$nonprintable[key]" class="help-block">
                     This secret value contains non-printable characters and is displayed as a Base64-encoded string.
                   </div>
                 </div>
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -9218,19 +9218,25 @@ n[e.key] = e.value;
 } ]), function() {
 angular.module("openshiftConsole").component("editEnvironmentFrom", {
 controller: [ "$attrs", "$filter", "keyValueEditorUtils", "SecretsService", function(e, t, n, a) {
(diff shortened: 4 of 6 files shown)


spadgett:
For these dialogs, I'd really like to see the raw values that will actually be injected into my pod as the environment variable. If that's a JSON blob, we should show the JSON.

jhadvig:
So in case of an object thats key/values cant be properly displayed we should show the json ?

spadgett:
We should show the same value that would be injected if I checked from the shell using `echo $MY_VAR`. Whatever the raw decoded content of the secret is without any special handling for these types.

spadgett:
This one is sort of strange since `.dockercfg` would actually be skipped. It's not a valid env var name.

jhadvig:
So have been thinking about the fact that the `.dockercfg` would be skipped, since its not a valid env var name.
So besides the warning that the secret wont take place in the pod, which wont help the user much, we could do:
1, modify the envVar key so it fulfill the envVar naming conventions, by creating a filter that could remove any unwanted chars:
dc-object-env
Here im just not sure if a json object inside a envvar would be a good user-exp..

2, for the case of image secrets we could just process the `auths` and print them in following fashion:
dc-auths
The key naming is just an idea how it could look like...


Those are just suggestions at this point Im ok with the warning and skipping the image secret, but would be good to have a option how to get those secrets into the pod.

spadgett:
@jhadvig Since `envFrom` is a k8s feature where you can only specify the secret name, I'm not sure there's much we can do in the console other than let the user know these keys won't be injected.

The way around is to use `valueFrom` for the keys that aren't valid env variable names. Then you can pick the name. So it's not impossible to inject those values.

jhadvig:
@spadgett PR update based on our discussion.
Screens:
warn2
warn1

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -26,10 +26,24 @@
...
+      return false;
+    };
 
     ctrl.viewOverlayPanel = function(entry) {
       ctrl.decodedData = entry.data;
       ctrl.overlayPaneEntryDetails = entry;
+      if (ctrl.isImageSecret(ctrl.decodedData)) {
+        ctrl.alerts["cantInjectSecret"] = {
+          type: "warning",
+          message: "This secret keys won't be infected into the deployment pod."
```
spadgett: It's not just image secrets. It's secrets that have keys with names that are not valid env vars. See

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/addConfigToApplication.js#L63-L66

We should warn about what keys specifically won't be added using the regex. I think maybe a warning icon by the name in this dialog with tooltip text saying it's not a valid environment variable name.

jhadvig:
@spadgett I've update the PR.
Now when the secret will contain invalid keys user will be warned under the selected secret + the dialog will show which keys wont be added.
env

PTAL

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -24,13 +24,26 @@
     var canI = $filter('canI');
     var humanizeKind = $filter('humanizeKind');
     var uniqueId = _.uniqueId();
+    var keyValidator = new RegExp("^[A-Za-z_][A-Za-z0-9_]*$");
```
spadgett: Better to use a regex literal. It improves performance, so it's not just style. (I realize this was probably copied from elsewhere.)

```js
var keyValidator = /^[A-Za-z_][A-Za-z0-9_]*$/;
```

app/scripts/directives/editEnvironmentFrom.js
```diff
@@ -24,13 +24,26 @@
...
     ctrl.setFocusClass = 'edit-environment-from-set-focus-' + uniqueId;
 
+    ctrl.isEnvVarInvalid = function(keyName) {
+      return !keyValidator.test(keyName);
+    };
+
+    ctrl.hasInvalidEnvVar = function(secret) {
+      if (!secret) {
+        return false;
+      }
```
spadgett: This `if (!secret)` block can be removed. `_.some` handles null values.

app/views/directives/edit-environment-from.html
```diff
@@ -149,8 +152,11 @@
       <div ng-if="$ctrl.overlayPaneEntryDetails.data | size" class="table-responsive scroll-shadows-horizontal">
         <table class="table table-bordered table-bordered-columns config-map-table key-value-table">
           <tbody>
-            <tr ng-repeat="(prop, value) in $ctrl.decodedData">
-              <td class="key">{{prop}}</td>
+            <tr ng-repeat="(key, value) in $ctrl.decodedData">
+              <td class="key">
+                <span ng-if="$ctrl.isEnvVarInvalid(key)" class="pficon pficon-warning-triangle-o" style="cursor: help; vertical-align: middle;" data-toggle="popover" data-trigger="hover" dynamic-content="{{key}} is not a valid environment variable name and will not be added."></span>
```
spadgett: Better to have these styles in a CSS file somewhere than declaring them inline. I would guess we have a tooltip class somewhere. If not, we should add it. (I know we have inline tooltip styles other places, but I'd like to get out of the habit of doing that since it's not a good practice.)

cc @rhamilto @sg00dwin

spadgett:
Usually the icon comes after the name most places we show a warning icon. @jwforres @rhamilto Any strong opinion?

rhamilto:
> Usually the icon comes after the name most places we show a warning icon. 

Agree it should be consistent with other instances and come after the text.

jhadvig:
@spadgett @rhamilto comments addressed. PTAL

put the warning icon after the key name:
err2

app/styles/_add-config-to-application.less
```diff
@@ -52,6 +52,14 @@
     margin-bottom: 10px;
   }
 
+  .key {
+    white-space: nowrap;
```
spadgett: Key names can be quite long. I don't think we should add nowrap

app/views/directives/edit-environment-from.html
```diff
@@ -61,6 +61,9 @@
               </ui-select-choices>
             </ui-select>
           </div>
+          <div class="has-warning" ng-show="$ctrl.hasInvalidEnvVar(entry.selectedEnvFrom.data)">
+              <div class="help-block">Some of the keys for {{entry.selectedEnvFrom.kind | humanizeKind}} <strong>{{entry.selectedEnvFrom.metadata.name}}</strong> are not valid environment variable names and will not be added.</div>
```
spadgett: nit: indentation

app/styles/_add-config-to-application.less
```diff
@@ -52,6 +52,14 @@
...
+    .pficon-warning-triangle-o {
```
spadgett: Can you try reusing this style instead of adding a new one?

https://github.com/spadgett/origin-web-console/blob/0a97382207e3c1d1d85657abde47f01a43145c84/app/styles/_tooltip.less#L6-L9

I think we need one common tooltip style so we're consistent everywhere

app/styles/_kve.less
```diff
@@ -156,6 +156,9 @@
         float: left;
         width: 100%;
       }
+      .has-warning {
+        display:inline-block;
```
spadgett: nit: space after the `:` for consistent code style

app/views/directives/edit-environment-from.html
```diff
@@ -61,6 +61,9 @@
               </ui-select-choices>
             </ui-select>
           </div>
+          <div class="has-warning" ng-show="$ctrl.hasInvalidEnvVar(entry.selectedEnvFrom.data)">
```
spadgett: Any reason not to use `ng-if`  instead of `ng-show`?

jhadvig: Was looking on https://github.com/openshift/origin-web-console/blob/master/app/views/directives/osc-routing.html#L27 as an example no which attr are we using, but Now I've noticed that we are mostly using `ng-show`. Sorry for that

jhadvig:
@spadgett comments addressed.
After removing the extra css the dialog looks like:
err3

PR #2499, opened 2017-11-13
Update PersistentVolumeClaim controller to use getPreferredVersion

benjaminapetersen:


benjaminapetersen:
love how having a dist makes a 5 line change into a `size/L` with the new system 😄

benjaminapetersen:
rebased

PR #2504, opened 2017-11-14
[WIP] Upgrading Angular-Bootstrap to v2.5.0

### LINKED ISSUE #1861: Upgrade to angular-bootstrap v2.2.x is needed
Angular-patternfly uses angular-bootstrap v2.2.x.  If console is to achieve parity, it must also upgrade from angular-bootstrap v0.14.3 to v2.2.x.

uib-bootstrap update problems:
- [ ] Fix `?tab` links that don't switch to the tab
- [ ] Fix runtime error on events tab
- [ ] Fix empty membership page
- [ ] Update instances of class="uib-dropdown-menu" to uib-dropdown-menu in catalog

Some dedicated work and time is probably need to fix the visual issues.

rhamilto:
Fixes #1861 

@jwforres, @spadgett, thoughts on this approach to updating persist-tab-state to work with the new version of uib-tabset?

### DIFF
--- a/app/scripts/directives/util.js
+++ b/app/scripts/directives/util.js
@@ -238,23 +238,22 @@ angular.module('openshiftConsole')
       restrict: 'A',
       scope: false,
       link: function(scope) {
-        scope.selectedTab = scope.selectedTab || {};
+        scope.selectTab = function(tab){
+          scope.selectedTab = tab;
+        };
         scope.$watch(function() {
           return $routeParams.tab;
         }, function(tab) {
           if (!tab) {
             return;
           }
-
-          scope.selectedTab[tab] = true;
+          scope.selectedTab = tab;
         });
 
         scope.$watch('selectedTab', function() {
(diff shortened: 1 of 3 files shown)


spadgett:
The problem is that `persist-tab-state` has no idea what index to set for a tab name. It will require some rework.

I was thinking of doing something similar to what we do for wizards where we define tabs and their views like:

```js
$scope.tabs = [{
  id: 'details',
  label: 'Details',
  view: 'views/pod/_pod-details.html'
}, {
  id: 'logs',
  label: 'Logs',
  view: 'views/pod/_pod-logs.html'
}
// other tabs ...
];
```

Then we could either wrap `uib-tabset` with our own component that handles everything or pass the tabs object to `persist-tab-state` so it can figure out the right index for a tab name. But even that gets a bit tricky if we dynamically show and hide some tabs.

I think this would make things more manageable long term, but it's some work. This is part of why we punted on this particular dependency.

PR #2518, opened 2017-11-16
Update browseCategory to use getPreferredVersion

benjaminapetersen:
This one is not tested, how do we trigger this route?


spadgett:
You can go to /projects/my-project/create, although this page should be removed this release. I wouldn't worry about it.

benjaminapetersen:
Thought that might be the case.  Will close.

PR #2520, opened 2017-11-17
Update nextSteps controller to use getPreferredVersion

benjaminapetersen:
This one might also be on the chopping block...


benjaminapetersen:
rebased

PR #2526, opened 2017-11-17
Update util/oauth controller to use getPreferredVersion

benjaminapetersen:
Depends on https://github.com/openshift/origin-web-common/pull/250 before it can be fully resolved.


benjaminapetersen:
rebased, `users` is now in apiPreferredVersion, removed [WIP]

PR #2528, opened 2017-11-20
Configurable inactivity timeout in console

jhadvig:
@spadgett @jwforres here is the first draft on the automatic logout upon user inactivity.
I've hardcoded the the inactivity period to 10 seconds, after that a dialog will be shown. If user cancels or dismisses the dialog then the inactivity period will be refreshed. If user wont make any activity or clicks 'Log Out' all the openshift tabs will be logged out. 

DEMO:
nov 20 2017 4_33 pm

Will do additional refactoring next couple of days.

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -486,7 +486,113 @@ angular
       LabelFilter.readPersistedState();
     });
   })
-  .run(function(durationFilter, timeOnlyDurationFromTimestampsFilter) {
+  .run(function($location, $uibModal, AuthService, Constants) {
+    if (Constants.INACTIVITY_TIMEOUT_PERIOD <= 0) {
+      return;
+    }
+    var lastInteractionKey = 'origin-web-console-last-interaction-timestamp';
+    var inactivityLogoutKey = 'origin-web-console-inactivity-logout';
+    var isLocalLogoutModalShown = false;
+    var checkInteractionInterval;
+    var modalInstance;
+    // Interval that will check for user inactivity every minute.
+    // In case the last interaction is higher then  INACTIVITY_TIMEOUT_PERIOD constant a logout modal will be shown.
+    // Also check for 'origin-web-console-inactivity-logout' object in the local storage, that indicates that the
+    // user has been already logged out, in case of multiple opened console tabs.
+    var startInteractionIntervalCheck = function () {
+      checkInteractionInterval = setInterval(function(){
+        if (!AuthService.isLoggedIn()) {
+          return;
+        }
+        var lastInteraction = Date.parse(localStorage.getItem(lastInteractionKey));
+        if (isNaN(lastInteraction)) {
+          Logger.warn("Last interaction timestamp has been corrupted. The logout timeout will be restarted.");
+          lastInteraction = new Date();
+        }
+        var currentTime = new Date();
+        // Since the INACTIVITY_TIMEOUT_PERIOD is set in minutes it needs to be converted into milliseconds.
+        if (currentTime - lastInteraction > Constants.INACTIVITY_TIMEOUT_PERIOD * 60000) {
+          showLogoutModal();
+        }
+      }, 60000);
+    };
+
+    // In case of multiple opened tabs the interval needs to be restarted everytime there is any activity in any tab,
+    // so all intervals are synced, and they start the counting at the same time.
+    var restartCheckInteractionInterval = function() {
+      // Handles cases of user activity during the opened modal, where user clicks inside the modal but not on the buttons.
+      if (modalInstance) {
+        modalInstance.dismiss("User activity");
+        modalInstance = null;
+      }
+      clearInterval(checkInteractionInterval);
+      startInteractionIntervalCheck();
+    };
+    // Updates lastInteraction date and restarts interval
+    // restartCheckInteractionInterval() has to be called, cause because the listener on the `storage` event
+    // is not fired in the tab where the event happened.
+    var updateLastInteraction = function() {
+      restartCheckInteractionInterval();
+      localStorage.setItem(lastInteractionKey, new Date().toString());
+    };
+    // Updates Local Storage variable that indicates if all opened tabs should log out current user.
+    var logoutAllTabs = function(bool) {
+      localStorage.setItem(inactivityLogoutKey, bool.toString());
+    };
+    // Logout with passing reason.
+    var logout = function() {
+      var logoutURI = URI.expand("/logout{?cause*}", {
+        cause: 'timeout'
+      });
+      $location.url(logoutURI.toString());
+    };
+    // Show inactivity logout modal.
+    // Before the the modal is shown the handler for checking the changes in local storage is detached.
+    // After the the modal is closed or dismissed the handler will be attached again. This is because of 
+    // the fact that the handler is attached 
+    function showLogoutModal() {
+      if (isLocalLogoutModalShown) {
+        return false;
+      }
+      isLocalLogoutModalShown = true;
+      modalInstance = $uibModal.open({
+        animation: true,
+        templateUrl: 'views/modals/logout.html',
+        controller: 'LogoutModalController',
+      });
+      modalInstance.result.then(function(event) {
+        if (event === 'logout') {
+          logoutAllTabs(true);
+          logout();
+        } else if (event === 'cancel') {
+          // In case of multiple tabs, when the logoutModal would be canceled, the timer has to be restarted.
+          updateLastInteraction();
+          isLocalLogoutModalShown = false;
+        }
+      }, function () {
+        updateLastInteraction();
+        isLocalLogoutModalShown = false;
+      });
+    }
+
+    // Need to check for changes in last interaction so the interval can be restarted. 
+    $(window).on('storage', function (event) {
+      if (event.originalEvent.key === lastInteractionKey) {
+        restartCheckInteractionInterval();
+      } else if (event.originalEvent.key === inactivityLogoutKey) {
+        logout();
+      }
+    });
+
+    // Reset 'origin-web-console-inactivity-logout' key on login.
+    AuthService.onUserChanged(function() {
+      logoutAllTabs(false);
+    });
+    updateLastInteraction();
+    // Bind to click and keydown events so the in last interaction timestamp is updated.
+    $(document).bind("click keydown", _.throttle(updateLastInteraction, 500));
+  })
+  .run(function(durationFilter, timeOnlyDurationFromTimestampsFilter, countdownToTimestampFilter) {
     // Use setInterval instead of $interval because we're directly manipulating the DOM and don't want scope.$apply overhead
     setInterval(function() {
       // Set by duration-until-now directive.
@@ -501,6 +607,10 @@ angular
         else {
           return durationFilter(timestamp, null, omitSingle, precision) || existing;
         }
+      });
+      $('.countdown[data-timestamp]').text(function(i, existing) {
+        var endTimestamp = $(this).data("timestamp");
+        return countdownToTimestampFilter(endTimestamp);
       });
     }, 1000);
   })
--- a/app/scripts/constants.js
+++ b/app/scripts/constants.js
@@ -109,6 +109,11 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     namespace: "openshift"
   },
 
+
+  // Set to number of minutes after which a modal with countdown should appear.
+  // When set to anything else then positive integer the inactivity timeout won't be enable. 
+  INACTIVITY_TIMEOUT_PERIOD: 0,
+
   // only resources from the namespaces listed below can be utilized with create from url (/create)
   // 'openshift' should always be included
   CREATE_FROM_URL_WHITELIST: ['openshift'],
--- a/app/scripts/controllers/modals/logoutModal.js
+++ b/app/scripts/controllers/modals/logoutModal.js
@@ -1,0 +1,32 @@
+'use strict';
+/* jshint unused: false */
+
+/**
+ * @ngdoc function
+ * @name openshiftConsole.controller:LogoutModalController
+ * @description
+ * # LogoutModalController
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('LogoutModalController', function ($timeout, $location, $filter, $scope, $uibModalInstance, Constants) {
+    $scope.endTimestamp = moment().add(30, 'seconds').toString();
+
+    var logoutTimeout = $timeout(function(){
+      $scope.logout();
+    }, 30000);
+
+    $scope.logout = function() {
+      $timeout.cancel(logoutTimeout);
+      $uibModalInstance.close('logout');
+    };
+
+    $scope.cancel = function() {
+      $timeout.cancel(logoutTimeout);
+      $uibModalInstance.close('cancel');
+    };
+
+    $scope.$on('$destroy', function(){
+      $timeout.cancel(logoutTimeout);
+    });
+  });
--- a/app/views/modals/logout.html
+++ b/app/views/modals/logout.html
@@ -1,0 +1,12 @@
+<div class="modal-resource-action inactivity-modal">
+  <form>
+    <div class="modal-body">
+      <h1>Your session is about to expire due to inactivity.</h1>
+      <h2>You will be logged out in <strong><time-remaining-from-now end-timestamp="endTimestamp" countdown-duration=""></time-remaining-from-now></strong> seconds.</h2>
+    </div>
+    <div class="modal-footer">
+      <button class="btn btn-lg btn-default" type="submit" ng-click="logout();">Log Out</button>
+      <button class="btn btn-lg btn-primary" type="button" ng-click="cancel();">Continue Session</button>
+    </div>
+  </form>
+</div>
--- a/app/index.html
+++ b/app/index.html
@@ -296,6 +296,7 @@
         <script src="scripts/controllers/modals/confirmReplaceModal.js"></script>
         <script src="scripts/controllers/modals/processOrSaveTemplateModal.js"></script>
         <script src="scripts/controllers/modals/linkService.js"></script>
+        <script src="scripts/controllers/modals/logoutModal.js"></script>
         <script src="scripts/controllers/modals/jenkinsfileExamplesModal.js"></script>
         <script src="scripts/controllers/modals/aboutComputeUnitsModal.js"></script>
         <script src="scripts/controllers/about.js"></script>
--- a/app/scripts/controllers/util/logout.js
+++ b/app/scripts/controllers/util/logout.js
@@ -8,7 +8,7 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('LogoutController', function ($scope, $log, AuthService, AUTH_CFG) {
+  .controller('LogoutController', function ($scope, $routeParams, $log, AuthService, AUTH_CFG) {
     $log.debug("LogoutController");
 
     if (AuthService.isLoggedIn()) {
@@ -36,6 +36,11 @@ angular.module('openshiftConsole')
       window.location.href = AUTH_CFG.logout_uri;
     } else {
       $log.debug("LogoutController, not logged in, logout complete");
-      $scope.logoutMessage = 'You are logged out. Return to the <a href="./">console</a>.';
+
+      var logoutMessage = "You are logged out.";
+      if ($routeParams.cause === "timeout") {
+        logoutMessage = "You have been logged out due to inactivity."
+      }
+      $scope.logoutMessage = logoutMessage + ' Return to the <a href="./">console</a>.';
     }
   });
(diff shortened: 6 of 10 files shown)


app/views/modals/logout.html
```diff
@@ -0,0 +1,13 @@
+<div class="modal-resource-action">
+  <!-- Use a form so that the enter key submits when typing a project name to confirm. -->
+  <form>
+    <div class="modal-body">
+      <h1>You have been inactive for period of time.</h1>
+      <h2>You are going to be logged out in <strong>{{timeToLogout}}</strong></h2>
+    </div>
+    <div class="modal-footer">
+      <button class="btn btn-lg btn-danger" type="submit" ng-click="logout();">Log Out</button>
```
jwforres: I'm not sure what the appropriate order if for these buttons.  @ncameronbritt can you look into what the typical behavior is for these automated logout dialogs.  I don't want someone to panic and click logout when they meant to cancel the logout.

app/scripts/app.js
```diff
@@ -481,6 +481,68 @@ angular
...
+        isLocalLogoutModalShown = false;
+        updateCloseAllModals(true);
+      });
+    };
+
+    updateLastInteraction();
+    updateLogOut(false);
+    updateCloseAllModals(false);
+
+    $(document).bind("click keydown keypress", _.throttle(updateLastInteraction, 500));
```
jwforres: why both keydown and keypress?

app/scripts/app.js
```diff
@@ -481,6 +481,68 @@ angular
...
+
+    setInterval(function(){
+      var lastInteraction = Date.parse(localStorage.getItem('lastInteraction'));
+      var newDate = new Date();
+      console.log(newDate - lastInteraction);
+      if (newDate - lastInteraction > 10000) {
+        showLogoutModal();
+      }
+      if (localStorage.getItem('logOut') === 'true') {
+        logout();
```
jwforres: I'm trying to decide if its worth tracking this to force all the other tabs to go to the logout page immediately... the token is no longer valid as soon as any of the tabs go through logout. So any API interaction happening in the background on the other tabs would trigger a redirect to the login flow, and if no API interaction happened then they would just get to the logout page on their own when their own timer fired. 

@spadgett thoughts?

app/scripts/app.js
```diff
@@ -481,6 +481,68 @@ angular
...
+  .run(function($location, $rootScope, $uibModal) {
+    if ($location.path() === '/logout' || $location.path() === '/error') {
+      return false;
+    }
+    var isLocalLogoutModalShown = false;
+    var updateLastInteraction = function() {
+      localStorage.setItem('lastInteraction', new Date());
+    };
+    var updateLogOut = function(bool) {
+      localStorage.setItem('logOut', bool);
```
jwforres: logOut and closeAllModals aren't specific enough keys, even if you append origin-web-console- to them, it needs to be something like origin-web-console-inactivity-logout or something

that said, im not sure i want to track either of these, see other comments below

app/views/modals/logout.html
ncameronbritt: Not sure you need two buttons. The action should be "stay logged in".  The user will be logged out anyway if she does nothing, hence this dialog. So is there any compelling reason to offer a "log out" action? I think it could be as simple as something like this: 
screen shot 2017-11-22 at 8 54 47 am

Another alternative is that you don't need any buttons. I like what they do [here](https://www.cloudera.com/documentation/enterprise/5-7-x/topics/cm_auto_logout.html) with a kind of "press any key to continue" prompt. If you're being logged out due to inactivity, it makes sense that any activity should keep the user logged in. 

Another thing I like about the example in that link is the way the log in page tells the user why they're seeing it again: "You were logged out due to inactivity."

cc @openshift/team-ux-review

jwforres:
We should still have a button for mobile I would think

serenamarie125:
Just had a conversation with a bunch of UX folks ... we landed here with the design: 
screen shot 2017-11-22 at 10 25 14 am

The Continue Session button should be the default button, and the Log Out should not be red.

serenamarie125:
@jwforres after being logged out, can we present the Login Screen with an inline notification (warning) stating "You have been logged out due to inactivity"

ncameronbritt:
@jhadvig @jwforres Related to this, I opened this issue: https://github.com/openshift/origin-web-console/issues/2530

jwforres:
On our own logout page yeah we could. But there is a URL we optionally
redirect to afterward that can be configured by the cluster admin so that
they can logout of their SSO environment completely if that's what they
want. That page is out of our hands at that point.

app/scripts/app.js
```diff
@@ -481,6 +481,68 @@ angular
...
+      var lastInteraction = Date.parse(localStorage.getItem('lastInteraction'));
+      var newDate = new Date();
+      console.log(newDate - lastInteraction);
+      if (newDate - lastInteraction > 10000) {
+        showLogoutModal();
+      }
+      if (localStorage.getItem('logOut') === 'true') {
+        logout();
+      }
+    }, 1000);
```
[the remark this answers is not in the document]
jhadvig: yeah, this was just a example value. doesnt make sense to check every second... If the timeout will be in minutes then it make sense to set the check for every minute.

app/scripts/app.js
jhadvig: Maybe just change the the message in the dialog, after the timeout.. something like:

login

That would make the logic more simple and we will solve the "redirect and notify" user issue, cause user will see that he was logged out.
@jwforres @ncameronbritt @serenamarie125  thoughts ?

app/scripts/controllers/modals/logoutModal.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * @ngdoc function
+ * @name openshiftConsole.controller:LogoutModalController
+ * @description
+ * # LogoutModalController
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('LogoutModalController', function ($timeout, $scope, $uibModalInstance) {
+  	$scope.timeToLogout = 20;
```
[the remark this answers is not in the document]
jhadvig: You mean like a constant in app.js ?

app/scripts/app.js
```diff
@@ -481,6 +481,68 @@ angular
       LabelFilter.readPersistedState();
     });
   })
+  .run(function($location, $rootScope, $uibModal) {
+    if ($location.path() === '/logout' || $location.path() === '/error') {
+      return false;
+    }
+    var isLocalLogoutModalShown = false;
+    var updateLastInteraction = function() {
+      localStorage.setItem('lastInteraction', new Date());
```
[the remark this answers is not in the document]
spadgett: We've been using the prefix `openshift/` for other keys and kebab case names, so I'd suggest something like `openshift/last-interaction-time`.

serenamarie125:
@jhadvig @jwforres what is the inactivity time period being set to ?

spadgett:
@serenamarie125 It will be configurable (disabled by default).

serenamarie125:
great thanks!

jhadvig:
@spadgett @jwforres I've update the PR addressing comments

The modal will now contain following btns:
logout-modal

When user is logout by inactivity he will see it in the logout page:
logout-timeout

The default logout page looks like:
logout-def

Also I've made following variables configurable:
 - **is_active** - Set to true if the the console shall check for user inactivity.
 - **timeout_period**  -  Set to number of milliseconds after which a modal with countdown should appear.
 - **check_for_timeout_period** - Set to number of milliseconds that shall console check for timeout.
 - **modal_countdown_period** -  Set to number of milliseconds that will be used in a for visual countdown in the modal.

jhadvig:
@ncameronbritt @serenamarie125 PTAL as well

ncameronbritt:
I would change the first sentence in the modal to say "Your session is about to expire due to inactivity." (I.e. give the reason why the session is expiring.)

This is a little bit of a nit, but I think the message on the logout page reads a little better as "You have been logged out due to inactivity."

jhadvig:
@ncameronbritt comments addressed ! Thanks not it looks better :) :+1:

app/scripts/constants.js
```diff
@@ -109,6 +109,17 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     namespace: "openshift"
   },
 
+  INACTIVITY_TIMEOUT: {
```
jhadvig: @jwforres @spadgett here we need to align on the default time that will be set. Also I believe that the `is_active` variable should be set to **false**, right ?

spadgett: There probably only needs to be one value. I'd suggest letting the user set the timeout in minutes. Make any value that is not positive (undefined, null, 0, -1) could mean off? @jwforres opinion?

Default should be disabled, yes.

Ultimately this should go into the AssetConfig when we split the console into a pod and openshift-ansible updated to have a flag to set it. But that can happen in a later PR.

jwforres: Agreed, one variable. Any positive integer is on, anything else is off.

jhadvig: So based on https://github.com/openshift/origin-web-console/pull/2528#discussion_r152555338 I thought that we want to be able to configure multiple varibles like those mentioned in https://github.com/openshift/origin-web-console/pull/2528#issuecomment-352446268, so we don't have to hardcode it into code.
In case we really want just a single variable that would set the inactivity period in minutes(none positive value would disable the inactivity logout), we need to align on following values that would need to be hardcoded:
**check_for_timeout_period** - Seconds or that shall console check for timeout - I guess it would be enough to check every minute ?
**modal_countdown_period** - Set to number of milliseconds that will be used in a for visual countdown in the modal - I've set it to 20 seconds in demo. Guess thats OK ?

jwforres: sorry i think there is some confusion here. I was just saying that you don't need `is_active` since setting `timeout_period` to something other than a positive integer would indicate whether we should be checking a timeout or not. And I agree with @spadgett that the timeout period should be configured in minutes.

I don't think you need `check_for_timeout_period`, checking once a minute for inactivity is sufficient. We will just document that it works that way.

For `modal_countdown_period` I would default to 30 seconds, that is what I commonly see.  I would also configure it in seconds, not milliseconds.  @spadgett were suggesting that we didn't need this one?

spadgett: > For modal_countdown_period I would default to 30 seconds, that is what I commonly see. I would also configure it in seconds, not milliseconds. @spadgett were suggesting that we didn't need this one?

Personally I don't see a ton of value in making it configurable and might cause some confusion in how the two values are different when an admin first configures it.

jwforres: Ok i'm fine with that. Let's just make it 30 seconds then and we'll need to document it.

jhadvig:
@spadgett @jwforres comments addressed. Now we only have one configurable variable **INACTIVITY_TIMEOUT_PERIOD** which is defaulted to false, otherwise it takes amount of minutes.
PTAL

app/scripts/controllers/modals/logoutModal.js
```diff
@@ -0,0 +1,35 @@
...
+        if (localStorage.getItem('closeAllModals') === 'true') {
+          $scope.cancel();
+        }
+        logoutTimeout = $timeout(onTimeout,1000);
+        if ($scope.timeToLogout <= 0) {
+          $scope.logout();
+        }
+    }
+
+    var logoutTimeout = $timeout(onTimeout,1000);
```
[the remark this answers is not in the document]
spadgett: +1, I'd probably keep a timestamp for when the logout will happen and then use relative time directive in the dialog. I'm not sure you really need a $timeout every second here (but I haven't looked closely at the code).

app/scripts/app.js
```diff
@@ -485,6 +485,80 @@ angular
...
+    }
+    var isLocalLogoutModalShown = false;
+    var updateLastInteraction = function() {
+      localStorage.setItem('origin-web-console-last-interaction-timestamp', new Date());
+    };
+    var updateLogOut = function(bool) {
+      localStorage.setItem('origin-web-console-inactivity-logout', bool);
+    };
+    var updateCloseAllModals = function(bool) {
+      localStorage.setItem('origin-web-console-dismiss-logout-modals', bool);
```
jwforres: i think we can eliminate this local storage variable.  when the user clicks the button to dismiss the dialogs and remain active that is going to set the last interaction time.

you can then have your modal add a listener on local storage https://stackoverflow.com/questions/28230845/communication-between-tabs-or-windows

that listener can dismiss the modal if it sees the key for the last interaction time is updated

you probably want to make sure that clicking the logout link does not reset the last interaction time, stopping propagation of the event from the ng-click of the button would probably fix this.

app/scripts/app.js
```diff
@@ -485,6 +485,80 @@ angular
...
+  .run(function($location, $uibModal, AuthService, Constants) {
+    if (Math.sign(Constants.INACTIVITY_TIMEOUT_PERIOD) < 1 || $location.path() === '/logout' || $location.path() === '/error') {
+      return false;
+    }
+    var isLocalLogoutModalShown = false;
+    var updateLastInteraction = function() {
+      localStorage.setItem('origin-web-console-last-interaction-timestamp', new Date());
+    };
+    var updateLogOut = function(bool) {
+      localStorage.setItem('origin-web-console-inactivity-logout', bool);
```
jwforres: you need to make sure this localStorage gets reset on login, the auth service has something like an onUserChanged i think that you could listen to

jhadvig: How about clearing the localStorage on logout ? Wouldnt that be enough ?

jwforres: no because there may be other tabs that haven't yet seen the flag to logout, think if you have 3+ tabs open

app/scripts/app.js
```diff
@@ -485,6 +485,81 @@ angular
       LabelFilter.readPersistedState();
     });
   })
+  .run(function($location, $uibModal, AuthService, Constants) {
+    if (!Constants.INACTIVITY_TIMEOUT_PERIOD > 0 || $location.path() === '/logout' || $location.path() === '/error') {
+      return false;
+    }
+    var isLocalLogoutModalShown = false;
+    var updateLastInteraction = function() {
+      localStorage.setItem('origin-web-console-last-interaction-timestamp', new Date());
```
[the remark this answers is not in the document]
jhadvig: But the value of the latest interaction is set in different cases:
1, when the loading the page ,but it might happen that user wont interact with the page anymore so we need to set it by default
2, on user interaction
3, when dismissing the logout modal

Not really sure whats the benefit of setting a local var and updating the localStorage with it everytime, since the those cases might be mutual exclusive. Thought that having a simple helper that would just update the latestInteraction key in the local storage directly when any of those cases occur

app/scripts/app.js
jhadvig: That would also handle the case you mentioned bellow 
> you probably want to make sure that clicking the logout link does not reset the last interaction time, stopping propagation of the event from the ng-click of the button would probably fix this.

app/scripts/app.js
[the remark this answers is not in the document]
jwforres: I'm talking about the key string, not the value.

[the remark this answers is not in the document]
jwforres: Treat it like a constant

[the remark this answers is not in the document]
jwforres: As in `var lastInteractionKey = "origin-web-console-last-interaction-timestamp";`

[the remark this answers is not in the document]
spadgett: You should be passing a string as the value to `setItem` not a `Date` object.

jhadvig:
@jwforres comments addressed. 
Added `countdownFromTimestamp` filter as suggested per our conversation. Also added comments to the functions so the code is more understandable. 
I've tested the changes and all the cases should be handled (single and also multi tab). Sorry that it took a bit longer. But there have been more edge cases then I've expected.
cc'ing @spadgett

app/scripts/app.js
```diff
@@ -485,6 +485,110 @@ angular
...
+          logoutAllTabs(true);
+          logout();
+        } else if (event === 'cancel') {
+          // In case of multiple tabs, when the logoutModal would be canceled, the timer has to be restarted.
+          restartCheckInteractionInterval();
+          isLocalLogoutModalShown = false;
+        }
+      }, function () {
+        updateLastInteraction();
+        restartCheckInteractionInterval();
```
jwforres: because you call updateLastInteraction which updates the local storage, that means the event below is going to fire, so restartCheckInteractionInterval gets called twice.  Do you actually need it here?

app/scripts/app.js
```diff
@@ -485,6 +485,110 @@ angular
...
+      } else if (event.originalEvent.key === inactivityLogoutKey) {
+        logout();
+      }
+    });
+
+    // Reset 'origin-web-console-inactivity-logout' key on login.
+    AuthService.onUserChanged(logoutAllTabs(false));
+    updateLastInteraction();
+    startInteractionIntervalCheck();
```
jwforres: same question as above, isnt the storage event going to fire and start the listener?

app/scripts/controllers/modals/logoutModal.js
```diff
@@ -0,0 +1,38 @@
...
+ * # LogoutModalController
+ * Controller of the openshiftConsole
+ */
+angular.module('openshiftConsole')
+  .controller('LogoutModalController', function ($location, $filter, $scope, $uibModalInstance, Constants) {
+    $scope.endTimestamp = moment().add(30, 'seconds');
+    var lastInteraction = Date.parse(localStorage.getItem('origin-web-console-last-interaction-timestamp'));
+    var countdownInterval = setInterval(function() {
+      var timeRemaining = $filter('countdownToTimestamp')($scope.endTimestamp);
+      $('.countdown').text(function() {
```
jwforres: this shouldn't be in here, it should be in the same block in app.js that handles the other date / time counters

app/scripts/controllers/modals/logoutModal.js
```diff
@@ -0,0 +1,38 @@
...
+        return timeRemaining;
+      });
+      if (timeRemaining == 0) {
+        $scope.logout();
+      }
+    }, 1000);
```
jwforres: instead of an interval here you just need a single timeout of 30 seconds that runs the logout, clear the timeout below instead of clear interval

jhadvig: But the interval makes sure that the displayed countdown number will be count down, the same as in https://github.com/openshift/origin-web-console/blob/master/app/scripts/app.js#L491-L505

app/scripts/controllers/modals/logoutModal.js
jhadvig: Not really sure if what would be the benefit of having it in `app.js`. Isn't it better to have the have it separated inside the modal controller as part of the interval, since we have to update the DOM with the actual countdown number every second ?

jhadvig:
@jwforres comments addressed. PTAL

app/scripts/app.js
```diff
@@ -485,6 +485,118 @@ angular
...
+      if (modalInstance) {
+        modalInstance.dismiss("User activity");
+        modalInstance = null;
+      }
+      clearInterval(checkInteractionInterval);
+      startInteractionIntervalCheck();
+    };
+    // Updates lastInteraction date and restarts interval
+    var updateLastInteraction = function() {
+      restartCheckInteractionInterval();
```
jwforres: looks like another case where restartCheckInteractionInteral doesnt need to be called directly since it will get called from the event

app/scripts/app.js
```diff
@@ -485,6 +485,118 @@ angular
...
+          isLocalLogoutModalShown = false;
+        }
+      }, function () {
+        updateLastInteraction();
+        isLocalLogoutModalShown = false;
+      });
+    }
+
+    // Set interval for all timeRemainingFromNow directives.
+    setInterval(function() {
```
jwforres: I don't want this in this block, since this block is not run unless the inactivity timeout feature is enabled.  The directive / filter are independent of that feature. Just take the content of this function and add it to the existing interval that is handling the duration timestamps

jwforres: So I would expect it to be part of this same interval https://github.com/openshift/origin-web-console/blob/master/app/scripts/app.js#L504

app/scripts/app.js
jhadvig: We need to call it here because the listener on the `storage` event is not fired in the tab where the event happened.
Removing `updateLastInteraction()` from the the modal resolving block remove the cases where the `restartCheckInteractionInterval()` is called twice (tested it)

jwforres: Ok good to know

jhadvig:
@jwforres comments addressed. It really makes sense to have countdown in a single interval :+1:

app/scripts/constants.js
```diff
@@ -109,6 +109,11 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     namespace: "openshift"
   },
 
+
+  // Set to number of minutes after which a modal with countdown should appear.
+  // When set to anything else then positive integer the inactivity timeout won't be enable. 
+  INACTIVITY_TIMEOUT_PERIOD: 1,
```
jwforres: Think this accidentally got checked in enabled

jhadvig: ohh... was retesting and accidentally forgot to disable it :-/

jwforres:
Looks good now, can you add the dist files?

jhadvig:
@jwforres dist added

PR #2536, opened 2017-11-28
Add search catalog in project context

jeff-phillips-18:
image

image

image

image

@openshift/team-ux-review 

Bumps version of origin-web-catalog to 0.0.68


jeff-phillips-18:
On a click for the view all filtered items, the following line is forward looking to when we have the browse catalog in project context:

https://github.com/openshift/origin-web-console/pull/2536/files#diff-7a4ed295c8a63008584c40556e2777a2R441

ncameronbritt:
Seems a little odd to me that on desktop the typeahead results are so much wider than the search box. Because of that, what the user types and the results they get back seem disconnected. The visual connection between query and results is much better on the landing page and on mobile, imho. 
screen shot 2017-11-28 at 4 36 46 pm

Could the search box be wider, or the results narrower? If we don't want the search box to be wider, could it make sense to introduce some interaction where the search box expands when the user clicks into it?

I'll defer to @jennyhaines

serenamarie125:
I'm not sure that it's the width that bothers me, but I think if the dropdown was left aligned with the search box, it would be fine.

app/styles/_project-bar.less
```diff
@@ -19,48 +19,57 @@
...
+    margin: 0 5px 0 0;
+    @media (min-width: @screen-sm-min) {
+      margin: 5px 15px 0 5px;
     }
     > a {
       align-items: center;
       color: @navbar-os-project-menu-color;
       display: flex;
-      padding: 5px 0;
       flex: 1 1 auto;
```
rhamilto: This still necessary now that display: flex was removed from parent?

app/styles/_project-bar.less
```diff
@@ -19,48 +19,57 @@
...
       justify-content: center;
+      line-height: 1.9em;
```
rhamilto: To ensure cross-browser consistency, it might be more reliable to set this to 1 and adjust padding?

app/styles/_project-bar.less
```diff
@@ -19,48 +19,57 @@
...
+      padding: 3px 5px;
       &:active,
       &:focus,
       &:hover {
         color: #fff;
         text-decoration: none;
       }
+      @media (min-width: @screen-sm-min) {
+        background-color: @color-pf-black-900;
+        border: solid 1px @navbar-pf-bg-color;
```
rhamilto: Nit:  we typically put the border size first.

app/styles/_project-bar.less
```diff
@@ -19,48 +19,57 @@
...
+        padding: 0px 5px;
```
rhamilto: Nit:  zero is always zero regardless of unit (drop the px).

app/styles/_project-bar.less
```diff
@@ -19,48 +19,57 @@
...
+      }
+      > i {
+        line-height: 1.3em;
```
rhamilto: To ensure cross-browser consistency, it might be more reliable to set this to 1 and adjust padding?

app/styles/_project-bar.less
```diff
@@ -19,48 +19,57 @@
...
       justify-content: center;
+      line-height: 1.9em;
+      padding: 3px 5px;
       &:active,
       &:focus,
       &:hover {
         color: #fff;
         text-decoration: none;
       }
+      @media (min-width: @screen-sm-min) {
```
rhamilto: Nit:  move this rule up to line 36 since you're overriding the parent rule and not appending to it?

app/styles/_project-bar.less
```diff
@@ -197,8 +206,107 @@
...
   }
+
+  .catalog-search {
+    background: transparent;
+    border: 0;
+    margin-left: 15px;
+    padding: 5px 0;
+    position: relative;
+    width: inherit;
+
```
rhamilto: Nit:  empty line is inconsistent

app/styles/_project-bar.less
```diff
@@ -197,8 +206,107 @@
...
+    @media (min-width: @screen-sm-min) {
+      position: inherit;
+    }
+
```
rhamilto: Nit:  empty line is inconsistent

app/styles/_project-bar.less
```diff
@@ -197,8 +206,107 @@
...
+    @media (max-width: @screen-xs-max) {
+      margin-left: 0;
+      padding: 2px 5px;
+      border-right: solid 1px @navbar-pf-bg-color;
```
rhamilto: Nits:  alpha and we typically put the size first

app/styles/_project-bar.less
```diff
@@ -197,8 +206,107 @@
...
+      &:focus,
+      &:hover {
+        color: #fff;
+        text-decoration: none;
+      }
+    }
+    .dropdown-menu {
+      @media (min-width: @screen-sm-min) {
+        right: 0;
+        left: auto !important;
```
rhamilto: Nit:  alpha

app/styles/_project-bar.less
```diff
@@ -197,8 +206,107 @@
...
+        width: 550px;
+      }
+    }
+    .landing-search-form {
+      width: 200px;
+      .search-pf-input-group{
```
rhamilto: Nit:  space needed between p and {

app/styles/_project-bar.less
```diff
@@ -197,8 +206,107 @@
...
+        }
+        .form-control {
+          background-color: @color-pf-black-900;
+          border-color: @navbar-pf-bg-color;
+          color: @navbar-os-project-menu-color;
+          font-size: @font-size-base;
+          height: 2em;
+          &:focus {
+            border-color: @color-pf-black-700;
+            outline: thin dotted !important;
```
rhamilto: Use [.tab-focus()](https://github.com/openshift/origin-web-console/blob/master/app/styles/_mixins.less#L37)

rhamilto:
> Seems a little odd to me that on desktop the typeahead results are so much wider than the search box. Because of that, what the user types and the results they get back seem disconnected. The visual connection between query and results is much better on the landing page and on mobile, imho.

Since we have the real estate, any reason the search input doesn't get wider?

app/scripts/directives/nav.js
```diff
@@ -201,6 +207,8 @@ angular.module('openshiftConsole')
       link: function($scope, $elem) {
         var MAX_PROJETS_TO_DISPLAY = 100;
         var NAV_COLLAPSED_STORAGE_KEY = 'openshift/vertical-nav-collapsed';
+        var plansByServiceClassName = [];
```
spadgett: Shouldn't this be a map?

```js
var plansByServiceClassName = {};
```

app/scripts/directives/nav.js
```diff
@@ -342,11 +355,98 @@ angular.module('openshiftConsole')
               $scope.currentProject = projects[$scope.currentProjectName];
               updateOptions();
             });
+
+            var includeTemplates = !CatalogService.isTemplateServiceBrokerEnabled();
+            Catalog.getCatalogItems(includeTemplates).then(_.spread(function(items, errorMessage) {
```
spadgett: This is potentially a lot of data, so let's be careful about loading it when not needed.

*  We shouldn't reload all of this data when switching projects since it's cluster scoped and doesn't change often. It looks like we request it again on project changes.
* We should only load the catalog items when the user has authority to add to a project.
* I suggest we load the data once for the console and cache it in a service. Then it can be used for the catalog and nav search and anywhere else we need it.

app/scripts/directives/nav.js
```diff
@@ -342,11 +355,98 @@ angular.module('openshiftConsole')
...
+              });
+              plansByServiceClassName = Catalog.groupPlansByServiceClassName(plans);
+            });
+
           } else {
             _.set($rootScope, 'view.hasProject', false);
           }
         };
 
+        var addTemplateToRecentlyViewed = function() {
```
spadgett: Is this always a template?

app/views/directives/header/header.html
```diff
@@ -27,17 +27,20 @@
     <!-- Use title="" to prevent "Nothing selected" from briefly flashing while the projects are still loading. -->
     <select class="selectpicker form-control" data-selected-text-format="count>3" id="boostrapSelect" title=""></select>
   </div>
+  <catalog-search catalog-items="catalogItems" base-project-url="project" toggle-at-mobile="true" search-toggle-callback="onSearchToggle"></catalog-search>
```
spadgett: Only visible if I can add to project.

```html
ng-if="canIAddToProject"
```

app/scripts/directives/nav.js
```diff
@@ -242,11 +250,16 @@ angular.module('openshiftConsole')
...
         };
 
         $scope.showOrderingPanel = function(panelName) {
-          _.set($scope, 'ordering.panelName', panelName);
+          $scope.orderingPanelVisible = true;
+          $scope.orderKind = panelName;
+        };
+
+        $scope.onSearchToggle = function(showMobileSearch) {
+          _.set($rootScope, 'view.hasProjectSearch', showMobileSearch);
```
spadgett: Just curious what happens if I navigate to a different page with the search open, either at mobile or desktop. For instance, if I click the back button. (Not sure what the right behavior is.)

spadgett:
Not necessarily for this PR, but we might look at adding project templates and builders to the search result if in the context of a project.

spadgett:
> Since we have the real estate, any reason the search input doesn't get wider?

Throwing this out as a possibility: I've seen some search inputs grow larger when you focus them and start typing. That way it wouldn't dominate the project bar when you're not using it.

app/styles/_project-bar.less
jeff-phillips-18: Yes, it is still necessary for alignment and space inside borders.

app/scripts/directives/nav.js
jeff-phillips-18: When you come back, the search is still open.

jeff-phillips-18:
Addressed @rhamilto issues. Set the search input and dropdown results menu width to 300px at desktop resolutions.

serenamarie125:
@jeff-phillips-18 can you update screenshot to what it looks like with 300px menu width?

jeff-phillips-18:
@serenamarie125 Updated in the opening comment above to keep PR accurate.

app/scripts/directives/nav.js
jeff-phillips-18: Should we do the same for service plans? We would have to watch those though as they can change and get filtered by the 'removedFromBrokerCatalog' field.

jeff-phillips-18: We allow search when the user is not allowed to add. This gives the ability to at least see what catalog items are out there and get their info.

We can remove search in project context if we think this is not necessary.

spadgett: > Should we do the same for service plans?

Yes. Plans contain the JSON schema describing the parameters, so the response could get large if you have hundreds or thousands. QE has seen the request take up to 10s in their environment.

https://bugzilla.redhat.com/show_bug.cgi?id=1512858

> We would have to watch those though as they can change and get filtered by the 'removedFromBrokerCatalog' field.

I don't think it's worth watching since this happens infrequently. (I'm not sure offhand if watch is supported for plans.)

It might make sense to lazy load plans for just the service class when opening the dialog instead of loading all of them up front. (Not necessarily for this PR, though.)

spadgett: > We allow search when the user is not allowed to add. This gives the ability to at least see what catalog items are out there and get their info.

That seems a little odd to me. We do have the landing page if the user wants to browse / search.

jeff-phillips-18: Added `CatalogService.getCatalogItems` which caches the catalog items

>It might make sense to lazy load plans for just the service class when opening the dialog instead of loading all of them up front. (Not necessarily for this PR, though.)

Updated to only get the plans for the class when opening the dialog

>That seems a little odd to me. We do have the landing page if the user wants to browse / search.

Added the check

app/scripts/directives/nav.js
jeff-phillips-18: no, renamed.

app/views/directives/header/header.html
jeff-phillips-18: done.

app/scripts/services/catalog.js
```diff
@@ -21,10 +26,41 @@ angular.module("openshiftConsole")
...
+    };
+
+    AuthService.onUserChanged(clearCachedCatalogItems);
+    AuthService.onLogout(clearCachedCatalogItems);
+
     var isTemplateServiceBrokerEnabled = function() {
       return !!Constants.TEMPLATE_SERVICE_BROKER_ENABLED;
     };
 
+    var getCatalogItems = function(forceRefresh) {
```
spadgett: Looks good. Do you mind updating landing page to use this?

jeff-phillips-18: sure thing

app/views/directives/header/header.html
```diff
@@ -27,17 +27,20 @@
...
+      <span class="hidden-xs">Add to Project</span>
+      <span class="hidden-xs caret" aria-hidden="true" title="Add to Project"></span>
+    </button>
     <ul role="menu" uib-dropdown-menu class="dropdown-menu dropdown-menu-right">
       <li ng-if-start="!catalogLandingPageEnabled" role="menuitem"><a ng-href="project/{{currentProjectName}}/create?tab=fromCatalog">Browse Catalog</a></li>
       <li role="menuitem"><a ng-href="project/{{currentProjectName}}/create?tab=deployImage">Deploy Image</a></li>
       <li ng-if-end role="menuitem"><a ng-href="project/{{currentProjectName}}/create?tab=fromFile">Import YAML / JSON</a></li>
 
-      <li ng-if-start="catalogLandingPageEnabled" role="menuitem"><a href="./">Browse Catalog</a></li>
+      <li ng-if-start="catalogLandingPageEnabled" role="menuitem"><a href="project/{{currentProjectName}}/project-browse-catalog">Browse Catalog</a></li>
```
spadgett: We should wait on this until the in-project browse works, correct?

jeff-phillips-18: yes. I will remove this change.

jeff-phillips-18:
Bumped version of origin-web-catalog to 0.0.68

Removed WIP tag.

spadgett:
@jeff-phillips-18 can you rebase and rebuild dist?

jeff-phillips-18:
Rebased and rebuilt.

jennyhaines:
@jeff-phillips-18 - Thank you for the change. I think the updated design looks great!

PR #2537, opened 2017-11-29
Localization the page of dev-console/project/app/overview

fenfenraorao:


app/config.js
```diff
@@ -7,7 +7,7 @@
   //
   // To change configuration for local development, copy this file to
   // assets/app/config.local.js and edit the copy.
-  var masterPublicHostname = 'localhost:8443';
+  var masterPublicHostname = '192.168.1.151:8443';
```
xiaods: it should be localhost, don't not modify the variable.  masterPublicHostname

app/config.js
```diff
@@ -34,7 +34,7 @@
     loggingURL: "",
     metricsURL: ""
   };
-
+  window.OPENSHIFT_LANG = "zh-CN";
```
xiaods: need comments on it. the OPENSHIFT_LANG is global var, so it default value is en, it not zh-CN default.

Issue #2543, opened 2017-11-29
overview .actions-dropdown-kebab size has changed


sg00dwin:
<img width="927" alt="list-pf-actions" src="https://user-images.githubusercontent.com/1874151/33398937-b5f141d0-d51e-11e7-803b-982f18465d0e.png">

PR #2544, opened 2017-11-29
Update to make actions-dropdown-kebab larger

sg00dwin:
Fix https://github.com/openshift/origin-web-console/issues/2543

### DIFF
--- a/app/styles/_overview.less
+++ b/app/styles/_overview.less
@@ -358,7 +358,7 @@
       color: #4d5258;
     }
   }
-  .list-view-pf-actions .actions-dropdown-kebab {
+  .list-pf-actions .actions-dropdown-kebab {
     font-size: 18px;
     // Give a bigger click target (fill `list-view-pf-actions` width)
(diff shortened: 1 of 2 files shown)


app/styles/_overview.less
```diff
@@ -358,7 +358,7 @@
       color: #4d5258;
     }
   }
-  .list-view-pf-actions .actions-dropdown-kebab {
+  .list-pf-actions .actions-dropdown-kebab {
```
spadgett: There's a `.list-pf-actions` rule above we should combine this with.

I'm a little confused why this wasn't a problem in 3.7, though.

spadgett:
Looks like this change is what caused the problem:

https://github.com/openshift/origin-web-console/pull/2472/files#diff-c6dc5c38b82edaf574db6b9dbd4fa375L6

I'm unclear on why the markup structure is different for project list vs overview. Probably this should be a general style so that all of our kebabs match.

PR #2546, opened 2017-11-30
Enable valueFrom values for build config env vars

rhamilto:
https://trello.com/c/DgVzSFtw/994-2-buildconfigs-env-valuefrom-build

browse/builds/:bc
screen shot 2017-11-30 at 12 41 40 pm

edit/builds/:bc
screen shot 2017-11-30 at 12 41 58 pm

create/fromimage
screen shot 2017-11-30 at 12 43 05 pm

### DIFF
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -19,6 +19,7 @@ angular.module('openshiftConsole')
                        ModalsService,
                        NotificationsService,
                        ProjectsService,
+                       SecretsService,
                        keyValueEditorUtils) {
     $scope.projectName = $routeParams.project;
     $scope.buildConfigName = $routeParams.buildconfig;
@@ -59,7 +60,12 @@ angular.module('openshiftConsole')
 
     var buildConfigForBuild = $filter('buildConfigForBuild');
     var buildStrategy = $filter('buildStrategy');
+    var orderByDisplayName = $filter('orderByDisplayName');
+    var getErrorDetails = $filter('getErrorDetails');
     var watches = [];
+    var configMapDataOrdered = [];
+    var secretDataOrdered = [];
+    $scope.valueFromObjects = [];
 
     // copy buildConfig and ensure it has env so that we can edit env vars using key-value-editor
     var copyBuildConfigAndEnsureEnv = function(buildConfig) {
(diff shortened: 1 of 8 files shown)


app/scripts/controllers/buildConfig.js
```diff
@@ -19,7 +19,8 @@ angular.module('openshiftConsole')
                        ModalsService,
                        NotificationsService,
                        ProjectsService,
-                       keyValueEditorUtils) {
+                       keyValueEditorUtils,
+                       SecretsService) {
```
spadgett: Nit: we try to keep the dependencies alphabetized.

app/scripts/controllers/buildConfig.js
```diff
@@ -197,6 +203,44 @@ angular.module('openshiftConsole')
...
+            id: "create-builder-list-config-maps-error",
+            type: "error",
+            message: "Could not load config maps.",
+            details: getErrorDetails(e)
+          });
+        });
+
+        DataService.list("secrets", context, null, { errorNotification: false }).then(function(secretData) {
+          secretDataOrdered = orderByDisplayName(secretData.by("metadata.name"));
+          $scope.valueFromObjects = secretDataOrdered.concat(configMapDataOrdered);
```
spadgett: Let's try to keep a consistent ordering as above so we get the same result regardless of which loads first. (It will show the select groupings in different orders depending on how we concat the arrays.)

```js
$scope.valueFromObjects = configMapDataOrdered.concat(secretDataOrdered);
```

app/scripts/controllers/buildConfig.js
```diff
@@ -197,6 +203,44 @@ angular.module('openshiftConsole')
...
+          var secretsByType = SecretsService.groupSecretsByType(secretData);
+          var secretNamesByType =_.mapValues(secretsByType, function(secretData) {return _.map(secretData, 'metadata.name');});
+          // Add empty option to the image/source secrets
+          $scope.secretsByType = _.each(secretNamesByType, function(secretsArray) {
+            secretsArray.unshift("");
+          });
```
spadgett: I think lines 225-230 can be deleted. I don't believe we do anything with `secretsByType` on this page.

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -372,6 +377,44 @@ angular.module('openshiftConsole')
...
+            id: "create-builder-list-config-maps-error",
+            type: "error",
+            message: "Could not load config maps.",
+            details: getErrorDetails(e)
+          });
+        });
+
+        DataService.list("secrets", context, null, { errorNotification: false }).then(function(secretData) {
+          secretDataOrdered = orderByDisplayName(secretData.by("metadata.name"));
+          $scope.valueFromObjects = secretDataOrdered.concat(configMapDataOrdered);
```
```js
$scope.valueFromObjects = configMapDataOrdered.concat(secretDataOrdered);
```

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -372,6 +377,44 @@ angular.module('openshiftConsole')
...
+
+          NotificationsService.addNotification({
+            id: "create-builder-list-config-maps-error",
+            type: "error",
+            message: "Could not load config maps.",
+            details: getErrorDetails(e)
+          });
+        });
+
+        DataService.list("secrets", context, null, { errorNotification: false }).then(function(secretData) {
```
spadgett: We list secrets already elsewhere in the controller. We should only request them once.

app/scripts/controllers/edit/buildConfig.js
rhamilto: @spadgett, I gather https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/create/createFromImage.js#L169 should be updated, then, too?

rhamilto:
Comments addressed.

app/scripts/controllers/edit/buildConfig.js
spadgett: Yes, looks like it.

jorgemoralespou:
@rhamilto Is this merged?
I just tried "origin latest" with "oc cluster up" and this is not showing. I guess it can be because of the vendoring.

What I'm missing from your PR is the whole "envFrom" like in Deployments:

image

Is this implemented?

rhamilto:
Unfortunately, no, @jorgemoralespou.  This PR enables Value From values, not Environment From.  The similarity in name and function makes this confusing.  :(

spadgett:
@bparees Is `envFrom` supported for build config?

bparees:
@spadgett only ValueFrom:
https://docs.openshift.org/latest/dev_guide/builds/build_environment.html#using-build-fields-as-environment-variables

we have a card for envfrom:
https://trello.com/c/kemg2ILC/1258-3-support-envfrom-in-builds-techdebt

PR #2548, opened 2017-11-30
Update directives/oscPersistentVolumeClaim to use getPreferredVersion

benjaminapetersen:
Depends on https://github.com/openshift/origin-web-common/pull/253


benjaminapetersen:
`storageclasses` is in common, this is good.

PR #2549, opened 2017-11-30
Fix Bug 1518407 - A-MQ Components not visible in Java Console when using IE11 browser

akieling:
https://bugzilla.redhat.com/show_bug.cgi?id=1519573


dist.java/java/index.html
```diff
@@ -3,7 +3,7 @@
 
   <head>
     <title>openshift-jvm</title>
-    <base href="/">
+    <base href="/dev-console/java/">
```
spadgett: You need to reset this to `/` when committing. It will be set correctly by the server when index.html is served using the real context root.

spadgett:
Thanks, do you mind squashing to one commit?

PR #2552, opened 2017-12-01
Update directives/processTemplate to use getPreferredVersion

benjaminapetersen:
Depends on common https://github.com/openshift/origin-web-common/pull/256

### DIFF
--- a/app/scripts/directives/processTemplate.js
+++ b/app/scripts/directives/processTemplate.js
@@ -7,6 +7,7 @@
       '$q',
       '$scope',
       '$uibModal',
+      'APIService',
       'DataService',
       'Navigate',
       'NotificationsService',
@@ -34,6 +35,7 @@
                           $q,
                           $scope,
                           $uibModal,
+                          APIService,
                           DataService,
                           Navigate,
                           NotificationsService,
@@ -217,6 +219,12 @@
       return ProjectsService.create(newProjName, newProjDisplayName, newProjDesc);
     };
 
+    var getProcessTemplateVersionForTemplate = function(template) {
+      var rgv = APIService.objectToResourceGroupVersion(template);
+      rgv.resource = 'processedtemplates';
+      return rgv;
+    };
+
     ctrl.createFromTemplate = function() {
       ctrl.disableInputs = true;
       createProjectIfNecessary().then(function(project) {
@@ -226,7 +234,9 @@
         };
         ctrl.template.labels = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries(ctrl.labels));
 
-        DataService.create("processedtemplates", null, ctrl.template, context).then(
+        var rgv = getProcessTemplateVersionForTemplate(ctrl.template);
+
+        DataService.create(rgv, null, ctrl.template, context).then(
           function(config) { // success
             // Cache template parameters and message so they can be displayed in the nexSteps page
             ProcessedTemplateService.setTemplateData(config.parameters, ctrl.template.parameters, config.message);
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -5,7 +5,7 @@ var N = this, D = t("isIE")();
 e.projectName = a.project;
 var A = a.isHomePage;
 N.catalogLandingPageEnabled = !d.DISABLE_SERVICE_CATALOG_LANDING_PAGE;
-var B = t("annotation"), L = t("canI"), U = t("buildConfigForBuild"), V = t("deploymentIsInProgress"), O = t("imageObjectRef"), F = t("isJenkinsPipelineStrategy"), x = t("isNewerResource"), M = t("label"), q = t("podTemplate"), z = i.getPreferredVersion("deployments"), H = i.getPreferredVersion("horizontalpodautoscalers"), G = i.getPreferredVersion("servicebindings"), K = i.getPreferredVersion("clusterserviceclasses"), W = i.getPreferredVersion("serviceinstances"), Q = i.getPreferredVersion("clusterserviceplans"), J = i.getPreferredVersion("statefulsets"), Y = i.getPreferredVersion("replicasets");
+var B = t("annotation"), L = t("canI"), V = t("buildConfigForBuild"), U = t("deploymentIsInProgress"), O = t("imageObjectRef"), F = t("isJenkinsPipelineStrategy"), x = t("isNewerResource"), M = t("label"), q = t("podTemplate"), z = i.getPreferredVersion("deployments"), H = i.getPreferredVersion("horizontalpodautoscalers"), G = i.getPreferredVersion("servicebindings"), K = i.getPreferredVersion("clusterserviceclasses"), W = i.getPreferredVersion("serviceinstances"), Q = i.getPreferredVersion("clusterserviceplans"), J = i.getPreferredVersion("statefulsets"), Y = i.getPreferredVersion("replicasets");
 N.buildConfigsInstantiateVersion = i.getPreferredVersion("buildconfigs/instantiate");
 var Z, X, ee = {}, te = {}, ne = {}, re = N.state = {
 alerts: {},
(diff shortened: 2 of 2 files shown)


benjaminapetersen:
WIP removed/rebased

spadgett:
flake #1685

/retest

benjaminapetersen:
Errors with e2e tests:

```
**************************************************
*                    Failures                    *
**************************************************

1) User adds a template to a project after creating a new project using the Import YAML tab should process and create the objects in the template
  - Expected false to be true.
      at catalogPage.processTemplate.then (/data/src/github.com/openshift/origin-web-console/test/integration/features/user_adds_template_to_project.spec.js:40:80)
    at /data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/goog/base.js:1582:15
    at webdriver.promise.ControlFlow.runInNewFrame_ (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:1654:20)
    at notify (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:465:12)
    at notifyAll (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:442:7)
    at resolve (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:420:7)
    at fulfill (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:535:5)
    at /data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/goog/base.js:1582:15
    at webdriver.promise.ControlFlow.runInNewFrame_ (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:1654:20)
  - Expected false to be true.
      at catalogPage.processTemplate.then (/data/src/github.com/openshift/origin-web-console/test/integration/features/user_adds_template_to_project.spec.js:41:95)
    at /data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/goog/base.js:1582:15
    at webdriver.promise.ControlFlow.runInNewFrame_ (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:1654:20)
    at notify (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:465:12)
    at notifyAll (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:442:7)
    at resolve (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:420:7)
    at fulfill (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:535:5)
    at /data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/goog/base.js:1582:15
    at webdriver.promise.ControlFlow.runInNewFrame_ (/data/src/github.com/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:1654:20)
```

app/scripts/directives/processTemplate.js
```diff
@@ -225,7 +229,7 @@
         };
         ctrl.template.labels = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries(ctrl.labels));
 
-        DataService.create("processedtemplates", null, ctrl.template, context).then(
+        DataService.create(processedTemplatesVersion, null, ctrl.template, context).then(
```
benjaminapetersen: Not sure if this is correct, actually.  Should probably extract version from `ctrl.template` to ensure they match.

@spadgett ?

benjaminapetersen: Hmm. maybe is fine.  difference would be:
`{ group: "template.openshift.io", resource:"processedtemplates", version:"v1" }`
vs
`{ group: "template.openshift.io", resource:"templates", version:"v1" }`
if I switched to `APIService.objectToResourceGroupVersion(ctrl.template)`.  That doesn't seem correct.

benjaminapetersen: Using `preferredVersion` (first in above comment) causes:

screen shot 2018-01-04 at 12 23 45 pm

Using `templateRGV` (second in above) causes:

screen shot 2018-01-04 at 12 23 09 pm

Kind of interesting, update to either fails.  Not sure why `namespace` is brought in, this shouldn't alter the `context` object, diagnosing.

spadgett: Hm, it looks need to use the API version from the template itself. I wonder if we have a bug in 3.7 if the template is declared using the API group.

spadgett: Resource should be `processedtemplates`, otherwise you're creating a template, not processing one.

benjaminapetersen: Think its a bug. If I used the `preferredVersion` but clear the `group`:

```
{
  group: "",  // drop the 'template.openshift.io'
  resource: "processedtemplates",
  version: "v1"
}
```
Works fine.

benjaminapetersen: Take that back, some do, some don't.  `cakephp` won't work.  `nodejs + mongo` works.  So depends on the template.

benjaminapetersen: Running `oc new-app -f origin/examples/quickstarts/cakephp-mysql.json  --loglevel=8` gives me a little insight.  POST to `https://127.0.0.1:8443/apis/template.openshift.io/v1/namespaces/baz/processedtemplates`, and apparently it updates the request body itself from `apiVersion: v1` to `"apiVersion":"template.openshift.io/v1"`.  

I'm not sure we want to be meddling with the templates directly in the console tho?

app/scripts/directives/processTemplate.js
```diff
@@ -226,7 +230,13 @@
         };
         ctrl.template.labels = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries(ctrl.labels));
 
-        DataService.create("processedtemplates", null, ctrl.template, context).then(
+        // NOTE:
```
benjaminapetersen: This is worth some discussion, I'm guessing there may be some potential problems with this approach, however, it fixes the mismatch problem and appears to be what the CLI is doing.  However:

```JavaScript 
// given 
preferredVersion = {group: 'template.openshift.io',      version: 'v1',      resource: 'processedtemplates' };
// if template is:
apiVersion = 'v1';                                      // update it to template.openshift.io/v1 
apiVersion = 'template.openshift.io/v1'  // updating, but its the same, so no foul 
apiVersion = 'template.openshift.io/v2' //  hmm... our preferredVersion is v1 so
apiVersion = 'template.something.else/v5' // what to do with unexpected? should there ever be?
```
So this fixes the problem at the moment, is it the correct fix long term?

app/scripts/directives/processTemplate.js
```diff
@@ -226,7 +228,10 @@
         };
         ctrl.template.labels = keyValueEditorUtils.mapEntries(keyValueEditorUtils.compactEntries(ctrl.labels));
 
-        DataService.create("processedtemplates", null, ctrl.template, context).then(
+        var tplGroupVersion = APIService.objectToResourceGroupVersion(ctrl.template);
```
benjaminapetersen: This reflects the latest small change we discussed.

@sg00dwin this branch updated, you might want to pull again if you are still using it (but I imagine it works for what you are doing either way).

benjaminapetersen:
Related, I tested the import yaml dialog with [this template](https://github.com/openshift-qe/v3-testfiles/blob/f7361e254d5f24f547d639a6bca6185ce4b6f826/templates/test-api.yaml) to see if it complains.  

First pass: `apiVersion: v1` works.
Second pass, manually updated to `apiVersion: template.openshift.io/v1`, works.

This template creates a number of resources each with different API versions.

app/scripts/directives/processTemplate.js
benjaminapetersen: @spadgett I think this covers the small change from the last discussion we had.

spadgett: nit: `tpl` = template? I'd rather just call it `rgv` or `resourceGroupVersion`

spadgett: We should add a comment here explaining that the processedtemplates version must match the version of the template in the request since it's not obvioius why we're doing this.

benjaminapetersen: Good with a named function?

```JavaScript  
  //
  var rgv = getProcessTemplateVersionForTemplate(ctrl.template);

  DataService.create(rgv, null, ctrl.template, context).then(
  //  
```

spadgett:
LGTM but you have a dist mismatch

benjaminapetersen:
Ok cool, wanted to make sure you were ok with the named fn.  I'll build the dist.

benjaminapetersen:
builds good. ready to roll.

benjaminapetersen:
@spadgett good for the official LGTM?

PR #2554, opened 2017-12-01
Fix storage nav highlighting

spadgett:
* Highlight the storage vertical nav item as selected when looking at a PVC
* Change breadcrumb to "Storage" to be consistent with the PVC list and
   nav label

Before:

image 2017-12-01 15-18-53

After:

image 2017-12-01 15-19-58

/kind bug
/assign @jwforres 
/area usability
/cc @ncameronbritt


ncameronbritt:
lgtm

PR #2561, opened 2017-12-04
Localization the page of dev-console/project/app/overview

fenfenraorao:


spadgett:
/ok-to-test

Hi @fenfenraorao does this replace #2537?

app/config.js
```diff
@@ -34,7 +34,8 @@
     loggingURL: "",
     metricsURL: ""
   };
-
+  // Additional support only for Chinese now: zh-CN
+  window.OPENSHIFT_LANG = "en";
```
spadgett: Should this be based on the user's language settings in the browser instead of a server setting? It should be available in `navigator.languages` in most browsers.

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages

app/scripts/constants.js
```diff
@@ -5,6 +5,9 @@
 // Reference: https://docs.openshift.org/latest/install_config/web_console_customization.html#loading-custom-scripts-and-stylesheets
 
 // NOTE: Update extensions/examples/online-extensions.js if you add a new help link to this map.
+var gettext = function(s) {
```
jwforres: I assume this is here so the gettext compiler finds theses strings since constants.js lives outside of the angular scope.  Can you add comments here explaining that so that someone doesn't try to remove it later thinking its just a no-op

app/scripts/app.js
```diff
@@ -506,6 +506,14 @@ angular
       // content (e.g. using :before pseudo-elements).
       $('body').addClass('ios');
     }
+  })
+  .run(function($window, gettextCatalog){
+    gettextCatalog.debug = true;
```
jwforres: we can't have debug mode on by default, this needs to be triggered off something else like a window constant or a local storage value.  To me this is similar to our UI logging, which is turned on by changing local storage values, see: https://github.com/openshift/origin-web-console#enable--disable-console-log-output

app/config.js
```diff
@@ -34,7 +34,8 @@
     loggingURL: "",
     metricsURL: ""
   };
-
+  // Additional support only for Chinese now: zh-CN
```
jwforres: I wouldn't use the word support here since this is still going to be a work in progress, maybe say something like:
Internationalization is currently under development, the default supported language is English

app/config.js
jwforres: To add to @spadgett 's point, config.js in the real console server is generated at startup and then cached in the server's memory so we don't have to re-evaluate the internal template every time someone hits it, this means we can't include content in this file that changes per user.

app/views/deployments.html
```diff
@@ -28,8 +28,8 @@ <h1>
           <div ng-if="(showEmptyState || filterWithZeroResults) && deploymentConfigsLoaded">
             <div class="empty-state-message text-center">
               <div ng-if="showEmptyState">
-                <h2>No deployments.</h2>
-                <p>No deployments have been added to project {{projectName}}.</p>
+                <h2 translate>No deployments.</h2>
```
benjaminapetersen: Question re: consistency / moving forward to `components` over `directives`, would we want it to always be an element `<translate>` or are we ok with an attrib `<p translate>`?  Do we then switch these to `<p><translate>`?

app/views/overview.html
```diff
@@ -6,16 +6,15 @@
       <!-- Getting started -->
       <div class="empty-state-message text-center">
         <div ng-if="project.metadata.name | canIAddToProject">
-          <h2>Get started with your project.</h2>
-          <p class="gutter-top">
-            Use your source or an example repository to build an application
-            image, or add components like databases and message queues.
+          <h2 translate>Get started with your project.</h2>
+          <p class="gutter-top" translate>
+            Use your source or an example repository to build an application image, or add components like databases and message queues.
```
benjaminapetersen: I'm curious if there is an automated test for validating we don't break any translation strings. I could see us accidentally changing a word in a sentence like this & finding that the translation fails.

app/views/deployments.html
benjaminapetersen: This comment relates to a number of elements in the PR.

spadgett: Well it's from angular-gettext, so it will be directive. That's how they implemented it.

I'd lean towards using it as an attribute to avoid a bunch of extra elements in our markup, but it can be used as an element.

https://angular-gettext.rocketeer.be/dev-guide/annotate/

I'm a little concerned that angular-gettext doesn't support Angular 2. If we update Angular, we'd need to find a different solution and rework all of the translations (in addition to everything else).

benjaminapetersen: I tend to prefer it as a filter, like [angular-translate](https://angular-translate.github.io/docs/#/guide) handles it.  It seems more like the kind of thing a filter would do, and that avoids the issue with nodes.

That said, there is a lot of praise for `angular-gettext` being easier to work with.

app/views/overview.html
jwforres: long term the assumption has to be that mid-release there is no guarantee that the translations are correct. Then every 3.x release we would have a translation phase. But we cant even get to that point until we have everything going through gettext, and then we have to figure out where in the release process the translation phase would live, because it means adding a "string lockdown" date

app/config.js
xiaods: @jwboyer Agree.

app/views/overview.html
xiaods: @jwforres yes, we are came from china, we have always deploy many more instance for users, but user prefer native language interface, so we need quick build the translation process workflow,  let the translation cycle is working on process. thanks a lot.

app/scripts/app.js
fenfenraorao: This is the debug of the angular-gettext itself,which is not very good with openshiftLogLevel. How do you feel about adding global variables to config.js like OPENSHIFT_LANGUAGE?such as   
window.GettextCatalogDebug = true

app/views/deployments.html
fenfenraorao: @spadgett Angular-gettext dose not support Angular2. I haven't found a way to compatible Angular1 and Angular2 for the time being.If you upgrade to Angular2,we need to find other ways of translating.

xiaods:
need resolve conflict. @fenfenraorao

xiaods:
close by  https://github.com/openshift/origin-web-console/pull/2588

PR #2562, opened 2017-12-04
[WIP] Set Home Page

dtaylor113:
Requires: https://github.com/openshift/origin-web-common/pull/260
Requires: https://github.com/openshift/origin-web-catalog/pull/575

Trello Card: https://trello.com/c/sg1xw7gG

**Adds 'Set Home Page' option to User dropdown:**
image

**Set Home Page Modal: Multiple Projects**
image

**Set Home Page Modal: One Project**
image

**Set Home Page Modal: No Projects**
image

**Notification When Home Page project No Longer Exists**
image
_* User is brought to projects list._


ncameronbritt:
@spadgett had an argument for defaulting to the landing page rather than the projects list page if the project had been deleted. Basically it was that for users on clusters with lots of projects, the shorter list of projects may be of more relevant to the user than the full list. I'm good with either. Interested in what Sam and @jwforres think.

For the radio buttons when there's only one project, it bothers me that the project select dropdown is disabled. I feel like that sends mixed messages--the radio button is active but the dropdown isn't. What if when there's only one project the third radio button just lists that project, with no dropdown? 
```
o Project Overview for My Project
```
Alternatively, I think having the dropdown active, even with one option, is preferable to the disabled dropdown.
cc @openshift/team-ux-review

dtaylor113:
> .... it bothers me that the project select dropdown is disabled [when only one project].

Understood, however just FYI that is the standard behavior with the `<select-project>` component, which is also used in a few other wizards and modals.  When there is only one choice we default to it and disable the dropdown.  But I understand this may not be ideal when used in conjunction with radio buttons.

dtaylor113:
@ncameronbritt when there is only one project, how should we display it without the dropdown:

    (  ) Overview Page for:
         My Second Project - my-second-project

....or something inline, ex:

    (  ) Overview Page for 'My Second Project' (my-second-project)

Looking for direction on how to display the project Display name and the project-name.  -thanks

ncameronbritt:
My preference is for inline. For the free tier of Online and probably other scenarios too, it will always be the case that as user only has access to a single project, so imho having it inline makes it appear more deliberate, as opposed to being an edge case we're designing around. 

Happy to hear other opinions though.

serenamarie125:
Agreed on the inline suggestion

app/scripts/controllers/util/error.js
```diff
@@ -8,16 +8,18 @@
...
 angular.module('openshiftConsole')
-  .controller('ErrorController', function ($scope, $window) {
+  .controller('ErrorController', function ($scope, $location, $window, $uibModal, HomePagePreference, NotificationsService) {
     var params = URI(window.location.href).query(true);
     var error = params.error;
 
     switch(error) {
       case 'access_denied':
         $scope.errorMessage = "Access denied";
+        checkHomePagePreference(params.project_name);
```
benjaminapetersen: This would typically throw an error w/our `jshint` conf, using a function before it is declared.  Do we still enforce that rule?

app/scripts/controllers/util/error.js
```diff
@@ -40,4 +42,29 @@ angular.module('openshiftConsole')
     $scope.reloadConsole = function() {
       $window.location.href = "/";
     };
+
+    function checkHomePagePreference(projectName) {
```
benjaminapetersen: Curious if this function actually belongs within `HomePagePreference` rather than in `ErrorController`, even if it gets called here.  Should this module grow into a collection of error functions related to other modules?

app/scripts/controllers/modals/setHomePageModal.js
```diff
@@ -0,0 +1,74 @@
...
+    };
+
+    $scope.$on('no-projects-cannot-create', function() {
+      $scope.noProjects = true;
+    });
+
+    $scope.disableSaveButton = function() {
+      // if radiobuttons not changed
+      if ($scope.preferedHomePage === $scope.initialPreferedHomePage) {
+        if ($scope.preferedHomePage === 'project-overview') {
```
benjaminapetersen: This `if` block is duplicated below on line 52, and a lot of nested `if` statements.  Care to dedupe w/function?  Can also prob eliminate the `// comments` with named fn calls.

dist.java/java/404.html
```diff
@@ -38,10 +38,10 @@ <h4>Connected to <span ng-bind="containerName"></span></h4>
       <div ng-include src="viewPartial"></div>
     </div>
```
benjaminapetersen: Guessing you didn't change `dist.java`.

spadgett:
@dtaylor113 You have `dist.java` changes that shouldn't be there. I'd try

```
$ rm -rf dist.java
$ hack/install-assets
$ grunt build
```

app/scripts/controllers/modals/setHomePageModal.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+/* jshint unused: false */
```
spadgett: Can we fix whatever is causing the warning instead of ignoring it?

app/scripts/controllers/modals/setHomePageModal.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * @ngdoc function
+ * @name openshiftConsole.controller:SetHomePageModalController
+ * @description
+ * # SetHomePageModalController
+ * Controller of the openshiftConsole
+ */
```
spadgett: I don't think this comment is helpful. I understand a lot of our files have it, but I'd just take it out.

app/scripts/controllers/modals/setHomePageModal.js
```diff
@@ -0,0 +1,74 @@
...
+angular.module('openshiftConsole')
+  .controller('SetHomePageModalController', function ($scope, $uibModalInstance, HomePagePreference) {
+
+    $scope.initialPreferedHomePage = HomePagePreference.getHomePagePreference();
```
spadgett: s/prefered/preferred

here and many places below

app/scripts/controllers/modals/setHomePageModal.js
```diff
@@ -0,0 +1,74 @@
...
+    $scope.preferedHomePage = $scope.initialPreferedHomePage;
+
+    $scope.selectedProject = null;
+    $scope.noProjects = false;
+
+    $scope.preSelectedProjectName = HomePagePreference.getProjectOverviewPage();
```
spadgett: nit: "preselect" is one word, so `preselectedProjectName`

app/scripts/controllers/modals/setHomePageModal.js
```diff
@@ -0,0 +1,74 @@
...
+
+    $scope.$on('no-projects-cannot-create', function() {
+      $scope.noProjects = true;
+    });
+
+    $scope.disableSaveButton = function() {
+      // if radiobuttons not changed
+      if ($scope.preferedHomePage === $scope.initialPreferedHomePage) {
+        if ($scope.preferedHomePage === 'project-overview') {
+          if (!$scope.selectedProject) {
```
spadgett: Can we just check `form.$pristine` instead to avoid all of this logic? The nested ifs are really hard to follow.

app/scripts/controllers/util/error.js
```diff
@@ -8,16 +8,18 @@
  * Controller of the openshiftConsole
  */
 angular.module('openshiftConsole')
-  .controller('ErrorController', function ($scope, $window) {
+  .controller('ErrorController', function ($scope, $location, $window, $uibModal, HomePagePreference, NotificationsService) {
```
spadgett: nit: `$uibModal` before `$window` to keep it alphabetized.

app/scripts/controllers/util/error.js
```diff
@@ -40,4 +42,29 @@ angular.module('openshiftConsole')
...
+        NotificationsService.addNotification({
+          id: "invalid-home-page-preference",
+          type: "warning",
+          message: "Home page set to invalid project.",
+          details: "Project '" + projectName + "' no longer exists.",
+          links: [{
+            href: "",
+            label: "Set Home Page",
+            onClick: function() {
+              $uibModal.open({
```
spadgett: Unfortunately this is going to leak memory because of https://github.com/openshift/origin-web-common/issues/97. I'm not sure what to suggest though until that issue is fixed.

app/scripts/directives/nav.js
```diff
@@ -204,6 +204,19 @@ angular.module('openshiftConsole')
     return {
       restrict: 'EA',
       templateUrl: 'views/directives/header/header.html',
+      controller: function (AuthService, HomePagePreference) {
+        AuthService.withUser().then(function() {
```
spadgett: I think this breaks the log out page? I'm not sure we can make this change since it will require every page to have an authenticated user.

Can we just check if `$rootScope.user` is set and not try to do anything if it's not.

app/scripts/directives/nav.js
```diff
@@ -204,6 +204,19 @@ angular.module('openshiftConsole')
...
+          HomePagePreference.init($rootScope.user);
+          switch(HomePagePreference.getHomePagePreference()) {
+            case 'project-list':
+              $location.path("/projects");
```
spadgett: We can't unconditionally redirect. This will break direct linking to any URL in the web console! We should only use the value of path is empty.

I'm don't think the nav controller is the right place to make this check. Is it possible to do the check here?

https://github.com/openshift/origin-web-console/blob/master/app/scripts/app.js#L39-L59

Then add another `/catalog` route for the catalog that is used when you click the logo?

spadgett:
```js
Naviage.toProjectList();
```

app/scripts/directives/nav.js
```diff
@@ -204,6 +204,19 @@ angular.module('openshiftConsole')
...
+              break;
+            case 'project-overview':
+              $location.path("/project/" + HomePagePreference.getProjectOverviewPage() + "/overview");
```
spadgett:
```js
Navigate.toProjectOverview(HomePagePreference.getProjectOverviewPage()));
```

app/scripts/directives/util.js
```diff
@@ -149,6 +149,22 @@ angular.module('openshiftConsole')
       }
     };
   })
+  .directive('setHomePage', function($uibModal) {
+    return {
+      restrict: 'E',
+      replace: true,
```
spadgett: `replace` is deprecated. We should avoid it.

app/scripts/directives/util.js
```diff
@@ -149,6 +149,22 @@ angular.module('openshiftConsole')
...
+      template: '<a href="">Set Home Page</a>',
+      link: function($scope, element) {
+        element.bind('click', function() {
```
spadgett: Why not use `ng-click`?

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .service("HomePagePreference", function(){
+
+    var HOME_PAGE_PREF_KEY = "origin/home-page-pref/";
```
spadgett: We use `openshift/` as the prefix elsewhere. Let's be consistent here.

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,37 @@
...
+
+    var init = function(user) {
+      HOME_PAGE_PREF_KEY += user.metadata.name;
```
spadgett: This could append the user name multiple times if called more than once. I think it's better to add a method to AuthService that gets the current user (without triggering a new login).

```js
var HOME_PAGE_KEY_PREFIX = "openshift/home-page-preference/";
var getHomePagePreference = function() {
  var user = AuthService.getUser(); // add to method
  if (!user) {
    return null;
  }
  var key = HOME_PAGE_KEY_PREFIX + user;
  // ...
};

File: app/scripts/services/homePagePreference.js [hunk 19]
```diff
@@ -0,0 +1,37 @@
...
+    };
+
+    var getHomePagePreference = function() {
+      var homePagePref = JSON.parse(localStorage.getItem(HOME_PAGE_PREF_KEY));
```
spadgett: we should have a try/catch to for JSON parse errors just in case.

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,37 @@
...
+      return homePagePref ? homePagePref.type : "catalog-home";
+    };
+
+    var getProjectOverviewPage = function() {
+      var homePagePref = JSON.parse(localStorage.getItem(HOME_PAGE_PREF_KEY));
```
spadgett: try/catch

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,37 @@
...
+
+    var HOME_PAGE_PREF_KEY = "origin/home-page-pref/";
+
+    var init = function(user) {
+      HOME_PAGE_PREF_KEY += user.metadata.name;
+    };
+
+    var getHomePagePreference = function() {
+      var homePagePref = JSON.parse(localStorage.getItem(HOME_PAGE_PREF_KEY));
+      return homePagePref ? homePagePref.type : "catalog-home";
```
spadgett:
```js
return _.get(homePagePref, 'type', 'catalog-home');
```

app/views/modals/set-home-page-modal.html
```diff
@@ -0,0 +1,48 @@
+<div class="modal-resource-action">
+  <div class="modal-header">
+    <h1>Set Home Page</h1>
+  </div>
+  <div class="modal-body indent">
+    <p>
+      <div class="title">Set your preferred home page.  This is the page you will see at login.</div>
```
spadgett: It's not necessarily the page you see when you login. It's the page if there's no URL path. We need to let users link directly to other pages.

app/scripts/controllers/util/error.js
```diff
@@ -40,4 +42,29 @@ angular.module('openshiftConsole')
...
     };
+
+    function checkHomePagePreference(projectName) {
+      if (projectName && projectName === HomePagePreference.getProjectOverviewPage()) {
+        HomePagePreference.clear();
+        NotificationsService.addNotification({
+          id: "invalid-home-page-preference",
+          type: "warning",
+          message: "Home page set to invalid project.",
+          details: "Project '" + projectName + "' no longer exists.",
```
spadgett: "no longer exists or you do not have access to it."

app/scripts/controllers/util/error.js
spadgett: It would be nice if we can catch this somewhere before we redirect to the error page.

I think this will actually incorrectly show the message if you go to

```
/projects/my-project/pod/missing-pod
```

even if my-project exists

app/scripts/controllers/util/error.js
spadgett: > Unfortunately this is going to leak memory because of openshift/origin-web-common#97. I'm not sure what to suggest though until that issue is fixed.

Actually if you move the code that shows the notification into a service, I think it should be OK.

app/scripts/controllers/util/error.js
```diff
@@ -40,4 +42,29 @@ angular.module('openshiftConsole')
...
       $window.location.href = "/";
     };
+
+    function checkHomePagePreference(projectName) {
+      if (projectName && projectName === HomePagePreference.getProjectOverviewPage()) {
+        HomePagePreference.clear();
+        NotificationsService.addNotification({
+          id: "invalid-home-page-preference",
+          type: "warning",
+          message: "Home page set to invalid project.",
```
spadgett: "invalid" doesn't really seem right. Maybe

> Home page project not found.
> Project my-project no longer exists or you do not have access to it.

dist.java/java/404.html
dtaylor113: Fixed

app/views/modals/set-home-page-modal.html
dtaylor113: @ncameronbritt, could you please re-wordsmith this?  -thanks

ncameronbritt: I feel like just removing that second sentence covers it. I believe that users would understand what a home page means.

spadgett:
Should clicking on the OpenShift logo take you to the home page or the catalog page? If it takes you to your home page, do we need another way to get to the catalog page?

app/scripts/directives/nav.js
dtaylor113: How would you recommend navigating to '/project/my-project/overview' in app.js's .config block when the preferred home page is set to a user's project overview?  -thanks

dtaylor113: This seems to work:
```
.run(function(HomePagePreference, Navigate) {
    var homePagePref = HomePagePreference.getHomePagePreference();
    if(homePagePref === "project-overview"){
      Navigate.toProjectOverview(HomePagePreference.getProjectOverviewPage());
    }
  })
```

spadgett: > How would you recommend navigating to '/project/my-project/overview' in app.js's .config block when the preferred home page is set to a user's project overview?

Check the home page preference once in that block (if the user is logged in) and then 

```js
var homePagePath = <determine home page path...>
if (homePagePath) {
  $routeProvider.when('/', { redirectTo: homePagePath } );
} else {
  $routeProvider.when('/', {
    templateUrl: 'views/landing-page.html',
    controller: 'LandingPageController',
    reloadOnSearch: false
  });
}
```

dtaylor113:
> Should clicking on the OpenShift logo take you to the home page or the catalog page? If it takes you to your home page, do we need another way to get to the catalog page?

With the latest code.  If homePage is set to project-list, then clicking on the logo brings you to the project list.   If homePage is undefined, 'catalog-home', or 'project-overview' -clicking on the logo brings you to the browse catalog landing page.  ...and that's only because I'm still trying to figure out how to default "/" to a specific project overview in app.js.

ncameronbritt:
That's a good question @spadgett. On the one hand, users have the expectation that clicking the logo will take them "home". But on the other hand, with Jeff's browse in project stuff, we potentially lose any ability to get to that landing page if OpenShift logo doesn't go there. And the application IA can look different depending on whether or not a user has specified something other than the landing page to be their home page.

It didn't occur to me that users already had the ability to specify where they land upon logging in by using bookmarks. Right now it would be easy enough to bypass the landing page simply by bookmarking another page. Is the idea with this design that *whenever* users log in they end up on their home page? If I have bookmarked different pages that are important to me, will those bookmarks still work if I have to log in?

app/scripts/services/homePagePreference.js
benjaminapetersen: or use lodash `_.attempt()`

app/scripts/directives/util.js
dtaylor113: The only problem is that without the `replace` the HTML is:
```
<dropdown-menu>
  <li>
       <set-home-page>   <--- not 'replaced'
           <a href...>
```
Which breaks the CSS dropdown selectors, which many have:
`.dropdown-menu > li > a`

So 'Set Home Page' in the user dropdown menu isn't styled like the other menu options.

dtaylor113: Not sure we want to override all the `.dropdown-menu > li > a` selectors

app/scripts/directives/util.js
dtaylor113: The directive is being used in an extension.  I did try ng-click in the nav-menu controller but the extensions seem to not trigger the ng-click function in the controller.

app/scripts/directives/util.js
spadgett: The directive is simple enough I'd almost rather not have it be a directive. We could have a helper in a service to open the modal. Then all it is is a link with a one line click handler.

app/scripts/directives/nav.js
dtaylor113: > I don't think the nav controller is the right place to make this check. Is it possible to do the check here?
> 
> https://github.com/openshift/origin-web-console/blob/master/app/scripts/app.js#L39-L59

Hi @spadgett,  the problem I'm running into is that the user is not logged in at the time app.js is executed.   I believed you and @jwforres indicated if getting the username was going to be an issue, then just store the homePagePref per browser and not use the userName in the key.   File a follow up issue to track.   Are we at that point? :-)

spadgett: I'm OK with that.

One thing you could try, though, is to register an `AuthService.onUserChanged` callback in app.js and update the route to redirect if the user has a home page preference.

app/scripts/directives/util.js
spadgett: OK I see the problem and it's what we do for copy login link. I'm OK with it.

app/scripts/directives/util.js
dtaylor113: I run into the same issue below where userDropdown.js is an extension where ng-click doesn't work, so I have to defer to using a directive.   The other `<a href=...`'s in the user and help dropdowns actually redirect to a URLs, but I don't think we want to do that for a modal.

app/scripts/directives/nav.js
dtaylor113: Ok, so I have `AuthService.onUserChanged` working in app.js's .run block.   The only issue is that when user sets a new home page, clicking on the logo will still goto the old route/prev. home page until the user logs in again or reloads "/" in the browser (so that app.js .run/.config blocks are rerun to reset the `$routeParams`.

dtaylor113:
Hi @spadgett & @ncameronbritt ,  I believe I have addressed all your review issues.  
I updated the screenshots in this PR's Description.

Only two issues remain:

1. After user sets a new home page, clicking on the logo will redirect to prev. home page setting until user reloads "\" or logs out/in.   We could have the Set Home Page modal on Save reload "/", not sure if that's good user experience though as they may initiate the Set Home Page modal from any location and may not want to goto the home page pref upon Saving.

2. @spadgett,  still using the `<set-home-page>` directive with `replace: true`.  Can you elaborate on how you might launch the modal from userDroppdown.js (extension) without using a directive?  Can't use `ng-click` as userDroppdown is an extension.  -thanks

dist/styles/fonts/openshift-logos-icon.svg
```diff
@@ -9,8 +9,8 @@
 	"license": "Apache 2.0",
 	"licenseURL": "https://www.apache.org/licenses/LICENSE-2.0",
 	"majorVersion": 1,
-	"minorVersion": 3,
-	"version": "Version 1.3",
+	"minorVersion": 2,
+	"version": "Version 1.2",
```
spadgett: Make sure you run `bower install` since this dependency was updated.

beanh66:
@ahardin-rh FYI, just wanted you to be aware of this feature!

beanh66:
@dtaylor113 Per the comment from @spadgett, I agree it is strange if we don't allow users any way to get back to the catalog landing page. In the original design we had said the logo in the masthead should still take users to the catalog landing page and "set home page" would only alter the page that appears on initial login.

dtaylor113:
@beanh66 , ok, I'll update it so that the logo always brings user to the Catalog.  -thanks

serenamarie125:
@dtaylor113 I think that makes sense for now, otherwise the user never has the opportunity to access the catalog with SaaS offerings.  Thanks!

dtaylor113:
Updated so that update it so that clicking on logo always brings user to the Catalog (unless catalog is disabled via config.js)

app/views/modals/set-home-page-modal.html
```diff
@@ -0,0 +1,50 @@
+<div class="modal-resource-action">
+  <div class="modal-header">
+    <h1>Set Home Page</h1>
+  </div>
+  <div class="modal-body indent">
+    <p>
+      <div class="title">Set your preferred home page.</div>
+      Note: This setting is browser specific and will not be maintained across browsers.
+    </p>
+    <ng-form name="form">
```
spadgett:
```html
<form name="form">
```

Give it a more specific name, though. Since this is a dialog shown on any page, the generic name `form` might already be used.

app/views/modals/set-home-page-modal.html
```diff
@@ -0,0 +1,50 @@
...
+                            hide-label="true">
+            </select-project>
+          </div>
+        </div>
+      </fieldset>
+    </ng-form>
+  </div>
+  <div class="modal-footer">
+      <button class="btn btn-lg btn-primary" type="button"
+              ng-disabled="(preferredHomePage === 'project-overview' && form.$invalid) || form.$pristine"
```
spadgett: It would be better to update `select-project` to take an optional required attribute that is set to true only when `preferredHomePage === 'project-overview'`. Then this changes to

```
ng-disabled="form.$invalid || form.$pristine"
```

app/views/overview/_service-bindings.html
```diff
@@ -19,7 +19,7 @@
   <p ng-if="($ctrl.refApiObject.kind !== 'ServiceInstance')  && !($ctrl.bindableServiceInstances | size)">
     <span>You must have a bindable service in your namespace in order to create bindings.</span>
     <div>
-      <a href="./">Browse Catalog</a>
+      <a href="./browse-catalog">Browse Catalog</a>
```
spadgett:
```html
<a ng-href="{{project | catalogURL}}">
```

app/scripts/app.js
```diff
@@ -35,6 +35,14 @@ angular
     'openshiftCommonUI',
     'webCatalog'
   ])
+  .run(function(AuthService, HomePagePreference) {
+    HomePagePreference.gotoHomePage();
```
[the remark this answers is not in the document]
dtaylor113: yes it does :-) sorry.  So we only want to go to the home page when path is empty.  `$location.$$url` seems to contain this info; path minus the baseHref:

```
When https://localhost:9000/dev-console/project/myproject/browse/stateful-sets/
$location.$$url =  /project/myproject/browse/stateful-sets/

When https://localhost:9000/dev-console/
$location.$$url = "/"
```
Seem reasonable? Unless there is a helper function which already tests this.
Thanks,
Dave

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,130 @@
...
+      var key = getHomePagePreferenceKey();
+
+      if (!key) {
+        return 'catalog-home';
+      }
+
+      try {
+        homePagePref = JSON.parse(localStorage.getItem(key));
+      } catch (e) {
+        return 'catalog-home';
```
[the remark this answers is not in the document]
benjaminapetersen: Yeah, swallowing the error entirely will probably be really frustrating to debug at some point.

app/scripts/directives/util.js
benjaminapetersen: Is a bug with extensions?  I believe there is a way to `ng-click`, but it's been a good while since I've looked @ the code.  Doesn't need to hold up this PR, but I can def take a look @ it if its causing trouble.

app/scripts/directives/util.js
benjaminapetersen: Its a little too bad that the selector is `.dropdown-menu > li > a` instead of `.dropdown-menu  .link > a` or something that would provide a little more flexibility...

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,130 @@
...
+      }
+
+      return (homePagePref && homePagePref.type === 'project-overview') ? homePagePref.project : null;
+    };
+
+    var verifyProjectOverviewPage = function(projectOverviewPage) {
+      var deferred = $q.defer();
+      var projects, projectNames;
+
+      ProjectsService.list().then(function(projectData) {
```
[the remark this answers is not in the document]
dtaylor113: duh..yes..sorry :-)

app/views/overview/_service-bindings.html
dtaylor113: Hmm...I don't see `project` defined in `/app/scripts/directives/overview/serviceBindings.js`

app/scripts/app.js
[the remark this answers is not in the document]
dtaylor113: Added code to only go to home page when path is empty.

app/scripts/app.js
```diff
@@ -57,6 +90,11 @@ angular
 
     $routeProvider
       .when('/', landingPageRoute)
+      .when('/browse-catalog', {
```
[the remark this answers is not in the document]
dtaylor113: I actually took this out of my PR and just limited this PR to only going to home page preference upon login.  I reverted all my previous changes to "Browse Catalog" links in favor of #2563.

app/views/directives/resource-service-bindings.html
```diff
@@ -19,7 +19,7 @@ <h3>Bindings</h3>
   <div ng-if="!$ctrl.apiObject.metadata.deletionTimestamp && !($ctrl.bindableServiceInstances | size)">
     <span>You must have a bindable service in your namespace in order to create bindings.</span>
     <div>
-      <a href="./">Browse Catalog</a>
+      <a href="./browse-catalog">Browse Catalog</a>
```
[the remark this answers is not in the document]
dtaylor113: Fixed

app/views/overview.html
```diff
@@ -12,7 +12,7 @@ <h2>Get started with your project.</h2>
             image, or add components like databases and message queues.
           </p>
           <p class="gutter-top">
-            <a ng-if="overview.catalogLandingPageEnabled" href="./" class="btn btn-lg btn-primary">
+            <a ng-if="overview.catalogLandingPageEnabled" href="./browse-catalog" class="btn btn-lg btn-primary">
```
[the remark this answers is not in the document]
dtaylor113: Reverted this change in favor of what was done in #2563

app/scripts/app.js
```diff
@@ -35,8 +58,18 @@ angular
     'openshiftCommonUI',
     'webCatalog'
   ])
+  .run(function(AuthService, HomePagePreference, Navigate) {
+    setHomePage(HomePagePreference, Navigate);
+    AuthService.onUserChanged(function(user) {
+      if (user !== null) {
```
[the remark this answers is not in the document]
dtaylor113: Fixed

app/scripts/app.js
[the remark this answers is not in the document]
dtaylor113: Arrggg...now if you set your home page pref to 'project-list' and log out and back in, it doesn't redirect to '/projects' even though the $location.path('projects') is called.   

As I test I have the code below, but even though `$location.path('projects')` is called during `onUserChanged` browser doesn't redirect to the Project List.

```
.run(function(AuthService, HomePagePreference, $location) {
    if ($location.$$url === "/") {
      HomePagePreference.gotoHomePage();
    }
    AuthService.onUserChanged(function(user) {
      if (user) {
        //HomePagePreference.gotoHomePage();
        $location.path('projects');
      }
    });
  })
```

It's weird because when log in to project which no longer exists, you do get the toast notification, **and a redirect to Project List!**.   Log in/out works fine with Catalog(Home) and Project Overview set as home page pref, just not the project-list.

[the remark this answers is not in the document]
dtaylor113: Ok, had to add a `$timeout` [here](https://github.com/openshift/origin-web-console/pull/2562/files#diff-f9201982127fd8efc3606dcbb9fa087fR110) in order to navigate.

app/views/modals/set-home-page-modal.html
dtaylor113: Hi @spadgett, sorry, I little confused on "...update select-project to take an optional required attribute..",  Can you please elaborate on 'optional required attribute`? -Thanks

app/scripts/services/homePagePreference.js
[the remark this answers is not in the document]
dtaylor113: Fixed

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,130 @@
...
+      var key = getHomePagePreferenceKey();
+
+      if (!key) {
+        return null;
+      }
+
+      try {
+        homePagePref = JSON.parse(localStorage.getItem(key));
+      } catch (e) {
+        return null;
```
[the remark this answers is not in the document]
dtaylor113: Fixed

app/scripts/services/homePagePreference.js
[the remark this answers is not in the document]
dtaylor113: Fixed

app/views/directives/header/header.html
```diff
@@ -40,7 +40,7 @@
       <li role="menuitem"><a ng-href="project/{{currentProjectName}}/create?tab=deployImage">Deploy Image</a></li>
       <li ng-if-end role="menuitem"><a ng-href="project/{{currentProjectName}}/create?tab=fromFile">Import YAML / JSON</a></li>
 
-      <li ng-if-start="catalogLandingPageEnabled" role="menuitem"><a href="./">Browse Catalog</a></li>
+      <li ng-if-start="catalogLandingPageEnabled" role="menuitem"><a href="./browse-catalog">Browse Catalog</a></li>
```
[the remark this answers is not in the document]
dtaylor113: Reverted, not using `/browse-catalog` or `catalog` route

app/views/modals/set-home-page-modal.html
dtaylor113: Fixed

spadgett:
@dtaylor113 I feel like we might be making this harder on ourselves than we need to. Is it possible to skip checking the user and just set up a redirect using...?

```js
$routeProvider.when('/', {
  redirectTo: myHomePage
});
```

app/views/modals/set-home-page-modal.html
spadgett: Sorry I phrased that poorly. I mean something like

```html
<select-project is-required="preferredHomePage === 'project-overview'" ...>
```

Then `form.$invalid` will be the correct value regardless of `preferredHomePage`

dtaylor113:
> $routeProvider.when('/', {
>   redirectTo: myHomePage
> });

What is `myHomePage`?  And in order to determine `myHomePage` won't we need to check the user's home page preference?

spadgett:
> What is myHomePage? And in order to determine myHomePage won't we need to check the user's home page preference?

Yes, but if we don't tie it to a specific user ID, it's just a check to localStorage

dtaylor113:
> Yes, but if we don't tie it to a specific user ID, it's just a check to localStorage
>> Ok, so your suggesting doing away with the user specific setting and just having a home page per browser (regardless of what user is logged in).

If so, then the only edge case maybe if userOne sets home to a project, then userTwo logs in but doesn't have access to said project, so userTwo will get the 'project no longer exists' toast notification where they can set home page to a project they have access to...so two users on same browser might keep stepping on each others toes.

dtaylor113:
Doesn't a user have to be logged in in order to do the `verifyProjectOverviewPage()` check where we do a `ProjectsService.get()..which uses....AuthService.withUser().then('...`  ?

Also having trouble using a promise in the `redirectTo`:

```
$routeProvider
      .when('/', {
        redirectTo: function() {
          HomePagePreferenceProvider.$get().getHomePageURL().then(function (url) {
            return url;
          });
          //returns before the .then(...)
        }
      })
```

spadgett:
@dtaylor113 We can try `resolveRedirectTo` instead of `redirectTo`, but it means we'd have to update to AngularJS 1.6.

https://code.angularjs.org/1.6.1/docs/api/ngRoute/provider/$routeProvider

dtaylor113:
> @dtaylor113 We can try resolveRedirectTo instead of redirectTo, but it means we'd have to update to AngularJS 1.6.

I can try that if you don't like the `.run` block.

app/scripts/app.js
```diff
@@ -35,6 +35,16 @@ angular
     'openshiftCommonUI',
     'webCatalog'
   ])
+  .run(function(AuthService, HomePagePreference, $location) {
+    if ($location.$$url === "/") {
```
spadgett: `$$url` is an internal angular property that we should avoid using.

app/scripts/app.js
```diff
@@ -35,6 +35,16 @@ angular
...
+      HomePagePreference.gotoHomePage();
+    }
+    AuthService.onUserChanged(function(user) {
+      if (user) {
+        HomePagePreference.gotoHomePage();
```
spadgett: It still seems like this could break direct linking? I think we need to avoid the `run()` block.

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,131 @@
...
+          if (valid) {
+            Navigate.toProjectOverview(homeProjectPage);
+          } else {
+            notifyInvalidProjectHomePage(homeProjectPage);
+            Navigate.toProjectList();
+          }
+        });
+      } else if (homePagePref === "project-list") {
+        // navigate on next digest cycle
+        $timeout(function(){
```
spadgett: Why is the timeout needed? Because we're calling from a `run()` block? I'd rather not have this in the service, but have it in the run block. But I'm guessing we're be able to remove this if we switch to `redirectTo`

app/styles/_core.less
```diff
@@ -978,6 +978,16 @@ a.disabled-link {
     margin-top: 0px;
     margin-bottom: 10px;
   }
+  .indent {
```
spadgett: These styles are specific to the home page dialog and shouldn't be in core.less or have such generic names.

app/styles/_core.less
```diff
@@ -978,6 +978,16 @@ a.disabled-link {
...
+    padding-left: 48px;
+    .title {
+      font-weight: 600;
+    }
+  }
+  .select-project-container {
+    padding-left: 20px;
+    width: 70%;
```
spadgett: How does this look at mobile?

Better to give it a max width in px instead of a percentage width. Then it can be 100% at mobile.

dist/styles/fonts/openshift-logos-icon.svg
```diff
@@ -35,7 +35,6 @@
...
...e="openshift" data-tags="logo-openshift" d="M300.992 496.168l-162.604-59.173c2.076-26.073 6.575-51.829 12.881-77.054l154.469 56.242c-5.021 26.108-6.801 53.029-4.746 79.985zM1019.614 675.975c-11.335 23.39-24.456 46.002-39.654 67.302l-162.576-59.166c18.919-19.357 34.795-41.101 47.775-64.364l154.455 56.228zM661.892 722.027c33.82-15.784 63.121-37.316 87.831-62.563l162.583 59.18c-45.035 63.184-106.376 115.987-181.53 151.058-232.413 108.375-509.698 7.458-618.065-224.913-35.077-75.204-48.107-155.090-41.991-232.427l162.597 59.166c2.698 35.226 11.306 70.529 27.055 104.377 70.416 150.994 250.527 216.516 401.521 146.121zM161.397 363.641l-154.483-56.264c14.202-56.334 38.813-109.836 72.648-157.463l162.237 59.046c-41.645 42.775-69.342 96.877-80.402 154.68zM835.28 424.94c-2.592-35.204-11.476-70.515-27.281-104.392-70.416-151.008-250.534-216.53-401.507-146.128-33.87 15.805-63.389 37.161-88.015 62.457l-162.244-59.060c44.922-63.184 106.206-116.016 181.403-151.1 232.413-108.36 509.662-7.436 618.037 224.962 35.098 75.169 48.050 155.027 41.85 232.307l-162.244-59.046zM875.259 623.433l-154.469-56.228c28.686-51.405 42.238-110.606 37.81-170.132l162.237 59.018c-4.647 58.227-20.176 115.055-45.579 167.343z" />
```
spadgett: Your PR has changes to logos that are unrelated.

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,131 @@
...
+    var clear = function() {
+      var key = getHomePagePreferenceKey();
+      if (!key) {
+        return;
+      }
+      localStorage.removeItem(key);
+    };
+
+    var notifyInvalidProjectHomePage = function(projectName) {
+      clear();
```
spadgett: This will `clear` your preference on any error. So if you lose network connectivity or something and the project is fine, it still clears your preference, which it shouldn't.

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,131 @@
...
+      return (homePagePref && homePagePref.type === 'project-overview') ? homePagePref.project : null;
+    };
+
+    var verifyProjectOverviewPage = function(projectOverviewPage) {
+      var deferred = $q.defer();
+
+      ProjectsService.get(projectOverviewPage).then(function() {
+        deferred.resolve(true);
+      }, function() {
+        deferred.resolve(false);
```
spadgett: We really need to check the status code. If it's not 403 or 404, the project might still exist.

dtaylor113:
Closing this WIP PR, too many comments.  I'll create another PR which hopefully is closer to the final solution.

PR #2563, opened 2017-12-04
Add Browse Catalog to Project context view.

jeff-phillips-18:
image

Waiting on final placement and visuals from @jennyhaines for the Catalog nav item.

@openshift/team-ux-review 

/hold

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -61,6 +61,10 @@ angular
         templateUrl: 'views/create-project.html',
         controller: 'CreateProjectController'
       })
+      .when('/project/:project/catalog', {
+        templateUrl: 'views/project-browse-catalog.html',
+        controller: 'ProjectBrowseCatalogController'
+      })
       .when('/project/:project', {
         redirectTo: function(params) {
           return '/project/' + encodeURIComponent(params.project) + "/overview";
--- a/app/scripts/constants.js
+++ b/app/scripts/constants.js
@@ -412,6 +412,14 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
       //     ]
       //   }
       // ]
+    },
+    {
+      label: "Catalog",
+      iconClass: "pficon pficon-service-catalog",
+      href: "/catalog",
+      canI: {
+        addToProject: true
+      }
     }
   ],
   CATALOG_CATEGORIES: [
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -4,6 +4,7 @@ angular.module('openshiftConsole').controller('OverviewController', [
   '$scope',
   '$filter',
   '$q',
+  '$location',
   '$routeParams',
   'AlertMessageService',
   'APIService',
@@ -35,6 +36,7 @@ angular.module('openshiftConsole').controller('OverviewController', [
 function OverviewController($scope,
                             $filter,
                             $q,
+                            $location,
                             $routeParams,
                             AlertMessageService,
                             APIService,
@@ -1089,6 +1091,10 @@ function OverviewController($scope,
     overview.clearFilter();
     updateFilter();
   });
+
+  $scope.browseCatalog = function() {
+    Navigate.toProjectCatalog($scope.projectName);
+  };
 
   LabelFilter.onActiveFiltersChanged(function() {
     $scope.$evalAsync(updateFilter);
--- a/app/scripts/controllers/projectBrowseCatalog.js
+++ b/app/scripts/controllers/projectBrowseCatalog.js
@@ -1,0 +1,55 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .controller('ProjectBrowseCatalogController',
+    function($scope,
+             $q,
+             $routeParams,
+             DataService,
+             AuthorizationService,
+             Catalog,
+             CatalogService,
+             Navigate,
+             NotificationsService,
+             ProjectsService) {
+      var currentProjectName = $routeParams.project;
+
+
+      ProjectsService.get(currentProjectName).then(function() {
+        AuthorizationService.getProjectRules(currentProjectName).then(function () {
+          var canIAddToProject = AuthorizationService.canIAddToProject(currentProjectName);
+          if (!canIAddToProject) {
+            Navigate.toProjectOverview(currentProjectName);
+            return;
+          }
+
+          var catalogItems;
+          var projectItems;
+
+          // Show all catalog items, project templates, and project image streams
+          var catalogItemsPromise = CatalogService.getCatalogItems().then(function (items) {
+            catalogItems = items;
+          });
+
+          var projectItemsPromise = Catalog.getProjectCatalogItems(currentProjectName).then( _.spread(function(catalogServiceItems, errorMessage) {
+            projectItems = catalogServiceItems;
+            if (errorMessage) {
+              NotificationsService.addNotification(
+                {
+                  type: "error",
+                  message: errorMessage
+                }
+              );
+            }
+          }));
+
+          $q.all([catalogItemsPromise, projectItemsPromise]).then(function() {
+            $scope.catalogItems = Catalog.sortCatalogItems(_.concat(catalogItems, projectItems));
+            if ($routeParams.filter) {
+              $scope.keywordFilter = $routeParams.filter;
+            }
+
+          });
+        });
+      });
+    });
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -475,6 +475,9 @@ angular.module('openshiftConsole')
                 Navigate.projectOverviewURL(''); // fail case will invoke default route (projects list)
     };
   })
+  .filter('catalogURL', function(Navigate) {
+    return Navigate.catalogURL;
+  })
   .filter('createFromSourceURL', function() {
     return function(projectName, sourceURL) {
       var createURI = URI.expand("project/{project}/catalog/images{?q*}", {
--- a/app/scripts/services/navigate.js
+++ b/app/scripts/services/navigate.js
@@ -82,6 +82,35 @@ angular.module("openshiftConsole")
         $location.path(this.membershipURL(projectName));
       },
 
+      /**
+       * Return the URL for the project catalog browse page
+       *
+       * @param {Object|String}     project - Can be a project object or the project's name (string)
+       * @returns {String} a URL string for the project catalog browse page
+       */
+      catalogURL: function(project) {
+        var projectName = angular.isString(project) ? project : _.get(project, 'metadata.name');
+        if (!projectName) {
+          return './';
+        }
+
+        return "project/" + encodeURIComponent(projectName) + "/catalog";
+      },
+
+      /**
+       * Navigate and display the project catalog browse page.
+       *
+       * @param {Object|String} project - Can be a project object or the project's name (string)
+       * @param {Object} search - optional search object (supports initial filters via a filter field)
+       * @returns {undefined}
+       */
+      toProjectCatalog: function(project, search) {
+        var loc = $location.path(this.catalogURL(project));
+        if (search) {
+          loc.search(search);
+        }
+      },
+
       quotaURL: function(projectName) {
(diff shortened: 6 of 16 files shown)


app/scripts/app.js
```diff
@@ -61,6 +61,10 @@ angular
         templateUrl: 'views/create-project.html',
         controller: 'CreateProjectController'
       })
+      .when('/project/:project/project-browse-catalog', {
```
spadgett: Personally, I like the shorter `/project/:project/catalog`

app/scripts/controllers/overview.js
```diff
@@ -1090,6 +1092,10 @@ function OverviewController($scope,
     updateFilter();
   });
 
+  $scope.browseCatalog = function() {
+    $location.url('project/' + $scope.projectName + '/project-browse-catalog');
```
spadgett: Suggest adding this to the `Navigate` service so we only ever have one place where we build the URL.

app/scripts/controllers/projectBrowseCatalog.js
```diff
@@ -0,0 +1,34 @@
...
+
+angular.module('openshiftConsole')
+  .controller('ProjectBrowseCatalogController',
+    function($scope,
+             $location,
+             $routeParams,
+             AuthorizationService,
+             AuthService,
+             CatalogService) {
+      AuthService.withUser().then(function() {
```
spadgett: We should really be calling `ProjectsService.get()` to make sure the project exists. In that case, you don't need `AuthService.withUser()`.

app/scripts/controllers/projectBrowseCatalog.js
```diff
@@ -0,0 +1,34 @@
...
+
+        AuthorizationService.getProjectRules(currentProjectName).then(function () {
+          var canIAddToProject = AuthorizationService.canIAddToProject(currentProjectName);
+          if (!canIAddToProject) {
+            $location.url('project/' + currentProjectName);
+            return;
+          }
+
+          CatalogService.getCatalogItems().then(function (items) {
+            $scope.catalogItems = items;
```
spadgett: It would be nice to show the image streams and templates from the current project as well since we have a project context.

app/styles/_project-browse-catalog.less
```diff
@@ -0,0 +1,37 @@
+.project-browse-catalog {
+  .services-view-container {
+    background: @color-pf-black-200;
+    min-height: calc(100vh ~"-" 101px);
```
rhamilto: We should avoid using vh units since they aren't reliable on mobile devices.

jeff-phillips-18:
Addressed all issues except for showing the image streams and templates from the current project.

Updated the nav item icon to the pficon-service-catalog per latest designs.

image

app/scripts/controllers/projectBrowseCatalog.js
jeff-phillips-18: We don't currently support browsing and choosing image streams. I can add project templates though.

spadgett: > We don't currently support browsing and choosing image streams.

We support builder images in the catalog itself.

spadgett:
> Addressed all issues except for showing the image streams and templates from the current project.

We can leave that as a follow on.

app/views/directives/header/header.html
```diff
@@ -40,7 +40,7 @@
       <li role="menuitem"><a ng-href="project/{{currentProjectName}}/create?tab=deployImage">Deploy Image</a></li>
       <li ng-if-end role="menuitem"><a ng-href="project/{{currentProjectName}}/create?tab=fromFile">Import YAML / JSON</a></li>
 
-      <li ng-if-start="catalogLandingPageEnabled" role="menuitem"><a href="./">Browse Catalog</a></li>
+      <li ng-if-start="catalogLandingPageEnabled" role="menuitem"><a href="project/{{currentProjectName}}/catalog">Browse Catalog</a></li>
```
spadgett: Ideally we'd add a filter to build the URL using the `Navigate` service, similar to

https://github.com/openshift/origin-web-console/blob/master/app/scripts/filters/resources.js#L469

jeff-phillips-18:
Added project templates and image streams to the list of items being browsed.

jeff-phillips-18:
Set the icon back to fa-book for the catalog item on the nav menu.

jeff-phillips-18:
That completes the UI visuals updates per @jennyhaines

Still WIP for https://github.com/openshift/origin-web-catalog/pull/587

@jwforres Are we OK with having 'Browse Catalog' on the 'Add to Project' dropdown still? It does the same thing as clicking on the nav item. While it is redundant, it may help the novice. Easily removable if we want to avoid the redundancy.

jwforres:
Yeah I wouldn't remove it from add to project, our existing users expect it
there.

spadgett:
@jeff-phillips-18 I merged openshift/origin-web-catalog#587. We should be ready to cut a v3.9.0-alpha.0 release of catalog.

jeff-phillips-18:
openshift/origin-web-catalog#587 is available, removed WIP

/hold cancel

app/scripts/controllers/projectBrowseCatalog.js
```diff
@@ -0,0 +1,49 @@
...
+          var canIAddToProject = AuthorizationService.canIAddToProject(currentProjectName);
+          if (!canIAddToProject) {
+            Navigate.toProjectOverview(currentProjectName);
+            return;
+          }
+
+          // Show all catalog items, project templates, and project image streams
+          CatalogService.getCatalogItems().then(function (items) {
+            Catalog.getProjectCatalogItems(currentProjectName, false, true).then( _.spread(function(catalogServiceItems, errorMessage) {
+              DataService.list(imageStreamsVersion, {namespace: currentProjectName}, function(isData) {
```
spadgett: Pass `true` for `includeImages` to `getProjectCatalogItems` instead?

https://github.com/openshift/origin-web-catalog/blob/master/src/services/catalog.service.ts#L116-L125

app/scripts/filters/resources.js
```diff
@@ -475,6 +475,15 @@ angular.module('openshiftConsole')
                 Navigate.projectOverviewURL(''); // fail case will invoke default route (projects list)
     };
   })
+  .filter('catalogURL', function(Navigate) {
+    return function(project) {
+      return angular.isString(project) ?
+        Navigate.catalogURL(project) :
+        angular.isObject(project) ?
+          Navigate.catalogURL(project.metadata && project.metadata.name):
+          Navigate.catalogURL(''); // fail case will invoke default route (projects list)
```
spadgett: > fail case will invoke default route (projects list)

Looks like that's not the case looking at the implementation below.

app/scripts/services/navigate.js
```diff
@@ -82,6 +82,17 @@ angular.module("openshiftConsole")
         $location.path(this.membershipURL(projectName));
       },
 
+      catalogURL: function(projectName) {
+        return "project/" + encodeURIComponent(projectName) + "/catalog";
```
spadgett: Maybe

```js
if (!projectName) {
  return "./";
}

return "project/" + encodeURIComponent(projectName) + "/catalog";
```

app/scripts/filters/resources.js
spadgett: I realize this was copied from the overview method, but I think something like this is better:

```js
if (!_.isString(project)) {
  project = _.get(project, 'metadata.name');
}
return Navigate.catalogURL(project);
```

Or possibly move that logic into `Navigate.catalogURL`

app/scripts/controllers/projectBrowseCatalog.js
jeff-phillips-18: The CatalogService passes true to Catalog if i!sTemplateServiceBrokerEnabled

app/scripts/filters/resources.js
jeff-phillips-18: I've moved it into Navigate.catalogURL

app/scripts/controllers/projectBrowseCatalog.js
spadgett: > The CatalogService passes true to Catalog if i!sTemplateServiceBrokerEnabled

I think you might be looking at a different method / argument? Unless I'm misreading the code, there's one for image streams for the `getProjectCatalogItems` method you're calling directly. You're passing `false` just above.

jeff-phillips-18: Yeah, I was looking at the CatalogService.getCatalogItems call.

However, the Catalog.getProjectCatalogItems defaults the includeImages and includeTemplates to true:

`  public getProjectCatalogItems(projectName: string, includeImages: boolean = true, includeTemplates: boolean = true, partialObjectMetadataList: boolean = false ) {
`

so I think this is OK, right? When testing I do see both project templates and project image streams when browsing in project context which I do not see from the landing page.

app/scripts/controllers/projectBrowseCatalog.js
```diff
@@ -0,0 +1,49 @@
...
+          var canIAddToProject = AuthorizationService.canIAddToProject(currentProjectName);
+          if (!canIAddToProject) {
+            Navigate.toProjectOverview(currentProjectName);
+            return;
+          }
+
+          // Show all catalog items, project templates, and project image streams
+          CatalogService.getCatalogItems().then(function (items) {
+            Catalog.getProjectCatalogItems(currentProjectName).then( _.spread(function(catalogServiceItems, errorMessage) {
+              DataService.list(imageStreamsVersion, {namespace: currentProjectName}, function(isData) {
```
spadgett: I think we can remove this call? Aren't we fetching it twice now since `getProjectCatalogItems` also request image streams?

app/scripts/filters/resources.js
```diff
@@ -475,6 +475,11 @@ angular.module('openshiftConsole')
                 Navigate.projectOverviewURL(''); // fail case will invoke default route (projects list)
     };
   })
+  .filter('catalogURL', function(Navigate) {
+    return function(project) {
+      return Navigate.catalogURL(project);
+    };
+  })
```
spadgett: nit: this is fine, but I think it could just be

```js
.filter('catalogURL', function(Navigate) {
  return Navigate.catalogURL;
})
```

app/scripts/services/navigate.js
```diff
@@ -82,6 +82,22 @@ angular.module("openshiftConsole")
...
+      catalogURL: function(project) {
+        var projectName = angular.isString(project) ? project : _.get(project, 'metadata.name');
+        if (!projectName) {
+          return './';
+        }
+
+        return "project/" + encodeURIComponent(projectName) + "/catalog";
+      },
+
+      toProjectCatalog: function(projectName, search) {
```
spadgett: nit: You might change the parameter name to `project` since it can be an object, too. (Probably worth a comment as well.)

app/scripts/controllers/projectBrowseCatalog.js
spadgett: nit: We could request these in parallel and use `$q.all` so the page load slightly faster.

app/scripts/services/navigate.js
jeff-phillips-18: renamed and added comments

serenamarie125:
I'd suggest the icon to be set back to pficon-service-catalog, this way, when PatternFly updates the icon, OpenShift will not need to update the code to get the new icon.

app/scripts/constants.js
```diff
@@ -412,6 +412,14 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
       //     ]
       //   }
       // ]
+    },
+    {
+      label: "Catalog",
+      iconClass: "fa fa-book",
```
serenamarie125: I'd suggest the icon to be set back to pficon-service-catalog, this way, when PatternFly updates the icon, OpenShift will not need to update the code to get the new icon.  @jeff-phillips-18

jeff-phillips-18:
Thanks for you diligence @spadgett 

I've addressed those issues.

PR #2567, opened 2017-12-05
Convert alerts directive to component and add spec test

TheRealJon:
## Description
Converted Angular directive `app/scripts/directives/alerts.js` into component at `app/components/alerts/alerts.component.js`. Added a spec test to the test suite which unit tests the new alerts component.

## Changes
- Create alerts component
- Add call to `close(alert)` function when an alert's link `onClick` function returns true. This ensures that the alert's `onClose` callback is called when it is dismissed by a link.
- Refactor alert template for use in component rather than directive.
- Remove alerts directive
- Load component in index.html
- Add spec test for component
- Add external html template to javascript module conversion to karma config (so that external templates can be loaded in spec tests without the http request).
- Load external template module in `spec-helper.js`
- Move alert component, template, and spec test files to `app/components/alerts` folder
- Adjust grunt and karma config for new alerts component file locations.
- Remove alerts component `toast` property
- Remove alerts component `animateSlide` property
- Change `filter` and `hideCloseButton` properties to one-way data binding in alerts component


app/scripts/components/alerts.component.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
```
benjaminapetersen: I'm a fan of moving things to a `scripts/components` directory, but would like to see a step further.  Can we do something like:

```
- scripts 
   - components 
     - alerts
        - alert.component.js
        - alert.view.html
```
I'm 150% into a component based heirarchy, its so much easier to derive relationships.  I realize this will require a little bit of tweaking to some of our build tools, but I doubt its much more than tweaking the globbing patterns.

@spadgett @jwforres may not agree, so I'll defer to their preference.

app/scripts/components/alerts.component.js
```diff
@@ -0,0 +1,45 @@
...
+
+angular.module('openshiftConsole')
+  .component('alerts', {
+    bindings: {
+      // Map (or array) of alert objects, each with properties
```
benjaminapetersen: Maybe just put an exampe of the object here:

```JavaScript
alerts: {
  type: '',            // text....
  message: '',    // text....
  links: [{
     href: ''           // text....
     // etc.
  }],
  onClose: ''      // text....
}
```
Might be easier to consume

app/scripts/components/alerts.component.js
```diff
@@ -0,0 +1,45 @@
...
+      //   links:    an array of link objects (optional, see below)
+      //   onClose:  callback function when the user dismisses the alert (optional)
+      //
+      // Each link can have the following properties:
+      //   href:     link URL (optional if onClick is set)
+      //   label:    the link text
+      //   onClick:  handler to call when the link if clicked (optional)
+      //             If onClick returns true, dismisses the alert
+      alerts: '=',
+      // Filter function (optional).
```
benjaminapetersen: `Filter function` is a bit too obvious of a comment here.  Can you update it to describe what/why this would be used?

app/scripts/components/alerts.component.js
```diff
@@ -0,0 +1,45 @@
...
+      ctrl.close = function(alert) {
+        alert.hidden = true;
+        if (_.isFunction(alert.onClose)) {
+          alert.onClose();
+        }
+      };
+
+      ctrl.onClick = function(alert, link) {
+        if (_.isFunction(link.onClick)) {
+          // If onClick() returns true, also hide the alert.
```
benjaminapetersen: Which `onClick` returns `true`?  `ctrl.onClick` or `links.onClick`?  This comment is a bit confusing.

benjaminapetersen: (preference for no comment if the code can be altered a bit to make that clear)

app/scripts/components/alerts.component.js
```diff
@@ -0,0 +1,45 @@
...
+      //             If onClick returns true, dismisses the alert
+      alerts: '=',
+      // Filter function (optional).
+      filter: '=?',
+      animateSlide: '=?',
+      hideCloseButton: '=?',
+      toast: '=?'
+    },
+    templateUrl: 'views/_alerts.html',
+    controller: function() {
```
benjaminapetersen: We have a convention for using named functions for controllers for components.  That said, I prefer inline anonymous, I don't think we gain anything out of the named functions other than the need for IIFE's around all of these files.  

@spadgett I think may have a preference for the named function however, so I defer to his call on it.

app/views/_alerts.html
```diff
@@ -1,28 +1,16 @@
...
-    <div class="alert word-break" ng-class="{
-      'alert-danger': alert.type === 'error',
-      'alert-warning': alert.type === 'warning',
-      'alert-success': alert.type === 'success',
-      'alert-info': !alert.type || alert.type === 'info',
-      'toast-pf': toast,
-      'mar-left-sm': toast
-    }">
-      <button ng-if="!hideCloseButton" ng-click="close(alert)" type="button" class="close">
+<div ng-attr-row="{{$ctrl.toast}}"  ng-attr-wrap="{{$ctrl.toast}}">
```
benjaminapetersen: See above comment, but I'd love to see this file moved to `app/scripts/components/alerts/alert.view.html`.  Colocation is 😄  😀

app/views/_alerts.html
```diff
@@ -1,28 +1,16 @@
...
+  <div ng-repeat="(alertID, alert) in ($ctrl.alerts | filterCollection : filter) track by (alertID + (alert.message || alert.details))" ng-if="!alert.hidden" class="alert-wrapper animate-repeat" ng-class="{'animate-slide': $ctrl.animateSlide}">
+    <div class="alert word-break" ng-class="{ 'alert-danger': alert.type === 'error', 'alert-warning': alert.type === 'warning', 'alert-success': alert.type === 'success', 'alert-info': !alert.type || alert.type === 'info', 'toast-pf': $ctrl.toast, 'mar-left-sm': $ctrl.toast}">
```
benjaminapetersen: Not sure if we have a consensus on style in the console, but I def prefer the previous style:

```html 
<!-- preferred -->
<div
   id="foo"
   class="bar"
   ng-complicated-thing="break,it,up,again"></div>

<!-- not preferred, paragraph form gets hard to parse quickly -->
<div id="foo" class="bar" ng-complicated-thing="break,it,up,again"></div>
```
That and extremely long line lengths make it difficult to use multiple columns in a text editor.  Its nice to open JS `controller` next to HTML `view` and sometimes a JS `service` and be able to see everything.

We don't have a style guide tho, so I'll defer again to @jwforres @spadgett!

test/spec/components/alerts.component.spec.js
```diff
@@ -0,0 +1,99 @@
...
+          message: "An error happened and stuff.",
+          links: [
+            {
+              label: "test_link"
+            }
+          ]
+        },
+        warning: {
+          type: "warning",
+          message: "I'm going to count to five."
```
benjaminapetersen: hehe....

test/spec/components/alerts.component.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      };
+
+      // Create an instance of the component
+      element = angular.element('<alerts alerts="alerts"></alerts>');
+      element = $compile(element)($scope); // Compile the component
+      $scope.$digest(); // Update the HTML
+  }));
+
+  // Ensure four alert elements exist
+  it("contain alert elements", function () {
```
benjaminapetersen: We have had language like `it('should contain....')`.  Dunno if we care about that `should` or not.

test/spec/components/alerts.component.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    let allAlerts = element.find('.alert');
+    expect(allAlerts.length).toBe(4); // four alerts exist
+  });
+
+  // Ensure danger alert exists and is correct
```
benjaminapetersen: Generally pretty good with tests.  Do we need to test negative cases along with positive cases?  Example, `warning` should not have `alert-success`, etc.  That could be a bit tedious for this, tho I think its good practice to test both sides.

test/spec/components/alerts.component.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  it("show info alert", function(){
+    let info = $(element.find('.alert')[3]);
+    expect(info.hasClass('alert-info')).toBeTruthy(); // info alert has correct class
+    expect(info.text()).toContain("This is informational."); // info alert contains correct text
+  });
+
+  // Ensure `close` function is called when close button is clicked, and the
+  // hidden state is true.
+  it("call contoller close function", function(){
+    $scope.onCloseSpy = jasmine.createSpy('onCloseSpy');
```
benjaminapetersen: +1 to using a `spy`, I think this may be the first in our tests?

app/scripts/components/alerts.component.js
TheRealJon: I agree, I would probably go even one step further and get rid of the scripts folder:
```
app/
  components/
    alerts/
      alerts.component.js
      alerts.component.spec.js
      alerts.template.html
```
Also, should we think about breaking this out into a separate story? I feel like introducing file structure changes in increments that are too small (one component at a time) will be confusing.

app/views/_alerts.html
TheRealJon: I do like the idea of breaking these long lines down, but for some reason, it breaks the spec test. The compiled element does not get the proper class applied when it's formatted that way.

app/scripts/components/alerts.component.js
TheRealJon: Most of the comments were just copy/paste from directive, but I can go back and spruce them up if needed.

app/views/_alerts.html
benjaminapetersen: oh it breaks the test?  what output do you get?

app/scripts/components/alerts.component.js
benjaminapetersen: Agree on pulling it out of `scripts` since components are not 100% scripts.

We def prefer small PRs over big sweeping PRs, so I don't think this is a problem.  Rebases of giant structural PRs are terribly painful.

app/scripts/components/alerts.component.js
TheRealJon: I can clarify what this filter function does. Just realized that this probably needs a spec test case as well.

app/scripts/components/alerts.component.js
TheRealJon: Yeah, I was thinking it could still be incremental, just more than one file at a time. We could choose whole sections to do at a time and break it up into a few PRs.

test/spec/components/alerts.component.spec.js
TheRealJon: Just noticed a typo there too... "call contoller close function"

app/scripts/components/alerts.component.js
spadgett: It seems like a good time to do this.

app/scripts/components/alerts.component.js
```diff
@@ -0,0 +1,45 @@
...
+      //   message:  the message to display
+      //   links:    an array of link objects (optional, see below)
+      //   onClose:  callback function when the user dismisses the alert (optional)
+      //
+      // Each link can have the following properties:
+      //   href:     link URL (optional if onClick is set)
+      //   label:    the link text
+      //   onClick:  handler to call when the link if clicked (optional)
+      //             If onClick returns true, dismisses the alert
+      alerts: '=',
```
spadgett: I think we should switch these to `<` bindings unless there is a reason not to.

(Not sure if this causes a problem with how we handle close since we set the hidden property inside the component. We might need to leave `alerts` as a `=` binding.)

app/scripts/components/alerts.component.js
```diff
@@ -0,0 +1,45 @@
...
+      // Filter function (optional).
+      filter: '=?',
+      animateSlide: '=?',
+      hideCloseButton: '=?',
+      toast: '=?'
```
spadgett: I think we can remove this now. @TheRealJon can you sweep the code to check if it's still used anywhere?

Can you also check `animateSlide`?

app/scripts/components/alerts.component.js
spadgett: Can you update the comment to say `link.onClick()`?

I think the comment here is helpful.

test/spec/components/alerts.component.spec.js
spadgett: +1, I prefer having the "should" in there

test/spec/components/alerts.component.spec.js
```diff
@@ -0,0 +1,99 @@
+"use strict";
+
+describe("Component: Alerts", function(){
+
+  var $scope, $compile, element;
+  beforeEach(inject(function(_$compile_, _$rootScope_) {
+      $scope = _$rootScope_;
+      $compile = _$compile_;
+      $scope.alerts = {
```
spadgett: One of these should have a `details` property that we test

spadgett:
@jwforres any objections to moving things under `app/components` as we convert these?

jwforres:
thats fine, we just need to make sure grunt is doing the rights thing for both build and serve

TheRealJon:
@benjaminapetersen @spadgett @jwforres I believe I have addressed all of the comments. I moved the alerts component, template, and spec test to be collocated in `app/components` and updated grunt build and serve, as well as karma config. I also removed the unused properties in the alert component, and updated the rest to one-way data binding. Could someone take a close look at the grunt file to make sure I did everything that was needed there? I'm not intimately familiar with the build process yet and just want to make sure I didn't miss anything. Just looked back over the comments and realized I missed the one about the alert details property. I will add a test in there for that.

spadgett:
@TheRealJon Thanks, you should be able to run `grunt serve:dist` to test what's actually built.

TheRealJon:
I added a check for the details property of an alert and also added it to the comments in the alerts component. Also, ran `grunt server:dist` just to make sure the build was working correctly and everything seems to work.

app/components/alerts/alerts.component.spec.js
```diff
@@ -0,0 +1,126 @@
+"use strict";
+
```
benjaminapetersen: Also, I think I do like having the test co-located with the file.  If we improve the structure so that related bits are all together, it makes it very clear when something isn't properly tested.

TheRealJon:
@spadgett @jwforres What do you think? Can I remove WIP status? Just trying to nudge this through before the end of the week since I will be gone for 3 weeks. 😄

spadgett:
@TheRealJon were you able to try `grunt serve:dist`?

TheRealJon:
@spadgett Yes, I added a test alert to the overview controller and ran `grunt serve:dist` and it seemed to work fine.

test/karma.conf.js
```diff
@@ -102,11 +102,13 @@ module.exports = function(config) {
       "bower_components/origin-web-catalog/dist/vendor-bundle.js",
       "bower_components/origin-web-catalog/dist/origin-web-catalogs.js",
       'app/scripts/**/*.js',
-      'app/views/directives/**/*.html',
```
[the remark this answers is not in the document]
TheRealJon: We are not using any of the templates in that directory for our tests.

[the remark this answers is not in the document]
TheRealJon: As long as we plan on moving things to a new component based file structure and writing tests for our components/directives as we do so, we shouldn't need that glob pattern any more.

jwforres:
Ok, I don't have a strong opinion here. Interested in Ben and Sam's
thoughts.

benjaminapetersen:
I love sharing opinions about things that don't really matter 😜  

So my initial thought is that since everything is under `/components` as a directory, having the extension be `.component.js` is probably redundant.  Template files with `.template.html` seem a bit lengthy as well, unless we have `.html` files that are not templates. (partials?) 

But I do tend to like [this style guide](https://github.com/toddmotto/angularjs-styleguide), which gives the following example for an angular 1.6 app:

```
/components/calendar
- calendar.module.js
- calendar.component.js
- calendar.service.js
- calendar.directive.js
- calendar.filter.js
- calendar.spec.js
- calendar.html
- calendar.scss
```

So perhaps more structure is better than less?  I'm sure nobody wants to rename all these files later.

(I think I very helpfully did not answer the question)

TheRealJon:
I can concede the `template.html` part because it really isn't all that useful (I just did it for consistency), but I still think the `.component.js` part would be helpful. The `components` folder in this case represents a broader context (something like "application components") while the file name implies more specifically that this file is an Angular Component. Plus if we are going to have all of the js files for an application component collocated, they will need to be differentiated somehow. Does anyone have any alternative ideas for this?

benjaminapetersen:
Yup, I think that's exactly where I land.

jwforres:
Agreed, the .template.html is really the only one that bothered me.

TheRealJon:
@jwforres @spadgett @benjaminapetersen File names fixed, should I squash then remove WIP?

jwforres:
@TheRealJon yes please do

TheRealJon:
All done! Someone say the magic word! 😄 
@jwforres @spadgett

PR #2570, opened 2017-12-06
Bump ace-builds to v1.2.9, fixes copy/paste right menu in Chrome

jwforres:


jwforres:
@spadgett i was able to verify the YAML editor and your warning/error inline icon stuff, but not Jenkinsfile/Dockerfile editors because my system is very hosed right now

PR #2585, opened 2017-12-07
Bug 1520828 - Do not enable Bind button until data fetched

jeff-phillips-18:
Also, do not fetch all the service classes and plans when only a single service class and its plans are necessary.

### DIFF
--- a/app/scripts/directives/bindService.js
+++ b/app/scripts/directives/bindService.js
@@ -8,6 +8,7 @@
       'APIService',
       'ApplicationsService',
       'BindingService',
+      'Catalog',
       'DataService',
       'ServiceInstancesService',
       BindService
@@ -26,6 +27,7 @@
                        APIService,
                        ApplicationsService,
                        BindingService,
+                       Catalog,
                        DataService,
                        ServiceInstancesService) {
     var ctrl = this;
@@ -37,6 +39,9 @@
     var bindingWatch;
     var statusCondition = $filter('statusCondition');
     var enableTechPreviewFeature = $filter('enableTechPreviewFeature');
+    var serviceInstancesVersion = APIService.getPreferredVersion('serviceinstances');
+    var serviceClassesVersion = APIService.getPreferredVersion('clusterserviceclasses');
+    var servicePlansVersion = APIService.getPreferredVersion('clusterserviceplans');
 
     var preselectService = function(){
       var newestReady;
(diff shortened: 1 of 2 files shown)


spadgett:
I suggest adding package-lock.json in a separate PR if we want it. We might backport this fix. (I'm not sure if including package-lock was intentional.)

app/scripts/directives/bindService.js
```diff
@@ -152,28 +157,37 @@
...
-      var servicePlanName = ServiceInstancesService.getServicePlanNameForInstance(instance);
-      ctrl.plan = ctrl.servicePlans[servicePlanName];
-      ctrl.parameterSchema = _.get(ctrl.plan, 'spec.serviceBindingCreateParameterSchema');
-      ctrl.parameterFormDefinition = _.get(ctrl.plan, 'spec.externalMetadata.schemas.service_binding.create.openshift_form_definition');
-      bindParametersStep.hidden = !_.has(ctrl.parameterSchema, 'properties');
-      ctrl.nextTitle = bindParametersStep.hidden ? 'Bind' : 'Next >';
-      ctrl.hideBack = bindParametersStep.hidden;
+      ServiceInstancesService.fetchServiceClassForInstance(ctrl.serviceToBind).then(function(serviceClass) {
+        ctrl.serviceClass = serviceClass;
+        Catalog.getServicePlansForServiceClass(ctrl.serviceClass).then(function (plans) {
```
spadgett: If this isn't returning cached data, it's probably better to just request the plan by name.

```js
var servicePlansVersion = APIService.getPreferredVersion('clusterserviceplans');
var servicePlanName = ServiceInstancesService.getServicePlanNameForInstance(ctrl.serviceToBind);
DataService.get(servicePlansVersion, servicePlanName, {}).then(function(plan) {
  ctrl.plan = plan;
  // ...
});
```

spadgett: Although I guess we have to worry about a race if this can get called more than once before the first request completes :(

jeff-phillips-18:
Sorry, did not mean to include package-lock.json

app/scripts/directives/bindService.js
jeff-phillips-18: The list of plans is need only if/when we support podPresets and allow the binding to occur from an application and then I believe we need all plans so this list would be incorrect anyways. I will move the list of plans retrieval to where it is needed.

This would run into a race condition either way right? This, again, would only happen when binding from an application and if the user were to switch service instances quickly.

spadgett: I'm starting to think we want different components for the different bind dialogs (with any common logic for creating the binding in a service), but that's a bigger change :/

jeff-phillips-18: yeah, I think I agree with you there. It's gotten more complex and hardly seems worth it at this point.

jeff-phillips-18:
@spadgett PTAL, I refactored the code some more to only get the list of plans when binding from an application. This should be clean for binding a service at this point, but I do agree that binding from an application could be broken out to make this component more clear and concise.

PR #2586, opened 2017-12-08
Switch kebabs to buttons so they are sized correctly

sg00dwin:
Stateful sets and provisioned services kebabs were small

screen shot 2017-12-08 at 11 10 25 am


spadgett:
Thanks for fixing my bug :)

/lgtm

PR #2588, opened 2017-12-11
I18n

fenfenraorao:


spadgett:
@fenfenraorao It's better to keep pushing to the same branch and keep the same PR so we don't lose reviews and discussion. You can always force push if you need to.

xiaods:
@spadgett sorry for the inconvenience,  all modified code is push to this PR.

spadgett:
Closing as this repo is now only for bug fixes to previous 3.x release. New development is happening at openshift/console.

/close

PR #2596, opened 2017-12-14
Fix deploy button on deployment page

benjaminapetersen:
So this update works, but the Q here is:

```JavaScript 
// the request payload to startLatestDeployment
apiVersion: "apps.openshift.io/v1",
force: true,
kind: "DeploymentRequest",
latest: true,
name: "mongodb"

// the resource, group, version
group: "apps.openshift.io",
resource: "deploymentconfigs/instantiate",
version: "v1"
```
Does it work because these two objects happen to be compatible, but I should further adjust this to ensure they are long term compatible?

### DIFF
--- a/app/scripts/services/deployments.js
+++ b/app/scripts/services/deployments.js
@@ -22,7 +22,7 @@ angular.module("openshiftConsole")
       // increase latest version by one so starts new deployment based on latest
       var req = {
         kind: "DeploymentRequest",
-        apiVersion: "v1",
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
update: spoke with @jwforres, this seems correct.

spadgett:
I don't think `deploymentConfig.apiVersion` always matches `deploymentConfigsInstantiateVersion`.

benjaminapetersen:
Closing this one & rolling it into the other w/the $filter to build the string.

PR #2597, opened 2017-12-14
Fix start build & deploy

benjaminapetersen:
Match up api version w/payload in start build.

depends on [web-common 268](https://github.com/openshift/origin-web-common/pull/268)

### DIFF
--- a/app/scripts/services/builds.js
+++ b/app/scripts/services/builds.js
@@ -42,11 +42,12 @@ angular.module("openshiftConsole")
       var buildType = isJenkinsPipelineStrategy(buildConfig) ? 'pipeline' : 'build';
       var req = {
         kind: "BuildRequest",
-        apiVersion: "v1",
+        apiVersion: APIService.toAPIVersion(buildConfigsInstantiateVersion),
         metadata: {
           name: buildConfig.metadata.name
         }
       };
+
       var context = {
         namespace: buildConfig.metadata.namespace
       };
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -161,6 +161,7 @@ angular.module("openshiftConsole")
 
         var linkSecretToServiceAccount = function(secret) {
           var updatedSA = angular.copy($scope.serviceAccounts[$scope.newSecret.pickedServiceAccountToLink]);
+          var rgv = APIService.objectToResourceGroupVersion(updatedSA);
           switch ($scope.newSecret.type) {
           case 'source':
             updatedSA.secrets.push({name: secret.metadata.name});
@@ -170,7 +171,7 @@ angular.module("openshiftConsole")
             break;
           }
 
-          DataService.update(serviceAccountsVersion, $scope.newSecret.pickedServiceAccountToLink, updatedSA, $scope).then(function(sa) {
+          DataService.update(rgv, $scope.newSecret.pickedServiceAccountToLink, updatedSA, $scope).then(function(sa) {
             // Show a single success message saying the secret was both created and linked.
             NotificationsService.addNotification({
               type: "success",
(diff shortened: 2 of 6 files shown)


app/scripts/services/builds.js
```diff
@@ -42,7 +42,7 @@ angular.module("openshiftConsole")
       var buildType = isJenkinsPipelineStrategy(buildConfig) ? 'pipeline' : 'build';
       var req = {
         kind: "BuildRequest",
-        apiVersion: "v1",
+        apiVersion: buildConfig.apiVersion,
```
spadgett: This should be the `apiVersion` built from `buildConfigsInstantiateVersion`. It's not necessarily the same version as `buildConfig.apiVersion`

spadgett:
> The following test failed....

Flake #1685

app/scripts/services/builds.js
benjaminapetersen: So that is interesting because if I change it to:

```JavaScript 
apiVersion: buildConfigsInstantiateVersion.version,
// which is this:
// {group: "build.openshift.io", version: "v1", resource: "buildconfigs/instantiate"}
```
it will error:

```
apiVersion: "v1"
code: 400
kind: "Status"
message: "the API version in the data (v1) does not match the expected API version (build.openshift.io/v1)"
metadata: {}
reason: "BadRequest"
status: "Failure"
```
So we need a helper fn to concat `buildConfigInstantiateVersion.group + '/' + buildConfigInstantiateVersion.version` to ensure these are correct?

benjaminapetersen:
Updated both start build & deploy here w/the filter to build the correct string.  I think this is where we want to be?

benjaminapetersen:
looking @ `  DeploymentsService.prototype.retryFailedDeployment` now, think that needs adjusting.

app/scripts/filters/resources.js
```diff
@@ -893,6 +893,11 @@ angular.module('openshiftConsole')
   .filter('kindToResource', function (APIService) {
     return APIService.kindToResource;
   })
+  .filter('groupVersion', function() {
+    return function(preferredVersion) {
+      return preferredVersion.group + '/' + preferredVersion.version;
```
spadgett:
```js
if (preferredVersion.group) {
  return preferredVersion.group + '/' + preferredVersion.version;
}

return preferredVersion.version;
```

app/scripts/filters/resources.js
```diff
@@ -893,6 +893,11 @@ angular.module('openshiftConsole')
   .filter('kindToResource', function (APIService) {
     return APIService.kindToResource;
   })
+  .filter('groupVersion', function() {
+    return function(preferredVersion) {
```
spadgett: s/preferredVersion/resourceGroupVersion

app/scripts/filters/resources.js
```diff
@@ -893,6 +893,11 @@ angular.module('openshiftConsole')
   .filter('kindToResource', function (APIService) {
     return APIService.kindToResource;
   })
+  .filter('groupVersion', function() {
```
spadgett: Probably better if this is an `APIService` method

app/scripts/filters/resources.js
benjaminapetersen: ah, that makes sense as well.

benjaminapetersen:
Updated & removed WIP.  This one should take care of the broken "start build" and "deploy".  I'd like to get it in and fix those two glaring issues & follow-up PR any other mismatches.  I believe others will work due to coincidence like we discussed, but fixes will be future-proofing.

benjaminapetersen:
hmm

spadgett:
@benjaminapetersen I'd rebase and rebuild

benjaminapetersen:
@spadgett had a weird git issue 

```
error: update_ref failed for ref 'refs/remotes/origin/trello/api-gr...': cannot lock ref ...
```
resolved now, pushing in a min

PR #2606, opened 2017-12-15
Migrate EventsController to EventsPage component

benjaminapetersen:
A fun one to leave everyone chewing on over the break :)

@TheRealJon's PR compelled me to do a quick test re:something I've been considering for a while. Most advocate going all in on components these days.  It is not too difficult (at least with simple controllers like this one) to migrate our controllers to a component.  Now that we have the build tools setup to watch the new `/components` directory, I think a slow migration in this direction could be a good thing. The newer component APIs are nice to work with.


benjaminapetersen:
@jwforres @spadgett @TheRealJon

spadgett:
https://docs.angularjs.org/guide/component#components-as-route-templates

benjaminapetersen:
Exactly.  Tho I think our `page` components are complex enough with `DataService` that I don't think we would benefit using the `resolve` the way that is suggested. I also tend to prefer simpler route files... but could be persuaded otherwise I'm sure.

PR #2607, opened 2017-12-15
Fix bug where pficons do not display

rhamilto:
Fixes https://github.com/openshift/origin/issues/17745

Requires https://github.com/openshift/origin-web-common/pull/270 and https://github.com/openshift/origin-web-catalog/pull/603

### DIFF
--- a/app/scripts/directives/fromFileDialog.js
+++ b/app/scripts/directives/fromFileDialog.js
@@ -23,6 +23,7 @@
     var ctrl = this;
     var annotation = $filter('annotation');
     var imageForIconClass = $filter('imageForIconClass');
+    var normalizeIconClass = $filter('normalizeIconClass');
 
     ctrl.$onInit = function() {
       ctrl.alerts = {};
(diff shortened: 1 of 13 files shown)


app/scripts/directives/fromFileDialog.js
```diff
@@ -38,7 +38,7 @@
 
     function getIconClass() {
       var icon = _.get(ctrl, 'template.metadata.annotations.iconClass', 'fa fa-clone');
-      return (icon.indexOf('icon-') !== -1) ? 'font-icon ' + icon : icon;
+      return icon.indexOf('icon', 0) === 0 ? 'font-icon ' + icon : icon;
```
spadgett: `_.startWith(icon, 'icon')` might read a little better. Do you have an example of what we're trying to match? It would be good to add a comment.

This is in enough places that it would be good to have a `normalizeIconClass` filter so the code is reused. I suspect it affects catalog, too.

spadgett: https://lodash.com/docs/4.17.4#startsWith

rhamilto: > _.startWith(icon, 'icon') might read a little better.

Agreed.  Was originally trying to minimize changes, but I don't see any reason not to use that.

Trying to match starts with "icon".  The existing code was matching "icon-*" and "pficon-*", hence the bug.

> This is in enough places that it would be good to have a normalizeIconClass filter so the code is reused. I suspect it affects catalog, too.

Ok.  Indeed [it does affect catalog](https://github.com/openshift/origin-web-catalog/pull/603).  And we were doing the check a couple different ways there.

PR #2610, opened 2017-12-18
Set Home Page Preference

dtaylor113:
image

image

Trello Card: https://trello.com/c/sg1xw7gG

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -35,7 +35,7 @@ angular
     'openshiftCommonUI',
     'webCatalog'
   ])
-  .config(function ($routeProvider) {
+  .config(function ($routeProvider, HomePagePreferenceServiceProvider) {
     var landingPageRoute;
     var projectsPageRoute = {
       templateUrl: 'views/projects.html',
@@ -56,7 +56,12 @@ angular
     }
 
     $routeProvider
-      .when('/', landingPageRoute)
+      .when('/', {
+        redirectTo: function() {
+          return HomePagePreferenceServiceProvider.$get().getHomePagePath();
+        }
+      })
+      .when('/catalog', landingPageRoute)
       .when('/create-project', {
         templateUrl: 'views/create-project.html',
         controller: 'CreateProjectController'
--- a/app/scripts/services/homePagePreference.js
+++ b/app/scripts/services/homePagePreference.js
@@ -1,0 +1,89 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .service("HomePagePreferenceService", function($location, $timeout, $q, $uibModal, AuthService, Logger, Navigate, NotificationsService){
+
+    function getHomePagePreferenceKey() {
+      return "openshift/home-page-pref/";
+    }
+
+    var getHomePagePreference = function() {
+      var homePagePref;
+      var key = getHomePagePreferenceKey();
+
+      try {
+        homePagePref = JSON.parse(localStorage.getItem(key));
+      } catch (e) {
+        Logger.error('Could not parse homePagePref as JSON', e);
+        return 'catalog-home';
+      }
+
+      return _.get(homePagePref, 'type', 'catalog-home');
+    };
+
+    var getHomePageProjectName = function() {
+      var homePagePref;
+      var key = getHomePagePreferenceKey();
+
+      try {
+        homePagePref = JSON.parse(localStorage.getItem(key));
+      } catch (e) {
+        Logger.error('Could not parse homePagePref as JSON', e);
+        return null;
+      }
+
+      return (homePagePref && homePagePref.type === 'project-overview') ? homePagePref.project : null;
+    };
+
+    var clear = function() {
+      var key = getHomePagePreferenceKey();
+      localStorage.removeItem(key);
+    };
+
+    var notifyInvalidProjectHomePage = function(projectName) {
+      clear();
+      NotificationsService.addNotification({
+        id: "invalid-home-page-preference",
+        type: "warning",
+        message: "Home page project not found.",
+        details: "Project " + projectName + " no longer exists or you do not have access to it.",
+        links: [{
+          href: "",
+          label: "Set Home Page",
+          onClick: function() {
+            $uibModal.open({
+              animation: true,
+              templateUrl: 'views/modals/set-home-page-modal.html',
+              controller: 'SetHomePageModalController'
+            });
+            return true;
+          }
+        }]
+      });
+    };
+
+    var getHomePagePath = function() {
+      var homePagePref = this.getHomePagePreference();
+      if (homePagePref === "project-overview") {
+        var homeProjectPageName = this.getHomePageProjectName();
+        return '/' + Navigate.projectOverviewURL(homeProjectPageName) + "?isHomePage=true";
+      } else if (homePagePref === "project-list") {
+        return '/projects';
+      } else {
+        return '/catalog';
+      }
+    };
+
+    var setHomePagePreference = function(homePagePref) {
+      var key = getHomePagePreferenceKey();
+      localStorage.setItem(key, JSON.stringify(homePagePref));
+    };
+
+    return {
+      getHomePagePreference: getHomePagePreference,
+      setHomePagePreference: setHomePagePreference,
+      getHomePageProjectName: getHomePageProjectName,
+      getHomePagePath: getHomePagePath,
+      notifyInvalidProjectHomePage: notifyInvalidProjectHomePage
+    };
+  });
--- a/app/styles/_set-home-page.less
+++ b/app/styles/_set-home-page.less
@@ -1,0 +1,12 @@
+.set-home-page {
+  .modal-body-indent {
+    padding-left: 48px;
+    .title {
+      font-weight: 600;
+    }
+  }
+  .select-project-container {
+    padding-left: 20px;
+    max-width: 360px;
+  }
+}
--- a/app/views/modals/set-home-page-modal.html
+++ b/app/views/modals/set-home-page-modal.html
@@ -1,0 +1,49 @@
+<div class="modal-resource-action set-home-page">
+  <div class="modal-header">
+    <h1>Set Home Page</h1>
+  </div>
+  <div class="modal-body modal-body-indent">
+    <p>
+      <div class="title">Set your preferred home page.</div>
+      Note: This setting is browser specific and will not be maintained across browsers.
+    </p>
+    <form name="homePageForm">
+      <fieldset ng-disabled="loading">
+        <div class="radio">
+          <label>
+            <input type="radio" ng-model="homePagePreference" value="catalog-home">
+            Catalog Home (Default)
+          </label>
+        </div>
+        <div class="radio">
+          <label>
+            <input type="radio" ng-model="homePagePreference" value="project-list">
+            My Projects List
+          </label>
+        </div>
+        <div class="radio" ng-show="availableProjects.length >= 1">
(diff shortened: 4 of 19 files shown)


app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,105 @@
...
+        homePagePref = JSON.parse(localStorage.getItem(key));
+      } catch (e) {
+        Logger.error('Could not parse homePagePref as JSON');
+        return 'catalog-home';
+      }
+
+      return _.get(homePagePref, 'type', 'catalog-home');
+    };
+
+    var getProjectOverviewPage = function() {
```
spadgett: nit: It's getting a project name, not an overview page (which to me implies a URL). Maybe

```js
var getHomePageProjectName = function() ...
```

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,105 @@
...
+              templateUrl: 'views/modals/set-home-page-modal.html',
+              controller: 'SetHomePageModalController'
+            });
+            return true;
+          }
+        }]
+      });
+    };
+
+    var getHomePageURL = function() {
```
spadgett: nit: getHomePagePath is more precise

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,105 @@
...
+      } else if (homePagePref === "project-list") {
+        return '/projects';
+      } else {
+        return '/catalog';
+      }
+    };
+
+    var setHomePagePreference = function(homePagePref) {
+      var key = getHomePagePreferenceKey();
+      if (!key) {
```
spadgett: `key` is always set

app/scripts/services/homePagePreference.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+      localStorage.setItem(key, JSON.stringify(homePagePref));
+    };
+
+    return {
+      getHomePagePreference: getHomePagePreference,
+      setHomePagePreference: setHomePagePreference,
+      getProjectOverviewPage: getProjectOverviewPage,
+      getHomePageURL: getHomePageURL,
+      notifyInvalidProjectHomePage: notifyInvalidProjectHomePage,
```
spadgett: trailing comma

app/styles/_set-home-page.less
```diff
@@ -0,0 +1,10 @@
+.modal-body-indent {
```
spadgett: These classes are very specific to this dialog and should be scoped to just the home page modal.

app/views/modals/set-home-page-modal.html
```diff
@@ -0,0 +1,50 @@
+<div class="modal-resource-action">
+  <div class="modal-header">
+    <h1>Set Home Page</h1>
+  </div>
+  <div class="modal-body modal-body-indent">
+    <p>
+      <div class="title">Set your preferred home page.</div>
+      Note: This setting is browser specific and will not be maintained across browsers.
+    </p>
+    <ng-form name="homePageForm">
```
spadgett: `<form>` instead of `<ng-form>`. `ng-form` should really only be used if it's nested inside another `form` element.

app/views/modals/set-home-page-modal.html
```diff
@@ -0,0 +1,50 @@
...
+        <div class="radio">
+          <label>
+            <input type="radio" ng-model="preferredHomePage" value="project-list">
+            My Projects List
+          </label>
+        </div>
+        <div class="radio" ng-show="availableProjects.length >= 1">
+          <label>
+            <input type="radio" ng-model="preferredHomePage" value="project-overview">
+            Overview Page for {{availableProjects.length === 1 ? (selectedProject | displayName) : ':'}}
```
spadgett: this will put an extra space in front of the `:`

app/views/modals/set-home-page-modal.html
```diff
@@ -0,0 +1,50 @@
...
+          </label>
+          <div class="select-project-container">
+            <select-project ng-if="availableProjects.length > 1"
```
spadgett: nit: if you move `ng-if` to the parent div it avoids the empty div

app/views/modals/set-home-page-modal.html
```diff
@@ -0,0 +1,50 @@
...
+                            class="select-project"
```
spadgett: Does this class have any styles? I don't see it in this PR.

app/views/modals/set-home-page-modal.html
```diff
@@ -0,0 +1,50 @@
...
+                            hide-label="true">
+            </select-project>
+          </div>
+        </div>
+      </fieldset>
+    </ng-form>
+  </div>
+  <div class="modal-footer">
+      <button class="btn btn-lg btn-primary" type="button"
+              ng-disabled="(preferredHomePage === 'project-overview' && homePageForm.$invalid) || homePageForm.$pristine"
```
spadgett: I think this can just be... ?

```html
ng-disabled="homePageForm.$invalid || homePageForm.$pristine"
```

spadgett: Nevermind I see why it's not possible.

It would be better to add an `ng-required` attribute to the `select-project` component and then use...

```html
<select-project
  ng-required="preferredHomePage === 'project-overview'"
  ....
```

Then you can change this disabled check to the above.

dtaylor113: Thanks for the tip, I hadn't used `ng-required` before...very useful :-)

app/views/modals/set-home-page-modal.html
dtaylor113: Removed, not needed

app/scripts/app.js
```diff
@@ -67,7 +72,7 @@ angular
       })
       .when('/project/:project', {
         redirectTo: function(params) {
-          return '/project/' + encodeURIComponent(params.project) + "/overview";
+          return '/project/' + encodeURIComponent(params.project) + "/overview" + (params.isHomePage ? "?isHomePage=true" : "");
```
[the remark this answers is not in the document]
dtaylor113: Removed, it was not needed  -thanks

dtaylor113:
Hi @spadgett, thanks for the great review.  I believe I have addressed your last round of comments.

spadgett:
@openshift/team-ux-review Since we've changed clicking on the logo to always go to the landing page, there is no direct link to your preferred home page in the UI. You literally have to type in "/console/" as the URL in the browser location bar. Should we have a way to navigate to your home page somewhere in the UI?

cc @jwforres

dtaylor113:
> @openshift/team-ux-review Since we've changed clicking on the logo to always go to the landing page, there is no direct link to your preferred home page in the UI. You literally have to type in "/console/" as the URL in the browser location bar. Should we have a way to navigate to your home page somewhere in the UI?

Good question, cc'ing @beanh66

dtaylor113:
Hi, I also just noticed:

image

Since home page is now configurable to places other than the Catalog, I think this Help menu option should change to something like 'Tour Catalog".   I also noticed that selecting `Help -> Tour Home Page` doesn't do anything if you are on a page other than the Browse Catalog landing page, but I think I can fix that.

spadgett:
>  I also noticed that selecting Help -> Tour Home Page doesn't do anything if you are on a page other than the Browse Catalog landing page, but I think I can fix that.

Yeah seems to work in master, so it is due to the redirect. We should probably sweep the code for anywhere else we are linking to `./`

spadgett:
All of these need to be updated.

```
app/scripts/services/navigate.js
94:          return './';

app/scripts/controllers/membership.js
306:                      $location.url("./");

app/views/overview/_service-bindings.html
22:      <a href="./">Browse Catalog</a>

app/scripts/extensions/nav/helpDropdown.js
23:                node: '<li><a href="./?startTour=true">Tour Home Page</a></li>'
```

dtaylor113:
Updated those references to `./`. Also changed `Help -> Tour Home Page` to:

image

Since that's what we call it in the Set Home Page modal:
image

ncameronbritt:
@spadgett The designers chatted about the need for another way to get to the homepage and we feel like things are ok as is. Our feeling is that if the user has changed their homepage, they can still to navigate to whatever page they selected within a couple of clicks, regardless of where they are in the console.

spadgett:
> The designers chatted about the need for another way to get to the homepage and we feel like things are ok as is. Our feeling is that if the user has changed their homepage, they can still to navigate to whatever page they selected within a couple of clicks, regardless of where they are in the console.

It's not so much about getting back to your homepage. My question is how you get to your homepage *at all*?

Presumably someone sends you a link to the console the first time you use it. Homepage is implemented as a redirect currently, so you will be redirected to the landing page, the default homepage, after logging in the first time. The URL changes to `/console/catalog`. If you create a bookmark now, it will always go to the catalog even if you change your homepage. You're bookmarking `/console/catalog`, the catalog page, and not `/console`, your homepage. You have no way to get back to your selected homepage without finding that first link someone sent you.

If we want this to work differently, we can't use redirects. We might need to revisit our implementation (sorry @dtaylor113 this would be my fault).

I guess my question is still if you have a browser bookmark anyway, why not just bookmark the page you want?

dtaylor113:
> We might need to revisit our implementation

@spadgett, would it be possible to do this work in a follow up PR?

spadgett:
> would it be possible to do this work in a follow up PR?

I think even a link that bookmarks your homepage or goes to your homepage would make it more useful without needing change the implementation.

dtaylor113:
> I think even a link that bookmarks your homepage or goes to your homepage would make it more useful without needing change the implementation.

Do you mean updating app.js to include something like `/my-home-page', which would do the same as  `/` does, ie. `return HomePagePreferenceServiceProvider.$get().getHomePagePath();`?

dtaylor113:
I'm working on getting the integration tests to work on my Linux machine to debug the issue

spadgett:
@dtaylor113 can you check create from URL with your changes?

https://docs.openshift.org/latest/dev_guide/create_from_url.html

dtaylor113:
> @dtaylor113 can you check create from URL with your changes?

Ok, will try that.

> Jenkins/Integration Tests failure. 

I'm seeing this failure on my system:
```
visiting url: create-project
visiting url: project/template-dumpster/create

/home/dtaylor/Dev/repos/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:1654
      var result = fn();
                   ^
Error: URL hasn't changed to project/template-dumpster/create(is currently: undefined)
Wait timed out after 5190ms
    at /home/dtaylor/Dev/repos/openshift/origin-web-console/node_modules/selenium-webdriver/lib/webdriver/promise.js:1425:29
```

dtaylor113:
> @dtaylor113 can you check create from URL with your changes?
> 
> https://docs.openshift.org/latest/dev_guide/create_from_url.html
> 
> Those are the tests that are failing.

I used `https://localhost:9000/dev-console/create?imageStream=nodejs&imageTag=4&name=nodejs&sourceURI=https:%2F%2Fgithub.com%2Fopenshift%2Fnodejs-ex.git&sourceRef=master&contextDir=%2F` and go to this point:

image

`the API version in the data (v1) does not match the expected API version (project.openshift.io/v1)` ?

dtaylor113:
@spadgett, @benjaminapetersen,  PR passed Jenkis and Travis tests.  Also the 'Create From URL' test worked in local testing.

spadgett:
Thanks @dtaylor113

PR #2617, opened 2018-01-02
Make warning message more accurate when adding invalid secret/configMap

jhadvig:
This fix is based from https://bugzilla.redhat.com/show_bug.cgi?id=1529467 where we suggested that best fix would be just warn user that the envVar is invalid.
@jwforres @spadgett PTAL

### DIFF
--- a/app/views/directives/add-config-to-application.html
+++ b/app/views/directives/add-config-to-application.html
@@ -39,7 +39,7 @@
           <div class="has-warning" ng-if="ctrl.hasInvalidEnvVars">
             <div class="help-block">
               <span class="pf-icon pficon-warning-triangle-o" aria-hidden="true"></span>
-              Some of the keys for {{ctrl.apiObject.kind | humanizeKind}} <strong>{{ctrl.apiObject.metadata.name}}</strong> are not valid environment variable names and will not be added.
+              {{ctrl.apiObject.kind | humanizeKind | upperFirst}} <strong>{{ctrl.apiObject.metadata.name}}</strong> contains keys that are not valid environment variable names. Only {{ctrl.apiObject.kind | humanizeKind}} keys with valid names will be added as environment variables.
             </div>
           </div>
           <div class="form-group">
--- a/app/views/directives/edit-environment-from.html
+++ b/app/views/directives/edit-environment-from.html
@@ -62,7 +62,8 @@
             </ui-select>
           </div>
           <div class="has-warning" ng-if="$ctrl.hasInvalidEnvVar(entry.selectedEnvFrom.data)">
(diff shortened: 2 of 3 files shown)


app/views/directives/add-config-to-application.html
```diff
@@ -39,7 +39,7 @@ <h3>Add to Application</h3>
           <div class="has-warning" ng-if="ctrl.hasInvalidEnvVars">
             <div class="help-block">
               <span class="pf-icon pficon-warning-triangle-o" aria-hidden="true"></span>
-              Some of the keys for {{ctrl.apiObject.kind | humanizeKind}} <strong>{{ctrl.apiObject.metadata.name}}</strong> are not valid environment variable names and will not be added.
+              Some of the keys for {{ctrl.apiObject.kind | humanizeKind}} <strong>{{ctrl.apiObject.metadata.name}}</strong> are not valid environment variable names.
```
spadgett: The message was saying just the invalid keys won't be added, not that the secret won't be added. But maybe the wording was confusing.

I think we still need to tell the user what will happen. How about...?

```html
Some of the keys for {{ctrl.apiObject.kind | humanizeKind}} <strong>{{ctrl.apiObject.metadata.name}}</strong> are not valid environment variable names.
Only {{ctrl.apiObject.kind | humanizeKind}} keys with valid names will be added as environment variables.
```

cc @jwforres

jhadvig: that sounds good to me :+1:

jwforres: So I like @spadget's message better generally than what is there.  My point in the bug is that we say "Some" even when its really "All"

jwforres: If we say

```
{{ctrl.apiObject.kind | humanizeKind}} <strong>{{ctrl.apiObject.metadata.name}}</strong> contains keys that are not valid environment variable names.
Only {{ctrl.apiObject.kind | humanizeKind}} keys with valid names will be added as environment variables.
```

then one message works generally

jwforres: will need to update that first humanized string so the first letter is capitalized, i.e. "Config map"

spadgett: That message is better 👍 

 Will need `upperFirst` to capitalize the first word in the sentence.

```html
{{ctrl.apiObject.kind | humanizeKind | upperFirst}} <strong>{{ctrl.apiObject.metadata.name}}</strong> ...
```

jwforres: IIRC the capitalize option on humanizeKind filter capitalizes both words?  if so then you can just wrap this humanizeKind filter call in parens and pass it to _.capitalize

spadgett: @jwforres We were making the same comment at the same time :)

I think `upperFirst` filter should do what we want.

PR #2621, opened 2018-01-02
Fix incorrect apiGroup for roles causing dupes on membership page

### LINKED ISSUE #2619: We have a second "admin" role appearing in the roles list
This is new.  Not sure why I can see two 'admin' roles.

screen shot 2018-01-02 at 1 24 18 pm

@enj anything change with roles recently to cause a second admin?

benjaminapetersen:
Fixes #2619 

I'm surprised the page worked & the bug was this subtle with this error. 

@jwforres @spadgett

### DIFF
--- a/app/scripts/services/membership/roles.js
+++ b/app/scripts/services/membership/roles.js
@@ -7,15 +7,15 @@ angular
     APIService,
     DataService) {
 
-    var roleBindingsVersion = APIService.getPreferredVersion('rolebindings');
-    var clusterRoleBindingsVersion = APIService.getPreferredVersion('clusterroles');
+    var rolesVersion = APIService.getPreferredVersion('roles');
+    var clusterRolesVersion = APIService.getPreferredVersion('clusterroles');
 
     var listAllRoles = function(context) {
       return $q.all([
         DataService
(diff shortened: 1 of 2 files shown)


app/scripts/services/membership/roles.js
```diff
@@ -7,15 +7,15 @@ angular
     APIService,
     DataService) {
 
-    var roleBindingsVersion = APIService.getPreferredVersion('rolebindings');
-    var clusterRoleBindingsVersion = APIService.getPreferredVersion('clusterroles');
+    var rolesVersion = APIService.getPreferredVersion('roles');
```
benjaminapetersen: The only change is the top one, I just renamed the rest of the variables as they were not clear (prob led to this sliding by).

enj:
How did this ever work?

spadgett:
cluster roles were correct at least

benjaminapetersen:
@enj I know right?!  Far as I can tell, the roles list on this page is really only needed for the string names to populate the dropdown & happen to match up well enough to fool a human (me).  Since the underlying objects have such similar structure, no code broke.  It took a quirk to notice that things weren't quite right.

PR #2622, opened 2018-01-02
Check if user can add to existing project during create from url

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1530162

If user cannot add to project, add error message on select of project
screen shot 2018-01-02 at 3 50 08 pm

### DIFF
--- a/app/views/create-from-url.html
+++ b/app/views/create-from-url.html
@@ -15,24 +15,9 @@
               </div>
               <div ng-if="!noProjects && !canCreateProject">
                 <h2>Choose a Project</h2>
-                <ui-select ng-model="selected.project">
-                  <ui-select-match placeholder="Project name">
-                    {{$select.selected | displayName}}
-                  </ui-select-match>
-                  <ui-select-choices repeat="project in projects | searchProjects : $select.search">
-                    <span ng-bind-html="project | displayName | highlightKeywords : $select.search"></span>
-                    <span ng-if="project | displayName : true" class="small text-muted">
-                      &ndash;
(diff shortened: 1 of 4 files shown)


app/views/create-from-url.html
```diff
@@ -15,24 +15,9 @@ <h2>Create a New Project</h2>
...
-                    </span>
-                  </ui-select-choices>
-                </ui-select>
-              </div>
-              <div ng-if="!noProjects && canCreateProject">
-                <uib-tabset>
-                  <uib-tab>
-                    <uib-tab-heading>Choose Existing Project</uib-tab-heading>
-                    <ui-select ng-model="selected.project">
+                <ng-form name="forms.selectProjectForm">
```
spadgett: Use `<form>` here instead of `<ng-form>` if it's not already nested in another form element.

app/views/create-from-url.html
```diff
@@ -44,15 +29,57 @@ <h2>Choose a Project</h2>
...
+                      <a class="btn btn-default" href="#" back>Cancel</a>
                     </div>
+                  </div>
+                </ng-form>
+              </div>
+              <div ng-if="!noProjects && canCreateProject">
+                <uib-tabset>
+                  <uib-tab>
+                    <uib-tab-heading>Choose Existing Project</uib-tab-heading>
+                    <ng-form name="forms.selectProjectForm">
```
spadgett: `<form>`

rhamilto:
@spadgett's comments addressed.

PR #2627, opened 2018-01-03
Add titles to roles in dropdown list on membership page

### LINKED ISSUE #2614: Membership Add Role cannot see dropdown contents
When attempting to add a another role in openshift online membership screen you cannot see the description of the roles.

image

benjaminapetersen:
Partial Fix for 2614, at least as a short term until we decide on a better solution.  See the conversation for more details.

Adds titles to the menu items:

screen shot 2018-01-03 at 11 18 25 am

Re: #2614

### DIFF
--- a/app/views/membership.html
+++ b/app/views/membership.html
@@ -125,7 +125,7 @@
                       ng-model="subject.newRole"
                       theme="bootstrap"
                       search-enabled="true"
-                      title="Select a new role for {{subjectKind.name}}"
+                      title="Select a new role for {{subject.name}}"
                       class="select-role"
                       flex>
                       <ui-select-match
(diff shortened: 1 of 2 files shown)


app/views/membership.html
```diff
@@ -125,7 +125,7 @@ <h3>
                       ng-model="subject.newRole"
                       theme="bootstrap"
                       search-enabled="true"
-                      title="Select a new role for {{subjectKind.name}}"
+                      title="Select a new role for {{subject.name}}"
```
benjaminapetersen: Also, this change is clearer:

screen shot 2018-01-03 at 11 21 51 am

app/views/membership.html
```diff
@@ -134,8 +134,12 @@ <h3>
                       </ui-select-match>
                       <ui-select-choices
                         repeat="role as role in filteredRoles | filter: excludeExistingRoles(subject.roles) | filter: { metadata: { name: $select.search } } track by (role | uid)">
-                        <div ng-bind-html="role.metadata.name | highlight: $select.search"></div>
-                        <div ng-if="role | annotation : 'description'">
+                        <div
+                          ng-bind-html="role.metadata.name | highlight: $select.search"
+                          title="{{role.metadata.name}}"></div>
```
spadgett: Can we specify this once on the parent node instead of once on each child?

dist/scripts/vendor.js
```diff
@@ -73402,7 +73402,7 @@ resource: "deploymentconfigs/log"
 },
 endpoints: {
 version: "v1",
-resource: "endpoints"
+resoource: "endpoints"
```
spadgett: bower install

benjaminapetersen:
I'll update the comment so it doesn't close the issue.

app/views/membership.html
benjaminapetersen: Tried that, but you can't put it directly on the `ui-select-choices`.  So I could instead add a new `div`, but then we have a `div` just for a `title`, which isn't awesome either.

dist/scripts/vendor.js
benjaminapetersen: yup got it

PR #2629, opened 2018-01-04
[DO NOT MERGE] Remove CRO handling from console

spadgett:
@jwforres @smarterclayton This needs more testing, but it's about how much code we could remove if we got rid of cluster resource override handling in the console.

/hold


jwforres:
fyi @pweil-

PR #2630, opened 2018-01-04
[WIP] Fix bug 1529045: editing buildConfig env vars always displays banner

benjaminapetersen:
RE: [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1529045)

Initial fix is the `$setPristine()` in the `BuildConfig` controller.  I was thinking this might be all we need, if the save is successful, clear the latest version of the form & trust the watch will update the variables.

However, @spadgett pointed me at [this code](https://github.com/spadgett/origin-web-console/blob/c44c43e51863cd120d1b6af6e216f22ba46168ab/app/scripts/directives/editEnvironmentVariables.js#L43-L69) from `editEnvironmentVariables.js`.  Handling a merge of updates like we do elsewhere would be ideal.

So, in attempting to use the `component`, I had to do a couple hacks.  Pausing here to gather feedback and see if we want to continue down this direction.

- I had to edit `EnvironmentService.getContainers` to return the `buildStrategy` for a `buildConfig` to get at the envs.  Will have to do some work here to update this fn since this is not container data.
- Now the component appears, but 
    - it shows a container name 
    - it shows an `environmentFrom` editor. 

Here is a screenshot of the current state of hackery.  Is it worth it/the correct fix to continue & try to get this component functional for this use case?

screen shot 2018-01-04 at 5 19 25 pm

@jwforres @spadgett

### DIFF
--- a/app/scripts/services/images.js
+++ b/app/scripts/services/images.js
@@ -1,7 +1,13 @@
 'use strict';
 
 angular.module("openshiftConsole")
-  .factory("ImagesService", function($filter, ApplicationGenerator, DataService) {
+  .factory("ImagesService", function(
+    $filter,
+    APIService,
+    ApplicationGenerator,
+    DataService) {
+
+    var imageStreamImportsVersion = APIService.getPreferredVersion('imagestreamimports');
 
     // maps an env object: { key: 'val', key2: 'val2'}
     // to an array: [{},{},{}]
@@ -33,7 +39,7 @@ angular.module("openshiftConsole")
         status: {}
       };
 
-      return DataService.create('imagestreamimports',
+      return DataService.create(imageStreamImportsVersion,
                                 null,
                                 importImage,
                                 projectContext).then(function(response) {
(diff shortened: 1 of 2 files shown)


app/scripts/services/environment.js
```diff
@@ -8,6 +8,11 @@ angular.module("openshiftConsole")
       if (apiObject.kind === 'Pod') {
         return _.get(apiObject, 'spec.containers', []);
       }
+      // TODO: this isn't actually containers....
```
benjaminapetersen: So this just gets the component to display.  However, its incorrect.

If this component is the way to go, I'd prob go back up into the `editEnvironmentVariables` component & test for `BuildConfig`, then alter the display/template.  Looking into this now.

benjaminapetersen:
`$setPristine()` was already set in the `saveEnvVars()` fn.  Removing it + the hacks to `edit-environment-variables` results in an empty PR, so this was automatically closed.  Investigating alternatives...

benjaminapetersen:
Working out the details of this bug.  

1.  Adding & moving while kicking off new builds:



2.  Complains when there is partial text in a text box when you start a build:

PR #2637, opened 2018-01-08
Whitelist service catalog resources for templates

spadgett:
Don't warn when users instantiate a template with service instances or
service bindings.

/kind bug
/assign @jwforres


spadgett:
@jwforres bump :)

PR #2638, opened 2018-01-08
[WIP] Fix bug 1529045: Editing buildConfig env vars can incorrectly display error messages

benjaminapetersen:
This bug appears to be a race condition associated with slow internet connections.  The `update` response and `webSocket` data may hit the browser inconsistently, causing the reported problem.

The fix here is to immediately declare the form `$pristine` and submit the update, backing out & setting the form back to `$dirty` only if the update fails.

### DIFF
--- a/app/scripts/controllers/buildConfig.js
+++ b/app/scripts/controllers/buildConfig.js
@@ -98,6 +98,7 @@ angular.module('openshiftConsole')
       NotificationsService.hideNotification("save-bc-env-error");
       $scope.envVars = _.filter($scope.envVars, 'name');
       buildStrategy($scope.updatedBuildConfig).env = keyValueEditorUtils.compactEntries(angular.copy($scope.envVars));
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
Chatting w @spadgett about this.  The proposed fix may introduce a different race condition, giving it a bit more thought.  Per discussion, we may be better tracking the request in flight with a `requestInFlight` variable, rather than `$pristine` on the env editor form.  Then we need to work out how to handle updates from the websocket & what to do if the request fails.

This is more of an edge case fix PR for the current release and doesn't include the [merge env vars](https://github.com/spadgett/origin-web-console/blob/c44c43e51863cd120d1b6af6e216f22ba46168ab/app/scripts/directives/editEnvironmentVariables.js#L43-L69) or `editEnvironmentVariables` component that we probably need long term.

benjaminapetersen:
Added WIP as we aren't quite sure this is the correct approach yet.

PR #2640, opened 2018-01-08
Handle new build webhooks that use secretRefs instead of inline secrets

jhadvig:
Adding functionality to handle the new build webhooks that use secretRefs instead of inline secrets.
Now when user will want to add webhook to build config he will have to choose from available secrets that contain the `WebHookSecretKey` key.
Screen:
a032d8ea-df07-4192-yd08-fb5c7e9ffd9f

To be consistent with changes in CLI (introduced in https://github.com/openshift/origin/pull/17314) I've updated the `webhookURL` so it will show webhook urls as:
`https://localhost:9000/namespaces/myproject/buildconfigs/test-build/webhooks/<secret>/github`

Wasn't really sure if part of this [card](https://trello.com/c/lo6HwQgu/1106-2-build-webhooks-with-secretrefs) is also update the create-secret page to be able to create secrets that user could add to the webhook in the BC editor. 

@spadgett PTAL


jhadvig:
cc'ing @jwforres

jwforres:
I think its reasonable to allow them to pick an existing secret, but I think we definitely need a really simple "generate one for me" option.

jwforres:
Make sure you still handle if they have existing webhooks using the old embedded secret format. Don't let them add them that way, but don't break if they exist. They should be able to see and remove those webhooks.

Also this affects more than just the BC editor. You need to worry about the app generation code when you create an app from a builder image, that creates webhooks in the advanced settings (the code exists in two places, in both the catalog and in the console).  It might be better to handle app generation as a separate PR though.

jhadvig:
@jwforres I've update the PR.
In the create secret page I've added option to create webhook secret reference, that user can afterwards feed to the webhook trigger in the BS editor.
secref2


Also in the BC editor I've extracted the adding webhook trigger logic into a separate directive `oscWebhookTriggers` (similar to `oscSecrets` && `oscSourceSecrets`), so the user can create new webhook secret reference directly in editor if needed.
secref1


Already working on the app generation code as you mentioned in https://github.com/openshift/origin-web-console/pull/2640#issuecomment-356117314. Shall I open a separate PR or put it here ?

jwforres:
for creating a webhook secret key, it would be nice if there was a "generate" button next to the field similar to what some password managers do, that generates a value for you simillar to what we used to generate for webhooks before.  Since a lot of folks just want a random token in there.

jhadvig:
@jwforres I've update the PR
Added the 'Generate' btn:
generate

Also added logic to warn user that the URL is incomplete, since user doesnt have permissions to list Secrets. For that I've wrapped the `copy-to-clipboard` directive that we ued with the `webhookURL` filter into new `copyWebhookUrl` that contains the warning, to make it more clean and dont repeat the code.
In case the webhook trigger contains the old plain secret the webhook will be assembled as before.
cantshowwebhook

jwforres:
So I see the alert where you tried to fetch the secrets, you should only
try to fetch the secrets if the canI list secrets check passes.

jhadvig:
@jwforres added the `canI` check before fetching the secrets

spadgett:
Any reason not to just generate the secret name and webhook key instead of making the user pick?

jwforres:
@spadgett you might actually want to re-use a webhook secret across different build configs

ncameronbritt:
Per our discussion on IRC:
screen shot 2018-01-10 at 12 03 21 pm

jhadvig:
@jwforres I've update the PR per @ncameronbritt comment with the visual.
Here is a simple demo:
secref3


PTAL

ncameronbritt:
Looks good @jhadvig!

app/scripts/directives/oscWebhookTriggers.js
```diff
@@ -0,0 +1,153 @@
...
+          // After removing trigger check check other triggers for duplicates.
+          _.each($scope.webhookTriggers, function(trigger) {
+            if (trigger.isDuplicit) {
+              duplicityCheck(trigger);
+            }
+          })
+        };
+
+        // Check if new or modified webhook trigger is a duplicate. If so, show a warning under the appropriate trigger
+        var duplicityCheck = function(trigger) {
```
spadgett: checkIfDuplicate

app/scripts/directives/oscWebhookTriggers.js
```diff
@@ -0,0 +1,153 @@
...
+              type: ""
+            };
+          } else {
+            $scope.webhookTriggers.splice(index,1);
+          }
+          $scope.form.$setDirty();
+
+          // After removing trigger check check other triggers for duplicates.
+          _.each($scope.webhookTriggers, function(trigger) {
+            if (trigger.isDuplicit) {
```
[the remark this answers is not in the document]
spadgett: You always need to check duplicates since the `isDuplicate` will be false the first time it appears.

```
trigger A
trigger B (duplicate of A)
```

We need to unflag trigger B if trigger A is removed.

app/scripts/directives/oscWebhookTriggers.js
```diff
@@ -0,0 +1,153 @@
...
+              duplicityCheck(trigger);
+            }
+          })
+        };
+
+        // Check if new or modified webhook trigger is a duplicate. If so, show a warning under the appropriate trigger
+        var duplicityCheck = function(trigger) {
+          var matchingTriggers = _.filter($scope.webhookTriggers, function(wehbookTrigger) {
+            return _.isEqual(wehbookTrigger.data, trigger.data);
```
spadgett: Does `data` include the secret? Do we expect to have duplicates often?

I'm wondering if we really need to be checking duplicates. @jwforres opinion?

app/scripts/directives/oscWebhookTriggers.js
```diff
@@ -0,0 +1,153 @@
...
+          });
+
+          if (matchingTriggers.length > 1) {
+            _.first(matchingTriggers).isDuplicit = false;
```
spadgett:
```js
// Mark all except the first as duplicates.
_.each(matchingTriggers, function(trigger, i) {
  var first = i === 0;
  trigger.isDuplicate = !first;
});
```

app/scripts/directives/oscWebhookTriggers.js
```diff
@@ -0,0 +1,153 @@
...
+        $scope.triggerTypeChange = function(trigger) {
+          var lastTriggerType = _.toLower(trigger.lastTriggerType);
+          var newTriggerType = _.toLower(trigger.data.type);
+
+          trigger.data = _.mapKeys(trigger.data, function(value, key) {
+            if (key === lastTriggerType) {
+              return newTriggerType;
+            }
+            return key;
+          });
```
spadgett: I think we're getting fancier than we need to.

```js
trigger.data[newTriggerType] = trigger.data[lastTriggerType];
delete trigger.data[lastTriggerType];
trigger.lastTriggerType = trigger.data.type;
```

app/scripts/directives/oscWebhookTriggers.js
```diff
@@ -0,0 +1,153 @@
...
+              type: ""
+            }
+          });
+        };
+
+        // Check last trigger if it's type and secret are selected.
+        $scope.checkLastAndAddNew = function() {
+          var lastTrigger = _.last($scope.webhookTriggers);
+          var lastTriggerSecretData = $filter('getWebhookSecretData')(lastTrigger);
+          if (lastTrigger.data.type && (lastTriggerSecretData.secret || lastTriggerSecretData.secretReference.name)) {
```
spadgett: Isn't this a runtime error if secretReference isn't defined?

app/scripts/directives/oscWebhookTriggers.js
```diff
@@ -0,0 +1,153 @@
...
+          });
+          trigger.lastTriggerType = trigger.data.type;
+          duplicityCheck(trigger);
+        };
+
+        $scope.triggerSecretChange = function(trigger) {
+          duplicityCheck(trigger);
+        };
+
+        var addWebhookTrigger = function() {
```
spadgett: I'd probably name this `addEmptyWebhookTrigger`

app/scripts/directives/util.js
```diff
@@ -57,6 +57,28 @@ angular.module('openshiftConsole')
       }
     };
   })
+  .directive('copyWebhookUrl', function() {
```
spadgett: Should be a component.

app/scripts/filters/resources.js
```diff
@@ -172,16 +172,30 @@ angular.module('openshiftConsole')
...
-      });
+  .filter('webhookURL', function(DataService, SecretsService) {
+    return function(buildConfig, type, project, secret, webhookSecrets) {
+      secret = SecretsService.getWebhookSecretValue(secret, webhookSecrets);
+      if (secret) {
+        return DataService.url({
+          // arbitrarily many subresources can be included
+          // url encoding of the segments is handled by the url() function
+          // subresource segments cannot contain '/'
+          resource: "buildconfigs/webhooks/" + secret + "/" + type.toLowerCase(),
```
spadgett: encodeURIComponent(secret)

spadgett: Let's also encode the type to be safe

app/scripts/filters/util.js
```diff
@@ -322,6 +322,11 @@ angular.module('openshiftConsole')
       return limitToFilter(input, limit);
     };
   })
+  .filter("getWebhookSecretData", function() {
+    return function(webhookTrigger) {
+      return webhookTrigger.data[_.toLower(webhookTrigger.data.type)];
```
spadgett: Probably better to use _.get to avoid errors if webhookTrigger is not defined

```js
var type = _.get(webhookTrigger, 'data.type');
if (!type) {
  return null;
}
return _.get(webhookTrigger, ['data', _.toLower(type)]);
```

app/scripts/services/secrets.js
```diff
@@ -141,8 +145,22 @@ angular.module("openshiftConsole")
       return decodedSecret;
     };
 
+    var getWebhookSecretValue = function(secret, webhookSecrets) {
+      // 'secretReference' field is an indicator the the webhook is using SecretRef
+      // otherwise the secret is in the depricated plain form.
```
spadgett: typo: deprecated

app/views/browse/build-config.html
```diff
@@ -386,28 +386,28 @@ <h3>Triggers <a href="{{'build-triggers' | helpLink}}" target="_blank"><span cla
...
-                                <copy-to-clipboard clipboard-text="buildConfig.metadata.name | webhookURL : trigger.type : trigger.bitbucket.secret : project.metadata.name"></copy-to-clipboard>
+                                <copy-webhook-url build-config-name="buildConfig.metadata.name" trigger-type="trigger.type" project-name="project.metadata.name" secret="trigger.bitbucket" webhook-secrets="webhookSecrets"></copy-webhook-url>
                               </dd>
                           </div>
                           <div ng-switch-when="GitHub">
                               <dt>GitHub Webhook URL:
                               </dt>
-                              <dd>
-                                <copy-to-clipboard clipboard-text="buildConfig.metadata.name | webhookURL : trigger.type : trigger.github.secret : project.metadata.name"></copy-to-clipboard>
+                              <dd class="with-warning">
```
spadgett: I don't see where `with-warning` styles are added. Why only for GitHub webhooks?

app/views/directives/create-secret.html
```diff
@@ -47,268 +47,296 @@
...
       <div class="form-group">
-        <label for="username">Username</label>
-        <div>
+        <label for="webhookSecretKey" class="required">Webhook Secret Key</label>
+        <div class="input-group">
           <input class="form-control"
-            id="username"
-            name="username"
-            ng-model="newSecret.data.username"
+            id="webhookSecretKey"
```
spadgett: nit: we usually kebab case ids (webhook-secret-key)

app/views/directives/create-secret.html
```diff
@@ -47,268 +47,296 @@
...
-        Password or token for Git authentication. Required if a ca.crt or .gitconfig file is not specified.
-      </div>
-
-      <div class="form-group">
-        <div class="checkbox">
-          <label>
-            <input type="checkbox" ng-model="add.cacert">
-            Use a custom ca.crt file
-          </label>
+            aria-describedby="username-help"
```
spadgett: username-help ?

app/views/directives/create-secret.html
```diff
@@ -47,268 +47,296 @@
...
-          rendererOptions: {
-            fadeFoldWidgets: true,
-            showPrintMargin: false
-          }
-        }" ng-model="newSecret.data.privateKey" class="create-secret-editor ace-bordered" id="private-key-editor" required></div>
-        <div class="help-block">
-          Private SSH key file for Git authentication.
+        <div class="help-block" id="secret-name-help">
+          Value of the secret will be supplied when invoking the webhook.
+          <a href="{{'webhooks' | helpLink}}" target="_blank"><span class="learn-more-inline">Learn More&nbsp;<i class="fa fa-external-link" aria-hidden="true"></i></span></a>
```
spadgett: nit: ng-href

app/views/directives/next-steps.html
```diff
@@ -89,7 +89,7 @@ <h2 class="h3">Making code changes</h2>
         Your source does not appear to be a URL to a GitHub repository.  If you have a GitHub repository that you want to trigger this build from then use the following payload URL and specifying a <i>Content type</i> of <code>application/json</code>:
       </span>
     </p>
-    <copy-to-clipboard clipboard-text="$ctrl.createdBuildConfig.metadata.name | webhookURL : trigger.type : trigger.github.secret : $ctrl.projectName"></copy-to-clipboard>
+    <copy-to-clipboard clipboard-text="$ctrl.createdBuildConfig.metadata.name | webhookURL : trigger.type :$ctrl.projectName : trigger.github"></copy-to-clipboard>
```
spadgett: Should we be passing secrets here?

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
+<ng-form name="secretsForm" class="add-webhook">
+  <div ng-repeat="trigger in webhookTriggers">
+    <div class="add-webhook-row">
+      <ui-select 
+        ng-model="trigger.data.type"
+        ng-disabled="isDeprecated(trigger)"
+        on-select="triggerTypeChange(trigger)"
+        on-remove="test()"
```
spadgett: test?

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
...
+        </ui-select-choices>
+      </ui-select>
+
+      <ui-select
+        ng-if="!isDeprecated(trigger)"
+        ng-model="trigger.data[trigger.data.type.toLowerCase()].secretReference.name"
+        ng-disabled="!trigger.data.type"
+        ng-required="trigger.data.type"
+        on-select="triggerSecretChange(trigger)"
+        title="Select a webhooke secret reference"
```
spadgett: typo: webhook

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
+<ng-form name="secretsForm" class="add-webhook">
+  <div ng-repeat="trigger in webhookTriggers">
+    <div class="add-webhook-row">
+      <ui-select 
+        ng-model="trigger.data.type"
+        ng-disabled="isDeprecated(trigger)"
+        on-select="triggerTypeChange(trigger)"
+        on-remove="test()"
+        search-enabled="false"
+        title="Select a webhooke type"
```
spadgett: typo: webhook

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
...
+        ng-disabled="!trigger.data.type"
+        ng-required="trigger.data.type"
+        on-select="triggerSecretChange(trigger)"
+        title="Select a webhooke secret reference"
+        class="select-secret-ref"
+        flex>
+        <ui-select-match placeholder="Webhook secret reference">
+          {{ $select.selected.metadata.name }}
+        </ui-select-match>
+        <ui-select-choices repeat="webhookSecret.metadata.name as webhookSecret in (webhookSecrets | filter : $select.search)">
```
spadgett: track by (webhookSecret | uid)

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
...
+          disabled>
+        <span class="input-group-btn">
+          <a
+              role="button"
+              href=""
+              class="btn btn-default"
+              title="Toggle Password Visibility"
+              aria-label="Toggle Password Visibility"
+              ng-click="togglePassword($index)">
+            <i class="glyphicon glyphicon-eye-open"></i>
```
spadgett: aria-hidden="true"

We probably shouldn't use `glyphicon`

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
...
+          ng-model="trigger.data[trigger.data.type.toLowerCase()].secret"
+          class="form-control"
+          type="password"
+          autocorrect="off"
+          autocapitalize="none"
+          spellcheck="false"
+          select-on-focus
+          disabled>
+        <span class="input-group-btn">
+          <a
```
spadgett: button instead of a?

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
...
+              aria-label="Toggle Password Visibility"
+              ng-click="togglePassword($index)">
+            <i class="glyphicon glyphicon-eye-open"></i>
+            <span class="sr-only">Toggle Password Visibility</span>
+          </a>
+        </span>
+      </div>
+
+      <div class="remove-webhook-trigger-action">
+        <a ng-click="removeWebhookTrigger($index)"  href="" role="button" class="btn-remove">
```
spadgett: button instead of a?

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
...
+      </div>
+    </div>
+    <div class="has-warning" ng-if="trigger.isDuplicit">
+      <span class="help-block">A {{trigger.data.type}} webhook trigger referencing the secret {{(trigger | getWebhookSecretData).secretReference.name}} already exists.</span>
+    </div>
+
+  </div>
+
+  <div class="add-webhook-action-btns">
+    <a href="" role="button" ng-click="checkLastAndAddNew()">Add Webhook</a>
```
spadgett:
```html
<button class="btn btn-link" ng-click="checkLastAndAddNew()">Add Webhook</button>
```

app/views/directives/osc-webhook-triggers.html
```diff
@@ -0,0 +1,82 @@
...
+    <span ng-if="secretsVersion | canI : 'create'">
+      <span class="action-divider" aria-hidden="true"> | </span>
+      <a href="" role="button" ng-click="openCreateWebhookSecretModal()">Create New Webhook Secret</a>
```
spadgett: button class="btn btn-link"

spadgett:
@rhamilto do you mind looking at the CSS / markup?

app/views/directives/create-secret.html
```diff
@@ -47,268 +47,296 @@
...
-    <div ng-if="newSecret.type === 'source'">
+    <div ng-if="newSecret.type != 'webhook'">
       <div class="form-group">
-        <div class="checkbox">
-          <label>
-            <input type="checkbox" ng-model="add.gitconfig">
-            Use a custom .gitconfig file
-          </label>
-        </div>
+        <label for="authentificationType">Authentication Type</label>
```
rhamilto: I see you didn't create this, but can you fix it?  Change to kebab case and add an input-id to the ui-select with the same kebab cased value.  It won't actually work because of https://github.com/angular-ui/ui-select/issues/1912, but at least the code will be right once the bug is resolved.

app/views/directives/create-secret.html
```diff
@@ -47,268 +47,296 @@
...
-            fadeFoldWidgets: true,
-            showPrintMargin: false
-          }
-        }" ng-model="newSecret.data.gitconfig" class="create-secret-editor ace-bordered" id="gitconfig-editor" required></div>
-      </div>
-    </div>
+      <div ng-if="newSecret.authType === 'kubernetes.io/basic-auth'">
+        <div class="form-group">
+          <label for="username">Username</label>
+          <div>
```
rhamilto: Extraneous div?

app/views/directives/create-secret.html
```diff
@@ -47,268 +47,296 @@
...
+              Use a custom ca.crt file
+            </label>
+          </div>
         </div>
-      </div>
-      <div ng-show="secretForm.dockerUsername.$error.required && secretForm.dockerUsername.$touched" class="has-error">
-        <div class="help-block">
-          Username is required.
+        <div class="form-group" ng-if="add.cacert" id="cacert">
+          <label class="required" for="cacert">CA Certificate File</label>
```
rhamilto: for attribute value does not match id value of input

app/views/directives/create-secret.html
```diff
@@ -47,268 +47,296 @@
...
-            name="dockerPassword"
-            ng-model="newSecret.data.dockerPassword"
-            type="password"
-            autocorrect="off"
-            autocapitalize="none"
-            spellcheck="false"
-            required>
+      <div ng-if="newSecret.authType === 'kubernetes.io/ssh-auth'">
+        <div class="form-group" id="private-key">
+          <label for="privateKey" class="required">SSH Private Key</label>
```
rhamilto: for attribute value does not match id value of input

app/views/directives/osc-webhook-triggers.html
rhamilto: I think glyphicons are fine since they ship with Bootstrap/PatternFly.

app/styles/_forms.less
```diff
@@ -7,6 +7,15 @@
   vertical-align: -2px;
 }
 
+.copy-webhook-url {
+  display: inline-flex;
+  .webhook-secrets-warning {
+    cursor: help;
+    padding: 5px;
+    font-size: 15px;
```
rhamilto: Nit: alphabetization

app/styles/_core.less
```diff
@@ -1044,30 +1044,32 @@ a.disabled-link {
...
+      table-layout: fixed;
+      width: 100%;
+      .select-webhook-type,
+      .select-secret-ref {
+        float: left;
+        margin-bottom: 0;
+        padding-right: 5px;
+        width: 50%;
+      }
+      .remove-webhook-trigger-action {
```
rhamilto: Nit: alpha

app/views/directives/create-secret.html
rhamilto: Need to check all id/for attributes as most of them are camel and not kebab.

app/scripts/directives/oscWebhookTriggers.js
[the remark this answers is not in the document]
jhadvig: Thats what Im doing in the `checkDuplicates` (renamed `duplicityCheck`) function which I feed with trigger I want to check, the one that is removed or the one user modifies 
Basically what Im doing there filtering an array of `matchingTriggers`  where all are basically duplicates, and I set the first ones `isDuplicate` field to `false` and on all others I set the field to `true`

app/scripts/directives/oscWebhookTriggers.js
jhadvig: yes it does.
After speaking with @jwforres and @ncameronbritt we decided to notify user about the dups.

app/scripts/directives/oscWebhookTriggers.js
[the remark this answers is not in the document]
spadgett: I see. I missed that you are looping over all of them and not just checking the is duplicate flag on the one you're removing in this spot.

app/scripts/directives/oscWebhookTriggers.js
jhadvig: The case shouldnt happen, since `lastTriggerSecretData.secret` or `lastTriggerSecretData.secretReference.name` should be set. But changing the check to `_.has()`

app/views/directives/osc-webhook-triggers.html
jhadvig: button would submit the form in which we put the directive/component
we are using the anchor tag in the [copy-to-clipboard](https://github.com/openshift/origin-web-console/blob/master/app/views/directives/_copy-to-clipboard.html#L14) directive as well

app/views/directives/osc-webhook-triggers.html
jhadvig: Was inpired by what we are using in envVar editor and on other places like https://github.com/openshift/origin-web-console/blob/master/app/views/directives/osc-source-secrets.html#L100-L102

app/views/directives/osc-webhook-triggers.html
spadgett: We're trying to move to buttons when it makes sense because they're better for accessibility.

Does adding `type="button"` fix the form submit problem?

app/scripts/directives/oscWebhookTriggers.js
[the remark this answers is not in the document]
jhadvig: yes, already updated the logic and will only loop on the trigger the action is on

app/views/directives/next-steps.html
jhadvig: we are, I've just have swapped the argument place, cause was adding additional arguman, but that wasnt necessary, will turn it back

app/views/directives/osc-webhook-triggers.html
jhadvig: yes, updated !

app/views/directives/create-secret.html
jhadvig: It is... under the `label` tag there is a `osc-file-input` directive to which we pass appropriate id in `drop-zone-id`

app/views/directives/create-secret.html
jhadvig: It is... under the `label` tag there is a `osc-file-input` directive to which we pass appropriate id in `drop-zone-id`

jhadvig:
@spadgett @rhamilto thanks for comments. I've addressed them.
Rewrite `oscWebhooTriggers` and `copyWebhookUrl` into components.
Regarding `stringData` usage, I've just changed the creation of the webhook secters. For other secret types I will open a card and assign it to myself, but didnt wanted to make this PR any more complex.

app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -0,0 +1,81 @@
...
+          autocorrect="off"
+          autocapitalize="none"
+          spellcheck="false"
+          select-on-focus
+          disabled>
+        <div class="input-group-btn">
+          <button
+            type="button"
+            class="btn btn-default toggle"
+            title="Toggle Password Visibility"
```
jwforres: its a nit, but this isnt really a password, i would probably call it a token

jhadvig:
@jwforres comment addressed. PTAL :)

app/components/copy-webhook-url/copy-webhook-url.component.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .component('copyWebhookUrl', {
+    bindings: {
+      buildConfigName: "=",
+      triggerType: "=",
+      projectName: "=",
+      secret: "=",
+      webhookSecrets: "="
```
spadgett: For future reference, these should probably be one way bindings with `<` instead of `=`. But I don't want to hold the PR on that.

PR #2641, opened 2018-01-08
prevent wrapping of text in terminal

Schnitzel:
I'm encountering a weird bug with some of our terminals with a big fancy prompts, it basically breaks at a wrong place:

openshift_web_console

see the `$PS1` output, it should not be breaking after `[idg-standard-public-` but somehow it does.

It's clearly the dash that causes the browser to think it should break the line. if I remove the dash it will not break:

openshift_web_console2

After setting the `white-space: nowrap;` it all works perfectly
openshift_web_console

Browser: Chrome 63
OpenShift Container Platform v3.6.173.0.49

### DIFF
--- a/app/styles/_container-terminal.less
+++ b/app/styles/_container-terminal.less
@@ -97,4 +97,5 @@ kubernetes-container-terminal {
   font-family: "Monospace Regular", "DejaVu Sans Mono", Menlo, Monaco, Consolas, monospace;
   font-size: 12px;
   line-height: 1em;
+  white-space: nowrap;
 }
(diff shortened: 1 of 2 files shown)


spadgett:
@Schnitzel Sorry for the slow review.

How does it this affect normal line wrapping for lines longer than the terminal window?

We might want to raise this as an issue instead against https://github.com/xtermjs/xterm.js if it's not anything specific to our styles.

cc @rhamilto @jwforres

jwforres:
Should also see if this is still an issue with v3 of xterm since we intend to move up soon

Schnitzel:
Good question!

Here without the patch:
openshift_web_console

Here with the patch:
openshift_web_console

Re xterm.js

This is actually already fixed there, see: https://github.com/xtermjs/xterm.js/issues/327

and proof:
xterm_js_demo

spadgett:
@Schnitzel thanks, do you mind running `grunt build` and adding the `dist` folder to your commit?

/assign @rhamilto

rhamilto:
This PR will be obsoleted by https://github.com/openshift/origin-web-console/pull/2849.

rhamilto:
Obsoleted by https://github.com/openshift/origin-web-console/pull/2849, which bumped xterm.js to v3.x

PR #2644, opened 2018-01-08
Add additional add to project options to overview empty state

### LINKED ISSUE #2631: Project overview empty state Add to Project button links to catalog page, but other add options not visible on the catalog page
screen shot 2018-01-05 at 10 21 46 am

The result of clicking "Add to Project" in the previous screenshot brings you here, and the page has no mention of the other add methods:
screen shot 2018-01-05 at 10 23 21 am

The old create page had tabs at the top with the other add methods:
screen shot 2018-01-05 at 10 24 49 am

And the landing page has the other add options listed:
screen shot 2018-01-05 at 10 25 25 am

It seems like we should be consistent?

rhamilto:
Fixes #2631

localhost_9000_dev-console_project_myproject_overview ipad pro
localhost_9000_dev-console_project_myproject_overview iphone 6 1

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -165,6 +165,38 @@
   .btn {
     margin: 10px 0 20px;
   }
+  .empty-state-message-main-action {
+    margin-top: @line-height-computed;
+    .btn {
+      margin-bottom: 0;
+    }
+  }
+  .empty-state-message-secondary-action {
+    margin-top: @line-height-computed;
+    @media(max-width: (@screen-xs-min - 1)) {
+      align-items: center;
+      flex-direction: column;
+    }
+    .btn {
+      margin-bottom: 8px;
+      @media(min-width: @screen-xs-min) {
+        margin-right: 5px;
+        &:last-of-type {
+          margin-right: 0;
+        }
+      }
+    }
+  }
+  .empty-state-message-main-action,
+  .empty-state-message-secondary-action {
+    display: flex;
+    justify-content: center;
+    .btn {
+      flex: 0 1 auto;
+      margin-top: 0;
+      min-width: 33.333333%;
+    }
+  }
   h2, .h2 {
     // Avoid clipped descenders with class truncate or overflow: hidden
     line-height: 1.4;
--- a/app/views/overview.html
+++ b/app/views/overview.html
@@ -7,15 +7,25 @@
       <div class="empty-state-message text-center">
         <div ng-if="project.metadata.name | canIAddToProject">
           <h2>Get started with your project.</h2>
-          <p class="gutter-top">
-            Use your source or an example repository to build an application
-            image, or add components like databases and message queues.
+          <p>
+            Add content to your project from the catalog of web frameworks, databases, and other components. You may also deploy an existing image, create or replace resources from their YAML or JSON definitions, or select an item shared from another project.
           </p>
-          <p class="gutter-top">
-            <button class="btn btn-lg btn-primary" ng-click="browseCatalog()">
-              Add to Project
+          <div class="empty-state-message-main-action">
+            <button class="btn btn-primary btn-lg" ng-click="browseCatalog()">
+              Browse Catalog
             </button>
-          </p>
+          </div>
+          <div class="empty-state-message-secondary-action">
+            <button class="btn btn-default btn-sm" ng-click="showOrderingPanel('deployImage')">
+              Deploy Image
+            </button>
+            <button class="btn btn-default btn-sm" ng-click="showOrderingPanel('fromFile')">
+              Import YAML / JSON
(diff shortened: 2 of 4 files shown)


rhamilto:
@openshift/team-ux-review, there are some reservations regarding the three secondary action buttons, so I wanted to make sure y'all see the screenshots above.

jwforres:
This doesn't lay out well when the buttons have realistic content in them, i.e. actions of different string lengths.

beanh66:
@rhamilto I would suggest making all buttons on the empty state pattern the same width.  Can we make that change here and contribute back to PF? Trying to figure out the best way to implement. I am happy to update the design documentation on PF as well. I agree it could look better so thanks for bringing it up with these screenshots. Independently, should we stack all 4 buttons on mobile? IMO, we should. Something like this: 
screen shot 2018-01-10 at 12 31 15 pm

serenamarie125:
I agree @beanh66 .  Would be best to have this in PF first, but unsure priority of fix.  I like vertically stacking the buttons, and fits well with the current PF Empty state pattern.

jennyhaines:
@beanh66 - I think that vertically stacking the buttons would help the mobile layout, too. I like this solution for both desktop and mobile.

rhamilto:
I'm not sure fixing the width of the buttons is a tenable solution.  Making the buttons the same width in PatternFly is problematic since we have no idea what the labels will be; how do you decide what the right width is?  And in OpenShift, we will one day likely need to internationalize, and it seems likely that label length will vary across languages.

So spitballing another idea... what if we use a [button group](https://getbootstrap.com/docs/3.3/components/#btn-groups) for the secondary actions at resolutions greater that 479px, and stack the secondary actions vertically at mobile less than 480px?

screen shot 2018-01-10 at 1 41 50 pm

screen shot 2018-01-10 at 1 56 45 pm

spadgett:
Can we use flexbox to make the buttons the same width without needing a fixed size?

rhamilto:
> Can we use flexbox to make the buttons the same width without needing a fixed size?

That's a possibility at resolutions greater than 479px.  I'm not sure the resulting appearance is desirable, however.  To my eye, the secondary action buttons become dominant.
screen shot 2018-01-10 at 2 14 07 pm

rhamilto:
And I made the secondary actions size small in the screenshot above.

rhamilto:
Sorry, @spadgett.  I inadvertently deleted your comment:

> Can we use flexbox to make the buttons the same width without needing a fixed size?

ncameronbritt:
@rhamilto Is there some kind of flexbox trickery you could do for the primary action button that would increase the padding like the secondary actions? So that all four buttons seem more related... Just curious what that might look like.

rhamilto:
> @rhamilto Is there some kind of flexbox trickery you could do for the primary action button that would increase the padding like the secondary actions? So that all four buttons seem more related... Just curious what that might look like.

screen shot 2018-01-10 at 2 38 10 pm

rhamilto:
Sorry.  In those last two screenshots, "Deploy Image" was a btn-sm, but the other two were not.  Fixed it so they're all btn-sm.

screen shot 2018-01-10 at 2 42 34 pm

cshinn:
@rhamilto that looks pretty good! Does it hold up in on large window sizes?

beanh66:
Agreed, looks great @rhamilto

rhamilto:
> Does it hold up in on large window sizes?

Yeah.  It never gets larger than it appears in https://github.com/openshift/origin-web-console/pull/2644#issuecomment-356714686 due to a max-width set on the empty-state-message.

This might be a good compromise:
http://recordit.co/TvaV5b6HCM

Because the labels vary in length, I can't keep the buttons the exact same size at the smallest viewport resolution of 320px, but I can let them grow to equal size as the viewport gets bigger.

jennyhaines:
@rhamilto - I think that button size transition between viewport widths looks good!

rhamilto:
PR updated with changes recorded in http://recordit.co/TvaV5b6HCM.

ncameronbritt:
Flexbox ftw

beanh66:
@rhamilto Looks like a great solution. Is it possible to contribute this code back to PatternFly so other products can use the Empty State in a similar way? Thoughts @dtaylor113?

rhamilto:
>  Is it possible to contribute this code back to PatternFly so other products can use the Empty State in a similar way?

I can, but understand this fix is specific to **three** secondary action buttons.  I think the PatternFly solution needs to be more robust in its approach to a variable number of buttons, and I'm not sure how to go about doing that.  It seems like this should become an upstream issue to be solved by the community at large given the complexity of the problem and the issues with the existing empty state solution in PatternFly.

PR #2650, opened 2018-01-11
Fix potential API mismatch in BuildsService

benjaminapetersen:
Ran into another mismatch & realized a few methods in BuildsService are probably not future proofed.

### DIFF
--- a/app/scripts/services/builds.js
+++ b/app/scripts/services/builds.js
@@ -9,7 +9,6 @@ angular.module("openshiftConsole")
                     Navigate,
                     NotificationsService) {
 
-    var buildsVersion = APIService.getPreferredVersion('builds');
     var buildConfigsInstantiateVersion = APIService.getPreferredVersion('buildconfigs/instantiate');
     var buildsCloneVersion = APIService.getPreferredVersion('builds/clone');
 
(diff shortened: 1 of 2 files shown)


app/scripts/services/builds.js
```diff
@@ -88,8 +88,10 @@ angular.module("openshiftConsole")
         namespace: build.metadata.namespace
       };
       var canceledBuild = angular.copy(build);
+      var rgv = APIService.objectToResourceGroupVersion(canceledBuild);
       canceledBuild.status.cancelled = true;
-      return DataService.update(buildsVersion, canceledBuild.metadata.name, canceledBuild, context).then(function() {
```
spadgett: Should be able to remove `buildsVersion` var now

benjaminapetersen: Not sure why jshint didn't catch that when I did my build, just removed.

app/scripts/services/builds.js
```diff
@@ -108,16 +109,18 @@ angular.module("openshiftConsole")
     var cloneBuild = function(originalBuild, buildConfigName) {
       var buildType = isJenkinsPipelineStrategy(originalBuild) ? 'pipeline' : 'build';
       var originalDisplayName = getBuildDisplayName(originalBuild, buildConfigName);
+
       var req = {
         kind: "BuildRequest",
-        apiVersion: "v1",
+        apiVersion: originalBuild.apiVersion,
```
spadgett: This should be the api version for `buildsCloneVersion`, which might not match `originalBuild.apiVersion`

benjaminapetersen: Ah, so `APIService.toAPIVersion(buildsCloneVersion)`, ill update.

benjaminapetersen: I bet I'm gonna be doing a # of bug fixes around this.  Now it's getting fiddly.

benjaminapetersen: updated

PR #2652, opened 2018-01-12
Bump angular-patternfly version to 4.12.2

dtaylor113:
Bumped angular-patternfly version from 4.11.8 to 4.14.2.

Fixes [catalog/#601](https://github.com/openshift/origin-web-catalog/issues/601)  
Tested: 'Clear Filter' link no longer redirects to Catalog Home/Landing page, stays on Project's Catalog

**Angular-Patternfly Tests**
angular-patternfly changelog between 4.11.8 - 4.14.2:
-  Wizard parameters from two way, to single way, binding.
   Tested in console, seems to be working.  Didn't find any console/catalog code which was watching for changes of a-pf wizard parameters.
-  Filter Panel: updated filter results and mobile.  
   Tested `catalog-filter`'s filter panel displays correctly at mobile.
- List View: not used by console
- Table View: not used by console
- Card: not used by console


dtaylor113:
Jenkins failure:
```
**************************************************
*                    Failures                    *
**************************************************

1) User adds a template to a project after creating a new project using the Import YAML tab should save the template in the project catalog
  Error: URL hasn't changed to project/console-test-project-1515784105411/create(is currently: undefined)
```
Using a new laptop with Fedora 27, can't seem to downgrade my FF to v39 (missing a lib).  Looking into it, but at the moment I can't run the integration tests locally.

bower.json
```diff
@@ -13,7 +13,7 @@
     "angular-touch": "1.5.11",
     "angular-route": "1.5.11",
     "angular-bootstrap": "0.14.3",
-    "angular-patternfly": "4.11.8",
+    "angular-patternfly": "4.14.2",
```
spadgett: Do we need to bump patternfly to the same version?

PR #2655, opened 2018-01-12
Fix membershipService.mapRolebindingsForUI to properly list custom Roles in the output

benjaminapetersen:
Fixes [bugzilla 1533753](https://bugzilla.redhat.com/show_bug.cgi?id=1533753)

- Maps custom `Roles` correctly now by checking the `Role` kind, not just assuming the presence of a `namespace` means the `Role` is not a `ClusterRole`.
- Updates tests to ensure a `Role` is covered
- Fixes (my) poorly written tests that made this harder to debug than it should have been, considering it was such a tiny update.

### DIFF
--- a/app/scripts/services/membership/membership.js
+++ b/app/scripts/services/membership/membership.js
@@ -2,7 +2,7 @@
 
 angular
   .module('openshiftConsole')
-  .factory('MembershipService', function($filter, Constants) {
+  .factory('MembershipService', function($filter, APIService, Constants) {
 
     var annotation = $filter('annotation');
 
@@ -88,8 +88,8 @@ angular
       var mapForUI = _.reduce(
                       rolebindings,
                       function(result, rolebinding) {
-                        // hmm. dont like this here
-                        var roleKey = uniqueKey(rolebinding.roleRef.namespace ? 'Role' : 'ClusterRole', rolebinding.roleRef.name);
+                        // matched keyedRoles below for lookups
+                        var roleKey = uniqueKey(rolebinding.roleRef.apiGroup, rolebinding.roleRef.kind, rolebinding.roleRef.name);
                         // each subject, (user, bob; group: hobbits)
                         _.each(rolebinding.subjects, function(subject) {
                           var subjectKey = uniqueKey(subject.namespace, subject.name);
@@ -128,7 +128,9 @@ angular
       return _.reduce(
               roles,
               function(result, role) {
-                result[uniqueKey(role.kind, role.metadata.name)] = role;
+                var group = APIService.parseGroupVersion(role.apiVersion).group;
+                var roleKey = uniqueKey(group, role.kind, role.metadata.name);
+                result[roleKey] = role;
                 return result;
               }, {});
     };
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -2419,13 +2419,13 @@ t[e.tag] = t[e.tag] || {}, t[e.tag].name = e.tag, t[e.tag].status = angular.copy
 }), t;
 }
 };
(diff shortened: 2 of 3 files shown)


benjaminapetersen:
Fixing, Jenkins doesn't like `() => { }` in tests, node version must be old.

app/scripts/services/membership/membership.js
```diff
@@ -84,12 +84,12 @@ angular
...
     };
 
+
     var mapRolebindingsForUI =  function(rolebindings, roles) {
       var mapForUI = _.reduce(
                       rolebindings,
                       function(result, rolebinding) {
-                        // hmm. dont like this here
-                        var roleKey = uniqueKey(rolebinding.roleRef.namespace ? 'Role' : 'ClusterRole', rolebinding.roleRef.name);
+                        var roleKey = uniqueKey(rolebinding.roleRef.kind, rolebinding.roleRef.name);
```
spadgett: Can we ever have role references from other namespaces? If so, the key won't be unique.

benjaminapetersen: I don't know if you can reference a role for namespace `foo` from namespace `bar`.   I think you can't, if its global it has to be a `ClusterRole`.

@enj true?

enj: Correct, but I would use all the data:

```go
// RoleRef contains information that points to the role being used
type RoleRef struct {
	// APIGroup is the group for the resource being referenced
	APIGroup string `json:"apiGroup" protobuf:"bytes,1,opt,name=apiGroup"`
	// Kind is the type of resource being referenced
	Kind string `json:"kind" protobuf:"bytes,2,opt,name=kind"`
	// Name is the name of resource being referenced
	Name string `json:"name" protobuf:"bytes,3,opt,name=name"`
}
```

Granted I am not sure what your code is actually doing...

spadgett: We might put it in anyway if it ever might be possible to make sure the key is always unique.

spadgett: OK so no namespace (which I guess is why we had the bug). Agree with @enj, though, I'd add the group.

benjaminapetersen: This is really a random property... I could probably make it a `Date.now()` even & skip all the actual values.  I think I used bits of the Role to keep it human readable for debugging.  Let me check & ensure & I can add anything to make it certainly unique.

benjaminapetersen: added `rolebinding.metadata.namespace` which will generate keys something like ' clusterrole-edit-myproject'.

spadgett: But that is always the same value for all bindings, so it doesn't help with uniqueness

benjaminapetersen: `roleRef` doesn't necessarily have a `namespace` (usually doesn't).  I can swap it for just a random number then.

spadgett: We're overthinking this... Just use the fields that are there: group + kind + name

benjaminapetersen: I believe `apiGroup` will always be the same, `rbac.authorization.k8s.io`, so I think we are probably coming back full circle to more or less what we have just with longer keys.

spadgett: That's true right now, but we can't assume. There's a property there that could be set to something else.

benjaminapetersen:
@spadgett ok should be ready for a final checkoff.  The test updates ended up being a tad tedious.  I did a lot of verification since they all share fixtures.

I think the fact that this was a 4 line change but has a +282 / -160 diff due to tests implies that the tests are too dependent on shared fixtures. 🙁

PR #2656, opened 2018-01-12
Add link to re-enable percentage slider when editing service weights on the create route page

TheRealJon:
A link now appears next to the alternate service action links to re-enable the slider once it is disabled. Fixes [Bugzilla Bug 1533818](https://bugzilla.redhat.com/show_bug.cgi?id=1533818)

screenshot-localhost-9000-2018-01-12-17-31-45-372

### DIFF
--- a/app/views/directives/osc-routing.html
+++ b/app/views/directives/osc-routing.html
@@ -160,19 +160,27 @@
           </div>
         </div>
       </div>
-
+      
       <div ng-repeat="alternate in route.alternateServices" class="form-group">
-        <osc-routing-service model="alternate"
-                             service-options="alternateServiceOptions"
-                             all-services="servicesByName"
-                             is-alternate="true"
-                             show-weight="route.alternateServices.length > 1 || controls.hideSlider">
+        <osc-routing-service 
+          model="alternate"
+          service-options="alternateServiceOptions"
+          all-services="servicesByName"
+          is-alternate="true"
+          show-weight="route.alternateServices.length > 1 || controls.hideSlider">
         </osc-routing-service>
-        <a href="" ng-click="route.alternateServices.splice($index, 1)">Remove Service</a>
-        <span ng-if="$last && route.alternateServices.length < alternateServiceOptions.length">
-          <span class="action-divider">|</span>
-          <a href="" ng-click="addAlternateService()">Add Another Service</a>
-        </span>
+        <div class="row">
+          <div class="col-sm-6">
+            <button type="button" class="btn btn-link" ng-click="route.alternateServices.splice($index, 1)">Remove Service</button>
+            <span ng-if="$last && route.alternateServices.length < alternateServiceOptions.length">
+              <span class="action-divider">|</span>
+              <button type="button" class="btn btn-link" ng-click="addAlternateService()">Add Another Service</button>
+            </span>
+          </div>
+          <div ng-if="route.alternateServices.length === 1 && controls.hideSlider" class="col-sm-6">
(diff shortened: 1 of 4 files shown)


TheRealJon:
The link is in a `col-sm-6' div, which aligns it with the **Weight** fields of the services above it. In xs viewports, the link wraps down below the other two service action links.

app/views/directives/osc-routing.html
```diff
@@ -160,19 +160,26 @@ <h3>Alternate Services</h3>
...
+        <div class="row">
+          <div class="col-sm-6">
+            <a href="" ng-click="route.alternateServices.splice($index, 1)">Remove Service</a>
+            <span ng-if="$last && route.alternateServices.length < alternateServiceOptions.length">
+              <span class="action-divider">|</span>
+              <a href="" ng-click="addAlternateService()">Add Another Service</a>
+            </span>
+          </div>
+          <div ng-if="route.alternateServices.length === 1 && controls.hideSlider" class="col-sm-6">
+            <a href="" ng-click="controls.hideSlider = false">Edit Weights Using Percentage Slider</a>
```
spadgett: We're trying to switch to `button` elements styled as links in new code for better accessibility. (I realize you've followed what was there.)

```html
<button type="button" ng-click="controls.hideSidebar = false" class="btn btn-link">Edit Weights Using Percentage Slider</button>
```

Please check that it doesn't break using enter to submit the form, however. If it does, we might just leave it an `<a>` for now.

TheRealJon:
One other issue that I'm finding is that when the slider is re-enabled, if the user edited the weights as integers, the slider will need to be updated to reflect that change. The only problem is that our function for translating the weights to percentages rounds to the nearest integer. This means that if one of the weight values translates to less than .5%, it will be rounded down to 0. There are a lot of different approaches we could take to solve this, but I wanted to get some input.

@jwforres @spadgett @benjaminapetersen

TheRealJon:
Nevermind, I think I figured this out using the `initializingSlider` variable.

TheRealJon:
@spadgett I changed the links to buttons and have the slider accurately updating when switching between slider and integer inputs.

app/views/directives/osc-routing.html
```diff
@@ -160,19 +160,26 @@ <h3>Alternate Services</h3>
           </div>
         </div>
       </div>
-
+      
       <div ng-repeat="alternate in route.alternateServices" class="form-group">
         <osc-routing-service model="alternate"
```
benjaminapetersen: nit: formatting consistency I might push `model="alternate"` down to line up with the rest of the attribs as well.

TheRealJon:
@benjaminapetersen Style issue fixed.

benjaminapetersen:
This is probably out of scope for the work you are doing, but do we think that the slider really needs to be full width?

screen shot 2018-01-25 at 4 52 50 pm

It seems like the slider could fit into the block on the right, replacing the previous input box.

screen shot 2018-01-25 at 4 53 34 pm

app/views/directives/osc-routing.html
benjaminapetersen: I was about to ask about this switch, glad I read your comment 😄

benjaminapetersen:
Otherwise I think looks good to me, this is a big directive!

app/views/directives/osc-routing.html
```diff
@@ -160,19 +160,27 @@ <h3>Alternate Services</h3>
           </div>
         </div>
       </div>
-
+
```
spadgett: nit: whitespace

Issue #2651, opened 2018-01-12
Secret creation should be using stringData 


jhadvig:
When [creating secrets](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/createSecret.js#L113-L131) we should use `stringData` and dont encode the data with `window.btoa()` function.

PR #2668, opened 2018-01-16
Secret creation should be using stringData

jhadvig:
Fixes https://github.com/openshift/origin-web-console/issues/2651

@spadgett PTAL

### DIFF
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -114,41 +114,41 @@ angular.module("openshiftConsole")
               name: $scope.newSecret.data.secretName
             },
             type: authType,
-            data: {}
+            stringData: {}
           };
 
           switch (authType) {
             case "kubernetes.io/basic-auth":
+
               // If the password/token is not entered either .gitconfig or ca.crt has to be provided
               if (data.passwordToken) {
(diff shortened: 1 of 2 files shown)


app/scripts/directives/createSecret.js
```diff
@@ -117,38 +117,38 @@ angular.module("openshiftConsole")
             data: {}
```
spadgett: nit: It's a little simpler to change `data` here to `stringData` and just set `secret.stringData` in each case block below.

app/scripts/directives/createSecret.js
```diff
@@ -117,38 +117,38 @@ angular.module("openshiftConsole")
...
           };
 
+          var secretData = {};
           switch (authType) {
             case "kubernetes.io/basic-auth":
               // If the password/token is not entered either .gitconfig or ca.crt has to be provided
               if (data.passwordToken) {
-                secret.data = {password: window.btoa(data.passwordToken)};
+                secretData['password'] = data.passwordToken;
```
spadgett: We typically don't use the bracket `[]` notation unless the key has special characters.

```js
secretData.password = data.passwordToken;
```

jhadvig:
@spadgett comments addressed

PR #2676, opened 2018-01-17
Fix bug where Init:Error failed pods displayed "Pod initializing"

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1512473

screen shot 2018-01-17 at 4 46 18 pm
screen shot 2018-01-17 at 4 46 24 pm

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -1022,7 +1022,8 @@ angular.module('openshiftConsole')
     };
   })
   .filter('podStatus', function() {
-    // Return results that match kubernetes/pkg/kubectl/resource_printer.go
+    // Return results that match
+    // https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/pkg/printers/internalversion/printers.go#L523-L615
     return function(pod) {
       if (!pod || (!pod.metadata.deletionTimestamp && !pod.status)) {
         return '';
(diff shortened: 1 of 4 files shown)


app/scripts/filters/resources.js
```diff
@@ -1038,7 +1038,8 @@ angular.module('openshiftConsole')
             signal,
             exitCode;
 
-        if (containerReason) {
+        // only show containerReason if pod.status.phase is not 'Failed'
+        if (reason !== 'Failed' && containerReason) {
```
jwforres: @spadgett can you think of a case where this isn't what we would want?

spadgett: I feel like there might be specific reasons you'd want to see even for failed pods. Evicted? Deadline exceeded?

This logic was modeled after the CLI, so it would be good to stay consistent. I'm curious what the CLI shows for these pods.

rhamilto:
```
NAME                READY     STATUS       RESTARTS   AGE
dancer-ex-1-build   0/1       Init:Error   0          1h
```

spadgett: @jwforres Do we need to iterate over `initContainerStatuses` here too?

spadgett: I believe this is the current CLI logic if we want to be consistent:

https://github.com/spadgett/origin/blob/5b71e8d8e441a0d39f9228a4bce4c7a6a13be08b/vendor/k8s.io/kubernetes/pkg/printers/internalversion/printers.go#L523-L615

jwforres: yeah i think that is probably the correct fix for this

spadgett: Looks like CLI has special initializing logic.

rhamilto:
Take 2.  I didn't add all the initializing logic since it seems we only care about errors.  Thoughts?

spadgett:
/hold
/assign

I'm not sure the logic is right. I'll follow up with a review, but hold for now.

app/scripts/filters/resources.js
```diff
@@ -1029,6 +1029,22 @@ angular.module('openshiftConsole')
         return 'Terminating';
       }
 
+      // If pod.status.initContainerStatuses include state.terminated.reason with
+      // a value of "Error", display it instead of pod.status.containerStatus.
+      var initReason;
+
+      angular.forEach(pod.status.initContainerStatuses, function(initContainerStatus) {
+        var initContainerReason = _.get(initContainerStatus, 'state.terminated.reason');
```
spadgett: It looks like there is a `state.terminated.reason` even for successful init containters. We probably need to check `exitCode`, too.

```yaml
  initContainerStatuses:
    - containerID: >-
        docker://695625642cd6a776c8fe86a0ee530e0acbe04cfbb7187552c55b068f60df1333
      image: 'docker.io/openshift/origin-sti-builder:v3.9.0-alpha.3'
      imageID: >-
        docker-pullable://docker.io/openshift/origin-sti-builder@sha256:1120189696907b0451dcd081b39406be02c14b988cc7fc655af39002a0953ced
      lastState: {}
      name: git-clone
      ready: true
      restartCount: 0
      state:
        terminated:
          containerID: >-
            docker://695625642cd6a776c8fe86a0ee530e0acbe04cfbb7187552c55b068f60df1333
          exitCode: 0
          finishedAt: '2018-01-19T23:18:35Z'
          reason: Completed
          startedAt: '2018-01-19T23:18:33Z'
```
If there are multiple init containers and only the first init container has it, it looks like `initContainerReason` will be unset since we don't break the loop early.

rhamilto:
Take 3.  No idea if this is right, but I gave it a go.

app/scripts/filters/resources.js
```diff
@@ -1029,32 +1029,66 @@ angular.module('openshiftConsole')
...
-            exitCode;
+      _.forEach(pod.status.initContainerStatuses, function(initContainerStatus) {
+        var initContainerState = _.get(initContainerStatus, 'state');
 
-        if (containerReason) {
-          reason = containerReason;
-          return;
+        if (initContainerState.terminated && initContainerState.terminated.exitCode === 0) {
+          // initialization is complete
+          return true;
```
spadgett: `return true` will short-circuit here and skip the remaining init containers. Since this container was successful, it's not what we want. We want to short0circuit only when we find one that is not complete or failed.

I think you want just `return;`

app/scripts/filters/resources.js
```diff
@@ -1029,32 +1029,66 @@ angular.module('openshiftConsole')
...
+      var initializing = false;
+      var reason;
 
-      // Print detailed container reasons if available. Only the last will be
-      // displayed if multiple containers have this detail.
-      angular.forEach(pod.status.containerStatuses, function(containerStatus) {
-        var containerReason = _.get(containerStatus, 'state.waiting.reason') || _.get(containerStatus, 'state.terminated.reason'),
-            signal,
-            exitCode;
+      _.forEach(pod.status.initContainerStatuses, function(initContainerStatus) {
```
spadgett: nit: We try to consistently use `_.each` (alias to `_.forEach`) so we don't have a mix of one and the other. They're the same, but if you don't know that, it might be confusing to see both in the code.

app/scripts/filters/resources.js
```diff
@@ -1029,32 +1029,66 @@ angular.module('openshiftConsole')
...
+          initializing = true;
         }
       });
 
+      if (!initializing) {
+        reason = pod.status.reason || pod.status.phase;
+
+        // Print detailed container reasons if available. Only the last will be
+        // displayed if multiple containers have this detail.
+        _.forEach(pod.status.containerStatuses, function(containerStatus) {
```
spadgett: nit: `_.each`

rhamilto:
Take 3 comments address.

app/scripts/filters/resources.js
```diff
@@ -1029,32 +1029,66 @@ angular.module('openshiftConsole')
...
+        if (initContainerState.waiting && initContainerState.waiting.reason && initContainerState.waiting.reason !== 'PodInitializing') {
+          reason = "Init " + initContainerState.waiting.reason;
+          initializing = true;
         }
       });
 
+      if (!initializing) {
+        reason = pod.status.reason || pod.status.phase;
+
+        // Print detailed container reasons if available. Only the last will be
```
spadgett: Update comment to say only first will be displayed if multiple containers have a reason. The comment no longer matches the code.

app/scripts/filters/resources.js
```diff
@@ -1029,32 +1029,66 @@ angular.module('openshiftConsole')
...
+
+          signal = _.get(containerStatus, 'state.terminated.signal');
+          if (signal) {
+            reason = "Signal: " + signal;
+            return true;
+          }
+
+          exitCode = _.get(containerStatus, 'state.terminated.exitCode');
+          if (exitCode) {
+            reason = "Exit Code: " + exitCode;
```
spadgett: We should `return true` after this line, too, so we stop looking at other containers like the other checks above. Otherwise it'll keep iterating.

rhamilto:
Round 4.  Thank you for your patience, @spadgett.

spadgett:
Sorry didn't realize this still had the hold label

/hold cancel

PR #2677, opened 2018-01-17
Remove obsolete extensions.js and extensions.css

spadgett:
- Remove requests from index.html that will now fail.
- Remove sample files that won't be loaded.

/assign @jwforres


jwforres:
@spadgett IIRC @benjaminapetersen 's tests for create from URL modify extension.js in order to check things related to allowed namespaces

spadgett:
OK I'll have to look for another way :/

jwforres:
hmm, could back up index.html and then inject extension.js into index.html, similar to what ben was doing with extension.js itself

i def want our built dist index.html to stop trying to load the files

benjaminapetersen:
@jwforres @spadgett they do, tho I think the tests can be updated if this is being dropped.  The code is `test/integration/helpers/extensions.js` and reads out the file.  I imagine we could write to `index.html` for this stuff, though I think there may be a timing issue to deal with.

test/integration/features/user_creates_from_url.spec.js
```diff
@@ -49,7 +41,7 @@ describe('authenticated e2e-user', function() {
 
   afterAll(function() {
     projectHelpers.deleteAllProjects();
-    resetExtensions();
+    browser.executeScript("window.OPENSHIFT_CONSTANTS.CREATE_FROM_URL_WHITELIST = ['openshift'];");
```
benjaminapetersen: Do we want symmetrical `setupEnv` and `tearDownEnv` for this?  No strong pref, otherwise everything looks good to me.  Yay removes a bunch of code!

benjaminapetersen: -135 lines, +3.  thats pretty good tradeoff 😄

spadgett: Unfortunately it doesn't work since we reload the page when we navigate :(

benjaminapetersen: Mmm. I have a bunch of the nav updated to use in-page navigation in the e2e test overhaul PR...

spadgett:
OK I think I found a not-terrible way to do this.

@jwforres @benjaminapetersen PTAL

app/index.html
```diff
@@ -408,8 +407,9 @@ <h1>JavaScript Required</h1>
         <script src="scripts/extensions/nav/dropdownMobile.js"></script>
         <!-- endbuild -->
 
-    <!-- Load any extensions last. -->
-    <script src="scripts/extensions.js"></script>
-
+        <!-- build:remove -->
```
benjaminapetersen: This is clever, seems like it will do the trick.

test/integration/features/user_creates_from_url.spec.js
```diff
@@ -20,13 +18,6 @@ describe('authenticated e2e-user', function() {
   let namespaceForFixtures = "template-dumpster";
 
   let setupEnv = function() {
```
benjaminapetersen: Want to leave a comment here as well?  I imagine we will forget about `extensions/test-extensions.js` in the future since its not really a real feature anymore.

app/index.html
benjaminapetersen: Wonder if we should drop the `extensions` directory && load this file from `test/integration/extensions` or something.

spadgett:
Thanks @benjaminapetersen, I added a comment to the test.

> aside from just the suggestion about putting the script in the test dir

The only reason I didn't is because I didn't want to set up serve static for the entire test directory, but maybe that is harmless.

benjaminapetersen:
@spadgett works for me!

app/index.html
```diff
@@ -32,7 +32,6 @@
     <!-- build:css(.tmp) styles/main.css -->
     <link rel="stylesheet" type="text/css" href="styles/main.css">
     <!-- endbuild -->
-    <link rel="stylesheet" type="text/css" href="styles/extensions.css">
```
jwforres: Why not do the same build exclusion for extensions.css so we can still do dev testing of extensions?

extensions/test-extensions.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+// Extensions used for local development and tests. These changes will not be
+// included in the built dist.
+
+// This is required for our integration tests, which tests whitespaced
+// namespaces for Create from URL.
+window.OPENSHIFT_CONSTANTS.CREATE_FROM_URL_WHITELIST = [
```
jwforres: so are we always running with this extension in place in dev mode then? i don't love creating this pattern, what if someone breaks the OOTB constants and doesn't realize it because the extension constants are overriding it in dev mode?

spadgett:
@jwforres thanks, updated

extensions/test-extensions.js
jwforres: i think this one would bother me less if we pushed template-dumpster onto the array, instead of overwriting the whole array

spadgett:
Let me see if I can't change that just for the integration tests and use array push

extensions/test-extensions.js
benjaminapetersen: I was considering suggesting we keep the `/helpers/extensions.js` and update the `addExtension()` fn to read `index.html` && write directly to the `head` of the document only when the test runs.

spadgett:
@jwforres Updated. I now have a separate extensions for development and integration tests, and I'm appending to the array. Tests are green.

spadgett:
Hold, I want to change the file names.

spadgett:
Updated. Renamed the files to local-extensions.{js,css}

benjaminapetersen:
So I just realized, I use the `extensions.js` file to occasionally do some hacks and overrides to `config.js` or other things.  How do we want to handle that at this point?  Tinkering with it a bit now.

spadgett:
I don't think it's any different. You just change local-extensions.js now. Or put your JS in config.local.js, which isn't tracked in git.

PR #2679, opened 2018-01-18
Webhook trigger editor should behave like envVar editor

jhadvig:
Updated the webhook trigger editor behavior to be the same as the one in envVar editor, which means that user can add as many triggers as he wants, but only those that have filled trigger type and secret will be save, rest wont be saved.

Will add the dist after review.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1535402

@jwforres PTAL

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -1104,7 +1104,7 @@ a.disabled-link {
       }
     }
     .add-webhook-action-btns {
-        margin-top: 15px;
+      margin-top: 15px;
     }
   }
 
--- a/app/components/copy-webhook-url/copy-webhook-url.component.js
+++ b/app/components/copy-webhook-url/copy-webhook-url.component.js
@@ -1,20 +1,23 @@
 'use strict';
-
-angular.module('openshiftConsole')
-  .component('copyWebhookUrl', {
+(function() {
+  angular.module('openshiftConsole').component('copyWebhookUrl', {
+    controller: CopyWebhookUrl,
+    controllerAs: '$ctrl',
     bindings: {
-      buildConfigName: "=",
-      triggerType: "=",
-      projectName: "=",
-      secret: "=",
-      webhookSecrets: "="
+      buildConfigName: "<",
+      triggerType: "<",
+      projectName: "<",
+      secret: "<",
+      webhookSecrets: "<"
     },
-    templateUrl: 'components/copy-webhook-url/copy-webhook-url.html',
-    controller: function() {
-      var ctrl = this;
-
-      ctrl.showSecretsWarning = function() {
-        return _.get(ctrl.secret, 'secretReference.name') && !ctrl.webhookSecrets;
-      };
-    }
+    templateUrl: 'components/copy-webhook-url/copy-webhook-url.html'
   });
+
+  function CopyWebhookUrl() {
+    var ctrl = this;
+
+    ctrl.showSecretsWarning = function() {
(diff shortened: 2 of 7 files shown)


jwforres:
yep this is definitely better. the last difference i see is that for env vars if you fill out the key but not the value you get a validation error on the key input

webhookvalidation

I think the same should be true for the webhooks, if you pick a secret but dont pick a type we should probably warn you.  I can't decide if the opposite should be true, if you pick a type but dont pick a secret. Would it be a bad experience if you pick the first dropdown and immediately get a validation warning about the second dropdown that you haven't even had a chance to fill out yet?  @spadgett @ncameronbritt thoughts?

ncameronbritt:
I don't think it would be a good experience for a user to get that validation warning before they've had a chance to make a selection--they haven't done anything wrong at that point. One possibility would be to use progressive disclosure and hide the select secret ref dropdown until the user chooses a webhook type. I guess that would differ from env vars, but it could make validation more straight-forward.

jwforres:
Hmm I dont think that actually helps the validation issue, it would be invalid the second we showed it to them. I think its possible to get the validation behavior we want which is we dont tell them there is an error until they focus leaves the input pair.

@jhadvig I think lets use this PR to make this main behavior change and look into validation in a followup

jhadvig:
@jwforres ok, so will add the dist in that case

ncameronbritt:
I would think that it wouldn't be invalid until that dropdown loses focus.

jhadvig:
@jwforres updated both components (copy-webhook-url and osc-webhook-triggers)with the structure we use in the directive folder.
Tested the new conde and also dist an we should be good now.

spadgett:
> I would think that it wouldn't be invalid until that dropdown loses focus.

I believe that's what we do for the env var keys. We only show the error after the field loses focus.

app/styles/_core.less
```diff
@@ -1105,6 +1105,9 @@ a.disabled-link {
     }
     .add-webhook-action-btns {
         margin-top: 15px;
+        .btn-link {
+          padding-left: 0px;
```
jwforres: instead of a custom style here, just add the pad-left-none class to the link

spadgett: We might want a generic style that removes the padding from `btn-link` so we can replace `<a>` elements

jhadvig:
@jwforres @spadgett @ncameronbritt comments addressed. Added a warning for both secret type and reference selectboxes (since both are required). When the user wont pick any of the values the warning under that particular selectbox will be displayed.
focus

jwforres:
hmm its not quite right, it shouldn't be warning at all unless you have actually set one of the two input boxes to something but not the other one

app/styles/_core.less
spadgett: > is bootstrap putting padding on btn-link? wonder why they would do that

As far as I can it's never set explicitly to 0 and `.btn` has padding.

cc @rhamilto

jwforres:
@jhadvig the issue with the component injection is affecting other parts of the UI ( see #2682 ), we need to get the original set of fixes in. Please stash your validation changes and move those to a followup PR as I originally suggested so that we can get the other fixes in.

jwforres:
/lgtm

going ahead with this set of changes to get the P1 fix in, we can talk about a generic fix for btn-link later, and the validation stuff will be a separate PR

PR #2680, opened 2018-01-18
Bug 1535976 - Fix input name in the create-secret form

jhadvig:
Fixed wrong `input` names in the `create-secret.html` 
@jwforres PTAL

### DIFF
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -26,6 +26,11 @@ angular.module("openshiftConsole")
       templateUrl: 'views/directives/create-secret.html',
       link: function($scope) {
         $scope.nameValidation = DNS1123_SUBDOMAIN_VALIDATION;
+        $scope.secretReferenceValidation = {
+          pattern: /^[a-zA-Z0-9\-_]+$/,
+          minLength: 8,
+          description: 'Secret reference key must consist of lower-case, upper-case letters, numbers, dash, and underscore.'
+        };
 
         $scope.secretAuthTypeMap = {
           image: {
--- a/app/views/directives/create-secret.html
+++ b/app/views/directives/create-secret.html
@@ -12,10 +12,10 @@
   <div ng-if="newSecret.type">
     <div class="form-group">
       <label for="secret-name" class="required">Secret Name</label>
-        <span ng-class="{'has-error': nameTaken || (secretForm.secretName.$invalid && secretForm.secretName.$touched)}">
+        <span ng-class="{'has-error': nameTaken || (secretForm.name.$invalid && secretForm.name.$touched)}">
           <input class="form-control"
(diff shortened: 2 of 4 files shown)


jwforres:
the bug says webhook secret key is not being validated, I dont see how this fixes that issue?

jhadvig:
@jwforres so since we are not doing any kind of validation of the secret key, but I've noticed that the secret name was not validated properly, since the input name was different from the name that that was checked on the form (and for some other input fields as well).
Sorry I've forgot to clarify it in the first comment.

jwforres:
ok i know on the original set of changes i said not to worry about the validation, but now we can probably look into it as a follow-on here. webhook secret key needs to be a valid URL path segment. Thats probably the only validation we should try to do.

jhadvig:
@jwforres PTAL

app/scripts/directives/createSecret.js
```diff
@@ -26,6 +26,11 @@ angular.module("openshiftConsole")
       templateUrl: 'views/directives/create-secret.html',
       link: function($scope) {
         $scope.nameValidation = DNS1123_SUBDOMAIN_VALIDATION;
+        $scope.secretReferenceValidation = {
+          pattern: /^[a-zA-Z0-9\-_]+$/,
+          minLength: 8,
+          description: 'Secret reference key must consist of lower-case, upper-case letters, numbers, bash and underscore.'
```
jwforres: dash not bash, also we use oxford commas here

"dash, and underscore"

app/views/directives/create-secret.html
```diff
@@ -70,6 +72,16 @@
...
           <a ng-href="{{'webhooks' | helpLink}}" target="_blank"><span class="learn-more-inline">Learn More&nbsp;<i class="fa fa-external-link" aria-hidden="true"></i></span></a>
         </div>
+        <div class="has-error" ng-show="secretForm.webhookSecretKey.$error.pattern && secretForm.webhookSecretKey.$touched">
+          <span class="help-block">
+            {{secretReferenceValidation.description}}
+          </span>
+        </div>
+        <div class="has-error" ng-show="secretForm.webhookSecretKey.$error.minlength && secretForm.webhookSecretKey.$touched">
+          <span class="help-block">
+            Minimum character count is {{secretReferenceValidation.minLength}}.
```
jwforres: looks like you got this message from the env editor, but I prefer what we say in other places which is something like:

`The key must have at least 8 characters.`

app/scripts/directives/createSecret.js
jhadvig: always forgetting about the commas. We dont use int with 'and' ... sorry

jhadvig:
@jwforres comments addressed. added dist as well.

PR #2683, opened 2018-01-18
Bug 1534316: Check for new templateServiceBrokerEnabled flag

spadgett:
Check for new flag from https://github.com/openshift/origin-web-console-server/pull/21

https://bugzilla.redhat.com/show_bug.cgi?id=1534316

/assign @jwforres

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -32,7 +32,8 @@
       logout_uri: ""
     },
     loggingURL: "",
-    metricsURL: ""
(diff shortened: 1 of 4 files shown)


jwforres:
@spadgett update the checked in config.js to include the field, that file is supposed to match a valid output of the server

spadgett:
@jwforres thanks, updated

PR #2685, opened 2018-01-18
Partial fix for bug 1440004

TheRealJon:
When build history bar chart is shown and a new build is added to the chart, sometimes the duration is initially set to 0 until the page is reloaded. This was mostly observed by setting the build config `source` property to an invalid url to generate a failed build with a very short duration.

Added a check for the `completionTimestamp` property in the `isIncompleteBuild` filter to ensure that any build without a completion timestamp is considered incomplete.

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -529,7 +529,7 @@ angular.module('openshiftConsole')
   })
   .filter('isIncompleteBuild', function(ageLessThanFilter) {
     return function(build) {
-      if (!build || !build.status || !build.status.phase) {
+      if (!build || !build.status || !build.status.phase){
         return false;
       }
 
@@ -539,6 +539,9 @@ angular.module('openshiftConsole')
         case 'Running':
           return true;
         default:
+          if (!build.status.completionTimestamp) {
+            return true;
+          }
           return false;
       }
     };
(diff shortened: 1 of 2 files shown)


app/scripts/filters/resources.js
```diff
@@ -529,8 +529,8 @@ angular.module('openshiftConsole')
   })
   .filter('isIncompleteBuild', function(ageLessThanFilter) {
     return function(build) {
-      if (!build || !build.status || !build.status.phase) {
-        return false;
+      if (!build || !build.status || !build.status.phase || !build.status.completionTimestamp) {
+        return true;
```
jwforres: part of this change is risky because this filter is used in so many places and may be expecting the behavior that a null/undefined build doesn't return true here.  I think everything except the `!build` check is ok to return true.  I would still return false for `!build`

If we actually have an issue in the build chart specifically with null / undefined builds, then we should fix the `completeBuilds` function in builds.js to specifically exclude those.

TheRealJon: Yeah, that makes sense. I didn't consider that this could cause null/undefined references to be returned. The problem wasn't that the builds are null/undefined, but that the `completionTimestamp` property was undefined. My train of thought was that this function should return true if that property is not defined (if it doesn't have a completionTimestamp then it's not complete). Maybe I should just add another check for that and leave the first conditional as it was?

jwforres: thats probably the safest change

PR #2686, opened 2018-01-19
add spec test for homePagePreference service

yapei:
Add spec test for homePagePreference service, could you please help review?

### DIFF
--- a/test/spec/services/homePagePreferenceSpec.js
+++ b/test/spec/services/homePagePreferenceSpec.js
@@ -1,0 +1,57 @@
+"use strict";
+
+describe("HomePagePreferenceService", function(){
+  var HomePagePreferenceService;
+  var store = {};
+
+  beforeEach(function(){
+    inject(function(_HomePagePreferenceService_){
+      HomePagePreferenceService = _HomePagePreferenceService_;
+    });
+    
+    // LocalStorage mock
+    spyOn(localStorage, 'getItem').and.callFake(function(key) {
+        return store[key];
+    });
+    
+    spyOn(localStorage, 'setItem').and.callFake(function (key, value) {
+    /* jshint boss:true */
+    return store[key] = value + '';
+    });
+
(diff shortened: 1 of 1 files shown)


spadgett:
@yapei 👍 Terrific, thank you.

I see one jshint error, which is causing Jenkins failure:

```
test/spec/services/homePagePreferenceSpec.js
  line 18  col 24  Did you mean to return a conditional instead of an assignment?
```

test/spec/services/homePagePreferenceSpec.js
```diff
@@ -0,0 +1,56 @@
...
+      HomePagePreferenceService = _HomePagePreferenceService_;
+    });
+    
+    // LocalStorage mock
+    spyOn(localStorage, 'getItem').and.callFake(function(key) {
+        return store[key];
+    });
+    
+    spyOn(localStorage, 'setItem').and.callFake(function (key, value) {
+    return store[key] = value + '';
```
spadgett: Try this to avoid the jshint error

```js
spyOn(localStorage, 'setItem').and.callFake(function (key, value) {
  /* jshint boss:true */
  return store[key] = value + '';
});
```

spadgett:
@dtaylor113 fyi

test/spec/services/homePagePreferenceSpec.js
yapei: Thanks Sam, updated! A little curious, do I need to change the assignment expression instead of just adding `/* jshint boss:true */` to avoid errors?

spadgett: I think what you have is fine. The linter is warning in case you meant `==` instead of `=` here.

yapei:
@dtaylor113  thank you for review, I could add them later

test/spec/services/homePagePreferenceSpec.js
yapei: Thank you Sam!

PR #2689, opened 2018-01-19
Webhook trigger editor should be more consistant with envVar editor

jhadvig:
Here is the Webhook trigger editor validation behavior update. Now the editor should behave like envVar editor:

- user can add as many rows as he wants but only those that have selected webhook Type and Secret  will be saved
- user will be alerted with a `help-block` that he is missing either webhook Type or Secret, see gif:
openshift web console


jhadvig:
@jwforres PTAL

jwforres:
screenshot looks like exactly what I would expect

benjaminapetersen:
If I don't have a secret, the input box is a little confusing:



It works well though if I first use the create link:



Should we message the user if there are no secrets available, or perhaps hide "add webhook" and only show the "create new webhook secret" link?

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
```diff
@@ -20,7 +21,7 @@
   });
 
 
-  function OscWebhookTriggers($scope, $uibModal, $filter, APIService) {
+  function OscWebhookTriggers($scope, $uibModal, $filter, $timeout, APIService) {
     var ctrl = this;
 
     ctrl.$onInit = function() {
```
benjaminapetersen: Not part of this PR but there is a `$scope.namespace` and `$scope.type` right here that seems to be unused.

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
```diff
@@ -95,22 +96,15 @@
       checkDuplicates(trigger);
     };
 
-    var addEmptyWebhookTrigger = function() {
+    ctrl.addEmptyWebhookTrigger = function() {
```
benjaminapetersen: if you update to `ctrl.addEmptyWebhookTrigger` here, you'll have to update the `addEmptyWebhookTrigger()` calls below, ~113.

app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -1,38 +1,53 @@
...
+      <div class="select-webhook-type">
+        <ui-select 
+          ng-model="trigger.data.type"
+          name="triggerType{{$index}}"
+          ng-disabled="$ctrl.isDeprecated(trigger)"
+          on-select="$ctrl.triggerTypeChange(trigger)"
+          search-enabled="false"
+          title="Select a webhook type"
+          ng-class="{'has-error': $ctrl.secretsForm.triggerType{{$index}}.$touched && !trigger.data.type && trigger.data[trigger.data.type.toLowerCase()].secretReference.name }"
+          focus-on="triggerTypeFocus{{$index}}"
```
benjaminapetersen: I don't think `triggerTypeFocus` exists as a function, though I see you broadcasting an `event` [up here](https://github.com/openshift/origin-web-console/pull/2689/files#diff-9d51eb270d91c5c6831894971b883c5cR107).  I can't quite tell that it is working.  It looks like the example just expects a [callback function](https://github.com/angular-ui/ui-select/wiki/ui-select#examples-set-focus).

app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -1,38 +1,53 @@
...
+          flex>
+          <ui-select-match placeholder="Webhook type">
+            {{ $select.selected.label }}
+          </ui-select-match>
+          <ui-select-choices repeat="option.label as option in $ctrl.webhookTypesOptions">
+            {{ option.label }}
+          </ui-select-choices>
+        </ui-select>
+        <div class="has-error" ng-show="$ctrl.secretsForm.triggerType{{$index}}.$touched && !trigger.data.type && trigger.data[trigger.data.type.toLowerCase()].secretReference.name">
```
benjaminapetersen: We might be able to reduce the duplication of `$ctrl.secretsForm.triggerType{{$index}}.$touched && !trigger.data.type && trigger.data[trigger.data.type.toLowerCase()].secretReference.name` with a named `$ctrl` function?

app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -1,38 +1,53 @@
...
-        </ui-select-match>
-        <ui-select-choices repeat="webhookSecret.metadata.name as webhookSecret in ($ctrl.webhookSecrets | filter : $select.search) track by (webhookSecret | uid)">
-          <div ng-bind-html="webhookSecret.metadata.name | highlight : $select.search"></div>
-        </ui-select-choices>
-      </ui-select>
+      <div class="select-secret-ref">
+        <ui-select
+          ng-if="!$ctrl.isDeprecated(trigger)"
+          ng-model="trigger.data[trigger.data.type.toLowerCase()].secretReference.name"
+          name="triggerSecretRef{{$index}}"
```
benjaminapetersen: This is probably a nit, but wondering if we would be helped by putting a prefix on our input names, like `inputTriggerSecretRef`.  I think for `keyValueEditor` we used utils that generated `names-with-hyphens` for the inputs, which helps a little bit as well.  Clues that help you understand you are dealing with form inputs rather than data variables are nice.

app/components/osc-webhook-triggers/osc-webhook-triggers.html
benjaminapetersen: Just for my understanding, what is the `.toLowerCase()` for?

app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -1,38 +1,53 @@
...
-        </ui-select-match>
-        <ui-select-choices repeat="option.label as option in $ctrl.webhookTypesOptions">
-          {{ option.label }}
-        </ui-select-choices>
-      </ui-select>
+      <div class="select-webhook-type">
+        <ui-select 
+          ng-model="trigger.data.type"
+          name="triggerType{{$index}}"
+          ng-disabled="$ctrl.isDeprecated(trigger)"
```
benjaminapetersen: What happens if a trigger is deprecated? Is there a reason to notify the user?

app/components/osc-webhook-triggers/osc-webhook-triggers.html
jhadvig: It should work, when user clicks on the `Add webhook` the new trigger type selectbox should be focused. Not really sure whats the issue, `focus-on` should take an event as a parameter. In the example ng-click triggers a function which calls the broadcast on the element with desired `focus-on` string.

app/components/osc-webhook-triggers/osc-webhook-triggers.html
jhadvig: Because of the [webhook trigger format](https://docs.openshift.org/latest/dev_guide/builds/triggering_builds.html#webhook-triggers):
```
type: "GitHub"
github:
  secretReference:
    name: "mysecret"
```

app/components/osc-webhook-triggers/osc-webhook-triggers.html
jhadvig: So webhook triggers that contain only plain inline secret, eg:
```
    - generic:
        secret: beb28e70e4b93d26
      type: Generic
```
Are deprecated and they cant be edited or added in the editor, just viewed and removed, Thats why both of their selectboxes are disabled.

app/components/osc-webhook-triggers/osc-webhook-triggers.html
benjaminapetersen: Ok cool, wanted to ask & make sure.  Have we used events like this before?  I wonder if it would be good to have a naming convention for the variable.

jhadvig:
@benjaminapetersen updated the PR and addressed comments together with answers to your questions :)
I've added an empty entry to the webhook secrets, so the select box has the same behavior as other secrets selectboxes. Demo:
openshift web console

app/components/osc-webhook-triggers/osc-webhook-triggers.html
benjaminapetersen: Meant for the `name="triggerSecretRef{{$index}}"`, something like `name="trigger-secret-ref-input-{{index}}"`  This is more of a thought/suggestion, not sure its necessary.

app/components/osc-webhook-triggers/osc-webhook-triggers.html
benjaminapetersen: Ah, so would it be helpful to have a message telling the user "this is deprecated & cannot be edited"?

benjaminapetersen:
@jhadvig great! How do you feel about hiding the "add webhook" link until there are secrets that can be used (user would be required to `create new webhook secret` first as they would only see that link).   Not sure if that would be a helpful funnel?

jhadvig:
@benjaminapetersen dont have any strong feelings cause the webhook trigger editor is designed in the same way the evnVar editor, meaning you can add as many triggers as you want but only those that have properly selected both values will be saved.
@jwforres @ncameronbritt any suggestions ? :)

ncameronbritt:
I don't think hiding the "add webhook" is a great idea because then it's not obvious that you can even add them. I think it would be better to include "create new webhook secret" in the secret dropdown and allow users to create the secret from there.

jhadvig:
@ncameronbritt I would rather keep the 'create new webhook secret' as it it since its consistent with the other secret types user can manipulate with. eg: source secret, build secret, ..

ncameronbritt:
I was suggesting to have it in that dropdown in addition to where it currently is. 

In the current implementation, it's not obvious what a user should do first. The user's goal is to create a webhook. I feel like by adding that action to the dropdown, a user can either begin by creating a secret, or by creating the webhook--we don't assume how they would think about it.

jhadvig:
Are we using this type of logic, to have an action together with options in a selectbox, somewhere in the project?
If there will be anything missing for the webhook secret creation the error help-block should notify user whats missing.

ncameronbritt:
We do that when you provision a service from the landing page:
screen shot 2018-01-22 at 11 29 42 am

This case is slightly different from the environment variable case:
- There is no "create secret" action as part of the env var page
- The "Add Value from Config Map or Secret" action on the env var page sort of implies that those things already exist. Nothing about "Add Webhook" gives that same sense.

Interested in others' thoughts here as well. @jwforres @openshift/team-ux-review

ncameronbritt:
One other thing that came up with the designers today @jhadvig: for the create a secret dialog, the buttons should be on the right side, with the primary action on the right.

jhadvig:
@ncameronbritt I've addressed the comment regarding the modal btns position, since rest will be part of effort to refactor BC/DC editors.
screen:
download

@benjaminapetersen based on [comment](https://github.com/openshift/origin-web-console/pull/2689#issuecomment-359555101) I've to update create-secret directive, by wrapping the modal body and footer into `modal-body` and `modal-footer` classes, which unfortunately created a big diff for the `create-secret.html` file :sob: 

PTAL

jhadvig:
cc'ing @jwforres

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
```diff
@@ -95,28 +96,34 @@
...
+        return ctrl.secretsForm[fullSelectBoxName].$touched && !type && secretReferenceName;
+      } else if (selectBoxName === 'triggerSecretRef') {
+        return ctrl.secretsForm[fullSelectBoxName].$touched && type && !secretReferenceName;
+      }
+      return false;
     };
 
     // If there are no webhook triggers create empty one, otherwise check for duplicates.
     // In case of deprecated secret format add a `secretInputType` field to the object so we can toggle secret visibility.
     if (_.isEmpty(ctrl.webhookTriggers)) {
```
benjaminapetersen: Quick Q, this block got pushed down into the middle of a whole set of `ctrl`, `ctrl`,`ctrl` functions, but looks like it runs when the component initializes.  Can we push it up to the top, by the `var ctrl = this;`?  Perhaps even put it in a named function && call it from within `$init` may be better?

app/scripts/controllers/edit/buildConfig.js
```diff
@@ -265,6 +265,8 @@ angular.module('openshiftConsole')
                 var secretsByType = SecretsService.groupSecretsByType(secrets);
                 var secretNamesByType =_.mapValues(secretsByType, function(secrets) {return _.map(secrets, 'metadata.name');});
                 $scope.webhookSecrets = SecretsService.groupSecretsByType(secrets).webhook;
+                // Add empty option to the webhook trigger secrets
```
benjaminapetersen: Consider switching this comment to `why` rather than `what`.  IE, I see the `.unshift('')`, so I can tell it's adding an option, but I'm guessing there was a reason you wanted to leave this comment.

jhadvig:
@benjaminapetersen comments addressed.
I've put the the if-else block into a function that will be called in the `$onInit`.
Also added the dist. PTAL :)

benjaminapetersen:
Great! Sorry to `nit`, just didn't realize that block existed till the very end. :)

jhadvig:
@jwforres can we merge this ? :)

app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -1,38 +1,53 @@
...
+          on-select="$ctrl.triggerTypeChange(trigger)"
+          search-enabled="false"
+          title="Select a webhook type"
+          ng-class="{'has-error': $ctrl.missingInputError('triggerType', $index, trigger.data.type, trigger.data[trigger.data.type.toLowerCase()].secretReference.name) }"
+          focus-on="triggerTypeFocus{{$index}}"
```
spadgett: It's unclear to me where the `focus-on` directive is defined. Is this a dependency?

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
```diff
@@ -20,9 +21,57 @@
   });
 
 
-  function OscWebhookTriggers($scope, $uibModal, $filter, APIService) {
+  function OscWebhookTriggers($scope, $uibModal, $filter, $timeout, APIService) {
```
spadgett: We're trying to keep these lists alphabetized since they can get long and hard to maintain

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
```diff
@@ -20,9 +21,57 @@
...
+
+    ctrl.addEmptyWebhookTrigger = function() {
+      ctrl.webhookTriggers.push({
+        lastTriggerType: "",
+        data: {
+          type: ""
+        }
+      });
+      var numberOfTriggers = ctrl.webhookTriggers.length - 1;
+      $timeout(function() {$scope.$broadcast('triggerTypeFocus' + numberOfTriggers);});
```
spadgett: style nit: please split onto multiple lines so it's easier to read

```js
$timeout(function() {
  $scope.$broadcast('triggerTypeFocus' + numberOfTriggers);
});
```

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
```diff
@@ -20,9 +21,57 @@
...
+    };
+
+    ctrl.addEmptyWebhookTrigger = function() {
+      ctrl.webhookTriggers.push({
+        lastTriggerType: "",
+        data: {
+          type: ""
+        }
+      });
+      var numberOfTriggers = ctrl.webhookTriggers.length - 1;
```
spadgett: Shouldn't numberOfTriggers be the same as length? I think you mean index?

```js
var newTriggerIndex = ctrl.webhookTriggers.length - 1;
```

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
```diff
@@ -95,42 +127,19 @@
...
-      });
-    };
-
-    // Check last trigger if it's type and secret are selected.
-    ctrl.checkLastAndAddNew = function() {
-      var lastTrigger = _.last(ctrl.webhookTriggers);
-      var lastTriggerSecretData = $filter('getWebhookSecretData')(lastTrigger);
-      if (lastTrigger.data.type && (_.has(lastTriggerSecretData, 'secret') || _.has(lastTriggerSecretData, 'secretReference.name'))) {
-        addEmptyWebhookTrigger();
+    ctrl.missingInputError = function(selectBoxName, rowIndex, type, secretReferenceName) {
```
spadgett: It's better if we can use the `ng-*` attributes for this. For instance, set `ng-required` on the secret ref if the trigger type is set or `ng-required` on the trigger type if secret ref is set. I don't think these checks need JS.

If you *do* need to set form validity in JS, we should call `$ctrl.formName.fieldName.$setValidity` and check when the value changes. For example,

https://github.com/openshift/origin-web-console/blob/832cec030ff015081965dd6675e5b0fe98ec02c2/app/scripts/directives/editRequestLimit.js#L117-L118

But it's better to avoid if not necessary.

app/components/osc-webhook-triggers/osc-webhook-triggers.html
jhadvig: Thats a ui-select directive

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
jhadvig: Originally the logic was in the `ng-*` attributes, but based on [conversation](https://github.com/openshift/origin-web-console/pull/2689#discussion_r162735611) we switched to JS, avoid dups in the html template, since for the `ng-class` needs to add/rm the `has-error` class and similar case was for the `ng-show` for the help blocks under the selectboxes.

Since we want to have similar experience then with the EnvVar editor, in which we dont block the `Save` btn if the any of the selectboxes is not selected. Only WH triggers that have selected type and secret will be saved.

spadgett: The original code is also different from what I had in mind. I think it might be straightforward to use the ng-required attributes on the select inputs themselves. Let me follow up.

> Since we want to have similar experience then with the EnvVar editor, in which we dont block the Save btn if the any of the selectboxes is not selected. Only WH triggers that have selected type and secret will be saved.

I don't agree with this behavior either :) If one of the inputs is filled in and the other is not, we should block save. That's invalid, and users might think they're saving changes they're not. I'd call this is a bug in the env editor if it's the current behavior. We should only allow save if both selects are empty (or both have values).

app/components/osc-webhook-triggers/osc-webhook-triggers.html
spadgett: OK, I see. Thanks!

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
jhadvig: @spadgett I understand your concerns and they make sense. I've discussed the design with @jwforres and @ncameronbritt and this what we align on. 
So I guess that before I rewrite the behavior we discuss whats the the proper way the editors should behave.

jwforres: I think there is some miscommunication happening in this thread :)
My opinion:
1) If there is an empty pair (or more than one pair) of inputs, it should not block Save
2) If a pair has partially filled out data, it should block Save. If either field is filled out but the other one is not, then it should not be possible to Save.
3) An empty pair should never be showing a warning
4) A partially filled out pair should not show a warning *until* the entire pair of inputs has lost focus.  This is to avoid filling out the first input, tabbing to the second input, and seeing a warning.

Notice I called out 2 & 4 separately. That is because as soon as you filled out the first input, you should block Save, even though you haven't warned the user yet.

spadgett: +1, this is also what I was thinking.

And I think we can implement it pretty simply with ng-required in the view without needing any JS in the controller.

ncameronbritt: That was my understanding. Thanks for clarifying, @jwforres.

jhadvig:
@spadgett comments addressed. The checks for adding the `has-error` classes are done in the template now. Also the behavior is corrected as @jwforres [described](https://github.com/openshift/origin-web-console/pull/2689#discussion_r164772974)
PTAL

app/components/osc-webhook-triggers/osc-webhook-triggers.component.js
```diff
@@ -20,9 +21,59 @@
...
+      });
+      var newTriggerIndex = ctrl.webhookTriggers.length - 1;
+      $timeout(function() {
+        $scope.$broadcast('triggerTypeFocus' + newTriggerIndex);
+      });
+    };
+
+    // Check if new or modified webhook trigger is a duplicate. If so, show a warning under the appropriate trigger
+    var checkDuplicates = function(trigger) {
+      var matchingTriggers = _.filter(ctrl.webhookTriggers, function(wehbookTrigger) {
```
spadgett: Check duplicates needs to avoid checking values when some of the data is not set. See the changes I had in the PR to your branch here: https://github.com/jhadvig/origin-web-console/pull/1/commits/0424279e9228b060b6e49469f7cd61cac9afe2e8#diff-9d51eb270d91c5c6831894971b883c5cR48

app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -1,38 +1,55 @@
...
-          {{ $select.selected.metadata.name }}
-        </ui-select-match>
-        <ui-select-choices repeat="webhookSecret.metadata.name as webhookSecret in ($ctrl.webhookSecrets | filter : $select.search) track by (webhookSecret | uid)">
-          <div ng-bind-html="webhookSecret.metadata.name | highlight : $select.search"></div>
-        </ui-select-choices>
-      </ui-select>
+      <div class="select-secret-ref">
+        <ui-select
+          ng-if="!$ctrl.isDeprecated(trigger)"
+          ng-model="trigger.data[trigger.data.type.toLowerCase()].secretReference.name"
```
spadgett: This could be problematic because you can set a value before trigger.data.type is set, so we'd be setting `trigger.data[''].secretReference.name`. I'm not sure what problems that might cause.

I think it's simpler just to disable the select until the trigger type is selected, which avoids a lot edge cases and unnecessary validation on the type field.

https://github.com/jhadvig/origin-web-console/pull/1/commits/0424279e9228b060b6e49469f7cd61cac9afe2e8#diff-08a15e28e59fecd178b2d160ea9a2362R31

app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -1,38 +1,55 @@
...
+        <ui-select 
+          ng-model="trigger.data.type"
+          name="triggerType{{$index}}"
+          ng-disabled="$ctrl.isDeprecated(trigger)"
+          on-select="$ctrl.triggerTypeChange(trigger)"
+          search-enabled="false"
+          title="Select a webhook type"
+          ng-class="{'has-error': $ctrl.secretsForm['triggerType{{$index}}'].$touched && !trigger.data.type && trigger.data[trigger.data.type.toLowerCase()].secretReference.name }"
+          focus-on="triggerTypeFocus{{$index}}"
+          ng-required="!trigger.data.type || trigger.data[trigger.data.type.toLowerCase()].secretReference.name"
```
spadgett: This logic doesn't look right. Type is required if type is not set? Isn't that the same thing as just saying required=true?

jhadvig: There was a digest loop when just checking for `trigger.data[trigger.data.type.toLowerCase()].secretReference.name` then the `trigger.data.type` wasnt set. But
> I think it's simpler just to disable the select until the trigger type is selected, which avoids a lot edge cases and unnecessary validation on the type field.

should fix it.

jhadvig:
@spadgett sorry for the confusion. Was checking for your PR but havent found it for some reason. Cherrypicked it and now its part of the PR... Not really sure if I should squash those two PRs since your commit dost not contain dist.

app/views/directives/create-secret.html
```diff
@@ -1,363 +1,367 @@
...
-    <button class="btn btn-primary"
-            type="button"
-            ng-disabled="secretForm.$invalid || secretForm.$pristine || invalidConfigFormat"
-            ng-click="create()">Create</button>
-    <button class="btn btn-default"
-            type="button"
-            ng-click="cancel()">Cancel</button>
+  <div class="modal-footer">
+    <div class="buttons gutter-top-bottom">
+      <button class="btn btn-lg btn-default"
```
spadgett: This should not be `btn-lg` any longer.

@rhamilto Can you review the changes to the dialog? It looks like we might have lost some of the changes from #2735. You probably want the diff w/o whitespace: https://github.com/openshift/origin-web-console/pull/2689/files?w=1

rhamilto: Because the [create secret directive gets used outside a modal](https://github.com/openshift/origin-web-console/blob/master/app/views/create-secret.html#L15-L20), we actually don't want to include any modal-specific markup in this directive.  I've [hacked the button layout for the modal with some CSS](https://github.com/openshift/origin-web-console/blob/784c0deb83d4ba9cc2479180f9abd26ad77281a5/app/styles/_core.less#L1004-L1015).

spadgett: @jhadvig Is there any reason you needed to change this file for this fix? I think it should be ok as-is in master

jhadvig:
@spadgett I've removed the changes that I've made in the modal, since they are not necessary anymore since they were made before @rhamilto's changes landed.
PTAL

PR #2690, opened 2018-01-19
Bug 1535374: Add console version to about page

spadgett:
Now that the console runs in a pod, it has its own version string.

Requires https://github.com/openshift/origin-web-console-server/pull/23
See https://bugzilla.redhat.com/show_bug.cgi?id=1535374

/assign @jwforres

### DIFF
--- a/app/views/about.html
+++ b/app/views/about.html
@@ -19,6 +19,8 @@
                   <dd>{{version.master.openshift || 'unknown'}}</dd>
                   <dt>Kubernetes Master:</dt>
                   <dd>{{version.master.kubernetes || 'unknown'}}</dd>
+                  <dt>OpenShift Web Console:</dt>
+                  <dd>{{version.console || 'unknown'}}</dd>
                 </dl>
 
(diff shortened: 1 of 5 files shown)


app/views/about.html
```diff
@@ -19,6 +19,8 @@ <h2 id="version">Version</h2>
                   <dd>{{version.master.openshift || 'unknown'}}</dd>
                   <dt>Kubernetes Master:</dt>
                   <dd>{{version.master.kubernetes || 'unknown'}}</dd>
+                  <dt>OpenShift Web Console:</dt>
```
jwforres: Web Console or Management Console? What are we actually calling it in the docs

spadgett: Hm, I thought we had switched everything to "Web Console." At least it's what we call it in the customization doc.

https://docs.openshift.org/latest/install_config/web_console_customization.html

spadgett: Yeah looking through the docs it seems to consistently be "Web Console"

PR #2691, opened 2018-01-19
Fix bug 14400044: Build chart legend hover issue

TheRealJon:
Add `order: null` config option to build history chart to prevent C3 from sorting legend items when `load()` API function is used.

[C3js Issue](https://github.com/c3js/c3/issues/1945)

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1440004.


jwforres:
@spadgett can you review since you wrote this

PR #2692, opened 2018-01-22
Remove System User, System Group from Membership page

benjaminapetersen:
This is a pretty simple change to remove the System User & System Group tabs.  I'll update the tests once I verify the other membership changes went it & won't conflict.  

screen shot 2018-01-21 at 9 47 06 pm

### DIFF
--- a/app/scripts/services/membership/membership.js
+++ b/app/scripts/services/membership/membership.js
@@ -60,26 +60,6 @@ angular
             "subjects":{
 
             }
-         },
-         "SystemUser":{
-            "kind":"SystemUser",
-            "sortOrder":4,
-            "description":"System users are virtual users automatically provisioned by the system.",
-            "helpLinkKey":"users_and_groups",
-            "name":"SystemUser",
-            "subjects":{
-
-            }
-         },
-         "SystemGroup":{
-            "kind":"SystemGroup",
-            "sortOrder":5,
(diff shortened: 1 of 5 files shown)


benjaminapetersen:
The only outstanding Q is whether or not we want to keep the tabs & have a deprecation notice for a release.  Chatting with @enj, we both tend to think this is probably unnecessary.  If the system versions are converted behind the scenes into normal users & groups, there should not be friction in the removal of the tabs.

jwforres:
Honestly I find this less confusing and less cluttered in general anyway.  As long as there is something in the docs that talks about where the system users/groups went, I'm good with just removing them completely.

jwforres:
@benjaminapetersen dist please

benjaminapetersen:
@jwforres yup, didn't bother updating the dist yet because I have to update the tests & the build would break anyway.  Will push an update shortly, now that the other membership PR went in.

benjaminapetersen:
Actually, this was a pretty trivial test update.

spadgett:
Do we need to remove the system user reference in role binding service?

benjaminapetersen:
Ah, yup, the `qualifySubject` fn, let me update that.  Thx @spadgett

benjaminapetersen:
Fixing test.

PR #2693, opened 2018-01-22
Remove extra 'reason' in membership prompt

benjaminapetersen:
Re [bugzilla 1535787](https://bugzilla.redhat.com/show_bug.cgi?id=1535787)

Initially just removing the extra `Reason:` in the string.  

Looking into @jwforres comment about:

     @ben looks like in one place you are not using errorDetailsFilter, not sure if this was an oversight or if there was a reason for it:

     https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/membership.js#L58

### DIFF
--- a/app/scripts/controllers/membership.js
+++ b/app/scripts/controllers/membership.js
@@ -55,7 +55,7 @@ angular
             exists: _.template('The role "<%= roleName %>" has already been granted to "<%= subjectName %>".')
           }
         },
-        errorReason: _.template('Reason: "<%= httpErr %>"')
+        errorReason: _.template('"<%= httpErr %>"')
       };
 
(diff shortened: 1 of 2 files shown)


jwforres:
/lgtm

letting this go in for the quick fix but definitely should figure out what's going on with errorDetailsFilter there

benjaminapetersen:
@jwforres just combed through the file, this is how that template is used:

```JavaScript 
messages.errorReason({httpErr: $filter('getErrorDetails')(err)})
```
So the `$filter('errorDetails')` output is always passed to `messages.errorReason`.  I think I can remove `messages.errorReason` entirely as it no longer does anything useful.  I'll PR this.

PR #2695, opened 2018-01-22
Trello/hpa/v2beta1/getHPAWarnings prep work

benjaminapetersen:
WIP because I only added a couple tests.  

@spadgett how do you feel about this change?  I'm aiming at breaking the large `getHPAWarnings` fn down into a couple smaller ones & adding tests as a foundation for my other changes. Pushing this up as its own PR in case you have reservations about the approach.

@jwforres

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -626,7 +626,7 @@ module.exports = function (grunt) {
                     // ['Nightmare'] is a good alt for a current headless
                     // FIXME: fix this, PhantomJS is deprecated
                     isMac ?
-                      ['Firefox', 'Chrome'] :
+                      ['Firefox'] :
                       ['PhantomJS']
       },
       unit: {
--- a/app/scripts/services/hpa.js
+++ b/app/scripts/services/hpa.js
@@ -73,6 +73,66 @@ angular.module("openshiftConsole")
     var humanizeKind = $filter('humanizeKind');
     var hasDeploymentConfig = $filter('hasDeploymentConfig');
 
+
+    var hasMetricsAvailableWarning = function(metricsAvailable) {
+      if (!metricsAvailable) {
+        return {
+          message: 'Metrics might not be configured by your cluster administrator. ' +
+                   'Metrics are required for autoscaling.',
+          reason: 'MetricsNotAvailable'
+        };
+      }
+    };
+
+
+    var hasCPURequestWarning = function(scaleTarget, limitRanges, project) {
+      var containers = _.get(scaleTarget, 'spec.template.spec.containers', []);
+      var kind;
+      if (!hasCPURequest(containers, limitRanges, project)) {
+        kind = humanizeKind(scaleTarget.kind);
+        return {
+          message: 'This ' + kind + ' does not have any containers with a CPU request set. ' +
+                   'Autoscaling will not work without a CPU request.',
+          reason: 'NoCPURequest'
+        };
+      }
+    };
+
+
+    var hasCompetingAutoscalersWarning = function(hpaResources) {
+      if (_.size(hpaResources) > 1) {
+        return {
+          message: 'More than one autoscaler is scaling this resource. ' +
+                   'This is not recommended because they might compete with each other. ' +
+                   'Consider removing all but one autoscaler.',
+          reason: 'MultipleHPA'
+        };
+      }
+    };
+
+
+    var hasCompetingDCAndAutoscalerWarning = function(scaleTarget, hpaResources) {
+      // Warn about replication controllers that have both an HPA and DC, but
+      // make sure an HPA targets the replication controller directly and
+      // not its parent DC.
+      var targetsRC = function() {
+        return _.some(hpaResources, function(hpa) {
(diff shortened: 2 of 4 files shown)


Gruntfile.js
```diff
@@ -626,7 +626,7 @@ module.exports = function (grunt) {
                     // ['Nightmare'] is a good alt for a current headless
                     // FIXME: fix this, PhantomJS is deprecated
                     isMac ?
```
benjaminapetersen: Ill squash this out, its just a tad annoying to see the output of all the tests twice while working on them

app/scripts/services/hpa.js
```diff
@@ -85,57 +146,13 @@ angular.module("openshiftConsole")
...
-            _.some(hpaResources, targetsRC)) {
-          warnings.push({
-            message: 'This deployment is scaled by both a deployment configuration and an autoscaler. ' +
-                     'This is not recommended because they might compete with each other.',
-            reason: 'DeploymentHasHPA'
-          });
-        }
-
-        return warnings;
+        return _.compact([
```
benjaminapetersen: This `compact` here is to eliminate a series of `if`'s/reduce complexity.

benjaminapetersen:
No longer WIP, covers the happy paths, but prob should test more as a follow later (it's a little time-consuming to write tests for older code).

@spadgett any objections to this change/adjustment?

spadgett:
Looks good. Can you run the same tests without your service changes to make sure we get the same results?

benjaminapetersen:
spadgett will do, ill give them a run now

benjaminapetersen:
Yup, ran against master, all pass.

Gruntfile.js
spadgett: Did you mean to leave this? (I'm actually OK taking it out, honestly, since I find it annoying, too. I don't think it gives much value, at least not for dev.)

test/spec/services/hpaServiceSpec.js
```diff
@@ -0,0 +1,205 @@
...
+  var defaultProject = {
+    metadata: {
+      name: ''
+    }
+  };
+
+
+
+  describe('#addCPURequestWarning', function() {
+
```
benjaminapetersen: `no no` bad boy 😄 

I'll fix this typo.

Gruntfile.js
benjaminapetersen: I'll leave this, the comments above clearly explain how to use `--browsers=Chrome,Firefox` to get multiple.

The coverage block is even more annoying, when you are trying to write tests, but I'm not sure how to fix that yet (keep it to annoy us, just not so much).

test/spec/services/hpaServiceSpec.js
benjaminapetersen: fixed

benjaminapetersen:
@spadgett minor typo fixed, agree we can leave the Gruntfile as is, ready to roll I think.

spadgett:
It looks like this and #2708 have some of the same changes? Which one did you want to go with?

benjaminapetersen:
@spadgett this one should go first. I branched #2708 off of this branch to build on it, and once this one merges I'll rebase it.  I wanted to leave room for clean breaks between reviews (though I realize this one is hard to review till I rebase post merge).

test/spec/services/hpaServiceSpec.js
```diff
@@ -0,0 +1,205 @@
...
+    describe('When there are no HPA resources', function() {
+      it('should return an empty set of messages (no errors)', function() {
+        var noHPA = [];
+        HPAService
+          .getHPAWarnings(defaultScaleTarget, noHPA, defaultLimitRanges, defaultProject)
+          .then(function(warnings) {
+            expect(_.size(warnings)).toEqual(0);
+          });
+        $rootScope.$digest();
```
spadgett: Why is the call to digest needed?

test/spec/services/hpaServiceSpec.js
```diff
@@ -0,0 +1,205 @@
...
+          .getHPAWarnings(defaultScaleTarget, defaultHpaResources, defaultLimitRanges, defaultProject)
+          .then(function(warnings) {
+            expect(_.head(warnings).reason).toEqual('MetricsNotAvailable');
+          });
+        $rootScope.$digest();
+      });
+    });
+
+
+    // TODO: we should be testing more than just the happy paths
```
spadgett: Some of the tests here might be helpful:

https://github.com/openshift/origin-web-console/blob/master/test/spec/services/limitsRangesServiceSpec.js

test/spec/services/hpaServiceSpec.js
benjaminapetersen: Force the promise resolution, I can check again if we can eliminate, but pretty sure `$q.when` doesn't move unless you force a digest in a test.

test/spec/services/hpaServiceSpec.js
benjaminapetersen: follow-on, or shall i flesh these out now?

PR #2697, opened 2018-01-22
Remove rolBindingsService.list

benjaminapetersen:
At one point `watch` was added & we stopped using `list`.  This PR removes the list function from the RoleBindings Service. We can eliminate the `qualifyBindingName` function by using the `generateName` prop.

@spadgett @jwforres

### DIFF
--- a/app/scripts/services/membership/roleBindings.js
+++ b/app/scripts/services/membership/roleBindings.js
@@ -9,26 +9,12 @@ angular
 
     var roleBindingsVersion = APIService.getPreferredVersion('rolebindings');
 
-    var cache = {};
-
-    // recurse until we get a unique binding name compared to what is cached
-    var qualifyBindingName = function(name, suffix) {
-      var search = suffix ?
-                name + suffix :
-                name;
-      return _.some(cache, _.matchesProperty('metadata.name', search)) ?
-              qualifyBindingName(name, _.uniqueId()) :
-              search;
-    };
-
     var bindingTPL = function(role, projectNamespace) {
-      var roleName = _.get(role, 'metadata.name');
-      var bindingName = roleName ? qualifyBindingName(roleName) : null;
       return {
         kind: 'RoleBinding',
         apiVersion: 'v1',
         metadata: {
-          name: bindingName,
+          generateName: _.get(role, 'metadata.name') + '-',
           namespace: projectNamespace,
         },
         roleRef: {
@@ -103,16 +89,7 @@ angular
         }));
     };
 
-    var list = function(context, fn, opts) {
-      return DataService
-              .list(roleBindingsVersion, context, function(resp) {
-                cache = resp.by('metadata.name');
-                fn(resp);
-              }, opts);
-    };
-
     return {
-      list: list,
       create: create,
       addSubject: addSubject,
       removeSubject: removeSubject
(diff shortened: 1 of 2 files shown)


app/scripts/services/membership/roleBindings.js
```diff
@@ -14,26 +14,12 @@ angular
...
-
     var bindingTPL = function(role, projectNamespace) {
-      var roleName = _.get(role, 'metadata.name');
-      var bindingName = roleName ? qualifyBindingName(roleName) : null;
       return {
         kind: 'RoleBinding',
         apiVersion: 'v1',
         metadata: {
-          name: bindingName,
+          generateName: _.get(role, 'metadata.name'),
```
benjaminapetersen: Tested, this will ensure a unique name if the name is taken.  For example, if there is a `rolebinding` that already exists named `view` (but for some reason has bindings for `admin` or something else), this will create a name like `view12345`.

spadgett: nit: usually generateName values have a trailing `-` (for instnance, `view-` so that the names are like `view-12345`)

benjaminapetersen: Ah, funny the api doesn't handle that.  Will update.

app/scripts/services/membership/roleBindings.js
```diff
@@ -9,26 +9,12 @@ angular
...
-
     var bindingTPL = function(role, projectNamespace) {
-      var roleName = _.get(role, 'metadata.name');
-      var bindingName = roleName ? qualifyBindingName(roleName) : null;
       return {
         kind: 'RoleBinding',
         apiVersion: 'v1',
         metadata: {
-          name: bindingName,
+          generateName: _.get(role, 'metadata.name').concat('-'),
```
benjaminapetersen: @spadgett got it here.

benjaminapetersen:
@spadgett should be good once tests pass.

app/scripts/services/membership/roleBindings.js
spadgett: It's generally better to use `+` rather than `concat()` to concat strings. See

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat

If we're going to do this, no reason to use `_.get` because it will fail if undefined anyway. So just

```js
role.metadata.name.concat('-')
```

But I'd still prefer

```js
role.metadata.name + '-'
```

spadgett:
LGTM, just one comment on concat

app/scripts/services/membership/roleBindings.js
benjaminapetersen: yeah think i failed a test too actually, updating.

app/scripts/services/membership/roleBindings.js
```diff
@@ -9,26 +9,12 @@ angular
...
-
     var bindingTPL = function(role, projectNamespace) {
-      var roleName = _.get(role, 'metadata.name');
-      var bindingName = roleName ? qualifyBindingName(roleName) : null;
       return {
         kind: 'RoleBinding',
         apiVersion: 'v1',
         metadata: {
-          name: bindingName,
+          generateName: _.get(role, 'metadata.name') + '-',
```
benjaminapetersen: Updated, had to keep the _.get() cause otherwise bunch of test fail, prob because I don't pass a role in the unit tests (dunno), and don't think there is reason to update them atm.

benjaminapetersen:
test passed, good to go @spadgett ?

spadgett:
> ERROR - Unable to start a WebDriver session.

/retest

PR #2705, opened 2018-01-23
Bump common version to resolve bugs

### LINKED ISSUE #2688: See All / Collapse link moving is weird
I understand why it's like it is, but I find the link moving strange. Here is the behavior for config maps, but it's a little weird in other places too like expanding a full command in a pod template.

screencast 2018-01-19 09-35-54

/kind bug
/priority p3
/area usability
/assign @rhamilto @sg00dwin

rhamilto:
Fixes #2688


spadgett:
@liggitt FYI

PR #2707, opened 2018-01-23
[3.7] Bump origin-web-common to 3.7.1

spadgett:

/assign @jwforres 
/cc @liggitt


jwforres:
/lgtm

I'm actually not sure if there will be more 3.8 builds, but doesn't hurt to have it there just in case

PR #2708, opened 2018-01-23
hpa v2beta1 api warning on DC & edit autoscaler pages

benjaminapetersen:
This PR depends on [common 285](https://github.com/openshift/origin-web-common/pull/285).

[WIP] notes:

`bower link origin-web-common` with the above PR or the tests will fail. 

On the DC page:

screen shot 2018-01-23 at 6 21 58 pm

- probably only want to show the v2beta1 message, will hide the other

On the edit autoscaler page:

screen shot 2018-01-23 at 6 25 12 pm

- shows a message w/a link to CLI
- hides the CPU request input as editing this is probably not what the user wants to do

TODO (perhaps follow-on PR):

- Update the warning message to show the user the data from the `metrics` annotation so they can easily see what the autoscaler is using to scale the resource.

### DIFF
--- a/app/scripts/controllers/edit/autoscaler.js
+++ b/app/scripts/controllers/edit/autoscaler.js
@@ -78,6 +78,7 @@ angular.module('openshiftConsole')
     var horizontalPodAutoscalerVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
     var limitRangesVersion = APIService.getPreferredVersion('limitranges');
 
+
     ProjectsService
       .get($routeParams.project)
       .then(_.spread(function(project, context) {
@@ -185,6 +186,8 @@ angular.module('openshiftConsole')
                                 value: val
                               };
                             });
+
+          $scope.usesV2Metrics = HPAService.usesV2Metrics(resource);
 
           // Are we editing an existing HPA?
           if ($routeParams.kind === "HorizontalPodAutoscaler") {
--- a/app/scripts/services/hpa.js
+++ b/app/scripts/services/hpa.js
@@ -2,6 +2,9 @@
 
 angular.module("openshiftConsole")
   .factory("HPAService", function($filter, $q, LimitRangesService, MetricsService) {
+
+    var annotation = $filter('annotation');
+
     // Checks if all containers have a value set for the compute resource request or limit.
     //
     // computeResource  - 'cpu' or 'memory'
@@ -99,6 +102,13 @@ angular.module("openshiftConsole")
     };
 
 
+    var hasV2HPAAnnotations = function(hpaResources) {
+      return _.some(hpaResources, function(hpa) {
+        return annotation(hpa, 'autoscaling.alpha.kubernetes.io/metrics');
+      });
+    };
+
+
     var hasCompetingAutoscalersWarning = function(hpaResources) {
       if (_.size(hpaResources) > 1) {
         return {
@@ -146,14 +156,18 @@ angular.module("openshiftConsole")
         return $q.when([]);
       }
       return MetricsService.isAvailable().then(function(metricsAvailable) {
+        var isV2HPA = hasV2HPAAnnotations(hpaResources);
         return _.compact([
           hasMetricsAvailableWarning(metricsAvailable),
-          hasCPURequestWarning(scaleTarget, limitRanges, project),
+          // we don't need to show this warning, but if we have it, we should also
+          // not show the cpuRequestWarning, until we update to the newer API
+          isV2HPA ? false : hasCPURequestWarning(scaleTarget, limitRanges, project),
           hasCompetingAutoscalersWarning(hpaResources),
           hasCompetingDCAndAutoscalerWarning(scaleTarget, hpaResources)
         ]);
       });
     };
+
 
     // Group HPAs by the object they scale.
     //
@@ -185,7 +199,13 @@ angular.module("openshiftConsole")
       return hpaByResource;
     };
 
+    // currently unsupported by the webconsole
+    var usesV2Metrics = function(hpa) {
+      return hasV2HPAAnnotations([hpa]);
+    };
+
     return {
+      usesV2Metrics: usesV2Metrics,
       hasCPURequest: hasCPURequest,
       filterHPA: filterHPA,
       getHPAWarnings: getHPAWarnings,
--- a/app/views/directives/osc-autoscaling.html
+++ b/app/views/directives/osc-autoscaling.html
@@ -99,7 +99,7 @@
       </span>
     </div>
   </div>
-  <div class="form-group">
+  <div ng-show="showRequestInput" class="form-group">
     <label>
       CPU Request Target
     </label>
--- a/app/views/edit/autoscaler.html
+++ b/app/views/edit/autoscaler.html
@@ -38,11 +38,21 @@
               <span ng-if="!('cpu' | isRequestCalculated : project)">request.</span>
             </div>
 
+            <!-- Warning: V2Beta1 HPA -->
+            <div ng-if="usesV2Metrics" class="alert alert-warning">
+              <span class="pficon pficon-warning-triangle-o" aria-hidden="true"></span>
+              <span class="sr-only">Warning:</span>
+              This autoscaler uses a newer API, consider editing it with the
+              <a href="command-line" target="_blank">command line tools</a>.
+            </div>
+
+
             <fieldset ng-disabled="disableInputs" class="gutter-top">
               <osc-autoscaling
                   model="autoscaling"
                   show-name-input="true"
-                  name-read-only="kind === 'HorizontalPodAutoscaler'">
+                  name-read-only="kind === 'HorizontalPodAutoscaler'"
+                  show-request-input="autoscaling.targetCPU">
               </osc-autoscaling>
 
               <label-editor
--- a/test/spec/services/hpaServiceSpec.js
+++ b/test/spec/services/hpaServiceSpec.js
@@ -2,8 +2,8 @@
 
 describe('HPAService', function() {
   var HPAService;
-  var $rootScope;
   var $q;
+  var $rootScope;
   var $window;
 
   var metricsAreAvailable;
@@ -23,8 +23,8 @@ describe('HPAService', function() {
 
 
   beforeEach(inject(function(_$q_, _$rootScope_, _$window_, _HPAService_) {
-    $rootScope = _$rootScope_;
     $q = _$q_;
+    $rootScope = _$rootScope_;
     $window = _$window_;
     HPAService = _HPAService_;
   }));
@@ -80,6 +80,7 @@ describe('HPAService', function() {
 
   describe('#addCPURequestWarning', function() {
 
+
     describe('When there are no HPA resources', function() {
       it('should return an empty set of messages (no errors)', function() {
         var noHPA = [];
@@ -107,58 +108,84 @@ describe('HPAService', function() {
 
 
     // TODO: we should be testing more than just the happy paths
-    describe('When there is a CPU request', function() {
-      // this is the first check in the hasCPURequest fn
-      // it should use $window && set OPENSHIFT_CONFIG.clusterResourceOverridesEnabled
-      // to trigger the LimitRangesService.hasClusterResourceOverrides function
-      // it('should not return a warning when cluster resource overrides are enabled', function() {
-      //
-      // });
-
-      // 2nd check in the hasCPURequest fn
-      // need to test this with some mock limit ranges passed to the fn
-      // it('should not return a warning when containers or limit ranges have a cpu request', function() {
-      //
-      // });
-
-      // 3rd check in the hasCPURequest fn
-      // more tests with mock limit ranges
-      // it('should not return a warning when the containers or limit ranges have a cpu limit', function() {
-      //
-      // });
-
-    });
+    // describe('When there is a CPU request', function() {
+    //   // this is the first check in the hasCPURequest fn
+    //   // it should use $window && set OPENSHIFT_CONFIG.clusterResourceOverridesEnabled
+    //   // to trigger the LimitRangesService.hasClusterResourceOverrides function
+    //   // it('should not return a warning when cluster resource overrides are enabled', function() {
+    //   //
+    //   // });
+    //
+    //   // 2nd check in the hasCPURequest fn
+    //   // need to test this with some mock limit ranges passed to the fn
+    //   // it('should not return a warning when containers or limit ranges have a cpu request', function() {
+    //   //
+    //   // });
+    //
+    //   // 3rd check in the hasCPURequest fn
+    //   // more tests with mock limit ranges
+    //   // it('should not return a warning when the containers or limit ranges have a cpu limit', function() {
+    //   //
+    //   // });
+    //
+    // });
 
     describe('When the scale target does not have a CPU request set', function() {
-      it('should return a warning with reason ≈', function() {
-        // NOTE: if the scale target has no containers, then the
-        // hasRequestOrLimit function in HPAService will return true.
-        // This could be a potential bug, though I imagine we never
-        // pass it an object w/o containers.
-        var scaleTargetWithoutLimit = {
-          kind: '',
-          spec: {
-            template: {
-              spec: {
-                containers: [
-                  {
-                    name: 'bob',
-                    image: '',
-                    env: [],
-                  }
-                ]
-              }
+      var scaleTargetWithoutLimit = {
+        kind: '',
+        spec: {
+          template: {
+            spec: {
+              containers: [
+                {
+                  name: 'bob',
+                  image: '',
+                  env: [],
+                }
+              ]
             }
           }
-        };
-        HPAService
-          .getHPAWarnings(scaleTargetWithoutLimit, defaultHpaResources, defaultLimitRanges, defaultProject)
-          .then(function(warnings) {
-            expect(_.head(warnings).reason).toEqual('NoCPURequest');
-          });
-        $rootScope.$digest();
+        }
+      };
+
+      describe('if a v2beta1 HPA is used', function() {
+        it('should return neither a warning with reason V2Beta1HPA or NoCPURequest', function() {
+          // this is a v2beta1 hpa represented as a v1 object
+          var hpaWithV2Beta1 = [{
+            metadata: {
+              annotations: {
+                'autoscaling.alpha.kubernetes.io/metrics': '[{ "type": "Pods" }]'
+              }
+            },
+            spec: {
+              scaleTargetRef: {
+                kind: 'ReplicationController'
+              }
+            }
+          }];
+          HPAService
+            .getHPAWarnings(scaleTargetWithoutLimit, hpaWithV2Beta1, defaultLimitRanges, defaultProject)
+            .then(function(warnings) {
+              expect(warnings.length).toEqual(0);
+            });
+          $rootScope.$digest();
+        });
       });
 
+      describe('if a v1 HPA is used', function() {
+        it('should return a warning with reason NoCPURequest', function() {
+          // NOTE: if the scale target has no containers, then the
+          // hasRequestOrLimit function in HPAService will return true.
+          // This could be a potential bug, though I imagine we never
+          // pass it an object w/o containers.
+          HPAService
+            .getHPAWarnings(scaleTargetWithoutLimit, defaultHpaResources, defaultLimitRanges, defaultProject)
+            .then(function(warnings) {
+              expect(_.head(warnings).reason).toEqual('NoCPURequest');
+            });
+          $rootScope.$digest();
+        });
+      });
     });
 
     describe('Warn when there are competing autoscalers', function() {
@@ -201,5 +228,4 @@ describe('HPAService', function() {
     });
 
   });
-
 });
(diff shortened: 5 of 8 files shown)


benjaminapetersen:
Updated to drop the cpuRequest warning if already shows the v2HPA warning:

screen shot 2018-01-23 at 9 28 34 pm

screen shot 2018-01-23 at 9 28 50 pm

Thinking about best wording for these messages.  Also looking to see if we can reduce the duplicate templating.  The browse `deployment-config.html`, `deployment.html` & `_replica-set-details.html` look like they have a good bit of duplicated html around the hpa warnings, and the `oscAutoscaler` directive has similar but slightly different set for the same warnings.

benjaminapetersen:
Tests will fail as I have not yet updated the ones I just added in that last PR.

spadgett:
I'm not convinced this should be a warning on the browse pages since there's not really anything wrong. It should be a warning in the editor since it doesn't support the new fields, though.

@jwforres opinion?

jwforres:
yeah warning should only have been in the editor

benjaminapetersen:
Makes sense to me. Let me tweak.

app/scripts/controllers/deploymentConfig.js
```diff
@@ -74,7 +74,8 @@ angular.module('openshiftConsole')
         var updateHPAWarnings = function() {
             HPAService.getHPAWarnings($scope.deploymentConfig, $scope.autoscalers, limitRanges, project)
                       .then(function(warnings) {
-              $scope.hpaWarnings = warnings;
+              // TODO: this should probably be a $filter we can reuse in other controllers?
```
benjaminapetersen: So I can pretty easily eliminate the warning from the controller, but thinking this should be a filter to share with other controllers.

app/scripts/controllers/deployment.js
```diff
@@ -68,7 +68,7 @@ angular.module('openshiftConsole')
         var updateHPAWarnings = function() {
             HPAService.getHPAWarnings($scope.deployment, $scope.autoscalers, limitRanges, project)
```
benjaminapetersen: I added this filter to each of the deployment pages, but now think I probably should have just added a `HPAService.getHPAWarningsForResource()` method & eliminated the duplication.

app/scripts/filters/resources.js
```diff
@@ -1354,4 +1354,11 @@ angular.module('openshiftConsole')
   })
   .filter('humanizePodStatus', function(humanizeReasonFilter) {
     return humanizeReasonFilter;
+  })
+  // not all of the HPA warnings should be surfaced on resources pages,
```
benjaminapetersen: Ben, you gonna finish the comment? or nah?

benjaminapetersen:
Test updated, but im `bower link origin-web-common` locally, so the build will not pass.

test/spec/services/hpaServiceSpec.js
```diff
@@ -22,8 +21,7 @@ describe('HPAService', function() {
   }));
 
 
-  beforeEach(inject(function(_$q_, _$rootScope_, _$window_, _HPAService_) {
-    $rootScope = _$rootScope_;
+  beforeEach(inject(function(_$q_, _$window_, _HPAService_) {
```
benjaminapetersen: @spadgett also did remove the `$rootScope` `$digest`.  Thought that was necessary to get `$q` to resolve in this env, but guess not!

app/scripts/services/hpa.js
```diff
@@ -146,15 +168,27 @@ angular.module("openshiftConsole")
         return $q.when([]);
       }
       return MetricsService.isAvailable().then(function(metricsAvailable) {
+        var v2HPAWarning = hasV2HPAScaler(scaleTarget, hpaResources);
         return _.compact([
           hasMetricsAvailableWarning(metricsAvailable),
-          hasCPURequestWarning(scaleTarget, limitRanges, project),
+          v2HPAWarning ? v2HPAWarning : hasCPURequestWarning(scaleTarget, limitRanges, project),
```
spadgett: I think this would be easier to read as...

```js
hasMetricsAvailableWarning(metricsAvailable),
// Only check for CPU request if not using v2 HPA metrics.
hasV2HPAScaler(scaleTarget, hpaResources) || hasCPURequestWarning(scaleTarget, limitRanges, project),
hasCompetingAutoscalersWarning(hpaResources),
// ...
```

app/scripts/services/hpa.js
```diff
@@ -99,6 +100,27 @@ angular.module("openshiftConsole")
...
 
 
+    var hasV2HPAAnnotations = function(hpaResources) {
+      return _.some(hpaResources, function(hpa) {
+        return $filter('hpaMetrics')(hpa);
+      });
+    };
+
+
+    var hasV2HPAScaler = function(scaleTarget, hpaResources) {
```
spadgett: hasV2HPAScalerWarning to be consistent with the other functions?

app/scripts/services/hpa.js
```diff
@@ -99,6 +100,27 @@ angular.module("openshiftConsole")
     };
 
 
+    var hasV2HPAAnnotations = function(hpaResources) {
+      return _.some(hpaResources, function(hpa) {
+        return $filter('hpaMetrics')(hpa);
```
spadgett: Since this is temporary and we shouldn't need it after we update to the newer API, I'd just inline the full annotation here.

```js
return $filter('annotation')(hpa, 'autoscaling.alpha.kubernetes.io/metrics');
```

I'd rather not do more than we need to with the annotations if the same values are supported in the v2beta1 API. And then you don't need the common PR.

nit: better to resolve the filter outside the loop

app/scripts/filters/resources.js
```diff
@@ -1355,4 +1355,9 @@ angular.module('openshiftConsole')
   })
   .filter('humanizePodStatus', function(humanizeReasonFilter) {
     return humanizeReasonFilter;
+  })
+  .filter('hpaWarningsForResource', function() {
```
spadgett: This is a very specific filter. I'd give it a name that makes it clear what it's really doing like `rejectV2HPAWarnings`. But we might not need it. (See my comments below.)

app/scripts/services/hpa.js
```diff
@@ -146,15 +168,27 @@ angular.module("openshiftConsole")
...
     };
 
+    // Get HPA warnings relevant to a Resource list page.
+    //
+    // This method returns the same set as the above getHPAWarnings but
+    // filters out the warnings that don't need to be fully surfaced to
+    // each individual resource.
+    var getHPAWarningsForResource = function(scaleTarget, hpaResources, limitRanges, project) {
+      return getHPAWarnings(scaleTarget, hpaResources, limitRanges, project).then(function(warnings) {
+        return $filter('hpaWarningsForResource')(warnings);
```
spadgett: nit: better to resolve the filter outside the loop

app/scripts/services/hpa.js
```diff
@@ -185,10 +219,17 @@ angular.module("openshiftConsole")
       return hpaByResource;
     };
 
+    // currently unsupported by the webconsole
+    var isUnsupportedAPI = function(hpa) {
```
spadgett: nit: `isUnsupportedAPI` sounds like the API is no longer supported by the server, but that's not the case. I think something like `usesV2Metrics` is more clear? The resource is technically v1 API since that's what we requested.

app/scripts/services/hpa.js
```diff
@@ -146,15 +168,27 @@ angular.module("openshiftConsole")
...
         ]);
       });
     };
 
+    // Get HPA warnings relevant to a Resource list page.
+    //
+    // This method returns the same set as the above getHPAWarnings but
+    // filters out the warnings that don't need to be fully surfaced to
+    // each individual resource.
+    var getHPAWarningsForResource = function(scaleTarget, hpaResources, limitRanges, project) {
```
spadgett: Are we filtering out the v2 warnings from all places we show them? For instance, the deployment donut on the overview?

I think it would be safer to opt in to the v2 editor warnings instead of always including editor warnings and having to filter them out. There's only one context where we need the extra warning: the editor. And that context performs the check in a different way (`isUnsupportedAPI()`).

app/views/directives/osc-autoscaling.html
```diff
@@ -99,7 +99,7 @@
       </span>
     </div>
   </div>
-  <div class="form-group">
+  <div ng-show="showCPURequestInput"  class="form-group">
```
spadgett: nit: extra space before class

app/scripts/services/hpa.js
spadgett: Unless I'm misunderstanding, we will be filtering this out every time we add it. The only place we want to show the warning is the editor page, but we specifically check `isUnsupportedAPI` there instead of calling the get warnings method. So I think we just avoid adding the warning in `getHPAWarnings`.

app/views/edit/autoscaler.html
```diff
@@ -38,11 +38,21 @@ <h1>
...
+            </div>
+
+
             <fieldset ng-disabled="disableInputs" class="gutter-top">
               <osc-autoscaling
                   model="autoscaling"
                   show-name-input="true"
-                  name-read-only="kind === 'HorizontalPodAutoscaler'">
+                  name-read-only="kind === 'HorizontalPodAutoscaler'"
+                  show-cpu-request-input="showV2Beta1Warning">
```
spadgett: nit: I'd call this scope var `usesV2Metrics` since we use the flag for more than just showing the warning.

benjaminapetersen:
@spadgett 

    Just confirming that not all HPAs have the metrics annotation we're checking, only ones that scale on metrics other than CPU request.

I've been going off of the docs page @ `kubernetes.io`, but not surprised if I missed something.  Have a link handy?

app/scripts/filters/resources.js
benjaminapetersen: Ok I can remove.

app/scripts/services/hpa.js
benjaminapetersen: This makes sense as well, closed the other PR.

app/scripts/services/hpa.js
benjaminapetersen: I can eliminate `hasV2HPAScalerWarning` from `getHPAWarnings`, though then I'll be back to showing this:

screen shot 2018-02-01 at 3 17 38 pm

Which I'll still need to check for `v2beta1` hpa annotations & filter out anyway.  I'm not sure if I can entirely remove the filtering problem.

app/scripts/services/hpa.js
benjaminapetersen: Ah, I didn't check the donut, thanks, this is problematic:

screen shot 2018-02-01 at 3 22 10 pm
screen shot 2018-02-01 at 3 22 32 pm

I think I may still need to do the `hasV2HPAScalerWarning` check in `getHPAWarnings`, but then just drop both `v2` && the `CPURequest` warning.

app/scripts/services/hpa.js
benjaminapetersen: I'll drop the `filter` per your other comment.

app/scripts/controllers/edit/autoscaler.js
```diff
@@ -78,6 +78,21 @@ angular.module('openshiftConsole')
     var horizontalPodAutoscalerVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
     var limitRangesVersion = APIService.getPreferredVersion('limitranges');
 
+    var getHPAWarnings = function(resource, limitRanges, project, context) {
+      var scaleTarget = resource.spec.scaleTargetRef;
+      var scaleTargetVersion = APIService.objectToResourceGroupVersion(scaleTarget);
+      // if the DC still has extensions in the verison, it will fail, so we need to
+      // get the preferred version.
```
benjaminapetersen: Wish we had es6 fat arrows && auto `return` 😄

benjaminapetersen: This gets the `edit` `autoscaler` page able to use `warnings` from the `hpaService` instead of the standalone `showCPURequestWarning` flags.

app/scripts/controllers/edit/autoscaler.js
```diff
@@ -78,6 +78,21 @@ angular.module('openshiftConsole')
...
+      var scaleTargetPreferredVersion = APIService.getPreferredVersion(scaleTargetVersion.resource);
+      return DataService.get(scaleTargetPreferredVersion, scaleTarget.name, context).then(function(target) {
+        return HPAService.getHPAEditorWarnings(target, [resource], limitRanges, project).then(function(warnings) {
+          return _.keyBy(warnings, function(warning) {
```
spadgett:
```js
return _.keyBy(warnings, 'reason');
```

app/scripts/controllers/edit/autoscaler.js
```diff
@@ -78,6 +78,21 @@ angular.module('openshiftConsole')
     var horizontalPodAutoscalerVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
     var limitRangesVersion = APIService.getPreferredVersion('limitranges');
 
+    var getHPAWarnings = function(resource, limitRanges, project, context) {
+      var scaleTarget = resource.spec.scaleTargetRef;
+      var scaleTargetVersion = APIService.objectToResourceGroupVersion(scaleTarget);
+      // if the DC still has extensions in the verison, it will fail, so we need to
+      // get the preferred version.
+      var scaleTargetPreferredVersion = APIService.getPreferredVersion(scaleTargetVersion.resource);
+      return DataService.get(scaleTargetPreferredVersion, scaleTarget.name, context).then(function(target) {
```
spadgett: I don't think we can always get the scale target because it might be a scale subresource that doesn't support GET?

app/scripts/directives/oscFileInput.js
```diff
@@ -127,6 +127,10 @@ angular.module('openshiftConsole')
             if (files.length > 0 ) {
               scope.file = _.head(files);
               addFile(scope.file);
+            } else {
+              scope.uploadError = true;
+              scope.cleanInputValues();
+              Logger.error("Could not read file", e);
```
spadgett: This change looks unrelated

app/views/edit/autoscaler.html
```diff
@@ -38,11 +38,21 @@ <h1>
...
+            </div>
+
+
             <fieldset ng-disabled="disableInputs" class="gutter-top">
               <osc-autoscaling
                   model="autoscaling"
                   show-name-input="true"
-                  name-read-only="kind === 'HorizontalPodAutoscaler'">
+                  name-read-only="kind === 'HorizontalPodAutoscaler'"
+                  show-request-input="!(usesV2Metrics)">
```
spadgett: I don't see where `usesV2Metrics` is set.

app/scripts/controllers/edit/autoscaler.js
benjaminapetersen: Is there a good example of that?   I'm thinking about things we have like `deploymentconfig/instantiate` or `pods/log` or `builds/clone` and can't think of a why an autoscaler would be attached.

app/scripts/directives/oscFileInput.js
benjaminapetersen: Correct, cherry picked jakub's PR for the file input, should have saw before my commit, removing.

app/scripts/controllers/edit/autoscaler.js
```diff
@@ -78,6 +78,7 @@ angular.module('openshiftConsole')
     var horizontalPodAutoscalerVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
     var limitRangesVersion = APIService.getPreferredVersion('limitranges');
 
+
```
benjaminapetersen: Rolled this back to minmal changes.

Note that  as it stands now & like `master`, the `showCPURequestWarning` will not trigger w/o hitting a certain set of conditionals, though the cpu request warning DOES appear on the DC page.

app/views/edit/autoscaler.html
benjaminapetersen: fixed

benjaminapetersen:
@spadgett rolled back, looking over my tests again now, think just some minimal tweaks there.

benjaminapetersen:
Ah, `$rootScope.$digest()` had to be introduced again, otherwise none of the test updates actually ran :frown:

app/scripts/services/hpa.js
```diff
@@ -99,6 +102,27 @@ angular.module("openshiftConsole")
...
+    var hasV2HPAAnnotations = function(hpaResources) {
+      return _.some(hpaResources, function(hpa) {
+        return JSON.parse(annotation(hpa, 'autoscaling.alpha.kubernetes.io/metrics'));
+      });
+    };
+
+
+    var hasV2HPAScalerWarning = function(scaleTarget, hpaResources) {
+      var containers = _.get(scaleTarget, 'spec.template.spec.containers', []);
+      var hasRequest = hasCPURequest(containers);
```
spadgett: It's not clear to how this affects whether we show the warning. Even if the container has a request, wouldn't we still show the warning if it's scaling on another metric we don't support in the editor?

app/scripts/services/hpa.js
```diff
@@ -146,15 +170,19 @@ angular.module("openshiftConsole")
         return $q.when([]);
       }
       return MetricsService.isAvailable().then(function(metricsAvailable) {
+        var isV2HPA = hasV2HPAScalerWarning(scaleTarget, hpaResources);
```
spadgett: Why not just use `hasV2HPAAnnotations` here and remove `hasV2HPAScalerWarning`?

app/scripts/services/hpa.js
benjaminapetersen: Well, via the [trello card description](https://trello.com/c/EuQHe0LT) I was under the impression that we are drawing a line between `cpuRequest` and `metrics` for how we show this warning.

benjaminapetersen: But, I belive the API does transfer `metrics[]` that are CPU related to a `cpuRequest` if requested via `v1` api, so your comment makes sense.

app/scripts/services/hpa.js
benjaminapetersen: Yeah that makes sense.

app/scripts/services/hpa.js
benjaminapetersen: Eliminating this `fn`, then, as we don't need it at all, the `autoscaler` editor does it inline.

benjaminapetersen: On that note then, in the editor, we could keep showing the cpu input box for `targetCPUUtilizationPercentage` if that value is set, even if there are other metrics.

benjaminapetersen:
So, editor at present-

With ONLY v2 metrics fields:

screen shot 2018-02-07 at 10 26 53 am

With v2 metrics && also a v2 metric on cpu that is converted back into a v1 `targetCPUUtilizationPercentage`:

screen shot 2018-02-07 at 10 38 33 am

With only the `targetCPUUtilizationPercentage` (just v1 fields):

screen shot 2018-02-07 at 10 39 50 am

Just hiding the input if there are ONLY v2 metrics present, otherwise showing it. Editing seems to work fine.  Viewing the objects under `other resources` & editing yaml appears correct with all 3 of these options.

benjaminapetersen:
& Tests pass again.

app/scripts/services/hpa.js
```diff
@@ -99,6 +102,13 @@ angular.module("openshiftConsole")
     };
 
 
+    var hasV2HPAAnnotations = function(hpaResources) {
+      return _.some(hpaResources, function(hpa) {
+        return JSON.parse(annotation(hpa, 'autoscaling.alpha.kubernetes.io/metrics'));
```
spadgett: Missed this before (sorry), but if the annotation is not valid JSON -- for instance, if edited by the user -- it will be a runtime error. We should add a try/catch here if we need JSON, but I would just skip the parse since we don't do anything with the value.

app/scripts/services/hpa.js
```diff
@@ -146,15 +156,19 @@ angular.module("openshiftConsole")
         return $q.when([]);
       }
       return MetricsService.isAvailable().then(function(metricsAvailable) {
+        var isV2HPA = hasV2HPAAnnotations(hpaResources);
```
spadgett: So looking at this again, it will skip CPU request validation if any of the resources uses v2 metrics. So if you have two:

1. v2 HPA
2. v1 HPA whose target has no CPU request

The CPU request warning will be dropped. Maybe that's OK since you really shouldn't have two HPA for the same resource, but pointing it out.

app/scripts/services/hpa.js
benjaminapetersen: Hmm. will the save be successful if the user edits but the JSON is invalid?  Prob my bad assuming we would reject the change...

spadgett: Either way we don't want the runtime error in the client that stops JS execution.

app/scripts/services/hpa.js
benjaminapetersen: Noting our discussion in case I come back to read these comments later, having 2 HPA is a problem that is more severe && MUST be fixed.  The CPU request warning might be a problem but is subordinate.

app/scripts/services/hpa.js
```diff
@@ -99,6 +102,13 @@ angular.module("openshiftConsole")
     };
 
 
+    var hasV2HPAAnnotations = function(hpaResources) {
+      return _.some(hpaResources, function(hpa) {
+        return annotation(hpa, 'autoscaling.alpha.kubernetes.io/metrics');
```
benjaminapetersen: Dropped the `JSON.parse`

benjaminapetersen:
Updated the `JSON.parse`, might be good now. maybe :)

PR #2710, opened 2018-01-24
Bump origin-web-common to 3.6.1

spadgett:

/assign @jwforres 
/cc @liggitt


spadgett:
@jwforres This pulled in an updated openshift-logos dependency, but it's something we probably want.

PR #2713, opened 2018-01-24
Bug 1504819: Obfuscate webhook trigger URL if user cant list secrets

jhadvig:
Removed the `copy-webhook-url` component and used `copy-to-clipboard` original since we decided to not show the warning tooltip and just obfuscate webhook URL is user doesnt have permissions to list secrets.
@jwforres PTAL


jhadvig:
@jwforres added the warning tooltip next to the Trigger section header:
download

PTAL

jhadvig:
@jwforres can we merge this ? Its the last bug on the [card](https://trello.com/c/lo6HwQgu/1106-2-build-webhooks-with-secretrefs)

jwforres:
yeah i think this is fine for now, I think we probably want to consider if there are better alternatives that we could look at in 3.10 @ncameronbritt that have a different representation for users that don't have access to Secrets.  I can't decide if its helpful to have the webhook URLs there when they don't contain the actual secret.

ncameronbritt:
That's certainly something we can look at. I'm not sure that using the warning icon is the right thing to do since there's not actually a problem with the platform. And if the user cannot use the webhook without the secret, I don't know that there is much value in showing the URL. Should we create an issue to follow up on this?

jwforres:
Yep plz open an issue

Issue #2696, opened 2018-01-22
When mini-pod-donut is visible it adds excess height to list-pf-item row, causing jump behavior when switching between deployment states. 


sg00dwin:
<img width="593" alt="screen shot 2018-01-22 at 11 59 31 am" src="https://user-images.githubusercontent.com/1874151/35233831-3f51fe88-ff6d-11e7-8f7e-fe7131142f7d.png">

![animatedgif](https://user-images.githubusercontent.com/1874151/35234033-d80e6120-ff6d-11e7-8d89-2e3b86ac5449.gif)

spadgett:
We should check with metrics enabled, too.

Issue #2105, opened 2017-09-18
Test Flake: User adds a template to a project [...] Error: Wait timed out after 5512ms


spadgett:
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_web_console/228

```
1) User adds a template to a project after creating a new project using the Import YAML tab should save the template in the project catalog
  - Error: Wait timed out after 5512ms

[...]

Error
    at describe (/data/src/github.com/openshift/origin-web-console/test/integration/features/user_adds_template_to_project.spec.js:14:3)
    at Object.<anonymous> (/data/src/github.com/openshift/origin-web-console/test/integration/features/user_adds_template_to_project.spec.js:12:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
```

PR #2718, opened 2018-01-24
Make the mini <pod-donut> positioned absolute, taking it out of flow,…

### LINKED ISSUE #2744: Idle text should not be shown inside mini donut
I don't think there's enough space. The text might be longer as well in other languages when we translate the console.

openshift web console 2018-01-30 18-36-50

/kind bug
/area styles
/priority p3
/assign @rhamilto @sg00dwin

sg00dwin:
… so that it doesn't alter the height of the pf-list-container row based on if it's displayed or not.
Note:
- expand and collapsed heights are now the same height.
- row height is slightly shorter since donut height no longer has an effect.

Fixes https://github.com/openshift/origin-web-console/issues/2696
Fixes #2744

Cross browser and Android, iOS tested.

mini-new-animated

**NEW**
new

**OLD**
old

### DIFF
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -84,25 +84,30 @@ code.command {
   cursor: pointer;
 }
 .pods {
-  display: inline-block;
+  display: block;
+  // mini donut circle diameter
+  height: @pod-mini-donut-diameter;
+  width: @pod-mini-donut-diameter;
   // align pod text with right outer edge of .list-pf-details text
-  margin-right: 25px;
+  @media(min-width: @screen-sm-min) {
+    margin-right: 32px;
+  }
 }
 pod-donut[mini] {
   display:inline-block;
-  // match size setting in podDonut.js
-  min-height: 45px;
-  min-width: 45px;
+  height: @pod-mini-donut-diameter;
   position: relative;
+  width: @pod-mini-donut-diameter;
 }
 .pod-donut {
   &.mini {
     display: inline-block;
-    // match size setting in podDonut.js
-    min-height: 45px;
-		min-width: 45px;
+    // Take mini donut out of flow; positioning is controlled by .pods parent element
+    height: @pod-mini-donut-diameter;
     position: absolute !important;
-		top: 2px;
+    // center donut within .pods element
+    transform: translate(-8px, -5px);
+    width: @pod-mini-donut-diameter;
   }
   .c3-defocused {
     // Adjusting the default hover "opacity: .3" to be slightly darker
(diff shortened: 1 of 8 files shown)


sg00dwin:
@spadgett @rhamilto ready for review
cc @jwforres

rhamilto:
Monitoring page at mobile needs some love.

screen shot 2018-01-25 at 9 08 36 am

app/styles/_components.less
```diff
@@ -101,8 +99,9 @@ pod-donut[mini] {
     // match size setting in podDonut.js
     min-height: 45px;
 		min-width: 45px;
+    // Take mini donut out of flow and position vertically centered in row
     position: absolute !important;
-		top: 2px;
+		transform: translate(0, -24px);
```
spadgett: It looks like these styles are pretty specific to list view. We might want to use mini donut in other contexts.

How does this look on the monitoring page? How does it look at mobile?

sg00dwin: This is positioning the mini donut within the `<pod-donut>` element. Any positioning of the mini-donut should be set by the container element outside the donut scope.

Random placement tests
screen shot 2018-01-29 at 4 55 28 pm
screen shot 2018-01-29 at 4 53 16 pm
screen shot 2018-01-29 at 4 33 20 pm

sg00dwin:
**Monitoring page**

screen shot 2018-01-29 at 3 50 08 pm
screen shot 2018-01-29 at 3 49 57 pm

sg00dwin:
@rhamilto updates pushed and ready for review.

app/styles/_overview.less
```diff
@@ -356,7 +360,18 @@
   }
   .metrics-collapsed {
     flex-grow: 1;
-    .text-center();
+    padding-top: 10px;
+    text-align: left;
+    @media(min-width:@screen-sm-min) {
```
rhamilto: Remove this empty media query?

sg00dwin:
Including fix for https://github.com/openshift/origin-web-console/issues/2744

screen shot 2018-01-31 at 1 56 52 pm

app/views/directives/pod-donut.html
```diff
@@ -4,16 +4,14 @@
...
-    Idle
-  </span>
 </div>
 <span ng-if="mini && total === 1 && !idled" class="donut-mini-text-name">pod</span>
 <span ng-if="mini && total !== 1 && !idled" class="donut-mini-text-name">
   <span ng-if="!idled && total > 99">
     {{total}}
   </span> pods
 </span>
+<span ng-if="idled" class="donut-mini-text-name">Idle</span>
```
rhamilto: Given the siblings of this node also check for mini, I think it's probably best to do so here, too.

e.g., ng-if="mini && idled"

rhamilto: Not only is it best, it's actually required.  :)  Note the stray "Idle" below "Idled due to inactivity."

screen shot 2018-01-31 at 2 17 08 pm

sg00dwin:
updated pr so "Idled" only shown with mini donut

screen shot 2018-02-01 at 10 28 07 am

sg00dwin:
@spadgett retests failing

spadgett:
This really looks like flake #2105. I had a PR that failed multiple times on this as well. Looking at the full PR history, it has passed the test.

https://openshift-gce-devel.appspot.com/pr/openshift_origin-web-console/2718

/retest

PR #2720, opened 2018-01-25
Fix bug #1538431, pod table triggers $digest() loop 

benjaminapetersen:
Fix [bug #1538431](https://bugzilla.redhat.com/show_bug.cgi?id=1538431), pod table triggers $digest() loop that surpasses the limit set by Angular.  This change is a backport of current code in the pod table.

### DIFF
--- a/app/scripts/directives/resources.js
+++ b/app/scripts/directives/resources.js
@@ -169,7 +169,7 @@ angular.module('openshiftConsole')
       templateUrl: 'views/directives/_probe.html'
     };
   })
-  .directive('podsTable', function() {
+  .directive('podsTable', function($filter) {
     return {
       restrict: 'E',
       scope: {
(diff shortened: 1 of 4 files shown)


benjaminapetersen:
This bug is already [fixed in 3.6](https://github.com/openshift/origin-web-console/blob/v3.6.0/app/scripts/directives/resources.js#L179), only one PR needed.

benjaminapetersen:
Jenkins fails, Travis passes, guessing because:

```
DEBUG [web-server]: serving (cached): /data/src/github.com/openshift/origin-web-console/test/spec/services/membership/roleBindingsSpec.js
```
Cached test might be newer than is expected for this PR.

benjaminapetersen:
You make me sad, Jenkins.

spadgett:
Cluster up is running latest, so it could be an incompatibility with the 3.5 console and latest API server, especially if it always fails the same way.

jwforres:
@spadgett i thought about that, but this is the spec tests not the e2es

jwforres:
ohhhh i was looking at the wrong thing, i totally missed the actual errors

jwforres:
@benjaminapetersen yeah this is because you are getting that same runtime error you saw locally when you were working on your fix.  The API server's types changed, and something we are requesting is blowing up.

jwforres:
this is a fundamental problem with our merge queue today, and is not easy to resolve, im going to merge this so that we can get the fix out

benjaminapetersen:
Ok great. Was thinking this was something I prob can't solve in any reasonable way.

jwforres:
@benjaminapetersen we actually have an idea for how to go fix this

PR #2726, opened 2018-01-26
Fixes incorrect focus on first "add from secret or config map" click

benjaminapetersen:
Re [bugzilla 1369315](https://bugzilla.redhat.com/show_bug.cgi?id=1369315)

Adds a slightly longer delay to the first "add from secret or config map".  Guessing it is a timing issue with instantiating the ui-select boxes & angular's $digest.  Oddly only happens the first time.

### DIFF
--- a/app/scripts/directives/keyValueEditor.js
+++ b/app/scripts/directives/keyValueEditor.js
@@ -234,8 +234,10 @@
                   utils.setFocusOn('.'+ $scope.setFocusKeyClass);
                 },
                 onAddRowWithSelectors: function() {
+                  var entriesCount = $scope.entries.length;
+                  var extraFocusDelay = (entriesCount === 1) ? 125 : null;
                   utils.addEntryWithSelectors($scope.entries);
-                  utils.setFocusOn('.'+ $scope.setFocusKeyClass);
+                  utils.setFocusOn('.'+ $scope.setFocusKeyClass, null, extraFocusDelay);
                 },
                 isValueFromReadonly: function(entry) {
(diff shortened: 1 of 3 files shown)


benjaminapetersen:
Thanks @rhamilto, @jwforres or @spadgett any objection?

app/scripts/directives/keyValueEditor.js
```diff
@@ -234,8 +234,10 @@
                   utils.setFocusOn('.'+ $scope.setFocusKeyClass);
                 },
                 onAddRowWithSelectors: function() {
+                  var entriesCount = $scope.entries.length;
+                  var extraFocusDelay = (entriesCount === 1) ? 125 : null;
```
spadgett: The magic number makes me nervous :(

Is there any way around that? It seems to have caused the bug to start.

benjaminapetersen: Agree. I think its revealing an underlying race condition *somewhere* that the original code does NOT trigger.  We only see it in this fn because rendering out the 2 `ui-select` nodes slows things down enough to see the focus set incorrectly.  I believe it is a race against `$digest`.  We could try replacing the `$timeout` with `$applyAsync`, though I think we didn't do that originally because the `setFocus` has to happen `$digest + 1` into the future, after the first render of the `ui-select`'s happens.

For `$digest +1`, I don't now if you can nest a `$applyAsync` inside an `$applyAsync`, and we don't want the delay perceptible to the user.  

Timing issues in Angular are hard. 😦 

(And I may be completely wrong, this is just my guess)

PR #2728, opened 2018-01-26
Fix loading of local-extensions(.css, .js)

benjaminapetersen:
Follow to #2677, realized that while we have this:

```html 
    <!-- build:remove -->
    <!-- Local / test stylesheets only. These files will not be included in the built dist. -->
    <link rel="stylesheet" type="text/css" href="extensions/local-extensions.css">
    <!-- endbuild -->

    <!-- build:remove -->
    <!-- Local / test scripts only. These files will not be included in the built dist. -->
    <script src="extensions/local-extensions.js"></script>
    <!-- endbuild -->
```
But these files aren't actually loading (I have a few things in `extensions.js` that didn't run).  Adding the extensions block back to `Gruntfile.js` fixes. 

I don't think the `.gitignore` works for the two extension files since they have already been committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ openshift-jvm
 /app/config.local.js
 .DS_Store
 test/junit/
+/extensions/local-extensions.css
+/extensions/local-extensions.js
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -73,6 +73,10 @@ module.exports = function (grunt) {
           }
         }
       },
+      extensions: {
+        files: ['extensions/local-extensions.js', 'extensions/local-extensions.css'],
+        tasks: ['copy:extensions']
+      },
       index: {
         files: ['<%= yeoman.app %>/index.html'],
         tasks: ['replace:index']


spadgett:
> (I have a few things in extensions.js that didn't run)

Did you use `local-extensions.js`?

spadgett:
I changed the name since it's local only.

.gitignore
```diff
@@ -10,3 +10,5 @@ openshift-jvm
 /app/config.local.js
 .DS_Store
 test/junit/
+/extensions/local-extensions.css
+/extensions/local-extensions.js
```
spadgett: If we add these to gitignore, they will 404 when you clone the repo.

Gruntfile.js
```diff
@@ -73,6 +73,10 @@ module.exports = function (grunt) {
           }
         }
       },
+      extensions: {
+        files: ['extensions/local-extensions.js', 'extensions/local-extensions.css'],
+        tasks: ['copy:extensions']
```
spadgett: This task no longer exists.

benjaminapetersen:
@spadgett yup, i copied my `extensions.js` to `extensions-local.js`.  I may have still had the `extensions` task from not doing an `npm install`.

Gruntfile.js
benjaminapetersen: Oh, i think we just need to ensure these are served in dev mode, shouldn't need a copy task.  i can update.

spadgett:
> yup, i copied my extensions.js to extensions-local.js

So that's the wrong name :)

local-extensions.js

It's working for me:

monosnap 2018-01-26 16-13-45

benjaminapetersen:
Huh, I believe you :)
not sure why mine didn't, ill close this out & give it another check.

PR #2736, opened 2018-01-30
Fixes null != undefined error in homePagePreferenceSpec.js

benjaminapetersen:
I tried this with a few versions of node & npm:

```
# first
node -v 6.10.2; npm -v 3.10.10
# then
node -v 9.4.0; npm -v 5.6.0
```

But either way the test failed due to `null != undefined`.  This tiny change gets the test passing again, though I'm not sure it actually fixes the test.  `getHomePagePath - should notify invalid project` is confusing to me in light of the `expect` statement.

Thoughts @dtaylor113 ?

### DIFF
--- a/test/spec/services/homePagePreferenceSpec.js
+++ b/test/spec/services/homePagePreferenceSpec.js
@@ -8,12 +8,12 @@ describe("HomePagePreferenceService", function(){
     inject(function(_HomePagePreferenceService_){
       HomePagePreferenceService = _HomePagePreferenceService_;
     });
-    
+
     // LocalStorage mock
     spyOn(localStorage, 'getItem').and.callFake(function(key) {
       return store[key];
     });
-    
+
     spyOn(localStorage, 'setItem').and.callFake(function (key, value) {
       /* jshint boss:true */
       return store[key] = value + '';
@@ -21,7 +21,7 @@ describe("HomePagePreferenceService", function(){
 
     spyOn(localStorage, 'removeItem').and.callFake(function (key) {
       delete store[key];
-    }); 
+    });
 
     spyOn(localStorage, 'clear').and.callFake(function () {
       store = {};
@@ -36,14 +36,14 @@ describe("HomePagePreferenceService", function(){
       var homePagePreference = HomePagePreferenceService.getHomePagePath();
       expect(homePagePreference).toEqual("/catalog");
     });
-        
+
     it("set project list", function(){
       var homePagePref = {"type":"project-list"};
       HomePagePreferenceService.setHomePagePreference(homePagePref);
       var homePagePreference = HomePagePreferenceService.getHomePagePath();
       expect(homePagePreference).toEqual("/projects");
     });
-   
+
     it("set to project overview",function(){
       var homePagePref = {
           "type":"project-overview",
@@ -61,7 +61,7 @@ describe("HomePagePreferenceService", function(){
       };
       HomePagePreferenceService.setHomePagePreference(homePagePref);
       HomePagePreferenceService.notifyInvalidProjectHomePage(homePagePref.project);
-      expect(localStorage.getItem("openshift/home-page-pref/")).not.toBeDefined();
+      expect(localStorage.getItem("openshift/home-page-pref/")).toBeFalsy();
     });
   });
 });


benjaminapetersen:
I'm a little bit inclined towards updating the `describe()` and `it()` to be more like:

```javascript 
describe('#getHomePagePath', function() {
  describe('when given an empty config object', function() {
    it('should set the default home page to /catalog', function() {
        // expect
    });
  });

   // etc.
});
```
It reads more clearly & makes it easier to fix a test when you aren't familiar with the domain.

dtaylor113:
> getHomePagePath - should notify invalid project is confusing to me in light of the expect statement.

When you have an invalid (non-existent) project as a home page preference, notifyInvalidProjectHomePage(..) (a) Clears the homePagePref local storage, then (b) pops up a warning toast notification, so the `expect` in the unit test is trying to verify that the local storage/home page pref has been cleared/removed.

test/spec/services/homePagePreferenceSpec.js
```diff
@@ -61,7 +61,7 @@ describe("HomePagePreferenceService", function(){
       };
       HomePagePreferenceService.setHomePagePreference(homePagePref);
       HomePagePreferenceService.notifyInvalidProjectHomePage(homePagePref.project);
-      expect(localStorage.getItem("openshift/home-page-pref/")).not.toBeDefined();
+      expect(localStorage.getItem("openshift/home-page-pref/")).toBeFalsy();
```
spadgett: I'm unclear why this should be necessary since we spyOn localStorage.removeItem to delete the key. It should always be undefined?

https://github.com/openshift/origin-web-console/pull/2736/files#diff-2e146cc351b88485c09870a5cc2f86f9R23

https://github.com/openshift/origin-web-console/blob/4e1c80cf375170e976c77ee8910d0aafacf62e16/app/scripts/services/homePagePreference.js#L38-L44

benjaminapetersen:
@dtaylor113 but the unit test doesn't test the notification, so this language is misleading in this context.  I think the notification part would make sense in an e2e test, or in another test for the page itself that verifies the notification appears when the service returns `null`/`undefined`.

test/spec/services/homePagePreferenceSpec.js
benjaminapetersen: @dtaylor113  thoughts?

@spadgett good Q, I'm only aiming to remove the failing test blocker with this PR.

test/spec/services/homePagePreferenceSpec.js
```diff
@@ -8,20 +8,20 @@ describe("HomePagePreferenceService", function(){
     inject(function(_HomePagePreferenceService_){
       HomePagePreferenceService = _HomePagePreferenceService_;
     });
-    
+
```
benjaminapetersen: We should also double check that everyone is using the same `.editorconfig` so we have consistent white space settings.

benjaminapetersen: (just to avoid unnecessary diffs)

spadgett:
@benjaminapetersen I agree the test description isn't quite right

test/spec/services/homePagePreferenceSpec.js
spadgett: @yapei FYI, it would be good to set up editorconfig for your editor

http://editorconfig.org/

test/spec/services/homePagePreferenceSpec.js
spadgett: I'm able to reproduce locally, but I worry it means `spyOn` is not working. Strange we don't see this in travis or jenkins. Maybe it's a chrome vs firefox thing.

benjaminapetersen:
@spadgett @dtaylor113 my own tests burned me on the last updates to `Membership`, so I'd really like to see our wording be as descriptive & specific as possible with each unit test. Granted, I think this service is probably small enough it's not going to cause us a ton of pain, but we can set the trajectory for future more complex tests.

spadgett:
@benjaminapetersen FYI @dtaylor113 didn't write the original test in case you weren't aware :)

benjaminapetersen:
Ha oh? I thought he owned this feature so I've been pinging him.  Oops, my bad 😄

dtaylor113:
> Ha oh? I thought he owned this feature so I've been pinging him. Oops, my bad

lol..no worries @benjaminapetersen

benjaminapetersen:
@spadgett yup, ill look into the spy issue.

It is very strange that this works in travis/jenkins but not locally.

dtaylor113:
> It is very strange that this works in travis/jenkins but not locally.

The unit tests work for me locally, Fedora 27, PhantomJS 2.1.1

spadgett:
I think it is a difference in native behavior of removeItem. If spyOn is not working, we can have different behavior across browser

test/spec/services/homePagePreferenceSpec.js
benjaminapetersen: @yapei & thanks for taking the time to add these tests!  Quirk with `jasmine` aside, every little bit helps!

test/spec/services/homePagePreferenceSpec.js
benjaminapetersen: Works locally if I switch the browser to Chrome.
It's a reported bug that spyOn localStorage doesn't work in Firefox.  Figuring out a workaround.

benjaminapetersen: `spyOn` wont work, mocking `window.localStorage` or `$window.localStorage` also does not work as `window.localStorage`as a data property as not writable and not configurable. 

Neat 😄

spadgett: OK I am good with the change if spyOn is working on other browsers and it makes the tests pass on Firefox.

test/spec/services/homePagePreferenceSpec.js
yapei: @benjaminapetersen  @spadgett  I'm really sorry for the inconsistency in my PR, I should setup editorconfig before opening a PR as documented in our repo [before opening a PR](https://github.com/openshift/origin-web-console#before-opening-a-pull-request)

yapei:
> getHomePagePath - should notify invalid project is confusing to me in light of the expect statement.
@benjaminapetersen  Thanks for pointing it out. I will update descriptions, it's my fault giving unclear statement.

>  I'm a little bit inclined towards updating the describe() and it() to be more like:
@benjaminapetersen  @dtaylor113  Thanks for your comments, will update also.

test/spec/services/homePagePreferenceSpec.js
benjaminapetersen: No worries, not a big deal.  Appreciate it!

yapei:
@benjaminapetersen  Actually I saw you already have an PR #2741 updating them, thank you so much! I would pay more attention to the language/description in my following PR.

benjaminapetersen:
Yup we got it @yapei, thanks again! Def appreciate the added tests, we have plenty of room for more. 😄

PR #2738, opened 2018-01-30
Bug 1538943: Image carry out the deployment should be optional for Custom deployment strategy and should have unique ID

jhadvig:
For Custom deployment strategy, the Image Name which will carry out the deployment should be optional per docs at https://docs.openshift.com/container-platform/3.7/dev_guide/deployments/deployment_strategies.html#custom-strategy

As part of this PR I also updated the `id`'s in the template to match `id's` of the `help-block`
@spadgett PTAL

### DIFF
--- a/app/views/edit/deployment-config.html
+++ b/app/views/edit/deployment-config.html
@@ -44,18 +44,17 @@
 
                     <div ng-if="strategyData.type === 'Custom'" >
                       <div class="form-group">
-                        <label for="imageName" class="required">Image Name</label>
+                        <label for="image-name">Image Name</label>
                         <div>
                           <input class="form-control"
-                            id="imageName"
+                            id="image-name"
                             name="imageName"
                             ng-model="strategyData.customParams.image"
                             type="text"
                             autocorrect="off"
                             autocapitalize="none"
                             spellcheck="false"
-                            aria-describedby="image-name-help"
-                            required>
+                            aria-describedby="image-name-help">
                         </div>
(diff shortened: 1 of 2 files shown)


app/views/edit/deployment-config.html
```diff
@@ -92,7 +91,7 @@ <h3>Deployment Strategy</h3>
                               aria-describedby="strategyTimeout">
```
spadgett:
```html
aria-describedby="strategy-timeout-help"
```

app/views/edit/deployment-config.html
```diff
@@ -127,7 +126,7 @@ <h3>Deployment Strategy</h3>
                                   aria-describedby="updatePeriod">
```
spadgett:
```html
aria-describedby="update-period-help"
```

app/views/edit/deployment-config.html
```diff
@@ -159,7 +158,7 @@ <h3>Deployment Strategy</h3>
                                   aria-describedby="interval">
```
spadgett:
```html
aria-describedby="interval-help"
```

app/views/edit/deployment-config.html
```diff
@@ -212,7 +211,7 @@ <h3>Deployment Strategy</h3>
                                 class="form-control"
                                 aria-describedby="maxSurge">
```
spadgett: max-surge-help

jhadvig:

PR #2739, opened 2018-01-30
Bug 1539840: Drag and Drop zone area should use absolute positioning

jhadvig:
Also adjust the size of the drop and drop zone area to not include the label height.
With the label height:
incorrect

Without it:
correct

Tested the change on osc-file-input used in modal and non-modal html and both Chrome and FF 

@jwforres PTAL


jhadvig:
@rhamilto thanks ! Dist added :)

PR #2741, opened 2018-01-30
Update homePagePreferenceSpec test describe() and it() statements

benjaminapetersen:
Had this as part of some local changes I was going to push to #2736 in an attempt to figure out the `localStorage` `spyOn` quirks in Firefox.  I'm good to punt on that, seems a big hassle, but prob the updated test language is worth salvaging.

### DIFF
--- a/test/spec/services/homePagePreferenceSpec.js
+++ b/test/spec/services/homePagePreferenceSpec.js
@@ -29,39 +29,59 @@ describe("HomePagePreferenceService", function(){
 
    });
 
-  describe("#getHomePagePath", function(){
-    it("default set to catalog-home",function(){
-      var homePagePref = {};
-      HomePagePreferenceService.setHomePagePreference(homePagePref);
-      var homePagePreference = HomePagePreferenceService.getHomePagePath();
-      expect(homePagePreference).toEqual("/catalog");
-    });
+   describe("#getHomePagePath", function(){
+     describe('when given an empty config object', function() {
+       it("should set the default home page to /catalog",function(){
+         var homePagePref = {};
+         HomePagePreferenceService.setHomePagePreference(homePagePref);
+         var homePagePreference = HomePagePreferenceService.getHomePagePath();
 
-    it("set project list", function(){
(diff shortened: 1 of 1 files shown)


test/spec/services/homePagePreferenceSpec.js
```diff
@@ -29,39 +29,57 @@ describe("HomePagePreferenceService", function(){
...
+
+         expect(homePagePreference).toEqual("/project/test-pro/overview?isHomePage=true");
+       });
+     });
+
+     describe('when given a config object with an invalid `type` option', function() {
+       // in the UI, this will also trigger a notification that the project is
+       // invalid.  This additional functionality should be verified either with
+       // a unit test or an e2e test against the appropriate controller.
+       it("should not set the home page", function(){
```
spadgett: This is really testing that `notificationInvalidProjectHomePage` clears the value. The description is better, but not quite accurate.

benjaminapetersen: True, this should be in a separate `describe('#notifyInvalidProjectHomePage')`.  I'll update.

benjaminapetersen: Updated.

test/spec/services/homePagePreferenceSpec.js
```diff
@@ -29,39 +29,59 @@ describe("HomePagePreferenceService", function(){
...
+         expect(homePagePreference).toEqual("/project/test-pro/overview?isHomePage=true");
+       });
+     });
+  });
+
+  describe('#notifyInvalidProjectHomePage', function() {
+    describe('when given a config object with an invalid `type` option', function() {
+      // in the UI, this will also trigger a notification that the project is
+      // invalid.  This additional functionality should be verified either with
+      // a unit test or an e2e test against the appropriate controller.
```
yapei: Yeah, that's what I planned also, to cover the notifications checking in SetHomePageModalController unit/e2e testing, but I didn't add them in statement, as well as inappropriate language/description which cause confusion, sorry for this.

Issue #2723, opened 2018-01-26
Stateful sets can be scaled from overview but not stateful set page


spadgett:
We should be consistent. I'm not sure there's a reason any longer to keep users from scaling them.

/assign @benjaminapetersen 
/kind bug
/priority p2

@jwforres FYI

benjaminapetersen:
Agree, that doesn't make sense. I can look at this.

benjaminapetersen:
Actually, I don't know if we are on the Overview page, I can scale my DC's but not my StatefulSets:

![screen shot 2018-01-30 at 10 36 48 pm](https://user-images.githubusercontent.com/280512/35604017-28d0e0e6-060e-11e8-9322-d2a96c02b3ba.png)

Investigating.

benjaminapetersen:
Related to #2733, I'm trying to decide how to approach this/which to follow, since we have the `replicas` editor with & without the donut.

spadgett:
Sometimes it can apparently?

![openshift web console 2018-01-31 08-45-06](https://user-images.githubusercontent.com/1167259/35626279-384ca0c8-0663-11e8-9ac1-1bca13952abc.png)

benjaminapetersen:
Any idea why....?  I'll swing your way on that.

spadgett:
I'm not sure why it wouldn't always be scalable.

https://github.com/spadgett/origin-web-console/blob/c5bb9ada6e7b73ba9355c7aad7b56cb6981af440/app/scripts/directives/overview/listRow.js#L142-L157

@jwforres any objection to just making stateful sets scalable now?

spadgett:
Actually, nevermind. It doesn't work: "Reason: Cannot scale kind StatefulSet."

spadgett:
I suspect it's the canI check that hides the arrows, but the check is skipped since I was logged in as cluster-admin.

jwforres:
so despite you having authority to get the scale subresource on statefulsets, it still throws an error? what error code is it, 405?

spadgett:
/assign

I have a fix as part of my fix for bug 1540916.

PR #2748, opened 2018-02-01
Bug 1540916 - Use correct HPA scaleRef apiVersion

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1540916

/assign @jwforres 
/cc @benjaminapetersen

### DIFF
--- a/app/views/browse/_replica-set-details.html
+++ b/app/views/browse/_replica-set-details.html
@@ -198,7 +198,7 @@
          ng-href="project/{{projectName}}/edit/autoscaler?kind=ReplicationController&name={{replicaSet.metadata.name}}"
          role="button">Add Autoscaler</a>
       <a ng-if="replicaSet.kind === 'ReplicationController' && deploymentConfigName"
-         ng-href="project/{{projectName}}/edit/autoscaler?kind=DeploymentConfig&name={{deploymentConfigName}}"
+         ng-href="project/{{projectName}}/edit/autoscaler?kind=DeploymentConfig&name={{deploymentConfigName}}&group=apps.openshift.io"
          role="button">Add Autoscaler</a>
     </span>
     <span ng-if="!({resource: 'horizontalpodautoscalers', group: 'autoscaling'} | canI : 'create')">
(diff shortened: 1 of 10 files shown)


spadgett:
`apps.openshift.io` has a `deploymentconfigs/scale` subresource. (It used to be in the extensions API group.) I'm checking all of the others to be sure.

openshift web console 2018-02-01 08-29-55

spadgett:
We should change our handling of the scale subresource in DeploymentsService. Related to https://github.com/openshift/origin-web-console/issues/2723#issuecomment-362097290

spadgett:
Updated DeploymentsService, which makes scaling stateful sets on the overview work (#2723).

spadgett:
Note that by default project admins don't have authority to update the statefulsets/scale subresource, but it works for cluster admins.

app/views/browse/_replica-set-details.html
```diff
@@ -198,7 +198,7 @@ <h3>Autoscaling</h3>
          ng-href="project/{{projectName}}/edit/autoscaler?kind=ReplicationController&name={{replicaSet.metadata.name}}"
          role="button">Add Autoscaler</a>
       <a ng-if="replicaSet.kind === 'ReplicationController' && deploymentConfigName"
-         ng-href="project/{{projectName}}/edit/autoscaler?kind=DeploymentConfig&name={{deploymentConfigName}}"
+         ng-href="project/{{projectName}}/edit/autoscaler?kind=DeploymentConfig&name={{deploymentConfigName}}&group=apps.openshift.io"
```
benjaminapetersen: Wondering if we should have things like `&group=apps.openshift.io` in our templates, or if that makes it easy to miss things later.  Could we potentially keep this in a variable in the controller near the `preferredVersion` stuff?

Just trying to think if we can avoid potential mismatches in the future, though I'm sure group changes will be fairly minimal (this is more of a question than a request for change).

app/views/browse/stateful-set.html
```diff
@@ -63,14 +63,16 @@ <h1>
               <div class="row" style="max-width: 650px;">
                 <div class="col-sm-4 col-sm-push-8 browse-deployment-donut">
                   <!--
```
benjaminapetersen: Assuming `autoscalers` would be part of issue #2723 work?

spadgett:
I'm going to break this down into smaller fixes.

/close

PR #2764, opened 2018-02-03
Bug 1540783 - Keep hostname updates in route editor

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1540783

### DIFF
--- a/app/scripts/controllers/edit/route.js
+++ b/app/scripts/controllers/edit/route.js
@@ -51,6 +51,7 @@ angular.module('openshiftConsole')
     var routesVersion = APIService.getPreferredVersion('routes');
     var servicesVersion = APIService.getPreferredVersion('services');
 
+    var host;
 
     ProjectsService
       .get($routeParams.project)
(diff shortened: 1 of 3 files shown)


spadgett:
Rebased.

benjaminapetersen:
Testing this now, just saw this error:

screen shot 2018-02-07 at 1 24 02 pm

But I'm not sure its related to your changes. Digging into it.

spadgett:
Good catch. Looks like it might be related to #2656. @TheRealJon can you look at the error @benjaminapetersen found?

I'd rather fix it in a separate PR since I need to backport this fix.

spadgett:
@benjaminapetersen mind opening an issue for that error?

benjaminapetersen:
Can do

app/scripts/controllers/edit/route.js
```diff
@@ -126,6 +127,13 @@ angular.module('openshiftConsole')
             _.set(updated, 'spec.to.weight', weight);
           }
```
benjaminapetersen: I'm tempted to want to see this bit function broken into smaller pieces instead of making it longer, but given you mentioned this is going to be backported, I can be ok with it 😄

PR #2768, opened 2018-02-05
Rewrite istagSelect directive into component

jhadvig:
@spadgett PTAL


app/scripts/directives/istagSelect.js
```diff
@@ -19,123 +20,133 @@ angular.module("openshiftConsole")
...
-          $scope.$watch('istag.namespace', function(namespace) {
-            if (!namespace || $scope.isByNamespace[namespace]) {
-              // We already have the data (or nothing selected).
-              return;
-            }
+      ctrl.namespaces = ctrl.namespaces.sort();
+
+      // Fetch image streams when a new namespace is picked.
+      $scope.$watch('$ctrl.istag.namespace', function(namespace) {
+        console.log(ctrl.isNamesByNamespace);
```
spadgett: remove console.log

app/scripts/directives/istagSelect.js
```diff
@@ -19,123 +20,133 @@ angular.module("openshiftConsole")
...
-          // Fetch image streams when a new namespace is picked.
-          $scope.$watch('istag.namespace', function(namespace) {
-            if (!namespace || $scope.isByNamespace[namespace]) {
-              // We already have the data (or nothing selected).
-              return;
-            }
+      ctrl.namespaces = ctrl.namespaces.sort();
+
+      // Fetch image streams when a new namespace is picked.
+      $scope.$watch('$ctrl.istag.namespace', function(namespace) {
```
spadgett: We should be using `ng-change` for this instead of `$scope.$watch`. Ideally we remove all references to `$scope` if possible since it doesn't exist in Angular2

app/scripts/directives/istagSelect.js
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
-angular.module("openshiftConsole")
+(function() {
```
spadgett: As we convert these, we should move them to the new directory structure.

app/scripts/directives/istagSelect.js
jhadvig: was a bit surprised that `ng-change` was just setting istag fields to `null`. Will rework the logic in this PR.

app/scripts/directives/istagSelect.js
jhadvig: makes sense ! wasnt sure since we still have most of the components in the directives folder. Will move it in this PR

jhadvig:
@spadgett thanks ! comments addressed

spadgett:
@jwforres FYI, one problem with moving these files into a components directory as we convert them is that we can lose the diff when reviewing in Github :(

We might need to move the files in a separate commit from the other changes going forward.

app/components/istag-select/istag-select.html
```diff
@@ -1,11 +1,11 @@
 <ng-form name="istagForm">
-  <fieldset ng-disabled="selectDisabled">
+  <fieldset ng-disabled="$ctrl.selectDisabled">
     <div class="row">
       <div class="form-group col-sm-4">
         <label class="sr-only">Namespace</label>
-        <ui-select ng-required="selectRequired" ng-model="istag.namespace" ng-disabled="selectDisabled" ng-change="istag.imageStream = null; istag.tagObject = null;" append-to-body="appendToBody">
+        <ui-select ng-required="$ctrl.selectRequired" ng-model="$ctrl.istag.namespace" ng-disabled="$ctrl.selectDisabled" ng-change="$ctrl.istag.imageStream = null;$ctrl.istag.tagObject = null;$ctrl.namespaceChanged($ctrl.istag.namespace)" append-to-body="appendToBody">
```
spadgett: It would be better to make this just

```html
ng-change="$ctrl.namespaceChanged($ctrl.istag.namespace)"
```

and null the other values in the namespaceChanged() function in the controller.

app/components/istag-select/istag-select.component.js
```diff
@@ -0,0 +1,151 @@
...
+      IstagSelect
+    ],
+    controllerAs: '$ctrl',
+    bindings: {
+      istag: '=model',
+      selectDisabled: '=',
+      selectRequired: '=',
+      includeSharedNamespace: '=',
+      allowCustomTag: '=',
+      appendToBody: '='
```
spadgett: Most of these should be one-way bindings with `<` (although `=model` needs to be a two-way binding as you have it).

jhadvig:
> FYI, one problem with moving these files into a components directory as we convert them is that we can lose the diff when reviewing in Github :(

So one solution would be to keep the directives where they are and just replace them with their components in index.html. Other option would be to as @spadgett suggested to have a separate commit for moving them.

jwforres:
Second commit to move them seems reasonable. Keep it in the same PR.

jhadvig:
@spadgett comments addressed and split the PR in PR into two separate commits, based on our conversation.

app/components/istag-select/istag-select.html
```diff
@@ -1,11 +1,11 @@
 <ng-form name="istagForm">
-  <fieldset ng-disabled="selectDisabled">
+  <fieldset ng-disabled="$ctrl.selectDisabled">
     <div class="row">
       <div class="form-group col-sm-4">
         <label class="sr-only">Namespace</label>
-        <ui-select ng-required="selectRequired" ng-model="istag.namespace" ng-disabled="selectDisabled" ng-change="istag.imageStream = null; istag.tagObject = null;" append-to-body="appendToBody">
+        <ui-select ng-required="$ctrl.selectRequired" ng-model="$ctrl.istag.namespace" ng-disabled="$ctrl.selectDisabled" ng-change="$ctrl.istag.imageStream = null;$ctrl.istag.tagObject = null;$ctrl.namespaceChanged($ctrl.istag.namespace)" append-to-body="$ctrl.appendToBody">
```
spadgett: This should be moved into the top of ctrl.namespaceChanged() and out of the view

```
$ctrl.istag.imageStream = null;$ctrl.istag.tagObject = null
```

jhadvig:
@spadgett sorry for that... accidentally forgot to save changes in the `istag-select.html`. Updated

app/components/istag-select/istag-select.component.js
```diff
@@ -0,0 +1,158 @@
...
+      }
+
+      ctrl.namespaces = ctrl.namespaces.sort();
+      ctrl.namespaceChanged(ctrl.istag.namespace);
+    });
+
+    ctrl.namespaceChanged = function (namespace){
+      // In case the selectboxes dont have to be prepopulated, which means the namespace selectbox
+      // value was changed by user, clear the imageStream and tag.
+      if (!shouldPrepopulate) {
```
spadgett: Are we sure this is necessary? According to the ng-change doc, it should only fire if the user changes the value. But maybe ui-select isn't following that.

https://code.angularjs.org/1.5.11/docs/api/ng/directive/ngChange

jhadvig: yes, cause we are calling the `namespaceChanged()` method to initialize the imageStream and tag selectboxes on line ~111.
Ultimately we could drop the `shouldPrepopulate` var and add an extra bool argument that would handle it.

jhadvig:
@spadgett do we want to proceed with merging this ?

jhadvig:
@spadgett had to rebuild the dist, since the PR was created some time ago... pls can you check and add the lgtm label. Thanks !

PR #2771, opened 2018-02-06
[WIP] feat(quotas): Add quota warnings to config maps and secrets pages.

TheRealJon:
## Description
Add warning alert to config maps and secrets pages and disabled `create` button if quota is already met for those resources.

## Changes
- Add `QuotaService.getLatestQuotas(context)` function. Returns a promise which resolves to `[<array> quotas,  <array> clusterQuotas]`
- Add `QuotaService.isAnyCurrentQuotaExceeded(context, typesToCheck)` function which determines if current quotas are already exceeded. No quotas are passed to this function because it automatically retrieves the latest quotas for the given context and checks against those. Works exactly like `QuotaService.isAnyQuotaExceeded` except that the caller does not need to provide the quotas.
- Some style/structure updates in the files that were touched. Mainly moving variable declarations to the top.
- `QuotaService.getLatestQuotaAlerts` now uses `QuotaService.getLatestQuotas`.
- Update storage controller to use new `QuotaService.isAnyCurrentQuotaExceeded` function
- Add quota check to secrets and config maps controllers and update templates to include alerts and conditional disabled `create` button.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1504007

### DIFF
--- a/app/scripts/controllers/configMaps.js
+++ b/app/scripts/controllers/configMaps.js
@@ -11,19 +11,47 @@ angular.module('openshiftConsole')
   .controller('ConfigMapsController',
               function ($scope,
                         $routeParams,
+                        AlertMessageService,
                         APIService,
                         DataService,
                         LabelFilter,
-                        ProjectsService) {
-    $scope.projectName = $routeParams.project;
-    $scope.loaded = false;
-    $scope.labelSuggestions = {};
-    $scope.configMapsVersion = APIService.getPreferredVersion('configmaps');
-    $scope.clearFilter = function () {
-      LabelFilter.clear();
-    };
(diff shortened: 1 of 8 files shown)


TheRealJon:
Still looking at the quota service to see if I can come up with a good strategy for creating the quota alerts. Right now, the logic for creating the alerts is repeated in all three controllers, but I feel like we should be able to pretty easily move that into the quota service.

benjaminapetersen:
+1 for controller cleanup, if its not a big change now!

TheRealJon:
Closing since this is a new feature and we are focusing on maintenance-only tasks in this repo.

PR #2774, opened 2018-02-06
Fix Bug 1518407 - A-MQ Components not visible in Java Console when using IE11 browser

akieling:
Fixes the following issue:
[https://bugzilla.redhat.com/show_bug.cgi?id=1518407](https://bugzilla.redhat.com/show_bug.cgi?id=1518407)

It also fixes the following additional error found during testing:
`Object doesn't support property or method 'startsWith'`

This PR includes the latest code from hawtio/openshift-jvm as discussed with @jwforres


spadgett:
@akieling The same fix should be made for all enterprise-3.4 - enterprise-3.8 branches. (I believe master is already current?)

akieling:
@spadgett I increased the patch version due to a few small fixes. I'll send a PR for master well.

spadgett:
@akieling thanks sounds good!

PR #2782, opened 2018-02-07
Poll quota on the Quota page

### LINKED ISSUE #2405: Poll quota on the Quota page
We already poll quota on Overview, we should start polling quota on the Quota page with the typical 1 minute interval.  This is the only page that doesn't live update so its a slightly weird experience if the user expects everything to be updating.

dtaylor113:
Fixes #2405 

Polling quota on the Quota page with the typical 1 minute interval.

### DIFF
--- a/app/scripts/controllers/quota.js
+++ b/app/scripts/controllers/quota.js
@@ -32,6 +32,7 @@ angular.module('openshiftConsole')
     var limitRangesVersion = APIService.getPreferredVersion('limitranges');
 
     var watches = [];
+    var DEFAULT_POLL_INTERVAL = 60 * 1000; // milliseconds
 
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/quota.js
```diff
@@ -32,6 +32,7 @@ angular.module('openshiftConsole')
     var limitRangesVersion = APIService.getPreferredVersion('limitranges');
 
     var watches = [];
+    var DEFAULT_POLL_INTERVAL = 60 * 1000; // milliseconds
```
benjaminapetersen: We should prob make this a `.constant()` since we have this `DEFAULT_POLL_INTERVAL` in a few places around the app now.

PR #2783, opened 2018-02-07
fix(routes): Fix undefined reference in oscRouting directive

### LINKED ISSUE #2780: Type error in oscRouting
```JavaScript 
// lin 302, looks like we assume the array exists when it might not be an array
if(!hideSlider && scope.route.alternateServices.length === 1)
```

screen shot 2018-02-07 at 1 24 02 pm

@TheRealJon

TheRealJon:
Add check to ensure alternateServices array exists before checking length in oscRouting directive.

Fixes #2780

### DIFF
--- a/app/scripts/directives/oscRouting.js
+++ b/app/scripts/directives/oscRouting.js
@@ -296,7 +296,7 @@ angular.module("openshiftConsole")
         });
 
         scope.$watch('controls.hideSlider', function(hideSlider){
-          if(!hideSlider && scope.route.alternateServices.length === 1){
+          if(!hideSlider && _.size(scope.route.alternateServices) === 1){
             initializingSlider = true;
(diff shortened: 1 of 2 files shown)


app/scripts/directives/oscRouting.js
```diff
@@ -296,7 +296,7 @@ angular.module("openshiftConsole")
         });
 
         scope.$watch('controls.hideSlider', function(hideSlider){
-          if(!hideSlider && scope.route.alternateServices.length === 1){
+          if(!hideSlider && scope.route.alternativeServices && scope.route.alternateServices.length === 1){
```
benjaminapetersen: I think you can 
`if(!hideSlider && _.size(scope.route.alternativeServices) === 1)`

PR #2786, opened 2018-02-08
Fix Bug 1543396 - A-MQ Components not visible in Java Console when using IE11 browser

akieling:
Fixes the following bug on OpenShift Enterprise 3.4:
https://bugzilla.redhat.com/show_bug.cgi?id=1543396


dist.java/java/404.html
```diff
@@ -38,10 +38,10 @@ <h4>Connected to <span ng-bind="containerName"></span></h4>
       <div ng-include src="viewPartial"></div>
     </div>
 
-    <script src="libs-2db3f78110.js"></script>
```
spadgett: Should the old JS libs-* and apps-* files be deleted?

akieling: You are right. I've removed the old files.

PR #2788, opened 2018-02-08
Fix Bug 1543435 - A-MQ Components not visible in Java Console when using IE11 browser

akieling:
Fixes the following bug on OpenShift Enterprise 3.6:
https://bugzilla.redhat.com/show_bug.cgi?id=1543435


spadgett:
@akieling Unfortunately, you might need to use an older version of npm in our older release branches. We had an issue where you would generate different dist depending on the version you used.

```diff
Verifying that checked in built files under dist match the source...
diff --git a/dist/scripts/vendor.js b/dist/scripts/vendor.js
index 1057b572..d27002ff 100644
--- a/dist/scripts/vendor.js
+++ b/dist/scripts/vendor.js
@@ -8062,7 +8062,7 @@ delete c[b.id], delete d[a];
 http:80,
 https:443,
 ftp:21
-}, sf = b("$location"), tf = /^\s*[\\\/]{2,}/, uf = {
+}, sf = b("$location"), tf = /^\s*[\\/]{2,}/, uf = {
 $$absUrl:"",
 $$html5:!1,
 $$replace:!1,
```

akieling:
@spadgett I've rebuilt the project using npm 3.10.10 and it seems to have fixed the issue with vendor.js.

spadgett:
@akieling I think it's the node version that's the problem rather than npm

spadgett:
disregard i see you say it's fixed the problem :)

PR #2792, opened 2018-02-08
Fix Bug 1543532 - A-MQ Components not visible in Java Console when using IE11 browser

akieling:
Fixes the following bug on OpenShift Enterprise 3.9:
https://bugzilla.redhat.com/show_bug.cgi?id=1543532


spadgett:
Please submit this to master branch instead. enterprise-3.9 currently tracks master

PR #2796, opened 2018-02-08
Bug 1543324 - Strip resourceVersion in import YAML for new resources

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1543324

/kind bug
/assign @benjaminapetersen 
/cc @juanvallejo


juanvallejo:
lgtm
Resembles what we do on the cli

PR #2797, opened 2018-02-09
Create ability to read annotation accessMode from StorageClasses

zherman0:
Create ability to read annotation accessMode from StorageClasses so that users do not have the option to select an access mode that is different from how the storage class is defined. Annotation is in the metadata section under annotations:
kind: StorageClass
apiVersion: storage.k8s.io/v1beta1
metadata:
  name: gold
  annotations:
    description: Gold is a storage class for showing gold level
    storage.alpha.openshift.io/access-mode: ReadWriteMany
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
  zone: us-east-1d

Drop down now shows if accessMode is defined in storageClass annotation, also separated description to new line:
annotations2-newstoragedropdown

If annotation is defined, access mode pre selected and disabled:
annotations4-storageclassselected

If the storage class has no annotation for the access mode, the access mode will still be visible:
annotations3-nostorageclassselected

@erinboyd @spadgett @jwforres @ncameronbritt

### DIFF
--- a/app/views/directives/osc-persistent-volume-claim.html
+++ b/app/views/directives/osc-persistent-volume-claim.html
@@ -8,11 +8,11 @@
       </div>
     </div>
     <div ng-if="storageClasses.length" class="form-group">
-      <!--storage class-->
       <label>Storage Class</label>
       <div>
         <ui-select
           ng-model="claim.storageClass"
+          on-select="onStorageClassSelected($item)"
           theme="bootstrap"
           search-enabled="true"
           title="Select a storage class"
@@ -26,8 +26,8 @@
             repeat="sclass as sclass in storageClasses | toArray | filter : { metadata: { name: $select.search } } ">
             <div>
               <span ng-bind-html="sclass.metadata.name  | highlight : $select.search"></span>
-              <span ng-if="sclass.parameters.type || sclass.parameters.zone || (sclass | description)" class="text-muted">
-                <small>&ndash;
+              <div ng-if="sclass.parameters.type || sclass.parameters.zone || (sclass | description) || (sclass | storageClassAccessMode)" class="text-muted">
+                <small>
                   <span ng-if="sclass.parameters.type">
                     Type: {{sclass.parameters.type}}
                   </span>
@@ -35,12 +35,16 @@
                     <span ng-if="sclass.parameters.type">|</span>
                     Zone: {{sclass.parameters.zone}}
                   </span>
-                  <span ng-if="sclass | description">
+                  <span ng-if="(sclass | storageClassAccessMode)">
                     <span ng-if="sclass.parameters.type || sclass.parameters.zone">|</span>
+                    Access: {{sclass | storageClassAccessMode}}
+                  </span>
+                  <span ng-if="sclass | description">
+                    <span ng-if="sclass.parameters.type || sclass.parameters.zone || (sclass | storageClassAccessMode)">|</span>
                     {{sclass | description}}
                   </span>
                 </small>
-              </span>
+              </div>
             </div>
           </ui-select-choices>
         </ui-select>
@@ -96,20 +100,31 @@
       </div>
     </div>
 
-    <div class="form-group" >
-      <label class="required">Access Mode</label><br/>
-      <label class="radio-inline">
-        <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadWriteOnce" aria-describedby="access-modes-help" ng-checked="true" >
-        Single User (RWO)
-      </label>
-      <label class="radio-inline">
-        <input type="radio" id="accessModes" name="accessModes" ng-model="claim.accessModes" value="ReadWriteMany" aria-describedby="access-modes-help" >
-        Shared Access (RWX)
-      </label>
-      <label class="radio-inline">
-        <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadOnlyMany" aria-describedby="access-modes-help">
-        Read Only (ROX)
-      </label>
+    <div class="form-group mar-bottom-xl" >
+      <label class="required">Access Mode</label>
+      <span ng-if="claim.storageClass && (claim.storageClass | storageClassAccessMode)" class="static-form-value-large">
+        <small>(cannot be changed)</small>
+      </span>
+      <div>
+        <label class="radio-inline">
+          <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadWriteOnce"
+                 aria-describedby="access-modes-help" ng-true-value="’1’" ng-false-value="’0’"
+                 ng-disabled="(claim.storageClass | storageClassAccessMode)">
+          Single User (RWO)
+        </label>
+        <label class="radio-inline">
+          <input type="radio" id="accessModes" name="accessModes" ng-model="claim.accessModes" value="ReadWriteMany"
+                 aria-describedby="access-modes-help" ng-true-value="’1’" ng-false-value="’0’"
+                 ng-disabled="(claim.storageClass | storageClassAccessMode)">
+          Shared Access (RWX)
+        </label>
+        <label class="radio-inline">
+          <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadOnlyMany"
+                 aria-describedby="access-modes-help" ng-true-value="’1’" ng-false-value="’0’"
+                 ng-disabled="(claim.storageClass | storageClassAccessMode)">
+          Read Only (ROX)
+        </label>
+      </div>
       <div>
         <span id="access-modes-help" class="help-block">Permissions to the mounted volume.</span>
       </div>
--- a/app/scripts/directives/oscPersistentVolumeClaim.js
+++ b/app/scripts/directives/oscPersistentVolumeClaim.js
@@ -24,12 +24,15 @@ angular.module("openshiftConsole")
       templateUrl: 'views/directives/osc-persistent-volume-claim.html',
       link: function(scope) {
         var amountAndUnit = $filter('amountAndUnit');
+        var storageClassAccessMode = $filter('storageClassAccessMode');
         var usageValue = $filter('usageValue');
 
         scope.nameValidation = DNS1123_SUBDOMAIN_VALIDATION;
 
         scope.storageClasses = [];
         scope.defaultStorageClass = "";
+        // Default to ReadWriteOnce access mode for new PVCs.
+        scope.claim.accessModes = 'ReadWriteOnce';
         scope.claim.unit = 'Gi';
         scope.units = [{
           value: "Mi",
@@ -84,6 +87,14 @@ angular.module("openshiftConsole")
 
         scope.showComputeUnitsHelp = function() {
           ModalsService.showComputeUnitsHelp();
+        };
+
+        scope.onStorageClassSelected = function(storageClass) {
+          // Update to use the access mode from the storage class if set.
+          var accessMode = storageClassAccessMode(storageClass);
+          if (accessMode) {
+            scope.claim.accessModes = accessMode;
+          }
         };
 
         var validateLimitRange = function() {
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -7,6 +7,11 @@ angular.module('openshiftConsole')
       return annotationFilter(pvc, 'volume.beta.kubernetes.io/storage-class');
     };
   })
+  .filter('storageClassAccessMode', function(annotationFilter) {
+    return function(storageClass) {
+      return annotationFilter(storageClass, 'storage.alpha.openshift.io/access-mode');
+    };
+  })
   .filter('tags', function(annotationFilter) {
     return function(resource, /* optional */ annotationKey) {
       annotationKey = annotationKey || "tags";
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -10180,8 +10180,8 @@ projectName: "="
 },
 templateUrl: "views/directives/osc-persistent-volume-claim.html",
 link: function(t) {
-var d = e("amountAndUnit"), m = e("usageValue");
-t.nameValidation = i, t.storageClasses = [], t.defaultStorageClass = "", t.claim.unit = "Gi", t.units = [ {
+var d = e("amountAndUnit"), m = e("storageClassAccessMode"), p = e("usageValue");
+t.nameValidation = i, t.storageClasses = [], t.defaultStorageClass = "", t.claim.accessModes = "ReadWriteOnce", t.claim.unit = "Gi", t.units = [ {
 value: "Mi",
 label: "MiB"
 }, {
@@ -10200,9 +10200,9 @@ label: "GB"
 value: "T",
 label: "TB"
 } ], t.claim.selectedLabels = [];
-var p = [];
+var f = [];
 t.$watch("useLabels", function(e, n) {
-e !== n && (e ? t.claim.selectedLabels = p : (p = t.claim.selectedLabels, t.claim.selectedLabels = []));
+e !== n && (e ? t.claim.selectedLabels = f : (f = t.claim.selectedLabels, t.claim.selectedLabels = []));
 }), t.groupUnits = function(e) {
 switch (e.value) {
 case "Mi":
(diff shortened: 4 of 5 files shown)


zherman0:
Addresses bug https://bugzilla.redhat.com/show_bug.cgi?id=1492617

spadgett:
We should add a namespace prefix to the annotation. This probably needs api review on the annotation name.

cc @smarterclayton 

Can you set a similar read-only style for access mode as we use here?

https://github.com/openshift/origin-web-console/blob/master/app/views/directives/osc-persistent-volume-claim.html#L119-L125

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -111,6 +111,9 @@ angular.module('openshiftConsole')
         if ($scope.claim.storageClass && $scope.claim.storageClass.metadata.name !== "No Storage Class") {
           //we can only have one storage class per claim
           pvc.metadata.annotations["volume.beta.kubernetes.io/storage-class"] = $scope.claim.storageClass.metadata.name;
+          if ($scope.claim.storageClass.metadata.annotations.accessMode) {
```
spadgett:
```js
var annotation = $filter('annotation');
if (annotation($scope.claim.storageClass, 'accessMode')) {
```

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -26,7 +26,7 @@
             repeat="sclass as sclass in storageClasses | toArray | filter : { metadata: { name: $select.search } } ">
             <div>
               <span ng-bind-html="sclass.metadata.name  | highlight : $select.search"></span>
-              <span ng-if="sclass.parameters.type || sclass.parameters.zone || (sclass | description)" class="text-muted">
+              <span ng-if="sclass.parameters.type || sclass.parameters.zone || (sclass | description) || (sclass.metadata.annotations.accessMode)" class="text-muted">
```
spadgett: Use the `annotation` filter.

```
|| (sclass | annotation : 'accessMode')
```

Here and several times below.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -114,6 +118,9 @@
         <span id="access-modes-help" class="help-block">Permissions to the mounted volume.</span>
       </div>
     </div>
+    <div class="form-group" ng-if="claim.storageClass && claim.storageClass.metadata.annotations.accessMode && claim.storageClass.metadata.name !== 'No Storage Class'">
+      <label class="required">Access Mode (Set as {{claim.storageClass.metadata.annotations.accessMode}} by the selected storage class)</label><br/>
```
spadgett:
```html
<label>Access Mode</label>
<div class="static-form-value-large">
  {{claim | annotation : 'accessMode'}}
  <small>(cannot be changed)</small>
</div>
```

spadgett:
I would suggest something like `storage.openshift.io/access-mode` for the annotation.

ncameronbritt:
I think for the sake of consistency that I would rather see the Access Mode as a radio button, even if there is only one supported mode. In that case it should be selected by default. Then if two access modes were supported, we would show those two.

I'm also curious in the Storage Class dropdown if moving the descriptive info to its own line(s), rather than inline with the storage class name, would make it easier to parse. We do something similar with roles on the membership page. 
screen shot 2018-02-14 at 2 38 53 pm

cc @openshift/team-ux-review

erinboyd:
Yes, I think moving it out would make it more readable.

spadgett:
I'd like to target this fix for 3.9 if we can. It's low-risk and improves the user experience.

@sspeiche

jwforres:
+1 for 3.9 if it's low risk

sspeiche:
+1 for 3.9. Would like fixing it as soon as possible, which appears to be 3.9 at this point

erinboyd:
+1

On Feb 16, 2018 8:24 AM, "Steve Speicher" <notifications@github.com> wrote:

+1 for 3.9. Would like fixing it as soon as possible, which appears to be
3.9 at this point

—
You are receiving this because you were mentioned.

<https://github.com/openshift/origin-web-console/pull/2797#issuecomment-366265524>,
or mute the thread

spadgett:
@zherman0 It'd be good to target tues or wed next week for updates so we have time to get the change in for 3.9.

spadgett:
I talked to @jwforres and we agreed on `storage.alpha.openshift.io/access-mode` for the annotation

zherman0:
@spadgett - made the recommended changes, let me know if this works for you.  Thanks.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -19,28 +19,32 @@
           class="select-role">
           <ui-select-match placeholder="Select a storage class">
             <span>
-              {{$select.selected.metadata.name}}
+              {{$select.selected.metadata.name}}<br/>
```
spadgett: Why is the `br` needed?

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -19,28 +19,32 @@
...
             </span>
           </ui-select-match>
           <ui-select-choices
             repeat="sclass as sclass in storageClasses | toArray | filter : { metadata: { name: $select.search } } ">
             <div>
               <span ng-bind-html="sclass.metadata.name  | highlight : $select.search"></span>
-              <span ng-if="sclass.parameters.type || sclass.parameters.zone || (sclass | description)" class="text-muted">
-                <small>&ndash;
+              <div ng-if="sclass.parameters.type || sclass.parameters.zone || (sclass | description) || (sclass | annotation : 'storage.alpha.openshift.io/access-mode')" class="text-muted">
+                <small>&nbsp;
```
spadgett: Why is `&nbsp;` needed?

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -97,17 +101,30 @@
...
+      <label class="required">Access Mode
+        <span ng-if="claim.storageClass && (claim.storageClass | annotation : 'storage.alpha.openshift.io/access-mode')">
+          <small> (cannot be changed)</small>
+        </span>
+      </label><br/>
       <label class="radio-inline">
-        <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadWriteOnce" aria-describedby="access-modes-help" ng-checked="true" >
+        <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadWriteOnce"
+               aria-describedby="access-modes-help" ng-true-value="’1’" ng-false-value="’0’"
+               ng-checked="(claim.storageClass | annotation : 'storage.alpha.openshift.io/access-mode')=='ReadWriteOnce'"
```
spadgett: Please use the `===` operator here and below.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -19,28 +19,32 @@
...
+                <small>&nbsp;
                   <span ng-if="sclass.parameters.type">
                     Type: {{sclass.parameters.type}}
                   </span>
                   <span ng-if="sclass.parameters.zone">
                     <span ng-if="sclass.parameters.type">|</span>
                     Zone: {{sclass.parameters.zone}}
                   </span>
-                  <span ng-if="sclass | description">
+                  <span ng-if="(sclass | annotation : 'storage.alpha.openshift.io/access-mode')">
```
spadgett: You have this check repeated so many times in the view that it's probably better as a filter.

```js
filter('storageClassAccessMode', function(annotationFilter) {
  return function(storageClass) {
    return annotationFilter(storageClass, 'storage.alpha.openshift.io/access-mode');
  };
});
```

Then everywhere in the view you can use

```html
<span ng-if="sclass | storageClassAccessMode">
```

This also keeps the annotation name in one place if we ever change it.

app/views/directives/osc-persistent-volume-claim.html
zherman0: Removed

app/views/directives/osc-persistent-volume-claim.html
zherman0: I thought it looked better:
With blank -
withblank

Without Blank -
withoutblank

What do you think?

spadgett: Left-aligned is consistent with other similar messages, so I'd leave out the leading space. If we really want that, we should use margins instead anyway. But I like it better without.

zherman0:
@spadgett - does this version work for you?

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -111,6 +111,10 @@ angular.module('openshiftConsole')
         if ($scope.claim.storageClass && $scope.claim.storageClass.metadata.name !== "No Storage Class") {
           //we can only have one storage class per claim
           pvc.metadata.annotations["volume.beta.kubernetes.io/storage-class"] = $scope.claim.storageClass.metadata.name;
+          var annotation = $filter('storageClassAccessMode');
```
spadgett: `annotation` is no longer an appropriate var name since we're using a different filter.

```js
var storageClassAccessMode = $filter('storageClassAccessMode');
```

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -26,21 +26,25 @@
...
                     Type: {{sclass.parameters.type}}
                   </span>
                   <span ng-if="sclass.parameters.zone">
                     <span ng-if="sclass.parameters.type">|</span>
                     Zone: {{sclass.parameters.zone}}
                   </span>
-                  <span ng-if="sclass | description">
+                  <span ng-if="(sclass | storageClassAccessMode)">
                     <span ng-if="sclass.parameters.type || sclass.parameters.zone">|</span>
+                    Access: {{(sclass | storageClassAccessMode)}}
```
spadgett: nit: Parens aren't needed here. I think it's harder to read with them in this context since there are already curly braces.

```
Access: {{sclass | storageClassAccessMode}}
```

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -97,17 +101,30 @@
     </div>
 
     <div class="form-group" >
-      <label class="required">Access Mode</label><br/>
+      <label class="required">Access Mode
+        <span ng-if="claim.storageClass && (claim.storageClass | storageClassAccessMode)">
+          <small> (cannot be changed)</small>
```
spadgett: nit: remove space before `(`

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -97,17 +101,30 @@
...
+        </span>
+      </label><br/>
```
spadgett: remove `br`

spadgett:
@zherman0 I noticed this introduces a bug when you aren't using storage classes. An access mode is no longer preselected when it was, and the create button is enabled without having to pick one. 

openshift web console 2018-02-20 14-06-55

zherman0:
@spadgett - Nice pickup - the code would have filled in a default, but that was definitely a bug to not show that on original load

spadgett:
Won't this still be a problem for storage classes that don't have the access-mode annotation?

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -97,19 +101,34 @@
...
+      <label class="required">Access Mode
+        <span ng-if="claim.storageClass && (claim.storageClass | storageClassAccessMode)">
+          <small>(cannot be changed)</small>
+        </span>
       </label>
+      <div>
+        <label class="radio-inline">
+          <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadWriteOnce"
+                 aria-describedby="access-modes-help" ng-true-value="’1’" ng-false-value="’0’"
+                 ng-checked="((claim.storageClass | storageClassAccessMode)==='ReadWriteOnce') || !claim.storageClass || (claim.storageClass.metadata.name === 'No Storage Class')"
```
spadgett: Will this prevent me from selecting another value?

zherman0: @spadgett - You can select other values if the access mode is not defined.  You are right about a default not being selected for a storage class without a access mode.  I have updated my test to include a storage class like that and I have fixed the problem.

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -97,19 +101,36 @@
...
+          <small>(cannot be changed)</small>
+        </span>
       </label>
+      <div>
+        <label class="radio-inline">
+          <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadWriteOnce"
+                 aria-describedby="access-modes-help" ng-true-value="’1’" ng-false-value="’0’"
+                 ng-checked="((claim.storageClass | storageClassAccessMode)==='ReadWriteOnce')
+                  || !claim.storageClass || (claim.storageClass.metadata.name === 'No Storage Class')
+                  || !(claim.storageClass | storageClassAccessMode)"
```
spadgett: Can't this simply be

```
ng-checked="!(claim.storageClass | storageClassAccessMode) || (claim.storageClass | storageClassAccessMode)==='ReadWriteOnce')"
```

zherman0: It can now, since the "No Service Class" option is covered.  Let me test and change it.

zherman0: Ok, I had to leave in the '|| !claim.storageClass' for when nothing is chosen in the drop down.  Checking in now.

zherman0:
@spadgett - is there a way to handle that in the model if they never select a storage class (or even touch) from the dropdown?

spadgett:
> is there a way to handle that in the model if they never select a storage class from the dropdown (or even touch the drop down)?

Yes, you can initialize the model to the default value in the controller.

zherman0:
@spadgett - Can we go ahead and use this solution for now instead of moving more of the logic into the model? I think this solution has the bugs worked out and it is fairly easy to follow.

spadgett:
@zherman0 I opened a PR against your branch with the change:

https://github.com/zherman0/origin-web-console/pull/1

I'm wary of using both `ng-checked` and `ng-model` since you have two attributes setting the checked value of the checkbox and the angular docs specifically tell you not to do that. I feel like that might cause hard-to-find bugs down the road.

app/scripts/controllers/createPersistentVolumeClaim.js
```diff
@@ -111,6 +111,10 @@ angular.module('openshiftConsole')
         if ($scope.claim.storageClass && $scope.claim.storageClass.metadata.name !== "No Storage Class") {
           //we can only have one storage class per claim
           pvc.metadata.annotations["volume.beta.kubernetes.io/storage-class"] = $scope.claim.storageClass.metadata.name;
+          var storageClassAccessMode = $filter('storageClassAccessMode');
+          if (storageClassAccessMode($scope.claim.storageClass)) {
+            pvc.spec.accessModes = [storageClassAccessMode($scope.claim.storageClass)];
+          }
```
spadgett: @zherman0 Missed this in my PR, but I don't think this change is needed anymore. It should already be set since we handle it in oscPersistentVolumeClaim.js

app/views/directives/osc-persistent-volume-claim.html
```diff
@@ -97,19 +101,32 @@
...
-        <input type="radio" id="accessModes" name="accessModes" ng-model="claim.accessModes" value="ReadWriteMany" aria-describedby="access-modes-help" >
-        Shared Access (RWX)
-      </label>
-      <label class="radio-inline">
-        <input type="radio" name="accessModes" ng-model="claim.accessModes" value="ReadOnlyMany" aria-describedby="access-modes-help">
-        Read Only (ROX)
+      <label class="required">
+        Access Mode
+        <span ng-if="claim.storageClass && (claim.storageClass | storageClassAccessMode)">
+          <small>(cannot be changed)</small>
```
spadgett: Missed this as well, but the "cannot be changed" font style here is different here compare to size. It would be good to be consistent.

openshift web console 2018-02-26 12-48-09

zherman0: @spadgett - Remove code and here is the new screen shot:
screenshot from 2018-02-26 11-35-47

Issue #2775, opened 2018-02-07
BC's build secret destination dir label and help-text are not align properly 


jhadvig:
Both label and help-text are not align properly. Screen:
![download](https://user-images.githubusercontent.com/1668218/35924048-3c304c70-0c22-11e8-88b7-d27702386ad1.png)

/assign

PR #2799, opened 2018-02-12
BC's build secret destination dir label and help-text are not align properly

jhadvig:
Replacing `destination-dir` class with `secrets-location` that should be added on both `label` and `div.help-block` elements that are surrounding the *Destination Directory* and *Mount path* input fields.
Screen for both Source and Custom build configs:
download 2

download

Fixes https://github.com/openshift/origin-web-console/issues/2775

### DIFF
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -288,7 +288,7 @@
                               <div class="is-item-description">
                                 <dt>Paths:</dt>
                                 <div ng-repeat="(source, destination) in imageSourcesPaths[$index]" class="image-source-paths">
-                                  <dd><span class="source-path">{{source}}</span><i class="fa fa-long-arrow-right"></i><span class="destination-dir">{{destination}}</span></dd>
+                                  <dd><span class="source-path">{{source}}</span><i class="fa fa-long-arrow-right" aria-hidden="true"></i><span class="destination-dir">{{destination}}</span></dd>
                                 </div>
                               </div>
                             </dl>
--- a/app/views/directives/osc-source-secrets.html
+++ b/app/views/directives/osc-source-secrets.html
@@ -7,7 +7,7 @@
           <label class="input-label">
             Build Secret
           </label>
-          <label class="input-label">
+          <label class="input-label destination-dir-padding">
             Destination Directory
           </label>
         </div>
@@ -21,7 +21,7 @@
                 </ui-select-choices>
               </ui-select>
             </div>
-            <div class="destination-dir">
+            <div class="destination-dir destination-dir-padding">
               <input class="form-control"
                 id="destinationDir"
                 name="destinationDir"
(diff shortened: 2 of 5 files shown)


jhadvig:
cc'ing @spadgett

app/views/browse/build-config.html
```diff
@@ -288,7 +288,7 @@ <h4>{{imageSource.from | imageObjectRef : buildConfig.metadata.namespace}}</h4>
                               <div class="is-item-description">
                                 <dt>Paths:</dt>
                                 <div ng-repeat="(source, destination) in imageSourcesPaths[$index]" class="image-source-paths">
-                                  <dd><span class="source-path">{{source}}</span><i class="fa fa-long-arrow-right"></i><span class="destination-dir">{{destination}}</span></dd>
+                                  <dd><span class="source-path">{{source}}</span><i class="fa fa-long-arrow-right"></i><span class="secrets-location">{{destination}}</span></dd>
```
spadgett: Not new, but the icons should have `aria-hidden="true"`

app/styles/_core.less
```diff
@@ -281,7 +281,7 @@ registry-imagestream-meta dt:after {
 
 .image-sources {
   .source-path:after,
-  .destination-dir:before {
+  .secrets-location:before {
     content: "\00a0 ";
```
spadgett: I'm a bit confused by this style. `\00a0` appears to be a non-breaking space, but then we're adding a regular space right after? Doesn't that defeat the purpose of the non-breaking space?

app/views/browse/build-config.html
spadgett: I don't think we should use the same class for the read-only view and the form inputs. They're different presentations and won't necessarily have the same padding.

app/views/directives/osc-source-secrets.html
```diff
@@ -90,7 +90,7 @@
         </div>
         <div class="help-blocks">
             <div class="help-block">Source secret to mount into the builder pod at build time.</div>
```
spadgett: nit: indentation is off here

sg00dwin:
@jhadvig I'm reviewing your changes. Do you have a test file I can use to view [this screen](https://user-images.githubusercontent.com/1668218/36097515-96495508-0ffb-11e8-9f08-45a405264de3.png)

jhadvig:
@sg00dwin not sure what do you mean under test file. If you mean a BuildConfig, just use the ruby example app and then check the `Advanced Options` in the BC editor. There you will see `Source Secret` with the `Destination Directory` input field.

sg00dwin:
I'm referring to the screenshot with the label "Mount Path". I can't seem to locate that view?

I do see the one with Destination Directory

jhadvig:
@spadgett comments addressed.

PR #2802, opened 2018-02-14
Fix bugzilla 1545011: cpu input disappears in edit autoscaler if value set to 0

benjaminapetersen:
Part of the work to tolerate v2beta1 autoscalers.

- if the user sets the cpu input to `0`, the input would disappear
- this change makes the attribute bound one-way, which ensures we can hide it on load (if the appropriate conditions are met) but that the user input doesn't flip the `ng-show` to a falsy value while editing.  It should be set to false by the controller only if it is already falsy && metrics are in use.


jwforres:
lgtm but you are missing a dist file it looks like

benjaminapetersen:
great, thx. yup, there is my `dist/template` change.

PR #2810, opened 2018-02-19
Bump origin-web-common to 3.9.0-alpha.14

### LINKED ISSUE #2723: Stateful sets can be scaled from overview but not stateful set page

spadgett:
https://github.com/openshift/origin-web-common/releases/tag/v3.9.0-alpha.14

Closes #2723

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -46,7 +46,7 @@
     "angular-moment": "1.0.0",
     "angular-utf8-base64": "0.0.5",
     "file-saver": "1.3.3",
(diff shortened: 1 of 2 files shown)


spadgett:
@rhamilto ptal

spadgett:
I added the additional one line fix in common 3.9.0-alpha.14 for #2723

PR #2819, opened 2018-02-21
Update buildPipeline directive to use getPreferredVersion

benjaminapetersen:
Another.

As an aside, I was not able to create the pipeline template:

screen shot 2018-02-20 at 8 40 06 pm

Strange?


spadgett:
There was a bug for this that I think got fixed. Can you verify that you've pulled the latest images of everything (including service catalog) and clustered up with service catalog enabled?

PR #2822, opened 2018-02-21
update replica set actions to use getPreferredVersion for canI checks

benjaminapetersen:
Def missed some `canI` checks in some `includes` templates & directives as well.

### DIFF
--- a/app/scripts/controllers/replicaSet.js
+++ b/app/scripts/controllers/replicaSet.js
@@ -36,7 +36,6 @@ angular.module('openshiftConsole')
 
     var buildsVersion = APIService.getPreferredVersion('builds');
     var imageStreamsVersion = APIService.getPreferredVersion('imagestreams');
-    var horizontalPodAutoscalersVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
     var limitRangesVersion = APIService.getPreferredVersion('limitranges');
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/replicaSet.js
```diff
@@ -79,9 +78,10 @@ angular.module('openshiftConsole')
 
     $scope.deploymentsVersion = APIService.getPreferredVersion('deployments');
     $scope.deploymentConfigsVersion = APIService.getPreferredVersion('deploymentconfigs');
+    $scope.deploymentConfigsLogVersion = APIService.getPreferredVersion('deploymentconfigs/log');
     $scope.eventsVersion = APIService.getPreferredVersion('events');
-    // TODO: update common/constants/apiPreferredVersions for this
-    $scope.deploymentConfigsLogVersion = 'deploymentconfigs/log';
+    $scope.horizontalPodAutoscalersVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
+    $scope.replicationControllersVersion = replicationControllersVersion;
```
spadgett: Assign the value directly to the scope var and use the scope var in the controller if needed to be consistent with other preferred versions

benjaminapetersen: Can do, I did it this way because the value is used in a switch statement above already.  I can just move this whole block up.

PR #2824, opened 2018-02-21
Update replica set details include to use getPreferredVersion

benjaminapetersen:
Missing update to replicaSet details include.

### DIFF
--- a/app/scripts/controllers/replicaSet.js
+++ b/app/scripts/controllers/replicaSet.js
@@ -36,7 +36,6 @@ angular.module('openshiftConsole')
 
     var buildsVersion = APIService.getPreferredVersion('builds');
     var imageStreamsVersion = APIService.getPreferredVersion('imagestreams');
-    var horizontalPodAutoscalersVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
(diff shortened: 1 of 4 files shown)


app/scripts/controllers/replicaSet.js
```diff
@@ -79,9 +78,10 @@ angular.module('openshiftConsole')
 
     $scope.deploymentsVersion = APIService.getPreferredVersion('deployments');
     $scope.deploymentConfigsVersion = APIService.getPreferredVersion('deploymentconfigs');
+    $scope.deploymentConfigsLogVersion = APIService.getPreferredVersion('deploymentconfigs/log');
     $scope.eventsVersion = APIService.getPreferredVersion('events');
-    // TODO: update common/constants/apiPreferredVersions for this
-    $scope.deploymentConfigsLogVersion = 'deploymentconfigs/log';
+    $scope.horizontalPodAutoscalersVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
+    $scope.replicationControllersVersion = replicationControllersVersion;
```
spadgett: Just assign it directly to the scope var since that's what we do for other version above.

PR #2832, opened 2018-02-21
Bug 1547348 Use project annotation to determine logging data prefix

jcantrill:
The PR fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1547348

### DIFF
--- a/app/scripts/directives/logViewer.js
+++ b/app/scripts/directives/logViewer.js
@@ -427,7 +427,7 @@ angular.module('openshiftConsole')
                                         podname: name,
                                         containername: $scope.options.container,
                                         backlink: URI.encode($window.location.href)
-                                      }))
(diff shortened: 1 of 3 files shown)


app/scripts/directives/logViewer.js
```diff
@@ -427,7 +427,7 @@ angular.module('openshiftConsole')
                                         podname: name,
                                         containername: $scope.options.container,
                                         backlink: URI.encode($window.location.href)
-                                      }))
+                                      }, $scope.context.project.metadata.annotations['openshift.io/logging.data.prefix']))
```
spadgett: This will be a runtime error if the project doesn't have any annotations. Better to use the annotation filter.

```
$filter('annotation')($scope.context.project, 'openshift.io/logging.data.prefix')))
```

PR #2833, opened 2018-02-21
Fix potential mismatch in createSecret

benjaminapetersen:
Combing one more time for these potential mismatches as well after finding that I missed a bunch of includes/directive templates for the `canI` check.

### DIFF
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -113,7 +113,7 @@ angular.module("openshiftConsole")
 
         var constructSecretObject = function(data, authType) {
           var secret = {
-            apiVersion: "v1",
+            apiVersion: APIService.toAPIVersion(secretsVersion),
             kind: "Secret",
             metadata: {
               name: $scope.newSecret.data.secretName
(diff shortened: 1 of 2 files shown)


benjaminapetersen:
/retest

```
1) User adds a template to a project after creating a new project using the Import YAML tab should save the template in the project catalog
  - Error: Wait timed out after 5117ms
...
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
  - timeout: timed out after 60000 msec waiting for spec to complete
```

app/scripts/directives/createSecret.js
```diff
@@ -113,7 +113,7 @@ angular.module("openshiftConsole")
 
         var constructSecretObject = function(data, authType) {
           var secret = {
-            apiVersion: "v1",
+            apiVersion: APIService.toAPIVersion(secretsVersion),
```
spadgett: I suggest we also call `APIService.objectToResourceGroupVersion` before we call `DataService.create` below to make sure the versions always match.

app/scripts/directives/createSecret.js
```diff
@@ -242,7 +242,7 @@ angular.module("openshiftConsole")
         $scope.create = function() {
           hideErrorNotifications();
           var newSecret = constructSecretObject($scope.newSecret.data, $scope.newSecret.authType);
-          DataService.create(secretsVersion, null, newSecret, $scope).then(function(secret) { // Success
+          DataService.create(APIService.objectToResourceGroupVersion(newSecret), null, newSecret, $scope).then(function(secret) { // Success
```
benjaminapetersen: @spadgett ah, I agree.  This should be the correct approach, right?

Issue #2835, opened 2018-02-21
Add support for viewing mobileclient CRD in the project overview screen


maleck13:
As part of on going work to make mobile development as simple and appealing as possible using OpenShift, we would like to allow developers to view mobile client representations [mobileclient CRD](https://github.com/aerogear/mobile-cli/blob/master/artifacts/mobileclient_crd.yaml) in the OpenShift overview screen. 
If the  CRD is present , it will enable parts of the UI designed with mobile developers in mind. 

This is the first issue in a larger piece of work to expose a set of mobile focused services and extensions to the OpenShift UI that are designed around improving the experience of mobile developers looking to take advantage of or interested in using OpenShift for their backend services. The mobileclient will be used in the future to allow developers to choose which services they wish to expose to their mobile clients, retrieve the configuration necessary to communicate with those services, and perform native builds via a build service. 

ping @cshinn @spadgett 

Will have a PR for this shortly

PR #2838, opened 2018-02-21
Allow viewing mobile clients in the overview screen

maleck13:
This is an initial PR focused around show mobile clients in the overview screen. Covered under the following issue:
https://github.com/openshift/origin-web-console/issues/2835

**Note to reviewer**

Looking for guidance here on approach and changes made. Happy to change based on feedback

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -371,6 +371,7 @@
         <script src="scripts/directives/create/nextSteps.js"></script>
         <script src="scripts/directives/imageNames.js"></script>
         <script src="scripts/directives/serviceBinding.js"></script>
+        <script src="scripts/directives/overview/mobileClientRow.js"></script>
 
         <!-- New Overview -->
         <script src="scripts/directives/buildCounts.js"></script>
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -631,6 +631,31 @@ angular
       // content (e.g. using :before pseudo-elements).
       $('body').addClass('ios');
     }
+  })
+  .run(function ($rootScope, APIService) {
+    $rootScope.AEROGEAR_MOBILE_ENABLED = !!APIService.apiInfo({ resource: "mobileclients", group: "mobile.k8s.io" });
+    if ($rootScope.AEROGEAR_MOBILE_ENABLED) {
+      // Add 'Mobile' category and sub-categories to the Service Catalog UI
+      window.OPENSHIFT_CONSTANTS.SERVICE_CATALOG_CATEGORIES.push(
+        {
+          id: 'mobile',
+          label: 'Mobile',
+          subCategories: [
+            {
+              id: 'apps',
+              label: 'Apps',
+              tags: ['mobile'],
+              icon: 'fa fa-mobile'
+            },
+            {
+              id: 'services',
+              label: 'Services',
+              tags: ['mobile-service'],
+              icon: 'fa fa-database'
+            }]
+        });
+    }
+    Logger.info("AEROGEAR_MOBILE_ENABLED: " + $rootScope.AEROGEAR_MOBILE_ENABLED);
   });
 
 hawtioPluginLoader.addModule('openshiftConsole');
--- a/app/scripts/directives/overview/mobileClientRow.js
+++ b/app/scripts/directives/overview/mobileClientRow.js
@@ -1,0 +1,69 @@
+'use strict';
+
+(function () {
+  angular.module('openshiftConsole').component('mobileClientRow', {
+    controller: [
+      '$scope',
+      '$filter',
+      '$routeParams',
+      'APIService',
+      'AuthorizationService',
+      'DataService',
+      'ListRowUtils',
+      'Navigate',
+      'ProjectsService',
+      MobileAppRow,
+    ],
+    controllerAs: 'row',
+    bindings: {
+      apiObject: '<',
+      state: '<'
+    },
+    templateUrl: 'views/overview/_mobile-client-row.html'
+  });
+
+  function MobileAppRow($scope, $filter, $routeParams, APIService, AuthorizationService, DataService, ListRowUtils, Navigate, ProjectsService) {
+    var row = this;
+    row.installType = '';
+
+    _.extend(row, ListRowUtils.ui);
+
+    row.$onChanges = function(changes) {
+      if (changes.apiObject) {
+        row.bundleDisplay = row.apiObject.spec.appIdentifier;
+        row.clientType = row.apiObject.spec.clientType.toUpperCase();
+        switch (row.apiObject.spec.clientType) {
+          case 'android':
+            row.installType = 'gradle';
+            break;
+          case 'iOS':
+            row.installType = 'cocoapods';
+            break;
+          case 'cordova':
+            row.installType = 'npm';
+            break;
+        }
+      }
+    };
+
+
+    row.mobileclientVersion = {
+      group: "mobile.k8s.io",
+      version: "v1alpha1",
+      resource: "mobileclients"
+    };
+    row.actionsDropdownVisible = function () {
+      // no actions on those marked for deletion
+      if (_.get(row.apiObject, 'metadata.deletionTimestamp')) {
+        return false;
+      }
+
+      // We can delete mobileclients
+      return AuthorizationService.canI(row.mobileclientVersion, 'delete');
+    };
+    row.projectName = $routeParams.project;
+    row.browseCatalog = function () {
+      Navigate.toProjectCatalog(row.projectName);
+    };
+  }
+})();
--- a/app/views/overview.html
+++ b/app/views/overview.html
@@ -379,7 +379,16 @@
                 </overview-list-row>
               </div>
             </div>
-
+            <!-- Show Mobile Apps -->
+            <div ng-if="(overview.filteredMobileClients | size) && AEROGEAR_MOBILE_ENABLED">
+              <h2>Mobile Clients</h2>
+              <div class="list-pf">
+                <mobile-client-row
+                  ng-repeat="mobileapp in overview.filteredMobileClients track by (mobileapp | uid)"
+                  api-object="mobileapp"
+                  state="overview.state"></mobile-client-row>
+              </div>
+            </div>
             <!-- Show provisioned services. -->
             <div ng-if="overview.filteredServiceInstances.length && !overview.hidePipelineOtherResources">
               <h2>
--- a/app/views/overview/_mobile-client-row.html
+++ b/app/views/overview/_mobile-client-row.html
@@ -1,0 +1,56 @@
+<div class="list-pf-item" ng-class="{ active: row.expanded }">
+  <div class="list-pf-container" ng-click="row.toggleExpand($event)">
+    <div class="list-pf-chevron">
+      <div ng-include src=" 'views/overview/_list-row-chevron.html' " class="list-pf-content"></div>
+    </div>
+    <div class="list-pf-content">
+      <div class="list-pf-name">
+        <h3>
+          <div class="list-row-longname"><span>{{row.clientType}}</span></div>
+          <!--<span class="fa-2x {{row.apiObject.metadata.annotations.icon}}"></span>-->
+          <a ng-href="{{row.apiObject | navigateResourceURL}}"><span ng-bind-html="row.apiObject.spec.name | highlightKeywords : row.state.filterKeywords"></span></a>
+
+
+          <div class="list-row-longname">{{row.bundleDisplay}}</div>
+        </h3>
+      </div>
+    </div>
+    <div class="list-pf-actions">
+      <div class="dropdown-kebab-pf" uib-dropdown ng-if="row.actionsDropdownVisible()">
+        <button uib-dropdown-toggle class="btn btn-link dropdown-toggle">
+          <i class="fa fa-ellipsis-v" aria-hidden="true"></i>
+          <span class="sr-only">Actions</span>
+        </button>
+        <ul class="dropdown-menu dropdown-menu-right" uib-dropdown-menu role="menu">
+          <li ng-if="row.mobileclientVersion | canI : 'delete'">
+            <delete-link
+              kind="MobileClient"
+              group="mobile.k8s.io"
+              stay-on-current-page="true"
+              resource-name="{{row.apiObject.metadata.name}}"
+              project-name="{{row.projectName}}">
+            </delete-link>
+          </li>
+        </ul>
+      </div>
+    </div>
+  </div>
+  <div class="list-pf-expansion collapse" ng-if="row.expanded" ng-class="{ in: row.expanded }">
+    <div class="list-pf-container">
+      <div class="expanded-section">
+        <div class="empty-state-message text-center">
+          <p>Add a mobile service to your project. Or connect to external service.</p>
+          <div class="empty-state-message-main-action">
+            <!--TODO link through to the mobile category -->
+            <button class="btn btn-primary btn-lg" ng-click="row.browseCatalog()">
+              Browse Mobile Services
+            </button>
+          </div>
+        </div>
+        <div ng-if="loading">
+          Loading...
+        </div>
+      </div>
+    </div>
+  </div>
+</div>
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -205,7 +205,8 @@ function OverviewController($scope,
            _.size(overview.statefulSets) +
            _.size(overview.daemonSets) +
            _.size(overview.monopods) +
-           _.size(overview.state.serviceInstances);
+           _.size(overview.state.serviceInstances) +
+           _.size(overview.mobileClients);
   };
 
   // The size of all visible top-level items after filtering.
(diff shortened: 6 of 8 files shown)


app/scripts/app.js
```diff
@@ -631,6 +631,28 @@ angular
       // content (e.g. using :before pseudo-elements).
       $('body').addClass('ios');
     }
+  })
+  .run(function ($rootScope, APIService) {
+    $rootScope.AEROGEAR_MOBILE_ENABLED = !!APIService.apiInfo({ resource: "mobileclients", group: "mobile.k8s.io" });
+    if ($rootScope.AEROGEAR_MOBILE_ENABLED) {
+      // Add 'Mobile' category and sub-categories to the Service Catalog UI
+      window.OPENSHIFT_CONSTANTS.SERVICE_CATALOG_CATEGORIES.splice(
```
spadgett: If you're adding to the end, `Array.push` is probably better.

app/scripts/app.js
```diff
@@ -631,6 +631,28 @@ angular
...
+          id: 'mobile',
+          label: 'Mobile',
+          subCategories: [
+            { id: 'apps', label: 'Apps', tags: ['mobile'], icon: 'fa fa-mobile' },
+            {
+              id: 'services',
+              label: 'Services',
+              tags: ['mobile-service'],
+              icon: 'fa fa-database'
+            }]
```
spadgett: nit: Better to keep consistent formatting for these items. (I prefer splitting it across lines.)

app/scripts/filters/util.js
```diff
@@ -514,4 +514,22 @@ angular.module('openshiftConsole')
...
-  });
+  })
+  .filter('mobileClientIconClass', function() {
+    return function(mobileClientResource) {
+      var clientIcon = _.get(mobileClientResource, 'metadata.labels.icon');
+      var clientType = _.get(mobileClientResource, 'spec.clientType');
+      switch (clientType) {
+        case 'android':
+        case 'iOS':
+          clientIcon = 'fa ' + clientIcon;
```
spadgett: I don't think we should make assumptions about the kind of icon. It would be better if we can have the full icon class like `fa fa-foo` if possible in an annotation.

app/scripts/filters/util.js
```diff
@@ -514,4 +514,22 @@ angular.module('openshiftConsole')
     return function(feature) {
       return _.get(Constants, ['ENABLE_TECH_PREVIEW_FEATURE', feature], false);
     };
-  });
+  })
+  .filter('mobileClientIconClass', function() {
+    return function(mobileClientResource) {
+      var clientIcon = _.get(mobileClientResource, 'metadata.labels.icon');
```
spadgett: This really feels like it should be an annotation, not a label. Other resources use annotations for icons.

app/views/overview.html
```diff
@@ -359,7 +359,16 @@ <h2>Other Resources</h2>
                 </overview-list-row>
               </div>
             </div>
-
+            <!-- Show Mobile Apps -->
+            <div ng-if="(overview.filteredMobileClients | hashSize) && AEROGEAR_MOBILE_ENABLED">
```
spadgett: We're switching to the `| size` filter from `hashSize`

app/views/overview.html
```diff
@@ -359,7 +359,16 @@ <h2>Other Resources</h2>
...
+              <h2>Mobile Clients</h2>
+              <div class="list-pf">
+                <mobile-client-row
+                  ng-repeat="mobileapp in overview.filteredMobileClients"
```
spadgett:
```
track by (mobileapp | uid)
```

app/index.html
```diff
@@ -230,6 +230,7 @@ <h1>JavaScript Required</h1>
         <script src="scripts/services/listRowUtils.js"></script>
         <script src="scripts/services/ownerReferences.js"></script>
         <script src="scripts/services/serviceInstances.js"></script>
+        <script src="scripts/services/mobileClients.js"></script>
```
spadgett: I think you forgot to add these files.

maleck13: Sigh sorry. Yes I did, will update

app/scripts/filters/util.js
maleck13: okay I will make the change to the CRD

app/scripts/filters/util.js
maleck13: will add as annotation to the CRD

sedroche:
@maleck13 Associated icons PR https://github.com/openshift/openshift-logos-icon/pull/11

maleck13:
switch to annotations
https://github.com/aerogear/mobile-cli/pull/84

app/scripts/filters/util.js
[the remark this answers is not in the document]
maleck13: as per sams comments above this is changing to be an icon annotation with the full icon class so this code will be removed

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,77 @@
...
+      }
+      var context = {
+        namespace: _.get(apiObject, 'metadata.namespace')
+      };
+      NotificationsService.hideNotification("remove-mobileclient-error");
+      DataService.delete(mobileclientVersion, apiObject.metadata.name, context, {
+        gracePeriodSeconds: 0
+      }).then(
+        // success
+        _.noop,
```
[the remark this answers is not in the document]
maleck13: the resource is removed immediately as it has no need for a grace period, so I think it disappearing is enough rather than showing an alert / notification too

app/scripts/directives/overview/mobileClientRow.js
```diff
@@ -0,0 +1,72 @@
...
+    _.extend(row, ListRowUtils.ui);
+
+    row.bundleDisplay = "bundle id: " + row.apiObject.spec.appIdentifier;
+    switch (row.apiObject.spec.clientType) {
+      case 'android':
+        row.installType = 'maven';
+        row.bundleDisplay = "package name: " + row.apiObject.spec.appIdentifier;
+        break;
+      case 'iOS':
+        row.installType = 'cocoapods';
```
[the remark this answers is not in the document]
maleck13: that is the default value set above on line 33

app/scripts/directives/overview/mobileClientRow.js
```diff
@@ -0,0 +1,72 @@
...
+        break;
+      case 'cordova':
+        row.installType = 'npm';
```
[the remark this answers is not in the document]
maleck13: it uses the default value set on line 33

app/scripts/directives/overview/mobileClientRow.js
```diff
@@ -0,0 +1,72 @@
...
+      '$scope',
+      '$filter',
+      '$routeParams',
+      'ProjectsService',
+      'DataService',
+      'APIService',
+      'MobileClientsService',
+      'ListRowUtils',
+      'Navigate',
+      'AuthorizationService',
```
spadgett: We've found if we keep these lists alphabetized, they're easier to maintain.

app/scripts/directives/overview/mobileClientRow.js
```diff
@@ -0,0 +1,72 @@
...
+
+      // We can delete mobileclients
+      return AuthorizationService.canI(row.mobileclientVersion, 'delete');
+    };
+
+    $scope.browseCatalog = function () {
+      Navigate.toProjectCatalog($scope.projectName);
+    };
+    $scope.projectName = $routeParams.project;
+    $scope.renderOptions = $scope.renderOptions || {};
```
spadgett: renderOptions is not needed. You should be able to delete this.

app/scripts/directives/overview/mobileClientRow.js
```diff
@@ -0,0 +1,72 @@
...
+      // no actions on those marked for deletion
+      if (_.get(row.apiObject, 'metadata.deletionTimestamp')) {
+        return false;
+      }
+
+      // We can delete mobileclients
+      return AuthorizationService.canI(row.mobileclientVersion, 'delete');
+    };
+
+    $scope.browseCatalog = function () {
```
spadgett: Add to `row` instead of `$scope` if possible, here and below.

app/scripts/directives/overview/mobileClientRow.js
```diff
@@ -0,0 +1,72 @@
...
+    templateUrl: 'views/overview/_mobile-client-row.html'
+  });
+
+  function MobileAppRow($scope, $filter, $routeParams, ProjectsService, DataService, APIService, MobileClientsService, ListRowUtils, Navigate, AuthorizationService) {
+    var row = this;
+    row.installType = '';
+
+    _.extend(row, ListRowUtils.ui);
+
+    row.bundleDisplay = "bundle id: " + row.apiObject.spec.appIdentifier;
```
spadgett: Bundle display and install type might make good files.

If you keep them in the controller, you'd probably want to put them in `$onChanges`

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,77 @@
...
+      if (!$rootScope.AEROGEAR_MOBILE_ENABLED) {
+        Logger("mobile is not enabled");
+        return;
+      }
+      var context = {
+        namespace: _.get(apiObject, 'metadata.namespace')
+      };
+      NotificationsService.hideNotification("remove-mobileclient-error");
+      DataService.delete(mobileclientVersion, apiObject.metadata.name, context, {
+        gracePeriodSeconds: 0
```
spadgett: Why is gracePeriodSeconds needed?

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,77 @@
...
+            type: "error",
+            message: "An error occurred while deleting the mobile client" + _.get(apiObject, 'metadata.name') + ".",
+            details: $filter('getErrorDetails')(err)
+          });
+          Logger("An error occurred while deleting mobileclient " + _.get(apiObject, 'metadata.name') + ".", err);
+        });
+    };
+
+
+    var deleteClient = function (apiObject) {
```
spadgett: Any reason you're not using our standard delete dialog?

app/scripts/services/mobileClients.js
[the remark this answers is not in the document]
spadgett: We usually notify. There's some debate about whether that's the right thing, but I'd notify to be consistent.

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,77 @@
...
+        _.noop,
+        // failure
+        function (err) {
+          NotificationsService.addNotification({
+            id: "remove-mobileclient-error",
+            type: "error",
+            message: "An error occurred while deleting the mobile client" + _.get(apiObject, 'metadata.name') + ".",
+            details: $filter('getErrorDetails')(err)
+          });
+          Logger("An error occurred while deleting mobileclient " + _.get(apiObject, 'metadata.name') + ".", err);
```
spadgett: Logger.warn ?

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,77 @@
...
+    DataService,
+    Logger,
+    NotificationsService) {
+
+    var mobileclientVersion = {
+      group: "mobile.k8s.io",
+      version: "v1alpha1",
+      resource: "mobileclients"
+    };
+    var remove = function (apiObject) {
```
spadgett: I would put "delete" in the function name to make it clear the resource is really being deleted from the API server. `remove` doesn't convey that to me.

app/views/overview/_mobile-client-row.html
```diff
@@ -0,0 +1,49 @@
+<div class="list-pf-item" ng-class="{ active: row.expanded }">
+    <div class="list-pf-container" ng-click="row.toggleExpand($event)">
```
spadgett: nit: indentation

app/views/overview/_mobile-client-row.html
```diff
@@ -0,0 +1,49 @@
...
+      <div class="list-pf-chevron">
+        <div ng-include src=" 'views/overview/_list-row-chevron.html' " class="list-pf-content"></div>
+      </div>
+      <div class="list-pf-content">
+        <div class="list-pf-name">
+          <h3>
+            <span class="fa-2x {{row.apiObject.metadata.annotations.icon}}"></span>
+            <a ng-href="{{row.apiObject | navigateResourceURL}}" ng-bind-html="row.apiObject.spec.name"></a>
```
spadgett: We should only use `ng-bind-html` if we are using `highlightKeywords`

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,77 @@
...
+      if (!$rootScope.AEROGEAR_MOBILE_ENABLED) {
+        Logger("mobile is not enabled");
```
vchepeli: what logging level its used here?

app/scripts/services/mobileClients.js
vchepeli: what logging level its used here?

app/scripts/services/mobileClients.js
maleck13: changed to Logger.info

app/scripts/services/mobileClients.js
maleck13: changed to Logger.warn

app/scripts/services/mobileClients.js
maleck13: happy to remove, was basing on an example I saw in the serviceinstance service

app/scripts/services/mobileClients.js
maleck13: As above I was looking at how it was done for service instances service and tried to follow that. Is there an example you can point me to that is more standard

app/scripts/services/mobileClients.js
spadgett: Yeah, that's a special case. I wouldn't include that unless it's really needed.

app/scripts/services/mobileClients.js
spadgett: You should be able to use the `delete-link` directive and remove a lot of this code. Here's an example:

https://github.com/openshift/origin-web-console/blob/master/app/views/browse/pod.html#L25-L31

maleck13: Thanks will update to use that

maleck13: based on that I think I can remove the mobile service for now too

maleck13:
@spadgett I believe I have addressed all comments now and have rebased and squashed

maleck13:
screen shot 2018-02-26 at 19 42 47
screen shot 2018-02-26 at 19 44 23
screen shot 2018-02-26 at 19 44 45

maleck13:
@spadgett I see the travis-ci build is failing. When i run the ```./hack/verify-dist.sh``` script locally it passes fine, what can I do to debug it?

spadgett:
cc @cshinn

spadgett:
@maleck13 Try running `bower install` then `grunt build`

maleck13:
@spadgett ok so travis looks happy now, but openshift jenkins is unhappy. Looking through other PRs this seems a common trend. Is there anything further I should do here?

cshinn:
@maleck13 What does the package name for a mobile client represent? is it the kind of information that needs to be displayed at the highest level? Additionally, I wonder if we need both the platform icon as well as the label. Perhaps something more like this to bring mobile clients in-line with our other object types

screen shot 2018-02-27 at 9 14 38 am

maleck13:
@cshinn It is effectively a unique id the mobile client, it used in the Google play store (I believe) for example. While it would be significant to the mobile developer, I think it is something we could show on the mobile client details screen. So happy to update as per your recommendation

cshinn:
@maleck13 Given that it's a unique ID for the client, I think it probably does make sense to show it here, especially if that ID is how the developer might recognize their application. Much the same way that IDs for services are shown as well. @spadgett @ncameronbritt thoughts on this?

spadgett:
flake #2105

/retest

spadgett:
Package name seems reasonable to show in the collapsed row. I like the mockup from @cshinn for consistency with other overview rows.

maleck13:
@spadgett @cshinn thanks for feedback. I will update to reflect the mockup

cshinn:
@maleck13 do you happen to know what kind of format the package names usually take? Are they similar to the display name of the application, or are they more like a URL or something like that?

maleck13:
they are generally in the following format ```com.example.myapp```

maleck13:
Updated screen shots after requested changes.
screen shot 2018-02-28 at 12 02 44
screen shot 2018-02-28 at 12 03 13

maleck13:
@spadgett @cshinn happy with the changes?

cshinn:
@maleck13 yeah, looks good to me! I assume there are plans to add in other functionality within the expanded section at some point in the future?

maleck13:
@cshinn yes there will be a number of other prs to flesh out all of the UI

spadgett:
Ready to merge?

maleck13:
pr changed due to doing a new grunt build after rebase. Rechecked work and PR all looks good to me for merge

PR #2840, opened 2018-02-21
Fix potential apiVersion mismatch in directives/fromFile

benjaminapetersen:
Unrelated to `preferredVersion`, but this seems to be the correct way to make this request.


benjaminapetersen:
/unhold 
i think?

PR #2841, opened 2018-02-21
Update miniLog to use preferredVersion when possible

benjaminapetersen:
This isn't actually a mismatch, it just updates the `miniLog` to use a preferred version if possible.

### DIFF
--- a/app/scripts/services/deployments.js
+++ b/app/scripts/services/deployments.js
@@ -268,6 +268,13 @@ angular.module("openshiftConsole")
       return annotation(object, 'deployment.kubernetes.io/revision');
     };
 
+    DeploymentsService.prototype.getName = function(object) {
+      return object.kind === "ReplicationController" ?
+        $filter('annotation')(object, 'deploymentConfig') :
+        object.metadata.name;
+    };
+
+
     DeploymentsService.prototype.isActiveReplicaSet = function(replicaSet, deployment) {
       var replicaSetRevision = this.getRevision(replicaSet);
       var deploymentRevision = this.getRevision(deployment);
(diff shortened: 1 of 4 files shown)


benjaminapetersen:
/retest 

```

       Output:
         Testing connectivity to master API
       Error Output:
         curl: (28) timed out before SSL handshake
         curl: (35) You are attempting to import a cert with the same issuer/serial as an existing cert, but that is not the same cert.
         curl: (35) You are attempting to import a cert with the same issuer/serial as an existing cert, but that is not the same cert.
         curl: (35) You are attempting to import a cert with the same issuer/serial as an existing cert, but that is not the same cert.
```
flake

benjaminapetersen:
Updated, bot `logViewer` and `miniLog` were basically doing the same bits of logic, neither landing on a `rgv`, I think moving this to `DeploymentsService` is consistent with the `getScaleResource()` method already there.

app/scripts/services/deployments.js
```diff
@@ -268,6 +268,13 @@ angular.module("openshiftConsole")
       return annotation(object, 'deployment.kubernetes.io/revision');
     };
 
+    DeploymentsService.prototype.getName = function(object) {
```
spadgett: I'd rather have one function that returns both the log resource and the resource name (perhaps as an object). The way this is written, this assumes too much about what `getLogResource` will return.

benjaminapetersen: They seem decoupled to me, but  I'm pretty sure its ignorance on my part.  Let me take a stab & see if I follow:

- `getName()` alone seems standalone, but perhaps not very useful.
- `getLogResource()` seems to require the output of `getName()` to actually be useful.

So I could update the return value of `getLogResource()` to `{ resource: rgv, name: getName(obj)  }`.

spadgett: They're coupled in that you can't change one without changing the other. If we ever return something other than `DeploymentConfig` in getLogResource for replication controllers owned by a deployment config, then you have to change `getName` too or it will return the wrong value. We more of less have that same logic in two places.

Better to keep them as one method to avoid that problem.

> I could update the return value of getLogResource() to { resource: rgv, name: getName(obj) }.

Yeah, that's what I was thinking.

Issue #2836, opened 2018-02-21
Different spacing between action dividers in build config editor


spadgett:
I like the bottom better, although I believe we use the top margins in other places.

![openshift web console 2018-02-21 13-05-04](https://user-images.githubusercontent.com/1167259/36496969-e9ffa220-1707-11e8-8b1b-c8408b3f5b65.png)

/area styles
/kind bug
/priority p3
/assign @sg00dwin @rhamilto

PR #2844, opened 2018-02-21
Remove padding from form-group-action buttons so they will have spacing consistency with links that are separated with .action-divider

sg00dwin:
Fixes https://github.com/openshift/origin-web-console/issues/2836

----

### DIFF
--- a/app/components/osc-webhook-triggers/osc-webhook-triggers.html
+++ b/app/components/osc-webhook-triggers/osc-webhook-triggers.html
@@ -79,10 +79,10 @@
     </div>
   </div>
 
-  <div class="add-webhook-action-btns">
+  <div class="form-group-actions">
     <button class="btn btn-link pad-left-none" type="button" ng-click="$ctrl.addEmptyWebhookTrigger()">Add Webhook</button>
     <span ng-if="$ctrl.secretsVersion | canI : 'create'">
-      <span class="action-divider" aria-hidden="true"> | </span>
+      <span class="action-divider" aria-hidden="true">|</span>
       <button class="btn btn-link" href="" type="button" ng-click="$ctrl.openCreateWebhookSecretModal()">Create New Webhook Secret</button>
     </span>
   </div>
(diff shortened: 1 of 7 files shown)


app/components/osc-webhook-triggers/osc-webhook-triggers.html
```diff
@@ -80,10 +80,10 @@
   </div>
 
   <div class="add-webhook-action-btns">
-    <button class="btn btn-link pad-left-none" type="button" ng-click="$ctrl.addEmptyWebhookTrigger()">Add Webhook</button>
+    <a href="" role="button" ng-click="$ctrl.addEmptyWebhookTrigger()">Add Webhook</a>
```
spadgett: We want to use `<button>` instead of `<a>` because it's better for accessibility. It's also prevents "open in a new window" when we know it won't work.

sg00dwin: ok, should all the `<a>` be converted to `<button>`s

sg00dwin: or should I make the change so that .action-divider can work with either?

spadgett: I wouldn’t change them to buttons yet. A little risky so close to the end of the release. Eventually we want to though

sg00dwin:
@spadgett I updated this pr to make the `<a>` and `<button>` actions with `.action-divider` to have consistent spacing.

app/styles/_forms.less
```diff
@@ -7,6 +7,10 @@
   vertical-align: -2px;
 }
 
+// Generic class to wrap a single form element row that contains multiple actions separated by a divider
+.form-group-actions .btn {
+  padding: 0;
```
spadgett: We want to target `btn-link` specifically

sg00dwin: yep. good catch

sg00dwin:
@spadgett pr updated

PR #2845, opened 2018-02-21
update edit route controller to use objectToResourceGroupVersion

benjaminapetersen:
...when updating a route to prevent a mismatch.

### DIFF
--- a/app/scripts/controllers/edit/route.js
+++ b/app/scripts/controllers/edit/route.js
@@ -181,7 +181,7 @@ angular.module('openshiftConsole')
             hideErrorNotifications();
             $scope.disableInputs = true;
             var updated = updateRouteFields();
-            DataService.update(routesVersion, $scope.routeName, updated, context)
+            DataService.update(APIService.objectToResourceGroupVersion(updated), $scope.routeName, updated, context)
               .then(function() { // Success
                 NotificationsService.addNotification({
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/edit/route.js
```diff
@@ -181,7 +181,7 @@ angular.module('openshiftConsole')
             hideErrorNotifications();
             $scope.disableInputs = true;
             var updated = updateRouteFields();
-            DataService.update(routesVersion, $scope.routeName, updated, context)
+            DataService.update(APIService.objectToResourceGroupVersion(updated), $scope.routeName, updated, context)
```
spadgett: I'm less concerned about this when we request and then update the same object in one controller. I think in this case it might be good to leave well-enough alone.

benjaminapetersen: I think in these cases I'd like to get the pattern right & be consistent throughout the app, both for my own sake in the future as well as others who may copy/paste as we add features.  Its a tad of a pain to keep the mental model, unless we set the precedent to just always get the `resource`, `group`, `version` directly from the object being sent to the API.  I'm fairly confident this pattern will always be correct.

Having said that, I realize its more of a human problem than a code problem, so if you're not worried about this one I'm cool with closing it out!

PR #2860, opened 2018-02-27
Removing layout.attrs to reduce dependencies

### LINKED ISSUE #2614: Membership Add Role cannot see dropdown contents

rhamilto:
Also fixes #2614 by eliminating the three column layout for desktop resolutions
screen shot 2018-02-28 at 10 24 44 am

And updates volume-claim-template heading to the new style
Before:
screen shot 2018-02-28 at 9 45 29 am
After:
screen shot 2018-02-28 at 9 45 35 am

### DIFF
--- a/app/styles/_components.less
+++ b/app/styles/_components.less
@@ -35,12 +35,9 @@
     margin-top: 15px;
   }
   .pod-template {
-    // so that .word-break() on children works
-    [flex] {
-      min-width: 0;
-    }
     .icon-row {
       border-left: 3px solid @gray-lighter;
+      display: flex;
       padding: 2px 0 0 2px;
     }
     .fa,
@@ -50,6 +47,10 @@
     }
     .pod-template-build {
       .word-break();
+    }
+    .pod-template-detail {
+      flex: 1 1 0%;
+      min-width: 0; // so that .word-break() on children works
     }
     .pod-template-key {
       font-weight: 600;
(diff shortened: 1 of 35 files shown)


spadgett:
@TheRealJon FYI, this removes the layout.attrs dependency (which was bower only)

app/styles/_components.less
```diff
@@ -35,12 +35,9 @@
     margin-top: 15px;
   }
   .pod-template {
-    // so that .word-break() on children works
-    [flex] {
-      min-width: 0;
```
sg00dwin: wouldn't we need to keep `min-width: 0`?

rhamilto: Yep.  It moved to line 53.

app/views/directives/_build-pipeline-expanded.html
```diff
@@ -30,7 +30,7 @@
     </div>
     <div class="pipeline">
       <div class="pipeline-stage" ng-repeat="stage in jenkinsStatus.stages track by stage.id">
-        <div column class="pipeline-stage-column">
+        <div class="pipeline-stage-column">
```
sg00dwin: Shouldn't there be a `.pipeline-stage-column` rule with `flex-direction:column; display:flex` since it was removed from the markup?

rhamilto: To be completely parallel with what was there, yes.  But a div displays block by default, and that renders the same way as flex/column, so I opted to remove it for less code.

PR #2863, opened 2018-02-28
[WIP] Bower to Yarn migration

TheRealJon:
# TODO BEFORE MERGE

- [ ] Update origin-web-common dependency once https://github.com/openshift/origin-web-common/pull/321 is merged and new release is cut
- [ ] Update origin-web-catalog dependency once https://github.com/openshift/origin-web-catalog/pull/670 is merged and new release is cut
- [x] Update registry-image-widgets dependency once https://github.com/cockpit-project/registry-image-widgets/pull/16 is merged and new release is cut

# Description
Migrate dependency management from bower to yarn.

https://trello.com/c/guhAEGVH

# Changes
- Remove bower.json and .bowerrc
- Move bower dependencies into package.json
- Install as many dependencies from NPM as possible
- Install other dependencies from git instead of bower so they are not scoped in node_modules/@bower_components
- Update `hack/clean-deps.sh` and `hack/install-deps.sh` to use yarn
- Update `bower_components` path references to `node_modules` in several source and config files.
- Prune problematic, unused, or unnecessary dependencies:
  - datatables (all related packages)
  - hawtio-extension-service
  - layout.attrs
  - phantomjs
  - sifter
  - autoprefixer
  - json3
  - es5-shim
  - es5-dom-shim
  - bootstrap-sass
  - font-awesome-sass
  - microplugin
  - objectpath
  - selectize
  - angular-drag-and-drop-lists
  - angular-schema-form
  - angular-schema-form-bootstrap
  - bootstrap-datepicker
  - bootstrap-slider
  - angular-dragdrop
  - angular-gettext (pending PR for registry-image-widgets)
  - angular-scenario
  - angular-touch
  - ansi_up
  - bootstrap-hover-dropdown
  - bootstrap-touchspin
  - exports-loader
  - google-code-prettify
  - hopscotch
  - imports-loader
  - jquery-match-height
  - jquery-ui
  - js-yaml
  - patternfly-bootstrap-combobox
  - patternfly-bootstrap-treeview
  - tv4
  - webpack
  - clean-css
  - grunt-contrib-imagemin
  - grunt-postcss
  - grunt-wiredep
  - html-minifier
  - imagemin
  - karma-nightmare
  - less
- Update karma, travis, and grunt configuration to use xvfb for headless browser testing.


spadgett:
@benjaminapetersen I'm confused about the relationship between openshift/angular-extension-registry and hawtio-extensions-service. Do we still need the latter?

benjaminapetersen:
@spadgett @TheRealJon `openshift/angular-extension-registry` and `hawtio-extensions-service` have no relationship. `hawtio-extensions-service` came first, but was extremely basic, if its still in the repo anywhere, it is tech debt.

benjaminapetersen:
a quick `grep`/ripgrep of the `app` dir indicates we do not use `HawtioExtension` at this point.

spadgett:
@benjaminapetersen So can we remove it w/o breaking extensions or changing the API for registering extensions?

hack/install-deps.sh
```diff
@@ -29,25 +29,37 @@ function cmd() {
...
 # correct binaries.
 repo_root="$( dirname "${BASH_SOURCE}" )/.."
-export PATH="${PATH}:${repo_root}/node_modules/bower/bin:${repo_root}/node_modules/grunt-cli/bin"
+export PATH="${PATH}:${repo_root}/node_modules/grunt-cli/bin"
 
-# Install bower if needed
-if ! which bower > /dev/null 2>&1 ; then
-  cmd "npm install bower"
+# Install yarn if needed
+if ! which yarn > /dev/null 2>&1 ; then
```
benjaminapetersen: Quick Q, does this possibly change a user's existing environment, for example if `yarn` was installed via `brew install yarn`?

hack/install-deps.sh
```diff
@@ -29,25 +29,37 @@ function cmd() {
...
+# compatibility test, even when it should not.
+#
+# ignore-scripts flag causes post-install scripts to be skipped. The
+# hawtio-exetnsion-service package has a post-install script which fails
+# because it tries to use bower and gulp which are not dependencies of this
+# project.
+cmd "yarn --ignore-engines --ignore-scripts"
+
+# Work around. --ignore-scripts flag causes phantomjs-prebuilt
+# post install script not be executed which leads to test failure.
```
benjaminapetersen: I think we are not using `phantomjs` anymore.  This might be able to be dropped (or spawned as a follow-up issue to avoid getting into weeds right now)

package.json
```diff
@@ -56,14 +139,14 @@
...
     "protractor-screenshot-reporter": "^0.0.5",
     "serve-static": "1.10.2",
     "time-grunt": "0.3.2"
   },
   "engines": {
-    "node": ">=0.10.0"
+    "node": ">=4.0.0",
+    "yarn": ">= 1.0.0"
   },
   "scripts": {
```
benjaminapetersen: As far as you can tell from this change, any need to update the `scripts` block with new commands? Or is everything basically the same from the outside.

hack/install-deps.sh
TheRealJon: If we are not using it, can we remove the dependencies?

hack/install-deps.sh
TheRealJon: I may be misunderstanding what you mean, but this block should have no effect as long as the yarn bin is somewhere in the users PATH already (regardless of how it was installed). Using the downloaded install script was the most platform agnostic installation solution I could find.

benjaminapetersen:
@spadgett correct, we should test it & make sure the extensions we are running still show up (unexpected side effects?), but they should not be related in any way.  

@TheRealJon `app/scripts/extensions` has a few extensions files.  It shouldn't be too bad to run the app & check & make sure they still work.  IE, does the "set homepage" link still appear in the user drop down menu, does "documentation" still appear in the help drop down, etc.

I would be very surprised if there was a conflict.

package.json
TheRealJon: Everything should be the same from the outside. I adjusted the scripts and grunt tasks so that it should be seamless.

benjaminapetersen:
"This repository has been archived by the owner. It is now read-only." - [Hawtio Extension Sevice](https://github.com/hawtio/hawtio-extension-service)

hack/install-deps.sh
benjaminapetersen: Ah, true, no thats my mistake in reading it.  Should be good!

hack/install-deps.sh
```diff
@@ -29,25 +29,31 @@ function cmd() {
...
 repo_root="$( dirname "${BASH_SOURCE}" )/.."
-export PATH="${PATH}:${repo_root}/node_modules/bower/bin:${repo_root}/node_modules/grunt-cli/bin"
+export PATH="${PATH}:${repo_root}/node_modules/grunt-cli/bin"
 
-# Install bower if needed
-if ! which bower > /dev/null 2>&1 ; then
-  cmd "npm install bower"
+# Install yarn if needed
+if ! which yarn > /dev/null 2>&1 ; then
+  cmd "curl -o- -L https://yarnpkg.com/install.sh | bash"
```
spadgett: This seems to fail in Jenkins. Maybe `npm install --global yarn` ?

spadgett: That way it will also install just for this repo rather than installing it globally.

TheRealJon: I tried that but there is a write permissions error when trying to install yarn globally. We could try just a local `npm install yarn` and see if that works.

spadgett: Yeah I ninja edited my comment because I didn't mean `--global`. I don't think we'd want `--global` anyway.

TheRealJon:
Okay, I think I have reached a point where I am blocked. I have everything running and passing in Travis using ChromeHeadless browser and Firefox with the "--headless" flag. I also fixed the problem with yarn not working in Jenkins, but now Firefox will not launch. My guess is that there is no Firefox browser in the Jenkins environment, or that it is an older version that does not support headless mode. Any suggestions or further insight?

spadgett:
@TheRealJon It looks like we're still starting the virtual framebuffer. Can you try Firefox without passing the headless parameter?

We have Firefox 52.6.0 in Jenkins

https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_web_console/1992/s3/download/generated/installed_packages.log

spadgett:
It looks like we need Firefox 56 for headless. I'd just leave that flag off.

benjaminapetersen:
@TheRealJon our `jenkins` setup is [here](https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_branch_origin_web_console.yml#L19), tho it doesn't say anything about `Xvfb` in that file.  Guessing its installed & already fired up by the time it gets to us.

TheRealJon:
@spadgett I just noticed that the service names in the `services-view` are not centered for some reason. I have been looking around for what change might have caused it but can't seem to find any style changes that would have done this.

yarn.lock
```diff
@@ -0,0 +1,9065 @@
...
+
+camelcase@^2.0.0:
+  version "2.1.1"
+  resolved "https://registry.yarnpkg.com/camelcase/-/camelcase-2.1.1.tgz#7c1d16d679a1bbe59ca02cacecfb011e201f5a1f"
+
+camelcase@^4.1.0:
+  version "4.1.0"
+  resolved "https://registry.yarnpkg.com/camelcase/-/camelcase-4.1.0.tgz#d545635be1e33c542649c69173e5de6acfae34dd"
+
+caniuse-lite@^1.0.30000792, caniuse-lite@^1.0.30000805:
```
rhamilto: We only want caniuse-lite@^1.0.30000792 here.

rhamilto:
Assuming we want to maintain `npm install` functionality:  

`npm install` fails because the https:// endpoints aren't valid for npm.

`"angular-utf8-base64": "https://github.com/stranger82/angular-utf8-base64.git#0.0.5",`
`"kubernetes-object-describer": "https://github.com/kubernetes-ui/object-describer.git#1.0.4",`
should be 
`"kubernetes-object-describer": "https://github.com/kubernetes-ui/object-describer.git#1.0.4",`
`"kubernetes-object-describer": "kubernetes-ui/object-describer.git#v1.0.4",`

additionally, https://github.com/termi/ES5-DOM-SHIM lacks a package.json, so it won't install via NPM at all.  i question if we still need this dependency.

rhamilto:
@TheRealJon, @spadgett and I briefly discussed it this morning and determined we don't care about `npm install` functionality since we want yarn to be responsible for installing and locking, and we should remove ES5-DOM-SHIM unless there is a need to keep it.

rhamilto:
This PR should probably also include an update to [README.md](https://github.com/openshift/origin-web-console/blob/master/README.md).  

Install of bower should be yarn in:
https://github.com/openshift/origin-web-console#getting-started

`bower link` should be `yarn link` in:
https://github.com/openshift/origin-web-console#contributing-to-the-primary-repositories

Anything else?

spadgett:
We should be able to remove es5-shim and es5-dom-shim. Those are top-level dependencies and none of our other dependencies need them. Looking at the list of shims, I don't think we need any for IE11.

TheRealJon:
Does this include json3?

Gruntfile.js
```diff
@@ -248,57 +244,13 @@ module.exports = function (grunt) {
         }]
       }
     },
-
-    // Automatically inject Bower components into the app
-    wiredep: {
```
spadgett: I see you're removing wiredep. Do we need a replacement? Can you try `grunt serve:dist`?

TheRealJon: Well, something like Webpack would be the replacement for Wiredep. For now, we can just manually keep our vendor script imports in the Usemin block in app/index.html and they will still get concatenated into a vendor.js script in dist. The only thing we lose with Wiredep is the automatic population of the vendor script imports in app/index.html.

TheRealJon:
Right now, the vendor scripts imports in app/index.html are ordered incorrectly. This should be fixed by  #2914 after a rebase.

spadgett:
@TheRealJon looks like the dist doesn't match

TheRealJon:
Obsolete as this repo is now maintenance only.

PR #2867, opened 2018-03-02
Bug 1550797 - Should remove 'loading..' when browser non-existent image

jhadvig:
Also fixed case when imageStream tag is missing/non-existing.

@spadgett PTAL


spadgett:
flake #2105

/retest

PR #2869, opened 2018-03-02
Bug 1550138 - Remove 5 minute warning for pods

spadgett:
Remove the warning we show when a pod takes more than 5 minutes to
start. Some pods will take longer, particularly if they need to pull an
image. Since we now do a better job of surfacing container statuses when
a container is crash looping, the 5 minute warning is no longer needed
and sometimes misleading.

See discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1550138

/assign @benjaminapetersen


spadgett:
Rebased

PR #2870, opened 2018-03-05
Bug 1549450 - Better to use the same name for Clear filters, Clear Filter, Clear All Filters

jhadvig:
Reword all the `Clear Filter` into `Clear All Filters` since this wording is from patternfly.
As part of the work, it would be good also to align on the wording for case when the filters all hidding all the options, since we are using different working for the same case, eg:
1. The filter is hiding all persistent volume claims
2. The current filter is hiding all projects.
3. All pipelines are filtered.

I suggest to use the first one and modify it into "The filters are hiding all persistent volume claims", since the first one is used on most places and with this change we are moving from singular `Clear Filter` to plural  `Clear All Filters`.

@spadgett @beanh66 @cshinn PTAL

### DIFF
--- a/app/styles/_core.less
+++ b/app/styles/_core.less
@@ -126,6 +126,13 @@ body.ios {
   display: none;
 }
 
+.btn-link.inline-btn-link {
+  font-size: inherit;
+  font-weight: inherit;
+  vertical-align: initial;
+  padding: 0;
+}
+
 .empty-state-message {
   margin: 40px auto 60px;
   max-width: 600px;
--- a/app/views/browse/config-maps.html
+++ b/app/views/browse/config-maps.html
@@ -38,7 +38,7 @@
                 </p>
               </div>
               <div ng-if="filterWithZeroResults">
-                <h2>The filter is hiding all config maps. <a href="" ng-click="clearFilter()" role="button" class="nowrap">Clear Filter</a></h2>
+                <h2>The filter is hiding all config maps. <button type="button" class="btn btn-link inline-btn-link" ng-click="clearFilter()">Clear All Filters</button></h2>
               </div>
             </div>
           </div>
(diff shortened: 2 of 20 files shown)


jhadvig:
Or use the wording that we have in catalog: `The active filters are hiding all catalog items.`

app/views/browse/config-maps.html
```diff
@@ -38,7 +38,7 @@ <h2>No config maps.</h2>
                 </p>
               </div>
               <div ng-if="filterWithZeroResults">
-                <h2>The filter is hiding all config maps. <a href="" ng-click="clearFilter()" role="button" class="nowrap">Clear Filter</a></h2>
+                <h2>The filter is hiding all config maps. <a href="" ng-click="clearFilter()" role="button" class="nowrap">Clear All Filters</a></h2>
```
spadgett: If we're changing all of these anyway, I suggest we go ahead and make them buttons for better accessibility.

```html
<button type="button" class="btn btn-link nowrap" ng-click="clearFilter()">Clear All Filters</button>
```

(We should double check if `nowrap` is needed with `btn-link`.)

jhadvig:
@spadgett comment addressed. Checked the `nowrap` and its not necessary.
PTAL

app/styles/_core.less
```diff
@@ -164,6 +164,12 @@ body.ios {
   }
   .btn {
     margin: 10px 0 20px;
+    &.clear-filter {
+      font-size: inherit;
+      font-weight: inherit;
+      vertical-align: initial;
+      padding: 0px;
```
spadgett: No need for `px` for zero values.

```
padding: 0;
```

spadgett: I'd rather make this style generic, something like `inline-btn-link` and target `btn-link` specifically.

```css
.btn-link.inline-btn-link {
   ...
}
```

Then we can reuse it in other places.

jhadvig:
@spadgett updated :)

PR #2877, opened 2018-03-06
Bug 551667 - Overview support for daemon sets

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1551667

Depends on https://github.com/openshift/origin-web-common/pull/314

https://trello.com/c/jLJInavM

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -87,6 +87,7 @@ function OverviewController($scope,
   var buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
   var buildsVersion = APIService.getPreferredVersion('builds');
   var clusterResourceQuotasVersion = APIService.getPreferredVersion('appliedclusterresourcequotas');
+  var daemonSetsVersion = APIService.getPreferredVersion('daemonsets');
   var deploymentConfigsVersion = APIService.getPreferredVersion('deploymentconfigs');
   var deploymentsVersion = APIService.getPreferredVersion('deployments');
   var horizontalPodAutoscalersVersion = APIService.getPreferredVersion('horizontalpodautoscalers');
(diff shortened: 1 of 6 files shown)


spadgett:
@jwforres I have the conditional watch working for daemon sets.

/assign @benjaminapetersen

app/scripts/controllers/overview.js
```diff
@@ -203,6 +203,7 @@ function OverviewController($scope,
            _.size(overview.deployments) +
            _.size(overview.vanillaReplicaSets) +
            _.size(overview.statefulSets) +
+           _.size(overview.daemonSets) +
```
benjaminapetersen: Wondering, now that we are up to 8 of these that get repeated in various functions, if we could reduce the duplication.  Noticing its basically the same 8 resources in this first bunch, but `toCheck` only has 7.  Any worry these will get out of sync & cause a bug?  I imagine we would catch it quick so its prob low risk.

spadgett: Agree. I made the same comment to @jwforres 

We might need to backport this particular change, so I'm hesitant to do it in this PR.

The different number between this and the `toCheck` is expected since the latter list shouldn't include service instances. So unfortunately it's not always the same list of resources, which complicates things.

spadgett:
Removing WIP. I don't see a reason to wait on this.

app/scripts/controllers/overview.js
benjaminapetersen: That makes sense to me.  Thinking using named arrays/objects would be helpful to make the difference explicit.  

Cool with leaving that for another day.

benjaminapetersen:
Ok great, I'll give it a second pass here, but first time looked pretty good.

smarterclayton:
From a visualization perspective I kind of want the ring to be about nodes, not pods (like highlight nodes that have two pods, nodes that have no pods, etc).  But kind of an overkill

PR #2879, opened 2018-03-07
Create Opaque Secrets from File

dtaylor113:
https://trello.com/c/SANVwQnv

- Created a new Secret Type 'Opaque Secret'.  Not sure if this should be 'Input' or 'Builder' secret?
- 'Opaque Secret' form is same key-value map used in Create Config Map.
image

Not sure what code to use to Create this type of secret.  Creating a Secret is:
```
DataService.create(secretsVersion, null, newSecret, $scope).then(function(secret) { // Success
            // In order to link:
            // - the SA has to be defined
            // - defined SA has to be present in the obtained SA list
            // - user can update SA
            // Else the linking will be skipped
            if ($scope.newSecret.linkSecret && $scope.serviceAccountsNames.contains($scope.newSecret.pickedServiceAccountToLink) && AuthorizationService.canI('serviceaccounts', 'update')) {
              linkSecretToServiceAccount(secret);
```
- Do I need to show the 'Link secret to a service account.' checkbox for 'Opaque Secrets' ?

Creating a Config Map is:
```
DataService.create(createConfigMapVersion, null, $scope.configMap, context)
```
- where `$scope.configMap.data` is the key-value pairs map.

Looking into `DataService.create(secretsVersion...` to see if it supports `data.keyvaluepairs`

### DIFF
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -70,6 +70,15 @@ angular.module("openshiftConsole")
                 label: "Webhook Secret"
               }
             ]
+          },
+          generic: {
+            label: "Generic Secret",
+            authTypes: [
+              {
+                id: "Opaque",
+                label: "Generic Secret"
+              }
+            ]
           }
         };
 
@@ -95,7 +104,11 @@ angular.module("openshiftConsole")
           $scope.newSecret = {
             type: "source",
             authType: "kubernetes.io/basic-auth",
-            data: {},
+            data: {
+              genericKeyValues: {
+                data: {}
+              }
+            },
             linkSecret: false,
             pickedServiceAccountToLink: "",
           };
@@ -172,6 +185,9 @@ angular.module("openshiftConsole")
             case "Opaque":
               if (data.webhookSecretKey) {
                 secret.stringData.WebHookSecretKey = data.webhookSecretKey;
+              }
+              if (data.genericKeyValues.data) {
+                secret.stringData = data.genericKeyValues.data;
               }
               break;
           }
--- a/app/views/directives/create-secret.html
+++ b/app/views/directives/create-secret.html
@@ -85,7 +85,7 @@
       </div>
     </div>
 
-    <div ng-if="newSecret.type != 'webhook'">
+    <div ng-if="newSecret.type === 'source' || newSecret.type === 'image'">
       <div class="form-group">
         <label for="authentification-type">Authentication Type</label>
         <ui-select required input-id="authentification-type" ng-model="newSecret.authType" search-enabled="false">
@@ -349,6 +349,9 @@
           </div>
         </div>
       </div>
+    </div>
+    <div ng-if="newSecret.type === 'generic'">
+      <edit-config-map-or-secret model="newSecret.data.genericKeyValues" type="secret"></edit-config-map-or-secret>
     </div>
   </div>
   <div class="buttons gutter-top-bottom">
--- a/app/index.html
+++ b/app/index.html
@@ -312,7 +312,7 @@
         <script src="scripts/directives/createSecret.js"></script>
         <script src="scripts/directives/date.js"></script>
         <script src="scripts/directives/deleteLink.js"></script>
-        <script src="scripts/directives/editConfigMap.js"></script>
+        <script src="scripts/directives/editConfigMapOrSecret.js"></script>
         <script src="scripts/directives/editEnvironmentFrom.js"></script>
         <script src="scripts/directives/events.js"></script>
         <script src="scripts/directives/eventsSidebar.js"></script>
--- a/app/scripts/directives/editConfigMapOrSecret.js
+++ b/app/scripts/directives/editConfigMapOrSecret.js
@@ -1,16 +1,17 @@
 "use strict";
 
 angular.module("openshiftConsole")
-  .directive("editConfigMap",
+  .directive("editConfigMapOrSecret",
              function(DNS1123_SUBDOMAIN_VALIDATION) {
     return {
       require: '^form',
       restrict: 'E',
       scope: {
-        configMap: "=model",
-        showNameInput: "="
+        map: "=model",
+        showNameInput: "=",
+        type: "@"
       },
-      templateUrl: 'views/directives/edit-config-map.html',
+      templateUrl: 'views/directives/edit-config-map-or-secret.html',
       link: function($scope, element, attrs, formCtl) {
         $scope.form = formCtl;
 
@@ -32,7 +33,7 @@ angular.module("openshiftConsole")
         };
 
         // Load the data once when it is first set.
-        var clearWatch = $scope.$watch('configMap.data', function(data) {
+        var clearWatch = $scope.$watch('map.data', function(data) {
           if (!data) {
             return;
           }
@@ -53,14 +54,14 @@ angular.module("openshiftConsole")
 
           clearWatch();
 
-          // Update `$scope.configMap` as the form data changes.
+          // Update `$scope.map` as the form data changes.
           $scope.$watch('data', function(data) {
             var map = {};
             _.each(data, function(keyValuePair) {
               map[keyValuePair.key] = keyValuePair.value;
             });
 
-            _.set($scope, 'configMap.data', map);
+            _.set($scope, 'map.data', map);
           }, true);
         });
       }
--- a/app/views/create-config-map.html
+++ b/app/views/create-config-map.html
@@ -12,7 +12,7 @@
             </div>
             <form name="createConfigMapForm" class="mar-top-xl">
               <fieldset ng-disabled="disableInputs">
-                <edit-config-map model="configMap" show-name-input="true"></edit-config-map>
+                <edit-config-map-or-secret model="configMap" type="config-map" show-name-input="true"></edit-config-map-or-secret>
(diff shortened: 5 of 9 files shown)


spadgett:
Screenshots look good.

> Do I need to show the 'Link secret to a service account.' checkbox for 'Opaque Secrets' ?

No, shouldn't be needed.

> where $scope.configMap.data is the key-value pairs map.

Use `$scope.newSecret.stringData` as the key-value paris map.

dtaylor113:
> Use $scope.newSecret.stringData as the key-value paris map.

It looks like [createSecret](https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/createSecret.js#L133) is storing other data in `stringData`.  If I'm to store user-defined key/value pairs in `stringData`, I'll need a way to identify them and pull them out to put into the key value blocks (Remove Item, Add Item).  Should I prefix the key used in 'stringData', Ie. 
```
stringData['user-def-my-key1'] = myVal1
stringData['user-def-my-key2'] = myVal2
```
...where 'user-def-' is the prefix?

app/scripts/directives/createSecret.js
```diff
@@ -70,6 +70,15 @@ angular.module("openshiftConsole")
                 label: "Webhook Secret"
               }
             ]
+          },
+          opaque: {
+            label: "Opaque Secret",
```
spadgett: Let's call it `Generic Secret` to match the CLI command (`oc create secret generic ...`)

app/views/directives/create-secret.html
```diff
@@ -85,7 +85,7 @@
       </div>
     </div>
 
-    <div ng-if="newSecret.type != 'webhook'">
+    <div ng-if="newSecret.type != 'webhook' && newSecret.type != 'opaque'">
```
spadgett: I'd rather switch this to check specifically for source and image types.

```html
<div ng-if="newSecret.type === 'source' || newSecret.type === 'image'">
```

app/scripts/directives/editKeyValueMap.js
```diff
@@ -1,16 +1,17 @@
 "use strict";
 
 angular.module("openshiftConsole")
-  .directive("editConfigMap",
+  .directive("editKeyValueMap",
```
spadgett: I worry that this name is too close to `key-value-editor` and will cause confusion.

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/keyValueEditor.js#L6

Maybe make it clear that it is specific to config maps and secrets? `editConfigMapOrSecret` too on the nose?

app/scripts/directives/createSecret.js
```diff
@@ -98,6 +107,9 @@ angular.module("openshiftConsole")
             data: {},
             linkSecret: false,
             pickedServiceAccountToLink: "",
+            keyValuePairs: {
+              data: {}
```
spadgett: I don't see where this data is used when actually creating the secret. Still working on that I'm guessing?

spadgett:
Thinking more on this, we might have to base64 encode things ourselves if the user adds a binary file to a secret (which is possible). In that case, we wouldn't use `stringData`. We'd use `data` instead. So the final secret request would look like

```yaml
[...]
data:
  key1: <base64-encoded value 1>
  key1: <base64-encoded value 2>
```

Let's catch up on this.

app/scripts/directives/createSecret.js
dtaylor113: Bugzilla and Build Inputs documentation seem to use ```oc secrets new secret-npmrc .npmrc=~/.npmrc```

app/scripts/directives/createSecret.js
dtaylor113: Yes, not sure how to include it in the `DataServices.create(secret...`

app/scripts/directives/createSecret.js
spadgett:
```
❯ oc create secret generic -h
Create a secret based on a file, directory, or specified literal value.

A single secret may package one or more key/value pairs.

When creating a secret based on a file, the key will default to the basename of the file, and the value will default to
the file content. If the basename is an invalid key or you wish to chose your own, you may specify an alternate key.

When creating a secret based on a directory, each file whose basename is a valid key in the directory will be packaged
into the secret. Any directory entries except regular files are ignored (e.g. subdirectories, symlinks, devices, pipes,
etc).

Usage:
  oc create secret generic NAME [--type=string] [--from-file=[key=]source] [--from-literal=key1=value1] [--dry-run]
[options]

Examples:
  # Create a new secret named my-secret with keys for each file in folder bar
  oc create secret generic my-secret --from-file=path/to/bar
```

The command you reference is deprecated.

```
❯ oc secrets -h
Manage secrets in your project

Secrets are used to store confidential information that should not be contained inside of an image. They are commonly
used to hold things like keys for authentication to other internal systems like Docker registries.

Aliases:
secrets, secret

Usage:
  oc secrets [options]

Available Commands:
  add           DEPRECATED: secrets link
  link          Link secrets to a ServiceAccount
  new           DEPRECATED: create secret
```

app/scripts/directives/createSecret.js
spadgett: OK, see my comment just below (https://github.com/openshift/origin-web-console/pull/2879#issuecomment-371240087)

app/scripts/directives/editKeyValueMap.js
dtaylor113: How about `key-value-uploader`, since it does add/remove key-value-pairs, but also allows `value` to be uploaded from a file.

spadgett: It's pretty specific to config maps and secrets rather than something general. For instance, we perform validation of keys using specific regex and maxlength checks specific to those resources. I still feel like it would be good to convey that in the name.

dtaylor113:
Hi @spadgett, need a little more direction on how we want to save/edit Generic Secrets.   When editing a **Config Map** with these key value pairs we see:
image

However, editing a **Secret** we just have a YAML editor:
image

So, for Generic Secrets would it be just the YAML Editor, which would show something like:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-secret
type: Opaque
data:
  key1: <base64-encoded value 1>
  key2: <base64-encoded value 2>
```
Thanks, Dave

spadgett:
We could probably implement edit the same way for secrets as config maps. Ideally we'd have special support for source and image secrets, but a generic key/value editor for all secrets is better than nothing. It doesn't have to be part of this PR, however.

dtaylor113:
> We could probably implement edit the same way for secrets as config maps...a generic key/value editor for all secrets is better than nothing. It doesn't have to be part of this PR, however.

So for this PR we just store key/values in `data` and show the `Edit YAML` when editing a generic secret, correct?

spadgett:
I think that's good for now, and we can handle edit in a follow on.

dtaylor113:
Here a create `my-generic-secret` with two key-value pairs:
image

**Browsing/Editing the Secret:**
image
---
image

Edit YAML:
image


@spadgett, does this look correct?  -thanks

spadgett:
Looks correct, thanks.

dtaylor113:
Latest code pushed, ready for formal review -thanks

app/scripts/directives/createSecret.js
```diff
@@ -174,6 +187,11 @@ angular.module("openshiftConsole")
                 secret.stringData.WebHookSecretKey = data.webhookSecretKey;
               }
               break;
+            case "Generic":
+              if (data.genericKeyValues.data) {
+                secret.stringData = data.genericKeyValues.data;
```
spadgett: We'll need a follow on to handle binary content for secrets, but this is fine for now.

app/scripts/directives/createSecret.js
```diff
@@ -70,6 +70,15 @@ angular.module("openshiftConsole")
                 label: "Webhook Secret"
               }
             ]
+          },
+          generic: {
+            label: "Generic Secret",
+            authTypes: [
+              {
+                id: "Generic",
```
spadgett:
```
id: "Opaque",
```

Even if we call it generic secret in the UI, the ID should be Opaque

app/views/directives/create-secret.html
```diff
@@ -350,6 +350,9 @@
         </div>
       </div>
     </div>
+    <div ng-if="newSecret.type === 'generic'">
+      <edit-config-map-or-secret model="newSecret.data.genericKeyValues" type="secret"></edit-config-map-or-secret>
```
spadgett: Does this work if you use...?

```
model="newSecret.data"
```

Then you don't need the intermediate `genericKeyValues` property.

dtaylor113: Unfortunately, it does not work as the `editConfigMapOrSecret` directive expects a `model.data` to exist, and it replaces all the values in `data` to match the key/value pairs entered.  This effectively replaces `newSecret.data.secretName` (and any other values we may use in data).

app/scripts/directives/createSecret.js
dtaylor113: ok, fixed

app/scripts/directives/createSecret.js
dtaylor113: Ok, filed issue #2896

PR #2887, opened 2018-03-09
Display routes including annotation on overview instead of default route

### LINKED ISSUE #2886: Overview displays a max of 2 routes per resource regardless of the number of routes
Example:

Jenkins has 5 routes associated with its service
screen shot 2018-03-08 at 4 03 26 pm

Only 2 of those 5 routes show on the overview
screen shot 2018-03-08 at 4 03 15 pm

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1523393
Fixes #2886

Maintain existing route display behavior for routes that do not have the annotation.  For routes that have the annotation [1], display as many routes as are annotated.

[1] `console.alpha.openshift.io/overview-app-route: 'true'`

localhost_9000_dev-console_project_myproject_overview 1

Also changes expanded row to show all routes instead of limiting to two.

screen shot 2018-03-09 at 11 03 43 am

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -257,9 +257,16 @@ function OverviewController($scope,
     return AppsService.groupByApp(collection, 'metadata.name');
   };
 
-  var getBestRoute = function(routes) {
+  var getRoutesToDisplay = function(routes) {
+    var routesToDisplay = [];
     var bestRoute = null;
     _.each(routes, function(candidate) {
+      // If the the route has the annotation, display it
+      if (RoutesService.isOverviewAppRoute(candidate)) {
+        routesToDisplay.push(candidate);
+        return;
+      }
+
       if (!bestRoute) {
         bestRoute = candidate;
         return;
@@ -269,13 +276,18 @@ function OverviewController($scope,
       bestRoute = RoutesService.getPreferredDisplayRoute(bestRoute, candidate);
     });
 
-    return bestRoute;
+    // If no routes have been added and bestRoute exists, add bestRoute
+    if (!routesToDisplay.length && bestRoute) {
+      routesToDisplay.push(bestRoute);
+    }
+
+    return RoutesService.sortRoutesByScore(routesToDisplay);
   };
 
   // Debounce so we're not reevaluating this too often.
   var updateRoutesByApp = _.debounce(function() {
     $scope.$evalAsync(function() {
-      overview.bestRouteByApp = {};
+      overview.routesToDisplayByApp = {};
 
       if (!overview.routes) {
         return;
@@ -308,8 +320,7 @@ function OverviewController($scope,
             });
           });
         });
-
-        overview.bestRouteByApp[app] = getBestRoute(routesForApp);
+        overview.routesToDisplayByApp[app] = getRoutesToDisplay(routesForApp);
       });
     });
   }, 300, { maxWait: 1500 });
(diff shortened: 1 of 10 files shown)


app/scripts/services/routes.js
```diff
@@ -102,9 +102,17 @@ angular.module("openshiftConsole")
       return annotation(route, "openshift.io/host.generated") !== "true";
     };
 
+    var isOverviewAppRouteToDisplay = function(route){
```
benjaminapetersen: @rhamilto I realize you may have nothing to do with this, but gonna ask anyway :)

Any idea where this annotation name came from?  As far as I have understood from conversations, `annotations` shouldn't have the word `alpha` in them because they really can't change anyway.  Apparently this is a bad practice (meaning if we can discourage it before it gets out there, we probably should).

I'm not sure I'm a fan of the word `overview` in the name either.  While today we are using it to display extra routes on the overview, is it the best name long term?  Would we ever want to use this elsewhere?  Seems unnecessarily restrictive.

app/scripts/controllers/overview.js
```diff
@@ -264,13 +271,18 @@ function OverviewController($scope,
       bestRoute = RoutesService.getPreferredDisplayRoute(bestRoute, candidate);
     });
 
-    return bestRoute;
+    // If no routes have been added and bestRoute exists, add bestRoute
+    if (routesToDisplay.length === 0 && bestRoute !== null) {
```
spadgett: nit, this can be

```js
if (!routesToDisplay.length && bestRoute) {
```

app/scripts/services/routes.js
```diff
@@ -102,9 +102,17 @@ angular.module("openshiftConsole")
...
+      return annotation(route, "console.alpha.openshift.io/overview-app-route-to-display") === "true";
```
spadgett: Maybe just `console.alpha.openshift.io/overview-app-route` ?

app/views/overview.html
```diff
@@ -122,8 +122,8 @@ <h2>
                   <div class="component-label">Application</div>
                   <span ng-bind-html="app | highlightKeywords : overview.state.filterKeywords"></span>
                 </h2>
-                <div ng-if="route = overview.bestRouteByApp[app]" class="pull-right">
-                  <h3 class="overview-route">
+                <div class="overview-routes" ng-if="overview.routesToDisplayByApp[app].length > 0">
```
spadgett: We typically use the `size` filter since it works with both arrays and maps.

```html
<div class="overview-routes" ng-if="overview.routesToDisplayByApp[app] | size">
```

app/views/overview.html
```diff
@@ -122,8 +122,8 @@ <h2>
...
+                  <h3 class="overview-route" ng-repeat="route in overview.routesToDisplayByApp[app] track by $index">
```
spadgett: Use `route | uid` for `track by`. UID is a better identifier for that route than the index in the array since things might get reordered.

```html
<h3 class="overview-route" ng-repeat="route in overview.routesToDisplayByApp[app] track by (route | uid)">
```

app/styles/_overview.less
```diff
@@ -83,16 +83,25 @@
...
-      .overview-route {
-        .h4();
-        margin-top: -5px;
-        margin-bottom: 15px;
+    }
+    @media (min-width: @screen-md-min) {
+      align-items: flex-end;
+      flex-wrap: wrap;
+    }
+    .overview-route {
```
benjaminapetersen: should the style be portable, in case we ever do show stacked routes elsewhere?

app/scripts/services/routes.js
benjaminapetersen: && eliminate `.alpha.` (my other comment)?

rhamilto:
Thanks for the great feedback.  Ready for final review.

PR #2888, opened 2018-03-09
Bug 1551991 - Correctly clear key when another valueFrom resource selected

spadgett:
In our key/value editor, correctly clear the valueFrom key when a different
resource reference is selected.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1551991
/assign @benjaminapetersen


spadgett:
Rebased, dist should be fixed now

spadgett:
Rebuilt dist only

PR #2894, opened 2018-03-13
Console context switcher

rhamilto:
Add optional context switcher.

screen shot 2018-04-16 at 9 31 44 am


app/scripts/directives/nav.js
```diff
@@ -273,10 +294,54 @@ angular.module('openshiftConsole')
         };
 
         $scope.catalogLandingPageEnabled = !Constants.DISABLE_SERVICE_CATALOG_LANDING_PAGE;
-        var select = $elem.find('.selectpicker');
-        var options = [];
 
-        var updateOptions = function() {
+        var contextSelector = $elem.find('.contextselector');
+        var previousContextSelectorValue;
+        contextSelector.append($('<option value="1">Service Catalog</option>'));
```
spadgett: Better to give the values meaningful identifiers like "catalog", "application-console", "cluster-console"

app/scripts/directives/nav.js
```diff
@@ -273,10 +294,54 @@ angular.module('openshiftConsole')
...
+        contextSelector.append($('<option value="2">Application Console</option>'));
+        if (window.OPENSHIFT_CONSTANTS.TECTONIC_URL) {
```
spadgett: Works for me for now. The goal is to have the tectonic console running on the same domain, so we might be able to put it in a well-known place like `/cluster-console` (or wherever) with no config needed.

app/scripts/directives/nav.js
spadgett: If these values are always the same, can they be in the markup instead of adding them in JS?

We had to do this for projects since the options are dynamic.

app/scripts/directives/nav.js
```diff
@@ -273,10 +294,54 @@ angular.module('openshiftConsole')
...
+          contextSelector.append($('<option value="3">Cluster Console</option>'));
+        }
+        contextSelector.append($('<option data-divider="true"></option>'));
+        contextSelector.append($('<option value="4">Set Home Page</option>'));
```
spadgett: We might leave this out for the initial PR until we have the home page dialog updated to include the cluster console homepage option.

app/scripts/directives/nav.js
```diff
@@ -242,6 +243,26 @@ angular.module('openshiftConsole')
...
+          if (path === catalogPath) {
+            selector.selectpicker('val', "catalog");
+          }
+          if (_.includes(path, projectPath)) {
+            selector.selectpicker('val', "application-console");
+          }
+        };
+
+        var goToURL = function(url) {
+          $scope.$apply(function() {
```
spadgett: Use

```js
$scope.$evalAsync(...)
```

app/scripts/directives/nav.js
```diff
@@ -242,6 +243,26 @@ angular.module('openshiftConsole')
...
+        var catalogPath = "/catalog";
+        var projectPath = "/project";
+        var projectsPath = "/projects";
+
+        var setContextPickerSelectedOption = function(selector) {
+          var path = $location.path();
+          if (path === catalogPath) {
+            selector.selectpicker('val', "catalog");
+          }
+          if (_.includes(path, projectPath)) {
```
spadgett: `_.startsWith` is probably more accurate. This could be an `else if` as well.

What happens on pages without a project such as the about page? I think it makes sense to always set it to "application-console" if it's not catalog.

app/scripts/directives/nav.js
```diff
@@ -273,10 +294,47 @@ angular.module('openshiftConsole')
...
 
         $scope.catalogLandingPageEnabled = !Constants.DISABLE_SERVICE_CATALOG_LANDING_PAGE;
-        var select = $elem.find('.selectpicker');
-        var options = [];
 
-        var updateOptions = function() {
+        var contextSelector = $elem.find('.contextselector');
+        var previousContextSelectorValue;
+        $scope.clusterConsoleURL = window.OPENSHIFT_CONSTANTS.TECTONIC_URL;
+        contextSelector
```
spadgett: I'd wrap this in an

```js
if ($scope.clusterConsoleURL) { ... }
```

so you're not initializing the select when we don't need to. Then you can also change the `ng-show` in the view to an `ng-if`.

spadgett:
cc @ggreer

app/scripts/directives/nav.js
rhamilto: I believe the select initializes no matter what since initialization occurs as a result of https://silviomoreto.github.io/bootstrap-select/.  If i change to an ng-if, said initialization doesn't  occur since the <select class="selectpicker contextpicker"> isn't present in the markup.

app/scripts/directives/nav.js
```diff
@@ -273,10 +291,47 @@ angular.module('openshiftConsole')
         };
 
         $scope.catalogLandingPageEnabled = !Constants.DISABLE_SERVICE_CATALOG_LANDING_PAGE;
-        var select = $elem.find('.selectpicker');
-        var options = [];
 
-        var updateOptions = function() {
+        var contextSelector = $elem.find('.contextselector');
+        var previousContextSelectorValue;
+        $scope.clusterConsoleURL = window.OPENSHIFT_CONSTANTS.TECTONIC_URL;
```
spadgett: Let's add an `OPENSHIFT_CONFIG` fallback for the short-term so that we can enable this without an extension.

```js
$scope.clusterConsoleURL = window.OPENSHIFT_CONSTANTS.TECTONIC_URL || window.OPENSHIFT_CONFIG.tectonicURL;
```

app/scripts/directives/nav.js
```diff
@@ -273,10 +291,47 @@ angular.module('openshiftConsole')
...
+        contextSelector
+          .on('loaded.bs.select', function () {
+            setContextPickerSelectedOption(contextSelector);
+          })
+          .on('shown.bs.select', function() {
+            previousContextSelectorValue = $(this).val();
+          })
+          .change(function() {
+            // // unset focus on .dropdown-toggle so outline isn't present
+            // $('.contextselector .dropdown-toggle').blur();
```
spadgett: Remove commented out code? (I assume we want to leave it out for accessibility. Whatever we do, we should try to be consistent with the behavior in Tectonic console.)

app/scripts/directives/nav.js
```diff
@@ -273,10 +291,47 @@ angular.module('openshiftConsole')
...
+            var val = $(this).val();
+            switch (val) {
+              case "catalog":
+               goToURL(catalogPath);
+               break;
```
spadgett: nit: indentation looks off in this switch.

app/scripts/directives/nav.js
```diff
@@ -273,10 +291,47 @@ angular.module('openshiftConsole')
...
+              case "application-console":
+               goToURL(projectsPath);
+               break;
+              case "cluster-console":
+                window.location.assign($scope.clusterConsoleURL);
+                break;
+              case "set-home-page":
```
spadgett: I'd take the set-home-page stuff out for now since it's unclear when we will add that (both here and in the view)

app/scripts/constants.js
```diff
@@ -67,6 +67,8 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     "Latest Release":          "https://github.com/openshift/origin/releases/latest"
   },
 
+  TECTONIC_URL: "https://redhat.com",
```
spadgett: Let's remove this from constants.js for now so no one tries to set it with an extension.

app/styles/_navbar-vertical.less
```diff
@@ -8,11 +8,10 @@
   background-repeat: no-repeat;
   // Max logo dimensions width 196px / height 40px
   height: @navbar-os-header-height-mobile - 1; // 1px border-bottom on navbar-os-mobile
-  width: 196px;
-  @media(min-width: @screen-sm-min) {
+  width: 191px;
+  // @media(min-width: @screen-sm-min) {
```
spadgett: Did you mean to leave this commented out?

rhamilto: Yeah as it serves as documentation of sorts since the image used to/can get bigger as the viewport gets bigger.  I debated removing it, but opted to keep it since we don't have this documented anywhere...

PR #2897, opened 2018-03-15
adding service integrations to service overview row

philbrookes:
It is possible for various mobile services to be bound to each other, these changes to the UI will make it easier for a mobile developer using OpenShift to host the services for their mobile client to connect these services to each other.

Creating an integration between services will create a binding between the two services and also create a pod preset for the binding values to be connected to the correct pods. The deployment of the consuming service will then be updated to trigger a fresh deployment with the new config values present.

There are various states possible:
1. A service is running which can integrate with a service that is not running:
integrations-provision

1.1 A service is being provisioned, which can be integrated with:
integrations-provisioning

2. A service is running which can integrate with a service that is also running:
integrations-no-binding

3. A service is in the process of being integrated or de-integrated with another service:
integrations-pending-binding

4. A service is currently integrated with another service:
integrations-binding-active

4. A service which could be integrated with, is being deprovisioned:
integrations-service-deprovisioning

### DIFF
--- a/app/scripts/directives/overview/serviceInstanceRow.js
+++ b/app/scripts/directives/overview/serviceInstanceRow.js
@@ -4,6 +4,7 @@
   angular.module('openshiftConsole').component('serviceInstanceRow', {
     controller: [
       '$filter',
+      '$rootScope',
       'APIService',
       'AuthorizationService',
       'BindingService',
@@ -21,12 +22,14 @@
   });
 
   function ServiceInstanceRow($filter,
+                              $rootScope,
                               APIService,
                               AuthorizationService,
                               BindingService,
                               ListRowUtils,
                               ServiceInstancesService) {
     var row = this;
+
     var isBindingFailed = $filter('isBindingFailed');
     var isBindingReady = $filter('isBindingReady');
     var serviceInstanceFailedMessage = $filter('serviceInstanceFailedMessage');
@@ -38,6 +41,7 @@
 
     row.serviceBindingsVersion = APIService.getPreferredVersion('servicebindings');
     row.serviceInstancesVersion = APIService.getPreferredVersion('serviceinstances');
+    row.isMobileService = $filter('isMobileService');
 
     var getServiceClass = function() {
       var serviceClassName = ServiceInstancesService.getServiceClassNameForInstance(row.apiObject);
@@ -74,6 +78,13 @@
     row.$onChanges = function(changes) {
       if (changes.bindings) {
         row.deleteableBindings = _.reject(row.bindings, 'metadata.deletionTimestamp');
+      }
+      if (row.isMobileService(row.apiObject) && $rootScope.AEROGEAR_MOBILE_ENABLED) {
+        var serviceClass = getServiceClass();
+        var integrations = _.get(serviceClass, "spec.externalMetadata.integrations");
+        if (integrations) {
+          row.integrations = integrations.split(",");
+        }
       }
     };
 
--- a/app/scripts/directives/serviceIntegration.js
+++ b/app/scripts/directives/serviceIntegration.js
@@ -1,0 +1,272 @@
+'use strict';
+
+(function() {
+  angular.module('openshiftConsole').component('serviceIntegration', {
+    controller: [
+      '$filter',
+      '$scope',
+      'APIService',
+      'AuthorizationService',
+      'BindingService',
+      'Catalog',
+      'DataService',
+      'NotificationsService',
+      ServiceIntegration
+    ],
+    bindings: {
+      integration: '<',
+      consumerService: '<?'
+    },
+    templateUrl: 'views/directives/_service-integration.html'
+  });
+
+  function ServiceIntegration(
+                          $filter,
+                          $scope,
+                          APIService,
+                          AuthorizationService,
+                          BindingService,
+                          Catalog,
+                          DataService,
+                          NotificationsService) {
+    var ctrl = this;
+
+    var deploymentPreferredVersion = APIService.getPreferredVersion('deployments');
+    var bindingPreferredVersion = APIService.getPreferredVersion('servicebindings');
+    var instancePreferredVersion = APIService.getPreferredVersion('serviceinstances');
+    var podPresetPreferredVersion = APIService.kindToResourceGroupVersion({group: "settings.k8s.io", kind: "podpreset"});
+    
+    var isServiceInstanceReady = $filter('isServiceInstanceReady');
+    var isBindingReady = $filter("isBindingReady");
+    var getErrorDetails = $filter('getErrorDetails');
+
+    var integrationName = ctrl.integration.spec.externalMetadata.serviceName;
+
+    var watches = [];
+    var ppwatch = false;
+
+    $scope.$on('$destroy', function() {
+      DataService.unwatchAll(watches);
+      if(ppwatch) {
+        DataService.unwatch(ppwatch);
+      }
+    });
+
+    ctrl.$onInit = function() {
+      var context = {namespace: ctrl.consumerService.metadata.namespace};
+
+      //setup watch on service instances, looking for an instance which provides this integration
+      watches.push(DataService.watch(instancePreferredVersion, context, function(serviceInstancesData) {
+        var data = serviceInstancesData.by('metadata.name');
+        $scope.serviceInstance = _.find(data, function(serviceInstance) {
+          var clusterServiceClassExternalName = _.get(serviceInstance, "spec.clusterServiceClassExternalName");
+          return (clusterServiceClassExternalName === ctrl.integration.spec.externalName);
+        });
+      }));
+
+      //pod preset watch should only exist when we have a handle to the service instance
+      $scope.$watch('serviceInstance', function(serviceInstance){
+        if(!serviceInstance && ppwatch !== false){
+          DataService.unwatch(ppwatch);
+          ppwatch = false;
+          return;
+        }
+        //dont recreate the pod preset watch
+        if(ppwatch !== false) {
+          return;
+        }
+        // watch for the pod preset for this integration
+        ppwatch = DataService.watch(podPresetPreferredVersion, context, function(podPresets) {
+          var data = podPresets.by("metadata.name");
+          ctrl.podPreset = _.find(data, function(podPreset) {
+            return podPreset.metadata.name === _.get(ctrl.consumerService, 'metadata.name') + "-" + _.get($scope.serviceInstance, 'metadata.name');
+          });
+        });
+      });
+      
+      //setup watch on servincebindings, watch for bindings consumed by this service
+      watches.push(DataService.watch(bindingPreferredVersion, context, function(bindingData) {
+        var data = bindingData.by("metadata.name");
+        ctrl.binding = _.find(data, function(binding) {
+          var bindingProviderName = _.get(binding, ["metadata", "annotations", "integrations.aerogear.org/provider"]);
+          var bindingConsumerName = _.get(binding, ["metadata", "annotations", "integrations.aerogear.org/consumer"]);
+          var consumerName = _.get(ctrl, "consumerService.metadata.labels.serviceName");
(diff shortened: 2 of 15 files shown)


spadgett:
/ok-to-test
cc @cshinn

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+        }).pop();
+      });
+
+      //setup watch on servincebindings, watch for bindings consumed by this service
+      DataService.watch(bindingPreferredVersion, context, function(bindingData) {
+        var data = bindingData.by("metadata.name");
+        ctrl.binding = _.filter(data, function(binding){
+          var bindingProviderName = _.get(binding, "metadata.annotations.provider", "provider_not_found");
+          var bindingConsumerName = _.get(binding, "metadata.annotations.consumer", "consumer_not_found");
+          var consumerName = _.get(ctrl, "consumerService.metadata.labels.serviceName", "consumer_name_not_found");
```
spadgett: I'd rather just check for null / undefined instead of using something like "provider_not_found".

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+        var data = podPresets.by("metadata.name");
+        ctrl.podPreset = _.filter(data, function(podPreset) {
+          return podPreset.metadata.name === _.get(ctrl.consumerService, 'metadata.name') + "-" + _.get(ctrl.serviceInstance, 'metadata.name');
+        }).pop();
+      });
+
+      //setup watch on servincebindings, watch for bindings consumed by this service
+      DataService.watch(bindingPreferredVersion, context, function(bindingData) {
+        var data = bindingData.by("metadata.name");
+        ctrl.binding = _.filter(data, function(binding){
```
spadgett: Use `_.find` instead of `_.filter` / `pop`

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+          var bindingConsumerName = _.get(binding, "metadata.annotations.consumer", "consumer_not_found");
+          var consumerName = _.get(ctrl, "consumerService.metadata.labels.serviceName", "consumer_name_not_found");
+          return (bindingProviderName === integrationName && bindingConsumerName === consumerName);
+        }).pop();
+      });
+
+      //setup watch on service instances, looking for an instance which provides this integration
+      DataService.watch(instancePreferredVersion, context, function(serviceInstancesData) {
+        var data = serviceInstancesData.by('metadata.name');
+        ctrl.serviceInstance = _.filter(data, function(serviceInstance){
```
spadgett: _.find

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+
+    var isServiceInstanceReady = $filter('isServiceInstanceReady');
+    var isBindingReady = $filter("isBindingReady");
+
+    var integrationName = ctrl.integration.spec.externalMetadata.serviceName;
+
+    ctrl.$onInit = function() {
+      var context = {namespace: ctrl.consumerService.metadata.namespace};
+      // watch for the pod preset for this integration
+      DataService.watch(podPresetPreferredVersion, context, function(podPresets){
```
spadgett: You're leaking websockets since you're not calling `DataService.unwatch` on these watches.

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+          return (_.get(serviceInstance, "metadata.labels.serviceName") === integrationName && isServiceInstanceReady(serviceInstance));
+        }).pop();
+      });
+    };
+
+    var generatePodPresetTemplate = function(consumerService, providerService, binding) {
+      var consumerSvcName = _.get(consumerService, 'metadata.name');
+      var providerSvcName = _.get(providerService, 'metadata.name');
+      var podPreset = {
+        "apiVersion": "settings.k8s.io/v1alpha1",
```
spadgett: nit: Since the keys here and below don't need to be quoted

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+                          AuthorizationService,
+                          BindingService,
+                          DataService,
+                          NotificationsService) {
+    var ctrl = this;
+
+    var deploymentPreferredVersion = APIService.getPreferredVersion('deployments');
+    var bindingPreferredVersion = APIService.getPreferredVersion('servicebindings');
+    var instancePreferredVersion = APIService.getPreferredVersion('serviceinstances');
+    var podPresetPreferredVersion = {group: "settings.k8s.io", version: "v1alpha1", resource: "podpresets"};
```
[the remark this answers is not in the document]
spadgett: We've also made a statement not to use alpha APIs in the console. We need to be forward-compatible with an n+1 API server for upgrades when masters can be at different versions.

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+      };
+
+      //binding might not be ready currently, so
+      //watch binding to wait for it to be ready
+      var bindingReadyWatch = DataService.watchObject(bindingPreferredVersion, _.get(binding, 'metadata.name'), context, function(watchBinding){
+        if(isBindingReady(watchBinding)){
+          DataService.unwatch(bindingReadyWatch);
+          NotificationsService.addNotification({
+            type: "success",
+            message: "Binding has been set in " + _.get(ctrl, 'consumerService.metadata.labels.serviceName') + " Redeployed " + _.get(ctrl, 'consumerService.metadata.labels.serviceName')
```
spadgett: We try to use complete sentences with punctuation in these messages.

app/scripts/directives/serviceIntegration.js
spadgett: We typically namespace annotations. Something like `servicecatalog.openshift.io/consumer`.

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+              _.get(ctrl, 'consumerService.metadata.labels.serviceName'), 
+              deployment, 
+              context
+            );
+          })
+          .catch(function(err) {
+            NotificationsService.addNotification({
+              type: "error",
+              message: "Failed to integrate service binding",
+              details: err.data.message
```
spadgett: getErrorDetails filter

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+              deploymentPreferredVersion, 
+              _.get(ctrl, 'consumerService.metadata.labels.serviceName'), 
+              deployment, 
+              context
+            );
+          })
+          .catch(function(err) {
+            NotificationsService.addNotification({
+              type: "error",
+              message: "Failed to integrate service binding",
```
spadgett: Add punctuation.

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+      DataService.delete(podPresetPreferredVersion, ctrl.podPreset.metadata.name, context, deleteOptions)
+      .then(function(){
+        return DataService.get(
+          deploymentPreferredVersion, 
+          _.get(ctrl, 'consumerService.metadata.labels.serviceName'), 
+          context
+        );
+      })
+      .then(function(deployment) {
+        delete deployment.spec.template.metadata.labels[integrationName];
```
spadgett: Use `angular.copy` before changing the deployment. DataService does some caching and you don't want to modify the cached data.

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+          deploymentPreferredVersion, 
+          _.get(ctrl, 'consumerService.metadata.labels.serviceName'), 
+          deployment, 
+          context
+        );
+      })
+      .catch(function(error){
+        NotificationsService.addNotification({
+          type: "error",
+          message: "error removing integration",
```
spadgett: Should be a complete sentence.

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+          details: error.data.message
```
spadgett: getErrorDetails filter

app/scripts/filters/mobile.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+/* jshint unused: false */
+
+angular.module('openshiftConsole')
+  .filter('isMobileService', function() {
+    return function(serviceInstance) {
+      return _.get(serviceInstance, 'metadata.labels', {}).mobile === 'enabled';
```
spadgett:
```js
return _.get(serviceInstance, 'metadata.labels.mobile') === 'enabled';
```

app/scripts/directives/serviceIntegration.js
```diff
@@ -0,0 +1,242 @@
...
+      //binding might not be ready currently, so
+      //watch binding to wait for it to be ready
+      var bindingReadyWatch = DataService.watchObject(bindingPreferredVersion, _.get(binding, 'metadata.name'), context, function(watchBinding){
+        if(isBindingReady(watchBinding)){
+          DataService.unwatch(bindingReadyWatch);
+          NotificationsService.addNotification({
+            type: "success",
+            message: "Binding has been set in " + _.get(ctrl, 'consumerService.metadata.labels.serviceName') + " Redeployed " + _.get(ctrl, 'consumerService.metadata.labels.serviceName')
+          });
+          _.set(watchBinding, "metadata.annotations.consumer", ctrl.consumerService.metadata.labels.serviceName);
```
spadgett: Use `angular.copy` before modifying the binding.

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -75,6 +79,12 @@
       if (changes.bindings) {
         row.deleteableBindings = _.reject(row.bindings, 'metadata.deletionTimestamp');
       }
+      if(row.isMobileService(row.apiObject) && $rootScope.AEROGEAR_MOBILE_ENABLED){
```
spadgett: nit: spaces before `(` and `{`, here and below

app/scripts/directives/overview/serviceInstanceRow.js
```diff
@@ -75,6 +79,12 @@
...
+        var integrations = _.get(getServiceClass(), "spec.externalMetadata.integrations", false);
```
spadgett: You don't need to give a default here since undefined is falsey. If you do, I'd rather it be `''` since we're expecting a string, but better to leave it off.

```js
var serviceClass = getServiceClass();
var integrations = _.get(serviceClass, "spec.externalMetadata.integrations");
```

cshinn:
@philbrookes I'm having Visuals take a look at it, and I'll get back to you ASAP

cshinn:
@philbrookes Here are a few small suggestions.
group 3

philbrookes:
@spadgett @cshinn I have implemented the changes as you have suggested - as much as I could make them work.

Some thoughts:
- The podpreset needs to stay as a watch, rather than a watchObject as it can be deleted and recreated and the watchobject does not appear to handle that correctly.
- The pull-left does seem to be required, if I remove that the UI looks broken.
- The podpreset version is pulled from the APIService now, rather than hard-coding the API version.
- I've added a new state which reflects that the integratable service is being provisioned.

Let me know if you have any thoughts or questions.

philbrookes:
@spadgett  @cshinn I have rebased this now, and retested against a 3.9.0 cluster, can I get a fresh review please?

spadgett:
> I have rebased this now, and retested against a 3.9.0 cluster, can I get a fresh review please?

Sorry following up on this:

* We're past feature freeze for 3.10 (for a while now), so we'll need to wait until master opens for 3.11
* We'll need to look at the state of pod presets in 3.11 before we can add that alpha API. Alpha APIs cause a lot of rework and can break things during upgrades when masters are at different versions since alpha APIs are not stable

PR #2900, opened 2018-03-16
Fix bug "Open link in new tab" in right click menu jumps to home page

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1470998

And also changes Navigate.toPodsForDeployment to go to the replication
controller page rather than the filtered pods page.

### DIFF
--- a/app/views/monitoring.html
+++ b/app/views/monitoring.html
@@ -190,8 +190,7 @@
                     <div class="list-pf-additional-content">
                       <div class="list-pf-additional-content-item">
                         <div class="pods">
-                          <a href=""
-                            ng-click="viewPodsForSet(replicationController)"
+                          <a ng-href="{{replicationController | donutURL : podsByOwnerUID[replicationController.metadata.uid]}}"
                             class="mini-donut-link"
                             ng-class="{ 'disabled-link': !(podsByOwnerUID[replicationController.metadata.uid] | size) }">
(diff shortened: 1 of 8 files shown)


app/views/monitoring.html
```diff
@@ -190,12 +190,12 @@ <h3>
                     <div class="list-pf-additional-content">
                       <div class="list-pf-additional-content-item">
                         <div class="pods">
-                          <a href=""
+                          <span
```
spadgett: This fixes the right-click behavior, but unfortunately it breaks keyboard and screenreader accessibility :/

We could leave it a link and then implement a filter that creates the URL.

```html
<a ng-href="replicationController | donutURL : podsByOwnerUID[replicationController.metadata.uid]" ...>
```

The filter would do what the current Navigate method does, except it just returns a URL without changing location. Then open in new window should work.

rhamilto: How do we handle the case where we [navigate directly to the pod if there is only 1](https://github.com/rhamilto/origin-web-console/blob/e362dc49cbf613150b4ff1a84904bd2eb5cbdd59/app/scripts/services/navigate.js#L161-L163)?

spadgett: The filter can return a different URL if there's one pod.

rhamilto:
Take 2.

PR #2901, opened 2018-03-19
Handle binary files when creating generic secrets

### LINKED ISSUE #2896: Handle Binary Content for Generic Secrets
Related #2879 

In Create Generic Secret we are storing key-pair values in `secret.stringData = data.genericKeyValues.data;`.   Since the 'value' can be uploaded from a file which has base64, we need to store those values in `secret.data` as opposed to `secret.stringData`.

spadgett:
/assign @dtaylor113 

Closes #2896

### DIFF
--- a/app/scripts/directives/createSecret.js
+++ b/app/scripts/directives/createSecret.js
@@ -36,6 +36,15 @@ angular.module("openshiftConsole")
         };
 
         $scope.secretAuthTypeMap = {
+          generic: {
+            label: "Generic Secret",
+            authTypes: [
+              {
+                id: "Opaque",
+                label: "Generic Secret"
+              }
+            ]
+          },
           image: {
             label: "Image Secret",
             authTypes: [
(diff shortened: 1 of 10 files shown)


dtaylor113:
I uploaded a file with non-printable characters, and the Ace editor is hidden:
image

However, could we display help text indicating that the file contains non-printable chars, so the editor is unavailable?  Similar to how we do it in Secret Details:
image

spadgett:
> However, could we display help text indicating that the file contains non-printable chars, so the editor is unavailable?

Yeah, my first implementation had that, and I actually took it out since I thought it worked better without. But I don't have a strong opinion.

@ncameronbritt thoughts?

dtaylor113:
> Yeah, my first implementation had that, and I actually took it out since I thought it worked better without. But I don't have a strong opinion.

IMO we need it, it looks like a bug when there may be multiple items, but some do not show the editor:
image

@ncameronbritt

spadgett:
@dtaylor113 I spoke to @ncameronbritt, and we agreed to add the message. PTAL

PR #2903, opened 2018-03-20
Add and remove mobile client from service instance

sedroche:
A mobile developer will want to see which clients are associated with which service instances.
This PR adds a component to mobile aware service instances to add and exclude clients from the service.

### Single client
single-client

### Multiple clients
multiple-clients

### Add client button becomes enabled when excluded clients are available
enabled-button

### Add Client modal
add-client-modal

### DIFF
--- a/app/scripts/directives/mobileServiceClients.js
+++ b/app/scripts/directives/mobileServiceClients.js
@@ -1,0 +1,66 @@
+'use strict';
+
+(function() {
+  angular.module('openshiftConsole').component('mobileServiceClients', {
+    controller: [
+      '$filter',
+      'MobileClientsService',
+      'NotificationsService',
+      MobileServiceClientsCtrl
+    ],
+    bindings: {
+      project: '<',
+      serviceInstance: '<',
+      mobileClients: '<'
+    },
+    templateUrl: 'views/directives/mobile-service-clients.html'
+  });
+
+  function MobileServiceClientsCtrl(
+                       $filter,
+                       MobileClientsService,
+                       NotificationsService) {
+    var ctrl = this;
+    var context = {namespace: _.get(ctrl, 'project.metadata.name')};
+    var getErrorDetails = $filter('getErrorDetails');
+
+    ctrl.$doCheck = function() {
+      ctrl.associatedClients = MobileClientsService.filterNotExcluded(ctrl.serviceInstance, ctrl.mobileClients);
+      ctrl.excludedClients = MobileClientsService.filterExcluded(ctrl.serviceInstance, ctrl.mobileClients);
+      ctrl.hasExcludedClients = !_.isEmpty(ctrl.excludedClients);
+    };
+
+    ctrl.excludeClient = function(mobileClient) {
+      MobileClientsService.excludeClient(mobileClient, ctrl.serviceInstance, context)
+      .then(function() {
+          NotificationsService.addNotification({
+            type: 'success',
+            message: 'Mobile client ' + _.get(mobileClient, 'spec.name') + ' excluded from ' + _.get(ctrl.serviceInstance, 'metadata.name') + '.'
+          });
+        }).catch(function(error) {
+          NotificationsService.addNotification({
+            type: 'error',
+            message: 'Failed to exclude mobile client ' + _.get(mobileClient, 'spec.name') + '.',
+            details: getErrorDetails(error)
+          });
+        });
+    };
+
+    ctrl.addMobileClient = function(mobileClient) {
+      MobileClientsService.removeFromExcluded(mobileClient, ctrl.serviceInstance, context)
+        .then(function() {
+          NotificationsService.addNotification({
+            type: 'success',
+            message: 'Successfully added ' + mobileClient.metadata.name + ' client.'
+          });
+        })
+        .catch(function(error) {
+          NotificationsService.addNotification({
+            type: 'error',
+            message: 'Failed to add mobile client.',
+            details: getErrorDetails(error)
+          });
+        });
+    };
+  }
+})();
--- a/app/scripts/filters/mobile.js
+++ b/app/scripts/filters/mobile.js
@@ -1,0 +1,8 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .filter('isMobileService', function() {
+    return function(serviceInstance) {
+      return _.get(serviceInstance, 'metadata.labels.mobile') === 'enabled';
+    };
+  });
--- a/app/scripts/services/mobileClients.js
+++ b/app/scripts/services/mobileClients.js
@@ -1,0 +1,67 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .factory("MobileClientsService", function(DataService) {
+
+    var mobileclientVersion = {
+      group: "mobile.k8s.io",
+      version: "v1alpha1",
+      resource: "mobileclients"
+    };
+
+
+    var watch = function(context, callback, opts) {
+      return DataService.watch(mobileclientVersion, context, callback, opts);
+    };
+
+    var getMobileClients = function(namespace) {
+      return DataService.list(mobileclientVersion, {namespace: namespace})
+        .then(function(clients) {
+          return clients.by("metadata.name");
+        });
+    };
+
+    var filterExcluded = function(serviceInstance, mobileClients) {
+      var serviceName = _.get(serviceInstance, 'metadata.name', '');
+      return _.filter(mobileClients, function(client) {
+        var excludedServices = _.get(client, 'spec.excludedServices');
+        return _.includes(excludedServices, serviceName);
+      });
+    };
+
+    var filterNotExcluded = function(serviceInstance, mobileClients) {
+      var serviceName = _.get(serviceInstance, 'metadata.name', '');
+      return _.filter(mobileClients, function(client) {
+        var excludedServices = _.get(client, 'spec.excludedServices');
+        return !_.includes(excludedServices, serviceName);
+      });
+    };
+
+    var removeFromExcluded = function(mobileClient, serviceInstance, context) {
+      mobileClient = angular.copy(mobileClient);
+      var serviceName = _.get(serviceInstance, 'metadata.name', '');
+      _.remove(mobileClient.spec.excludedServices, function(service) {
+        return service === serviceName;
+      });
+      return DataService.update(mobileclientVersion, mobileClient.metadata.name, mobileClient, context);
+    };
+
+    var excludeClient = function(mobileClient, serviceInstance, context) {
+      mobileClient = angular.copy(mobileClient);
+      var excludedServices = _.get(mobileClient, 'spec.excludedServices') || [];
+      excludedServices.push(_.get(serviceInstance, 'metadata.name'));
+      _.set(mobileClient, 'spec.excludedServices', excludedServices);
+      return DataService.update(mobileclientVersion, mobileClient.metadata.name, mobileClient, context);
+    };
+
+    return {
+      watch: watch,
+      filterExcluded: filterExcluded,
+      filterNotExcluded: filterNotExcluded,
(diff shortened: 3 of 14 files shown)


app/scripts/directives/addMobileClient.js
```diff
@@ -0,0 +1,54 @@
...
+        .then(function() {
+          NotificationsService.addNotification({
+            type: "success",
+            message: "Successfully added " + mobileClient.metadata.name + " client."
+          });
+        })
+        .catch(function(error) {
+          NotificationsService.addNotification({
+            type: "error",
+            message: "Failed to add mobile client",
```
spadgett: The messages here and below should have punctuation.

"Failed to add mobile client."

app/scripts/filters/mobile.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+/* jshint unused: false */
```
spadgett: Why is this needed?

app/scripts/directives/mobileServiceClients.js
```diff
@@ -0,0 +1,60 @@
...
+    };
+
+    ctrl.showOverlayPanel = function(panelName, state) {
+      _.set(ctrl, 'overlay.panelVisible', true);
+      _.set(ctrl, 'overlay.panelName', panelName);
+      _.set(ctrl, 'overlay.state', state);
+    };
+
+    ctrl.canAddMobileClient = function() {
+      return !MobileClientsService.filterExcluded(ctrl.serviceInstance, ctrl.mobileClients).length;
```
spadgett: nit: prefer !_.isEmpty(...)

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,65 @@
...
+  .factory("MobileClientsService", function(DataService) {
+
+    var mobileclientVersion = {
+      group: "mobile.k8s.io",
+      version: "v1alpha1",
+      resource: "mobileclients"
+    };
+
+
+    var watch = function(context, callback, opts ) {
```
spadgett: nit: extra space before closing paren

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,65 @@
...
+      return DataService.list(mobileclientVersion, {namespace: namespace})
+        .then(function(clients) {
+          return clients.by("metadata.name");
+        });
+    };
+
+    var filterExcluded = function(serviceInstance, mobileClients) {
+      var serviceName = _.get(serviceInstance, 'metadata.name', '');
+      return _.filter(mobileClients, function(client) {
+        var excludedServices = _.get(client, 'spec.excludedServices') || [];
```
spadgett: `_.includes` can take null / undefined, so `|| []` isn't needed

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,65 @@
...
+        return _.includes(excludedServices, serviceName);
+      });
+    };
+
+    var filterNotExcluded = function(serviceInstance, mobileClients) {
+      var serviceName = _.get(serviceInstance, 'metadata.name', '');
+      return _.filter(mobileClients, function(client) {
+        var excludedServices = _.get(client, 'spec.excludedServices') || [];
```
spadgett: `|| []` not needed

app/scripts/services/mobileClients.js
```diff
@@ -0,0 +1,65 @@
...
+        return !_.includes(excludedServices, serviceName);
+      });
+    };
+
+    var removeFromExcluded = function(mobileClient, serviceInstance, context) {
+      var serviceName = _.get(serviceInstance, 'metadata.name', '');
+      _.remove(mobileClient.spec.excludedServices, function(service) {
```
spadgett: Call `angular.copy` before modifying the mobile client.

app/styles/_mobile-clients.less
```diff
@@ -0,0 +1,41 @@
...
+
+  .mobile-client {
+    height: 60px;
+    margin-bottom: 1em;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+
+    .client-icon {
+     vertical-align: bottom;
```
spadgett: indentation

app/styles/_mobile-clients.less
```diff
@@ -0,0 +1,41 @@
+.mobile-clients {
```
spadgett: Please keep the declarations sorted alphabetically in this file. We use this style guide:

http://codeguide.patternfly.org/#css-declaration-order

app/views/directives/add-mobile-client.html
```diff
@@ -0,0 +1,34 @@
...
+  <div class="panel-heading">
+    <h3 class="panel-title pull-left">Select Mobile Client to add to this service</h3>
+    <button class="panel-title btn pull-right" type="button" ng-click="ctrl.onClose()">
+      <span aria-hidden="true" class="pficon pficon-close"></span>
+    </button>
+    <div class="clearfix"></div>
+  </div>
+  <div class="panel-body">
+    <div class="list-pf">
+        <div ng-repeat="client in ctrl.clientsWhereExcluded" class="list-pf-item">
```
spadgett: indentation

app/views/directives/add-mobile-client.html
```diff
@@ -0,0 +1,34 @@
+<div class="panel panel-default">
```
spadgett: The dialog should follow Patternfly markup and styles.

Here is an example from the console:

https://github.com/openshift/origin-web-console/blob/master/app/views/modals/confirm-save-log.html

app/views/directives/add-mobile-client.html
```diff
@@ -0,0 +1,34 @@
...
+  <div class="panel-heading">
+    <h3 class="panel-title pull-left">Select Mobile Client to add to this service</h3>
+    <button class="panel-title btn pull-right" type="button" ng-click="ctrl.onClose()">
+      <span aria-hidden="true" class="pficon pficon-close"></span>
```
spadgett: Missing `sr-only` text.

app/views/directives/add-mobile-client.html
spadgett:
```html
<div ng-repeat="client in ctrl.clientsWhereExcluded try by (client | uid)" class="list-pf-item">
```

app/views/directives/add-mobile-client.html
```diff
@@ -0,0 +1,34 @@
...
+    </button>
+    <div class="clearfix"></div>
+  </div>
+  <div class="panel-body">
+    <div class="list-pf">
+        <div ng-repeat="client in ctrl.clientsWhereExcluded" class="list-pf-item">
+          <div class="list-pf-container" ng-click="ctrl.addMobileClient(client)">
```
spadgett: This is not accessible.

app/views/directives/add-mobile-client.html
```diff
@@ -0,0 +1,34 @@
...
+              <div class="order-service-details-top">
+                <div class="service-icon">
+                  <span class="fa-4x {{client.metadata.annotations.icon}}"></span>
+                </div>
+                <div class="service-title-area">
+                  <div class="service-title ng-binding">
+                    {{client.spec.name}}
+                  </div>
+                  <div class="order-service-tags ng-scope">
+                    <span class="tag ng-binding ng-scope">
```
spadgett: Why are we adding `ng-binding` and `ng-scope` classes?

app/views/directives/mobile-service-clients.html
```diff
@@ -0,0 +1,30 @@
+<div class="row mobile-clients">
+  <div class="col-sm-6">
+    <div class="client-title component-label section-label">Mobile Clients</div>
+    <div ng-repeat="mobileClient in $ctrl.filteredClients" class="col-sm-12 col-md-6 mobile-client">
```
spadgett: track by (mobileClient | uid)

app/views/directives/mobile-service-clients.html
```diff
@@ -0,0 +1,30 @@
...
+      <span class="fa-2x {{mobileClient.metadata.annotations.icon}} client-icon"></span>
```
spadgett: aria-hidden="true"

app/views/directives/mobile-service-clients.html
```diff
@@ -0,0 +1,30 @@
...
+      <div class="client-details">
+        <a ng-href="{{mobileClient | navigateResourceURL}}">{{mobileClient.spec.name}}</a>
+        <span class="fa fa-trash-o action-button exclude" ng-click="$ctrl.excludeClient(mobileClient)"></span>
```
spadgett: This is not accessible.

app/views/directives/mobile-service-clients.html
```diff
@@ -0,0 +1,30 @@
...
+        <div class="client-type">{{mobileClient.spec.appIdentifier}}</div>
+      </div>
+    </div>
+  </div>
+
+  <div class="col-xs-12 col-sm-6">
+    <button ng-disabled="$ctrl.canAddMobileClient()" class="btn btn-default btn-lg pull-right"
```
spadgett: I'm not sure this should have `btn-lg`?

app/views/overview/_service-instance-row.html
```diff
@@ -152,12 +152,22 @@ <h3>
...
+        </div>
+        <div class="row">
+          <div class="col-sm-12">
             <div ng-if="row.serviceClass.spec.externalMetadata.documentationUrl || row.serviceClass.spec.externalMetadata.supportUrl">
               <a ng-if="row.serviceClass.spec.externalMetadata.documentationUrl" ng-href="{{row.serviceClass.spec.externalMetadata.documentationUrl}}" target="_blank" class="learn-more-link">View Documentation <i class="fa fa-external-link" aria-hidden="true"></i></a>
               <a ng-if="row.serviceClass.spec.externalMetadata.supportUrl" ng-href="{{row.serviceClass.spec.externalMetadata.supportUrl}}" target="_blank" class="learn-more-link">Get Support <i class="fa fa-external-link" aria-hidden="true"></i></a>
             </div>
           </div>
         </div>
+        <div ng-if="row.isMobileEnabled && row.hasMobileClients && row.apiObject | isMobileService">
```
spadgett: You need parens around the filter

```html
<div ng-if="row.isMobileEnabled && row.hasMobileClients && (row.apiObject | isMobileService)">
```

spadgett:
cc @cshinn

cshinn:
@sedroche Looks good! Just a couple suggestions:

For consistency with other products, it would be better to use pficon-delete in place of the current trash can icon.

Is it possible to show a filtered view of the catalog when selecting a client to add to the project?
If it is not, that modal should use the styling of other OpenShift modals. @spadgett can you think of a good example in the code to base this off of?

sedroche:
@cshinn  Cheers for feedback. When we open the modal we are selecting already provisioned clients rather than provisioning a new instance so the modal makes more sense here.

cshinn:
@sedroche right, makes sense. Another option might be to do something like this
group 2

sedroche:
@cshinn Yes we can do that. I would defer to you on this one so am happy to update if you would like?

cshinn:
@sedroche That's excellent! I checked with some other designers, and the dropdown seems like the best way to go.

sedroche:
@cshinn :+1: I will make this change.

sedroche:
@spadgett Thanks for your review. All feedback makes sense and I will add the requested changes.

sedroche:
@cshinn I have updated the design.

#### New delete icon
new-delete-icon

#### New drop down selector
new-dropdown

sedroche:
@spadgett I believe I have addressed your feedback if you can take another look? Thanks

sedroche:
Closing this branch as we have taken a different direction with this UI

PR #2904, opened 2018-03-20
Provide link to create pull secret to private repo image on the deploy image.

sg00dwin:
Also, provide additional help text to help the user understand that a image pull secret must exist to deploy an image from a private repository.
- If no project exists then do not show create pull secret link option
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1489374

deploy-image-create-pull-secret-first-then-deploy2

[Updated] Help text slightly altered to the following.
screen shot 2018-03-21 at 1 56 18 pm
screen shot 2018-03-21 at 1 56 13 pm

### DIFF
--- a/app/scripts/directives/deployImage.js
+++ b/app/scripts/directives/deployImage.js
@@ -404,6 +404,17 @@ angular.module("openshiftConsole")
             });
           };
 
+          $scope.openCreateWebhookSecretModal = function() {
+            var dialogScope = $scope.$new();
+            dialogScope.type = "image";
+            dialogScope.namespace = $scope.input.selectedProject.metadata.name;
+            $uibModal.open({
+              templateUrl: 'views/modals/create-secret.html',
+              controller: 'CreateSecretModalController',
+              scope: dialogScope
+            });
+          };
+
           // When the deploy-image component is displayed in a dialog, the create
           // button is outside the component since it is in the wizard footer. Listen
           // for an event for when the button is clicked.
--- a/app/views/directives/deploy-image.html
+++ b/app/views/directives/deploy-image.html
@@ -2,7 +2,7 @@
   <select-project ng-if="!project" selected-project="input.selectedProject" name-taken="projectNameTaken"></select-project>
   <span ng-show="!noProjectsCantCreate">
     <p>
-      Deploy an existing image from an image stream tag or docker pull spec.
+      Deploy an existing image from an image stream tag or image registry.
     </p>
     <ng-form name="forms.imageSelection">
       <fieldset ng-disabled="loading">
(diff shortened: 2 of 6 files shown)


app/scripts/directives/deployImage.js
```diff
@@ -404,6 +404,16 @@ angular.module("openshiftConsole")
             });
           };
 
+          $scope.openCreateWebhookSecretModal = function() {
+            $scope.type="image";
+            $scope.namespace = $scope.input.selectedProject.metadata.name;
```
spadgett: Can you try this?

```js
var dialogScope = $scope.$new();
dialogScope.type = "image";
dialogScope.namespace = $scope.input.selectedProject.metadata.name;
$uibModal.open({
  templateUrl: 'views/modals/create-secret.html',
  controller: 'CreateSecretModalController',
  scope: dialogScope
});
```

sg00dwin:
@spadgett pr updated

app/views/directives/deploy-image.html
```diff
@@ -60,6 +60,10 @@
                 </button>
               </span>
             </div>
+            <div ng-if="input.selectedProject.metadata.uid && !import.namespace" class="help-block">To pull a secured container image that is not from the internal registry, you must
```
spadgett: @sg00dwin It's not clear to me why we're checking `!import.namespace` in the `ng-if`

sg00dwin: My original thought was to not display the help text with the link to create a pull secret, when the Image Name option is disabled. But thinking about it, we should probably just show it.

app/views/directives/deploy-image.html
```diff
@@ -60,6 +60,10 @@
...
+              <a href=""
+                ng-click="openCreateWebhookSecretModal()">create an image pull secret</a>
+               from your Docker credentials and add it to your service account. <a href="https://docs.openshift.org/latest/dev_guide/managing_images.html#using-image-pull-secrets" target="_blank"><span class="learn-more-inline">Learn More&nbsp;<i class="fa fa-external-link" aria-hidden="true"></i></span></a></div>
```
xingxingxia: Hmm, something related to wording. Seems it is better to use `your registry credentials` or `your image registry credentials` or whatever that avoids string `Docker` (similar to bug 1447203)?

spadgett: Thanks yeah, let's make this

"To deploy an image from a private repository, you must create an image secret with your image registry credentials."

app/views/directives/deploy-image.html
```diff
@@ -196,6 +200,7 @@ <h2>
         Could not load image metadata.
       </h2>
       <p>{{import.error | upperFirst}}</p>
+      <p>The image may not exist or it may be in a secure registry. Check that you have entered the image name correctly or <a href="" ng-click="openCreateWebhookSecretModal()">create an image pull secret</a> from your Docker credentials and try again.</p>
```
spadgett: "...with your image registry credentials and try again."

sg00dwin: updated

openshift-docker:
Note that email field is completely optional and no one uses it anymore
(even dockerhub).  It's basically a vestigial appendix.

ncameronbritt:
If that's the case, I'd support removing it. Is there any reason/case that it's useful to provide an email address?

Issue #2864, opened 2018-03-01
Delete project dialog no longer shows the name to type


benjaminapetersen:
Was this an intentional change?  You used to see the name in bold:

![screen shot 2018-03-01 at 10 11 50 am 1](https://user-images.githubusercontent.com/280512/36852197-187b983e-1d39-11e8-8c49-889fea0a2942.png)

@spadgett @rhamilto 

If this is expected, this issue can be closed.

spadgett:
@benjaminapetersen The name is there.... Is the bug that it's no longer bolded?

benjaminapetersen:
Right. not bold. and i think it was larger.  it was previously very clear & obvious what you should type to get the "delete" button to activate.  Memory might be foggy but i thought this modal was scarier.  😄

spadgett:
@openshift/team-ux-review 

/bug usability
/priority p3
/assign @rhamilto

rhamilto:
I believe the issues is because the all the text "Are you sure you want to delete the project 'My Project'?" is now bold now that it's part of the modal-title rather than a separate node within the content (it was an h1 with strong around the project name).  A possible fix would be to add the project name to the last sentence?

Type **My Project** to confirm deletion.

For reference, Github's looks like:
![screen shot 2018-03-01 at 10 29 27 am](https://user-images.githubusercontent.com/895728/36853082-7bac958c-1d3b-11e8-9669-3ab35a0d89f6.PNG)

benjaminapetersen:
Yup. Github even used to be a bit more clear than that. Currently, ours is too much of a plain modal, it no longer feels like an "oh crap, better not screw this up" moment.

beanh66:
@rhamilto I like your idea to change that last sentence to "Type **My Project** to confirm deletion."

rhamilto:
Fixed by origin-web-common version bump in #2905

PR #2905, opened 2018-03-20
Remove hawtio-core dependency

TheRealJon:
Bump origin-web-catalog and origin-web-common dependencies so that hawtio-core and hawtio-extension-service dependencies can be removed.


bower.json
```diff
@@ -64,7 +62,7 @@
     "jquery": "3.2.1",
     "lodash": "4.17.4",
     "matchHeight": "0.7.2",
-    "origin-web-common": "3.10.0-alpha.0"
+    "origin-web-common": "3.10.0-alpha.1"
```
spadgett: I think this resolution can be removed now.

rhamilto:
Fixes #2864

PR #2906, opened 2018-03-21
Bug 1558863 - Fix bug editing empty config map

spadgett:
Don't push items onto an undefined config map `data` array.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1558863

/assign @benjaminapetersen


spadgett:
WebDriver flake

/retest

PR #2908, opened 2018-03-21
Disable autoprefixer's removal of outdated prefixes

rhamilto:
in order to fix bug where autoprefixer is removing -webkit-box-orient
declarations needed for line truncation.  See https://github.com/postcss/autoprefixer#outdated-prefixes for details.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -236,7 +236,10 @@ module.exports = function (grunt) {
     postcss: {
       options: {
         processors: [
-          require('autoprefixer')({browsers: ['last 1 version']})
+          require('autoprefixer')({
+            browsers: ['last 1 version'],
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -236,7 +236,7 @@ module.exports = function (grunt) {
     postcss: {
       options: {
         processors: [
-          require('autoprefixer')({browsers: ['last 1 version']})
+          require('autoprefixer')({browsers: ['last 1 version'], remove: false})
```
TheRealJon: For the sake of readability, can we spread this object out over a few lines?

rhamilto:
Replaced by https://github.com/openshift/origin-web-console/pull/2910

PR #2909, opened 2018-03-22
[enterprise-3.9] Bug 1558863 - Fix bug editing empty config map

spadgett:
Don't push items onto an undefined config map `data` array.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1558863

Backport of #2906

/assign @benjaminapetersen 
/hold


spadgett:
flake #1685
/retest

PR #2910, opened 2018-03-22
Remove autoprefixer as we're not benefitting from it

rhamilto:
Since we’re not benefitting from it (removing it has no bearing on
dist), I suggest we remove it as a means of simplifying our
dependencies.  Less dependencies to install and less tasks to run
should mean faster builds.

Replaces https://github.com/openshift/origin-web-console/pull/2908

cc: @TheRealJon

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -232,23 +232,6 @@ module.exports = function (grunt) {
       server: '.tmp'
     },
 
-    // Add vendor prefixed styles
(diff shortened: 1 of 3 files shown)


TheRealJon:
Can we remove grunt-postcss along with this?

rhamilto:
> Can we remove grunt-postcss along with this?

Good call.  Will update the PR.

PR #2914, opened 2018-03-23
Load JS dependencies in correct order

### LINKED ISSUE #2913: Runtime errors loading built console
The console does not load due to these errors:

```
vendor.js:10909 Uncaught TypeError: ace.define(...) is not a function
    at vendor.js:10909
(anonymous) @ vendor.js:10909
templates.js:1 Uncaught ReferenceError: angular is not defined
    at templates.js:1
(anonymous) @ templates.js:1
scripts.js:574 Uncaught ReferenceError: angular is not defined
    at scripts.js:574
(anonymous) @ scripts.js:574
```

/assign
/kind bug
/priority p0

spadgett:
Closes #2913
/assign @TheRealJon


TheRealJon:
This was probably my fault. I didn't realize that wiredep used the order from bower.json. I ordered them alphabetically. I thought that wiredep was smart enough to order imports correctly regardless of the bower.json order.

spadgett:
Yeah I'd have guessed that, too. It should be a moot point though when we switch to yarn.

We should  start using `grunt:serve-dist` in our merge queue. We've had several bugs only in built dist in the past that would have been caught by of integration tests.

TheRealJon:
Agreed on the `grunt serve:dist`. I forgot to run this locally, so it would have been nice to have our CI catch it.

spadgett:
WebDriver flake

/retest

benjaminapetersen:
+1 to grunt serve dist as well, ideally we wouldn't have to run that locally & can let CI until it catches a bug 😄

PR #2916, opened 2018-03-23
Handle UTF-8 BOM when decoding secrets for display

spadgett:
UTF-8 BOM can show up when decoding secret data using `window.atob()`.
Prevent this by use `base64.decode()` when decoding for display, which
handles the BOM.

Note: We should not use `base64.encode()` to encode the secret because
it will not correctly handle binary data. Since we're only decoding the
secret to display it and we don't display data with non-printable
characters, there's no danger of losing data using `base64.decode()`.

Before:

openshift web console 2018-03-23 12-06-58

After:

openshift web console 2018-03-23 12-07-36

/assign @benjaminapetersen

### DIFF
--- a/app/scripts/services/secrets.js
+++ b/app/scripts/services/secrets.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module("openshiftConsole")
-  .factory("SecretsService", function($filter, Logger, NotificationsService){
+  .factory("SecretsService", function($filter, base64, Logger, NotificationsService){
     var isNonPrintable = $filter('isNonPrintable');
 
     var groupSecretsByType = function(secrets) {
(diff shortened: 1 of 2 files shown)


spadgett:
OK I've dug into this more: the characters are UTF-8 BOM, which can appear in some text files.

https://stackoverflow.com/questions/2223882/whats-different-between-utf-8-and-utf-8-without-bom

I've updated the description to more accurately reflect the change. Note we should be careful using `base64.encode()` for secret data since it can be binary data. I've confirmed that it will mangle some files like images. But it's safe to use `base64.decode()` for display.

I've also confirmed that the secrets we create match what the CLI does for binary files.

/hold cancel

PR #2921, opened 2018-03-27
Bug 1560927 - Limit file size the console will read

spadgett:
Don't try to read any file larger than 5 MiB. There is no limit in the
HTML5 file API, but the browser can become unresponsive handling larger
files. This limit is larger than what's allowed for secrets and config
maps (1 MB).

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1560927

/assign @benjaminapetersen 

screen shot 2018-03-27 at 9 08 11 am

### DIFF
--- a/app/scripts/directives/oscFileInput.js
+++ b/app/scripts/directives/oscFileInput.js
@@ -1,7 +1,7 @@
 'use strict';
 
 angular.module('openshiftConsole')
-  .directive('oscFileInput', function($filter, Logger) {
+  .directive('oscFileInput', function($filter, Logger, NotificationsService) {
     return {
       restrict: 'E',
       scope: {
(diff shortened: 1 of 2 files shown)


app/scripts/directives/oscFileInput.js
```diff
@@ -188,6 +188,17 @@ angular.module('openshiftConsole')
         }
 
         function addFile(file) {
+          // If the file is larger than 5 MiB, don't try to upload. Note that secrets and config
+          // maps have an even smaller limit (1 MB).
+          if (file.size > (5 * 1024 * 1024)) {
```
benjaminapetersen: Only thought would be to take this `(5 * 1024 * 1024)` and put it in a `.constant("FILE_UPLOAD_MAX",5 * 1024 * 1024)` in `app.js`.  Otherwise lgtm.

spadgett: Since we're only using this in one place, I think it's better to keep it near where it's used.

benjaminapetersen: Works for me.

PR #2923, opened 2018-03-27
[enterprise-3.9] Overview support for daemon sets

spadgett:
https://bugzilla.redhat.com/show_bug.cgi?id=1551667

Backport of #2877

/assign @benjaminapetersen 
/hold

@benjaminapetersen The only difference from #2877 is that we are inlining the daemon set version to avoid cutting a common release for each branch we need to backport to. (We're using preferred version in master.)


spadgett:
flake #2105

/retest

spadgett:
> ERROR - Unable to start a WebDriver session.

/retest

PR #2925, opened 2018-03-28
add mobile-client-config component

philbrookes:
Mobile clients have configurations made available to them based on the services available that they are not excluded from.

This work exposes the client configuration conveniently for the ease of the mobile developer. The copy-to-clipboard element is re-used to make copying the configuration out very simple.

The following screenshot gives an idea of what this looks like:
screenshot from 2018-03-28 15-00-12

### DIFF
--- a/app/scripts/directives/overview/mobileClientRow.js
+++ b/app/scripts/directives/overview/mobileClientRow.js
@@ -3,7 +3,6 @@
 (function () {
   angular.module('openshiftConsole').component('mobileClientRow', {
     controller: [
-      '$scope',
       '$filter',
       '$routeParams',
       'APIService',
@@ -11,7 +10,7 @@
       'DataService',
       'ListRowUtils',
       'Navigate',
-      'ProjectsService',
+      'ServiceInstancesService',
       MobileAppRow,
     ],
     controllerAs: 'row',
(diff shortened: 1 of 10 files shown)


app/scripts/directives/overview/mobileClientRow.js
```diff
@@ -22,12 +21,27 @@
...
+    var serviceInstancesVersion = APIService.getPreferredVersion('serviceinstances');
+    var isServiceInstanceReady = $filter('isServiceInstanceReady');
+    var isMobileService = $filter('isMobileService');
     row.installType = '';
 
     _.extend(row, ListRowUtils.ui);
 
+    row.$onInit = function() {
+      row.context = {namespace: _.get(row, 'apiObject.metadata.namespace')};
+      DataService.watch(serviceInstancesVersion, row.context, function (serviceinstances){
```
spadgett: nit: we use this style for functions (here and below)

```js
function(serviceInstances) {
```

app/scripts/directives/overview/mobileClientRow.js
```diff
@@ -22,12 +21,27 @@
...
+        row.services = _.filter(serviceinstances.by('metadata.name'), function(serviceInstance){
+          return ServiceInstancesService.fetchServiceClassForInstance(serviceInstance)
```
spadgett: This is going to repeatedly request service classes anytime the instance changes. We should only request it one time.

PR #2929, opened 2018-03-29
Virtual Machines in Overview

jniederm:
Offline Kubevirt Virtual Machines [1] are listed in Overview page.

This commit:
* lists offline VMs
* allows to start, stop, restart (by deleting corresponding
  virtualmachine entity) and delete offline VM
* shows state and guest operating system of the offline VM
* hides unassigned pods ("monopods") running the VM of the offline VM
* allows to jump to pod running the VM (by clicking offline VM name) if
the offline VM is running

[1]: https://github.com/kubevirt/kubevirt/blob/35390ffcaa4d128d95466d2b1a438657c1eae249/docs/devel/offline-virtual-machine.md

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -31,6 +31,7 @@ angular.module('openshiftConsole').controller('OverviewController', [
   'ResourceAlertsService',
   'RoutesService',
   'ServiceInstancesService',
+  'KubevirtVersions',
   OverviewController
 ]);
 
@@ -63,7 +64,8 @@ function OverviewController($scope,
                             PromiseUtils,
                             ResourceAlertsService,
                             RoutesService,
-                            ServiceInstancesService) {
+                            ServiceInstancesService,
+                            KubevirtVersions) {
   var overview = this;
   var limitWatches = $filter('isIE')();
   var DEFAULT_POLL_INTERVAL = 60 * 1000; // milliseconds
@@ -206,7 +208,8 @@ function OverviewController($scope,
            _.size(overview.daemonSets) +
            _.size(overview.monopods) +
            _.size(overview.state.serviceInstances) +
-           _.size(overview.mobileClients);
+           _.size(overview.mobileClients) +
+           _.size(overview.offlineVirtualMachines);
   };
 
   // The size of all visible top-level items after filtering.
@@ -219,7 +222,8 @@ function OverviewController($scope,
            _.size(overview.filteredDaemonSets) +
            _.size(overview.filteredMonopods) +
            _.size(overview.filteredServiceInstances) +
-           _.size(overview.filteredMobileClients);
+           _.size(overview.filteredMobileClients) +
+           _.size(overview.filteredOfflineVirtualMachines);
   };
 
   // Show the "Get Started" message if the project is empty.
@@ -407,6 +411,7 @@ function OverviewController($scope,
     overview.filteredPipelineBuildConfigs = filterItems(overview.pipelineBuildConfigs);
     overview.filteredServiceInstances = filterItems(state.orderedServiceInstances);
     overview.filteredMobileClients = filterItems(overview.mobileClients);
+    overview.filteredOfflineVirtualMachines = filterItems(overview.offlineVirtualMachines);
     overview.filterActive = isFilterActive();
     updateApps();
     updateShowGetStarted();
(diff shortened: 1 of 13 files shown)


app/scripts/controllers/overview.js
```diff
@@ -692,10 +697,82 @@ function OverviewController($scope,
     return true;
   };
 
+  // part of lodash since 4.0.0
```
spadgett: We're using lodash 4, so you should be able to use `_.keyBy`

app/scripts/controllers/overview.js
```diff
@@ -692,10 +697,82 @@ function OverviewController($scope,
...
+    if (!ownerReferences) {
+      return null;
+    }
+    return _(ownerReferences)
+      .filter(function (reference) {
+        return reference.kind === 'OfflineVirtualMachine';
+      })
+      .map(function (reference) {
+        return reference.uid;
+      })
```
spadgett: You should be able to use the shorthand here:

```js
.filter({ kind: 'OfflineVirtualMachine' })
.map('uid')
```

app/scripts/controllers/overview.js
```diff
@@ -692,10 +697,82 @@ function OverviewController($scope,
...
+    var ovmIdToVm = _(overview.virtualMachines)
+      .values()
+      .filter(function (vm) {
+        return !!getOvmId(vm)
+      })
+      .reduce(byKey(function (vm) {
+        return getOvmId(vm);
+      }), {});
+    var podsOfOvms = [];
+    _.forEach(overview.offlineVirtualMachines, function (ovm) {
```
spadgett: nit: we try to always use `_.each` for consistency (so we're not using `_.forEach` in some places and `_.each` in others)

app/scripts/directives/overview/virtualMachineRow.js
```diff
@@ -0,0 +1,178 @@
...
+      scope: {
+        action: '&',
+        enabled: '@'
+      },
+      link: function ($scope, $element, $attrs, ctrl, transcludeFn) {
+        var currentElement = $element;
+
+        function onAttrChange() {
+          var template = $scope.enabled === 'true'
+            ? '<li><a ng-click="action()" ng-transclude></a></li>'
```
spadgett: Needs an `href=""` for accessibility

app/scripts/directives/overview/virtualMachineRow.js
```diff
@@ -0,0 +1,178 @@
...
+    return {
+      scope: {
+        ovm: '<'
+      },
+      controller: controller,
+      templateUrl: 'views/overview/_vm-status.html'
+    }
+  });
+
+  angular.module('openshiftConsole').directive('optionalLink', function () {
```
spadgett: Some of these directives are generic and probably should be split out into separate files.

spadgett:
You'll want to run `grunt build` and include the dist in your commit.

jniederm:
Screenshot:
image

app/scripts/controllers/overview.js
jniederm: done

app/scripts/controllers/overview.js
jniederm: done

app/scripts/directives/overview/virtualMachineRow.js
jniederm: `optionalLink` and `dropdownItem` extracted to separate files

app/scripts/directives/overview/virtualMachineRow.js
jniederm: done

app/scripts/controllers/overview.js
jniederm: done

jniederm:
> You'll want to run grunt build and include the dist in your commit.

done

app/scripts/directives/dropdownItem.js
```diff
@@ -0,0 +1,27 @@
...
+    transclude: true,
+    scope: {
+      action: '&',
+      enabled: '@'
+    },
+    link: function ($scope, $element, $attrs, ctrl, transcludeFn) {
+      var currentElement = $element;
+
+      function onAttrChange() {
+        var template = $scope.enabled === 'true' ?
```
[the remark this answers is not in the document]
spadgett: I think it's expecting the `?` on the same line to avoid potential problems with automatic semicolon insertion in JavaScript. (Not a problem here, but could be for `return` statements and a few other places.) I usually do something like

```js
condition ? thenExpr
          : elseExpr;
```

which I think the linter allows

spadgett:
cc @openshift/team-ux-review

beanh66:
@spadgett Do you have thoughts around how virtual machines should show up on the Overview when the "List by Application" is selected? Seems like the current screenshot is really showing the "List by Resource" view. 

@jniederm When you get a chance can you provide the latest screenshots of the vm expanded items when in "List by Resource" and "List by Application" modes? Preferably with additional items in the project so we have an idea how it realistically looks?

spadgett:
@beanh66 Provisioned services and mobile clients aren't currently displayed by application. They're listed separately at the bottom. So this is consistent with those resources.

ncameronbritt:
Would it make sense to have a separate section for VMs when viewing by
application, as we do for provisioned services?

beanh66:
@ncameronbritt We do already have that section added, I just wanted to determine if it should exist for both view types (by application and by resource). Sounds like yes, which I believe is the direction @jniederm was going! 

@spadgett To be sure I understand....Pods today are exposed in the expanded state of a deployment config on the overview, giving options to scale up or down, etc. In this case when it's a VM, that scaling option will not be available. I guess the question is, are VMs associated with a deployment config at all?

beanh66:
@jniederm Are you able to add the actions inline in the expanded state in addition to the kebab menu? Also wondering if we can show the VM uptime in that area. See below: 
podvm-2

Do you expect to add more details below the Operating System? Seems like some of the pod details should come through to this expanded state, but not sure if that's already in the works.

jniederm:
@beanh66 

> Are you able to add the actions inline in the expanded state in addition to the kebab menu?

Lifecycle actions are planned to be added to the expandable area.

> Also wondering if we can show the VM uptime in that area

AFAIK we don't have reliable data to show uptime.

> Seems like some of the pod details should come through to this expanded state

Info shown below is low level purely pod-related data so we don't plan to show it in VM expandable area.
image

serenamarie125:
@jniederm can you verify about uptime?  IMO, this is something that the user would want to see.  Happy to talk about a phased approach.

ncameronbritt:
Just looking at the screenshot with the Image, Mount, etc details, the entry for Memory needs units. Looks like it's probably bytes, but is that how users are accustomed to seeing this? I would think that using MB or GB (or MiB/GiB) would make it easier for a user to make sense of it.

PR #2938, opened 2018-04-04
Add vm-images to origin-console

rthallisey:


rhamilto:
We should probably avoid including text in any of these "icons" (e.g., free_bsd.svg, rhel.svg, rhel_atomic.svg).  It become illegible when scaled down to a 60px square.  Red Hat brand created icons for the other products for this exact purpose (e.g., [3scale](https://github.com/openshift/origin-web-console/blob/master/dist/images/logos/3scale.svg).  Presumably we'd want to do the same here?

rhamilto:
@rthallisey, did you see my previous comment?

rhamilto:
> Can you explain what I need to change?

Words inside of images like these do not render well.  In most cases, they're not legible because the images render within the catalog at 60px x 60px at their smallest.  So for free_bsd.svg, you'd want to eliminate the wordmark (FreeBSD) *assuming* that is ok per their brand guidelines. For rhel.svg, rhel_atomic.svg, Red Hat brand should probably be engaged to create icons similar to the other products ones they previously created (e.g., [3scale](https://github.com/openshift/origin-web-console/blob/master/dist/images/logos/3scale.svg)).  @jennyhaines, do you know who in brand can help with the creation of new product icons for RHEL and RHEL Atomic?

jennyhaines:
@rhamilto - Libby Levi created the last set of brand-approved product icons, so she would likely be the one to handle the creation of new product icons. Libby does know that a new RHEL icon needs to be created, but the brand team is preoccupied at the moment with the rebranding project. Based on what I've learned about their timeline, they are looking at wrapping up the brand creation in early July.

rhamilto:
@rthallisey, given brand won't be able to create an icon until July, how would you like to proceed?  One option is to remove rhel.svg and rhel_atomic.svg from this PR so the others can be added without RHEL.  Another is to wait to add all of them until a RHEL icon can be created by brand.  I do not recommend adding the current rhel.svg and rhel_atomic.svg as is because of the legibility issue.  A standalone shadowman icon does exist in https://rawgit.com/openshift/openshift-logos-icon/master/demo.html, but brand has strongly discouraged using shadowman without the word mark.

jniederm:
Cherry-picked to demo branch as https://github.com/jniederm/origin-web-console/commit/abbc96499e005c8a3a6f4d3d178aac36e1279fef

rthallisey:
@rhamilto I'll remove the rhel icons and create and issue to re add them in july.  How can I track the progress of the rhel icons?

rhamilto:
> How can I track the progress of the rhel icons?

^ @jennyhaines

rhamilto:
@rthallisey, it looks like your dependencies are out of date.  Rebase on master, run `hack/clean-deps.sh`, `hack/install-deps.sh`, and then `grunt build` to resolve.  You should only see files you added or changed and their resulting dist (dist/images/logos/* and dist/scripts/scripts.js).

jennyhaines:
@serenamarie125 - I know that Libby Levi said before that she could work on RHEL/RHEL Atomic product icons after the open brand push. Looking at the notional schedule for the branding work, it seems that she'll be totally busy until mid July (the release of the new logo/brand). Mid July is (of course!) when I'm due. Is there someone within the OpenShift team that could coordinate this icon creation with Libby once the open brand project if complete?

jennyhaines:
@rhamilto @rthallisey FYI^

app/scripts/constants.js
```diff
@@ -605,13 +605,16 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
...
+    'icon-debian': 'debian.svg',
     'icon-decisionserver': 'decisionserver.svg',
     'icon-django': 'django.svg',
     'icon-dotnet': 'dotnet.svg',
     'icon-drupal': 'drupal.svg',
     'icon-eap': 'eap.svg',
     'icon-elastic': 'elastic.svg',
     'icon-erlang': 'erlang.svg',
+    'icon-fedora': 'fedora.svg',
+    'icon-free-bsd': 'free-bsd.svg',
```
rhamilto: I noticed when looking for a single-color FreeBSD logo that per https://www.freebsdfoundation.org/about/project/, FreeBSD is one word.  Mind changing "icon-free-bsd" to "icon-freebsd" and "free-bsd.svg" to "freebsd.svg" and rename the actual image as well?  I used "icon-freebsd" in https://github.com/openshift/openshift-logos-icon/pull/16, and the class names will need to match.

While making that change, can you also squash to a single commit?

Thanks, @rthallisey.

jennyhaines:
Update on icon creation: @rhamilto @serenamarie125 @rthallisey - I just spoke with branding about this, and they are okay with starting on this RHEL/RHEL Atomic icon work earlier than July. (Delivery date TBD.) Jessica Burkitt will be working on the design, and the kickoff for this work will start some time after Summit. We will be bringing in the RHEL Marketing manager for reviews, and I can include all of you on the review meetings if you'd like- (just let me know). I'll definitely include Serena for a secondary UXD contact (incase I need to take maternity leave earlier than expected.)

rthallisey:
@jennyhaines, thanks for pushing this along.  I'll follow up you near the end of June to get a status on the icons.  KubeVirt will need to swap over to the icons before 3.10 goes out the door.

PR #2944, opened 2018-04-10
[enterprise-3.9] Fall back to JSON when PartialObjectMetadataList not available

spadgett:
Avoids an error when the 3.9 console is talking to a 3.10 master.

Bumps origin-web-common to 3.9.2.

/assign @jwforres

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -46,8 +46,8 @@
     "angular-moment": "1.0.0",
     "angular-utf8-base64": "0.0.5",
     "file-saver": "1.3.3",
-    "origin-web-common": "3.9.1",
-    "origin-web-catalog": "3.9.0-rc.0"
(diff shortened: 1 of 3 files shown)


spadgett:
Yeah, it's bootstrap-datepicker, which we're not using directly (patternfly dependency)

```
│ └─┬ patternfly#3.37.4 (4.0.0-rc.1 available)
│   ├─┬ bootstrap#3.3.7 (latest is 4.1.0)
│   │ └── jquery#3.2.1 (3.3.1 available)
│   ├─┬ bootstrap-datepicker#1.8.0
```

Since it's added using `^` in patternfly, we get automatically bumped to 1.8. I'm not sure it's worth pinning.

https://github.com/patternfly/patternfly/blob/master/bower.json#L24

PR #2946, opened 2018-04-12
 Virtual Machine uptime added 

jniederm:
Uptime is based on start time of the 'virt-launcher' container.

Depends on #2945

### DIFF
--- a/app/scripts/directives/overview/virtualMachineRow.js
+++ b/app/scripts/directives/overview/virtualMachineRow.js
@@ -13,6 +13,7 @@
       'Navigate',
       'ProjectsService',
       'KubevirtVersions',
+      'moment',
       VirtualMachineRow
     ],
     controllerAs: 'row',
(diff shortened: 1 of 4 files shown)


app/scripts/directives/overview/virtualMachineRow.js
```diff
@@ -97,6 +99,16 @@
     };
   }
 
+  angular.module('openshiftConsole').filter('podUptime', function () {
+    return function (pod) {
+      var startTime = _.get(pod, 'status.startTime');
```
spadgett: Do we want to get the actual container start time?

I think if the container is crash looping, it will still count as "up" since the pod is running.

jniederm: Fixed, thanks.

app/scripts/directives/overview/virtualMachineRow.js
```diff
@@ -97,6 +99,20 @@
     };
   }
 
+  angular.module('openshiftConsole').filter('podUptime', function () {
+    return function (pod) {
+      var computeContainerStartTime = _(_.get(pod, 'status.containerStatuses'))
+        .filter({ name: "compute" })
```
spadgett: We should give the filter a more specific name since it assumes a specific container name for KubeVirt.

jniederm: done

PR #2955, opened 2018-04-12
Sanitize HTML output for log viewer

spadgett:
Fixes Fixes CVE-2017-7534
Bumps origin-web-common to 3.10.0-alpha.3


spadgett:
Version bump only. Changes approved in https://github.com/openshift/origin-web-common/pull/326

PR #2956, opened 2018-04-16
Bug 1563595 - Update wording when deploying image from private registry with incorrect image pull credentials

jhadvig:
As we shortly discussed, Im updating the working in the deploy-image modal, since we ruled out deleting of any secrets from the modal or any similar logic.

Screen:
download

@spadgett  PTAL

### DIFF
--- a/app/views/directives/deploy-image.html
+++ b/app/views/directives/deploy-image.html
@@ -203,7 +203,7 @@
         Could not load image metadata.
       </h2>
       <p>{{import.error | upperFirst}}</p>
-      <p>The image may not exist or it may be in a secure registry. Check that you have entered the image name correctly or <a href="" ng-click="openCreateWebhookSecretModal()">create an image pull secret</a> with your image registry credentials and try again.</p>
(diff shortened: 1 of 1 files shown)


app/views/directives/deploy-image.html
```diff
@@ -203,7 +203,7 @@ <h2>
         Could not load image metadata.
       </h2>
       <p>{{import.error | upperFirst}}</p>
-      <p>The image may not exist or it may be in a secure registry. Check that you have entered the image name correctly or <a href="" ng-click="openCreateWebhookSecretModal()">create an image pull secret</a> with your image registry credentials and try again.</p>
+      <p>The image may not exist or it may be in a secure registry. Check that you have entered the image name correctly. In case of private image repository, check that your image pull secret credentials are correct or <a href="" ng-click="openCreateWebhookSecretModal()">create an image pull secret</a> with your image registry credentials and try again.</p>
```
spadgett: Since the API server error already says the username/password is incorrect, I don't see much value in restating that. And if we always state it, we'll say it even when that's not the case, whereas the error message from the server will only appear when it's really a problem.

I'd rather close the bug as won't fix.

jhadvig: Seems good to me. Will close PR and bug as well.
Thanks !

PR #2960, opened 2018-04-17
Bug 1550797 - Should remove 'loading..' when browser non-existent image

jhadvig:
Chrry-pick of https://github.com/openshift/origin-web-console/pull/2867/commits/b256cc02ceaee87d836dcb9046a4d10b3bdaaab8 since the issue was is solved only in 3.10 and not in 3.9 due to https://bugzilla.redhat.com/show_bug.cgi?id=1550797

@spadgett PTAL


spadgett:
This is minor enough I think it's better to just change the target release in the Bugzilla bug.

PR #2961, opened 2018-04-17
Bug 1568272 - Fix scaling message for DaemonSet pod donut

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1568272


spadgett:
ping @benjaminapetersen 

It would be good to get this change backported since 3.6-3.9 now also have this bug :(

PR #2965, opened 2018-04-18
 Virtual Machines hidden when Overview listed by pipeline 

jniederm:
Change suggested by Serena (UXD).

Depends on #2963


app/views/overview.html
```diff
@@ -390,7 +390,7 @@ <h2>Mobile Clients</h2>
               </div>
             </div>
             <!-- Show Kubevirt Virtual Machines -->
-            <div ng-if="overview.filteredOfflineVirtualMachines | size">
+            <div ng-if="(overview.filteredOfflineVirtualMachines | size) && overview.viewBy !== 'pipeline' ">
```
spadgett: I think to be consistent with provisioned services, this should be...

```html
<div ng-if="(overview.filteredOfflineVirtualMachines | size) && !overview.hidePipelineOtherResources">
```

Also do you mind making the same change to mobile clients just above?

spadgett:
Looks like this also includes the change from #2963?

app/views/overview.html
jniederm: Sure, I can do the change for mobile clients as well.

If did the Bool logic exercise correnctly then `!overview.hidePipelineOtherResources` can be expanded as 

```js
overview.viewBy !== 'pipeline' 
|| 
(!overview.filterActive && !_.isEmpty(overview.pipelineBuildConfigs))
``` 

Such clause is true event for following setup:

```js
overview.viewBy === 'pipeline'
overview.filterActive is false-ish
overview.pipelineBuildConfigs is not empty
```

So under certain conditions VMs will be shown with "pipeline" sorting after all. Is it ok?

spadgett: Yeah, I think so. That's consistent with other resources we have. We should revisit it for everything if we want different behavior (although I prefer how it works currently, otherwise the pipeline view is not as useful).

jniederm: done

PR #2971, opened 2018-04-19
[enterprise-3.6] Fix scaling message for DaemonSet pod donut

spadgett:
3.6 backport of #2961
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1568272


spadgett:
> is the change to dist/styles/main.css expected

Yeah, we purposely leave the openshift-font-icons dependency a range to pick up new icons. That way we can update templates to use new icons freely, and they work for older versions

PR #2981, opened 2018-04-30
Change routes help link on browse page

damemi:
Relevant BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1502540#c3

This just changes the Routes page to link to the main [routes](https://github.com/openshift/origin-web-console/blob/d7ca56bbd4561bfc9b1fe26ac26b7913d5e87ecc/app/scripts/constants.js#L56) doc page instead of the [route-types](https://github.com/openshift/origin-web-console/blob/d7ca56bbd4561bfc9b1fe26ac26b7913d5e87ecc/app/scripts/constants.js#L25) sub-header, at the recommendation of QE based on the context of the link.

### DIFF
--- a/app/views/browse/service.html
+++ b/app/views/browse/service.html
@@ -91,7 +91,7 @@
                   <traffic-table ports-by-route="portsByRoute" routes="routesForService" services="services" show-node-ports="showNodePorts" custom-name-header="'Route'"></traffic-table>
                 </div>
                 <p>
-                  Learn more about <a ng-href="{{'route-types' | helpLink}}" target="_blank">routes</a> and <a ng-href="{{'services' | helpLink}}" target="_blank">services</a>.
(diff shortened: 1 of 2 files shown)


damemi:
Ran grunt build and pushed, but a lot of files were affected so I'm interested if I messed something up. Will see if tests pass

dist.java/java/404.html
```diff
@@ -38,10 +38,10 @@ <h4>Connected to <span ng-bind="containerName"></span></h4>
       <div ng-include src="viewPartial"></div>
     </div>
 
-    <script src="libs-301d12fa8a.js"></script>
+    <script src="libs-3239db0851.js"></script>
```
spadgett: Looks like you'll pulling in an old JVM console. You'll want to run

```
$ hack/clean-deps.sh && hack/install-deps.sh && grunt build
```

Please squash commits when finished :)

damemi: Thanks! That looks like it fixed it up

PR #2992, opened 2018-05-10
Mobile client view

witmicko:
Adds Mobile client view component, make use of directive in #2925. 
NOTE: needs to be verified after this change is released https://github.com/openshift/origin-web-common

path:
`.../project/myproject/browse/mobile-clients/{{mobile client id}}?tab=configuration`
Screenshot:
image


app/scripts/constants.js
```diff
@@ -687,5 +687,10 @@ angular.extend(window.OPENSHIFT_CONSTANTS, {
     'openshift-',
     'kubernetes-',
     'kube-'
-  ]
+  ],
+  MOBILE_CLIENT_VERSION: {
+    group: 'mobile.k8s.io',
+    version: 'v1alpha1',
+    resource: 'mobileclients'
```
spadgett: This should probably go in origin-web-common preferred versions.

app/scripts/controllers/mobileClients.js
```diff
@@ -0,0 +1,76 @@
...
+    Constants,
+    DataService,
+    ProjectsService
+  ) {
+      var ctrl = this;
+      var watches = [];
+      ctrl.projectName = $routeParams.project;
+      ctrl.emptyMessage = 'Loading...';
+      ctrl.alerts = {};
+      ctrl.redirectUrl = '/project/' + ctrl.projectName + '/overview';
```
spadgett: Should use `Navigate` service to build the overview URL

app/styles/_mobile-clients.less
```diff
@@ -0,0 +1,9 @@
+.mobile-clients-view {
+
+  .copy-to-clipboard {
```
spadgett: We should do this as one declaration if there are no other rules for `mobile-clients-view`

```less
.mobile-clients-view .copy-to-clipboard {
  /* ... */
}
```

app/views/browse/mobile-clients.html
```diff
@@ -0,0 +1,68 @@
...
+      <div>
+        <h1 class="contains-actions">
+          <div class="pull-right dropdown" ng-if="ctrl.mobileClient">
+            <button type="button" class="dropdown-toggle btn btn-default actions-dropdown-btn hidden-xs" data-toggle="dropdown">
+              Actions
+              <span class="caret" aria-hidden="true"></span>
+            </button>
+            <a href=""
+               class="dropdown-toggle actions-dropdown-kebab visible-xs-inline"
+               data-toggle="dropdown"><i class="fa fa-ellipsis-v"></i><span class="sr-only">Actions</span></a>
```
spadgett:
```html
<i class="fa fa-ellipsis-v" aria-hidden="true">
```

app/scripts/constants.js
witmicko: @spadgett I`ll update once this gets approved https://github.com/openshift/origin-web-common/pull/320

spadgett: Yikes, sorry that has sat so long :( Looks like that is right around when 3.10 feature freeze happened. We'll get that integrated as soon as we branch for 3.11.

witmicko: no worries at all, fantastic, thank you

PR #2993, opened 2018-05-11
Bug 1577359 - Don't show Create Project button if you cannot create a…

Oats87:
Bug 1577359 - Don't show Create Project button if you cannot create a project

Add the correct ng-if=\"canCreate\" to the <div> that surrounds the Create Project button in the template.


spadgett:
Sorry looks like we both made the same fix... (#2994)

Note that the change needs to be made in the app folder. The dist folder is generated using `grunt build`

Issue #2999, opened 2018-05-14
jshint seems to be broken


benjaminapetersen:
Conversation that spawned the issue:  https://github.com/openshift/origin-web-console/pull/2998/files/c806e7fbec2a71c00fdb1e8ef654543b9d10b43a#discussion_r188018947

I don't think we want to manually review for `;` & other things like this, our linter should pick it up.

benjaminapetersen:
@spadgett

benjaminapetersen:
Not sure how high to prioritize.  Not important until we miss something that causes an error in a random browser. :/

PR #3000, opened 2018-05-14
[WIP] Fix jshint task config to notice nested directories

benjaminapetersen:
Fixes issue #2999

screen shot 2018-05-14 at 2 50 03 pm 1

We don't have a ton of problems at the moment, and won't so long as we don't do a lot of nested directories...

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -199,8 +199,8 @@ module.exports = function (grunt) {
       all: {
         src: [
           'Gruntfile.js',
-          '<%= yeoman.app %>/scripts/{,*/}*.js',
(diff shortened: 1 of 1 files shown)


benjaminapetersen:
If no objections, I can put it on my list to scrub all the problems && remove the `[WIP]`.

spadgett:
That would do it :/

How many errors?

spadgett:
Let's go ahead and fix. We probably want to wait for 3.11 unless you find one that results in a user-visible bug. Thanks for catching this!

benjaminapetersen:
Yeah, I agree, I don't think there is a rush, just glad to catch early enough to avoid too much drift...

PR #3004, opened 2018-05-24
Added PVC Expand functionality in Storage tab

cloudbehl:
[trello Increase size of pv storage](https://trello.com/c/WhnpxDH8/1138-increase-size-of-pv-storage)

This patch will add another button "Expand PVC” which should appear in the Actions menu of the details page for PVCs that support expansion.

1. If the allowVolumeExpension is set to true then on the "Expand PVC" will appear.
2. Set default number greater one(+1) than the previously allocated capacity. (if the allocated capacity is 6GB currently and when you open the expand modal it will show you to expand to 7GB).
3. The dropdown  has these values(GB, Gib, TB, Tib)

**Validation on textbox**
1. The text entered should be a positive number.
2. It should be greater than the allocated capacity.
3. The textbox should not be empty.

Once then the number in the textbox is valid then you can click on expand and the toast notification will say that the Persistent volume claim 'demo' expand request has been submitted and the user can wait for an expansion to happen.
If the requested capacity is greater than the bricks available capacity it will generate an event in the events tab then we "Failed to allocate volume expansion: Minimum brick size limit reached. Out of space".

### DIFF
--- a/app/scripts/controllers/modals/editPvcModal.js
+++ b/app/scripts/controllers/modals/editPvcModal.js
@@ -1,0 +1,129 @@
+'use strict';
+
+angular.module('openshiftConsole')
+  .controller('EditPvcModalController', function(APIService, DataService, $filter, LimitRangesService, QuotaService, $scope, $uibModalInstance) {
+
+    var limitRangesVersion = APIService.getPreferredVersion('limitranges');
+    var resourceQuotasVersion = APIService.getPreferredVersion('resourcequotas');
+    var appliedClusterResourceQuotasVersion = APIService.getPreferredVersion('appliedclusterresourcequotas');
+
+    var amountAndUnit = $filter('amountAndUnit');
+    var usageWithUnits = $filter('usageWithUnits');
+    var usageValue = $filter('usageValue');
+
+    var allocatedAmountAndUnit = amountAndUnit($scope.pvc.spec.resources.requests['storage']);
+
+    $scope.projectName = $scope.pvc.metadata.namespace;
+    $scope.typeDisplayName = $filter('humanizeKind')($scope.pvc.metadata.name);
+    $scope.claim = {}
+    $scope.claim.capacity = Number(allocatedAmountAndUnit[0]);
+    $scope.claim.unit = allocatedAmountAndUnit[1];
+    $scope.disableButton = true;
+
+    $scope.currentCapacityUnits = angular.copy($scope.claim);
+
+    $scope.units = [{
+      value: "Mi",
+      label: "MiB"
+    }, {
+      value: "Gi",
+      label: "GiB"
+    }, {
+      value: "Ti",
+      label: "TiB"
+    }, {
+      value: "M",
+      label: "MB"
+    }, {
+      value: "G",
+      label: "GB"
+    }, {
+      value: "T",
+      label: "TB"
+    }];
+
+    $scope.groupUnits = function(unit) {
+      switch (unit.value) {
+      case 'Mi':
+      case 'Gi':
+      case 'Ti':
+        return 'Binary Units';
+      case 'M':
+      case 'G':
+      case 'T':
+        return 'Decimal Units';
+      }
+
+      return '';
+    };
+
+    $scope.expand = function() {
+      $scope.updatedCapacity = $scope.claim.capacity + $scope.claim.unit;
+      $uibModalInstance.close($scope.updatedCapacity);
+    };
+
+    $scope.cancel = function() {
+      $uibModalInstance.dismiss('cancel');
+    };
+
+    var validateLimitRange = function() {
+      // Use usageValue filter to normalize units for comparison.
+      var value = $scope.claim.capacity && usageValue($scope.claim.capacity + $scope.claim.unit);
+      var currentValue = $scope.currentCapacityUnits.capacity && usageValue($scope.currentCapacityUnits.capacity + $scope.currentCapacityUnits.unit);
+      var min = _.has($scope, 'limits.min') && usageValue($scope.limits.min);
+      var max = _.has($scope, 'limits.max') && usageValue($scope.limits.max);
+      var minValid = true;
+      var maxValid = true;
+      var requestValid = true;
+
+      minValid = value >= min;
+      maxValid = value <= max;
+      requestValid = value > currentValue;
+
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('limitRangeMin', minValid);
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('limitRangeMax', maxValid);
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('checkCurrentCapacity', requestValid);
+      $scope.expandPersistentVolumeClaimForm.capacity.$touched =  true;
+    };
+
+    var validateQuota = function() {
+      var newValue = $scope.claim.capacity && usageValue($scope.claim.capacity + $scope.claim.unit);
+      var oldValue = $scope.currentCapacityUnits.capacity && usageValue($scope.currentCapacityUnits.capacity + $scope.currentCapacityUnits.unit);
+      var outOfClaims = QuotaService.isAnyStorageQuotaExceeded($scope.quotas, $scope.clusterQuotas);
+      var willExceedStorage = QuotaService.willRequestExceedQuota($scope.quotas, $scope.clusterQuotas, 'requests.storage', (newValue - oldValue));
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('willExceedStorage', !willExceedStorage);
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('outOfClaims', !outOfClaims);
+    };
+
+    DataService.list(limitRangesVersion, { namespace: $scope.projectName }, function(limitRangeData) {
+      var limitRanges = limitRangeData.by('metadata.name');
+      $scope.disableButton = false;
+      if (_.isEmpty(limitRanges)) {
+        return;
+      }
+
+      $scope.limits = LimitRangesService.getEffectiveLimitRange(limitRanges, 'storage', 'PersistentVolumeClaim');
+      if ($scope.limits.min && $scope.limits.max) {
+        var minUsage = usageValue($scope.limits.min);
+        var maxUsage = usageValue($scope.limits.max);
+        if (minUsage === maxUsage) {
+          var requiredAmountAndUnit = amountAndUnit($scope.limits.max);
+          $scope.claim.capacity = Number(requiredAmountAndUnit[0]);
+          $scope.claim.unit = requiredAmountAndUnit[1];
+          $scope.capacityReadOnly = true;
+        }
+      }
+
+      $scope.$watchGroup(['claim.capacity', 'claim.unit'], validateLimitRange);
+    });
+
+    DataService.list(resourceQuotasVersion, { namespace: $scope.projectName }, function(quotaData) {
+      $scope.quotas = quotaData.by('metadata.name');
+      $scope.$watchGroup(['claim.capacity', 'claim.unit'], validateQuota);
+    });
+
+    DataService.list(appliedClusterResourceQuotasVersion, { namespace: $scope.projectName }, function(quotaData) {
+      $scope.clusterQuotas = quotaData.by('metadata.name');
+    });
+
+  });
--- a/app/scripts/controllers/persistentVolumeClaim.js
+++ b/app/scripts/controllers/persistentVolumeClaim.js
@@ -30,14 +30,29 @@ angular.module('openshiftConsole')
       }
     ];
 
+    $scope.storageClassesVersion = APIService.getPreferredVersion('storageclasses');
     $scope.pvcVersion = APIService.getPreferredVersion('persistentvolumeclaims');
     $scope.eventsVersion = APIService.getPreferredVersion('events');
+    $scope.isExpansionAllowed = false;
 
+    var storageClass = $filter('storageClass');
     var watches = [];
+
+    var updateStorageClass = function(pvc) {
+      var storageClassName = storageClass($scope.pvc);
+      if (storageClassName !== _.get(storageClass, 'metadata.name')) {
+        DataService.get($scope.storageClassesVersion, storageClassName, {}).then(function(sc) {
+          $scope.isExpansionAllowed = (!sc || sc.allowVolumeExpansion) && pvc.status.phase === 'Bound';
+        });
+      }
+    };
 
     var pvcResolved = function(pvc, action) {
       $scope.pvc = pvc;
       $scope.loaded = true;
+
+      updateStorageClass(pvc);
+
       if (action === "DELETED") {
         $scope.alerts["deleted"] = {
           type: "warning",
@@ -65,9 +80,8 @@ angular.module('openshiftConsole')
           };
         });
 
-      $scope.$on('$destroy', function(){
+      $scope.$on('$destroy', function() {
         DataService.unwatchAll(watches);
       });
-
     }));
 });
--- a/app/scripts/directives/editPvc.js
+++ b/app/scripts/directives/editPvc.js
@@ -1,0 +1,72 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .directive("editPvc",
+             function($uibModal,
+                      $filter,
+                      $routeParams,
+                      APIService,
+                      DataService,
+                      ProjectsService,
+                      NotificationsService,
+                      Logger) {
+
+    return {
+      restrict: "E",
+      scope: {
+        // pvc object
+        pvc: "<",
+      },
+      template: '<a href="" ng-click="openEditModal()" role="button">Expand PVC</a>',
+      replace: true,
+      link: function(scope) {
+
+        scope.openEditModal = function() {
+          // opening the modal with settings scope as parent
+          var modalInstance = $uibModal.open({
+            templateUrl: 'views/modals/edit-pvc-resource.html',
+            controller: 'EditPvcModalController',
+            scope: scope
+          });
+
+          var hideErrorNotifications = function() {
+            NotificationsService.hideNotification("expand-pvc-error");
+          };
+
+          scope.$on('$destroy', hideErrorNotifications);
+
+          modalInstance.result.then(function(updatedSize) {
+            hideErrorNotifications();
+            var updatedPvc = angular.copy(scope.pvc);
+            _.set(updatedPvc, 'spec.resources.requests.storage', updatedSize);
+            var kind = scope.pvc.kind;
+            var resourceName = scope.pvc.metadata.name;
+            var typeDisplayName = $filter('humanizeKind')(kind);
+            var formattedResource = typeDisplayName + ' ' + "\'"  + resourceName + "\'";
+
+            DataService.update({
+              resource: APIService.kindToResource(kind)
+            }, resourceName,
+               updatedPvc,
+               {namespace: scope.pvc.metadata.namespace })
+            .then(function() {
+              NotificationsService.addNotification({
+                type: "success",
+                message: formattedResource + " expand request has been submitted."
+              });
+            })
+            .catch(function(err) {
+             // called if failure to edit
+              NotificationsService.addNotification({
+                id: "expand-pvc-error",
+                type: "error",
+                message: "Could not save " + formattedResource,
+                details: $filter('getErrorDetails')(err)
+              });
+              Logger.error(formattedResource + " could not be expanded.", err);
+            });
+          });
+        };
+      }
+    };
+  });
--- a/app/views/browse/persistent-volume-claim.html
+++ b/app/views/browse/persistent-volume-claim.html
@@ -19,6 +19,12 @@
                 <a ng-href="{{pvc | editYamlURL}}" role="button">Edit YAML</a>
               </li>
               <li>
+                <edit-pvc
+                  ng-if="isExpansionAllowed && (pvcVersion | canI : 'update')"
+                  pvc="pvc">
+                </edit-pvc>
+              </li>
+              <li>
                 <delete-link
                   ng-if="pvcVersion | canI : 'delete'"
                   kind="PersistentVolumeClaim"
--- a/app/views/modals/edit-pvc-resource.html
+++ b/app/views/modals/edit-pvc-resource.html
@@ -1,0 +1,110 @@
+<div class="modal-resource-action">
+  <div class="modal-header">
+    <button type="button" class="close" data-dismiss="modal" aria-hidden="true" aria-label="Close" ng-click="cancel()">
+      <span class="pficon pficon-close" aria-hidden="true"></span>
+    </button>
+    <h1 class="modal-title">Expand Persistent Volume Claim</h1>
+  </div>
+
+  <div class="modal-body">
+    <h3>Increase the capacity of claim
+      <strong>{{typeDisplayName}}</strong>.
+      <a ng-href="{{'persistent_volumes' | helpLink}}" target="_blank">
+        <span class="learn-more-inline">Learn More
+          <i class="fa fa-external-link" aria-hidden="true"></i>
+        </span>
+      </a>
+    </h3>
+
+    <p>This can be a time-consuming process.</p>
+
+    <ng-form name="expandPersistentVolumeClaimForm">
+      <div ng-if="capacityReadOnly" class="form-group mar-bottom-xl">
+        <label>Size</label>
+        <div class="static-form-value-large">
+          {{claim.capacity}} {{claim.unit | humanizeUnit : 'storage'}}
+          <small>(cannot be changed)</small>
+        </div>
+      </div>
+      <div ng-if="!capacityReadOnly" class="form-group">
+        <fieldset class="form-inline compute-resource">
+          <label>
+            Capacity
+            <small ng-if="limits.min && limits.max">
+              {{limits.min | usageWithUnits : 'storage'}} min to {{limits.max | usageWithUnits : 'storage'}} max
+            </small>
+            <small ng-if="limits.min && !limits.max">
+              Min: {{limits.min | usageWithUnits : 'storage'}}
+            </small>
+            <small ng-if="limits.max && !limits.min">
+              Max: {{limits.max | usageWithUnits : 'storage'}}
+            </small>
+          </label>
+          <div class="resource-size" ng-class="{ 'has-error': expandPersistentVolumeClaimForm.capacity.$invalid && expandPersistentVolumeClaimForm.capacity.$touched}">
+            <div class="resource-amount">
+              <label for="claim-amount" class="sr-only">Amount</label>
+              <input type="number"
+                     name="capacity"
+                     id="claim-amount"
+                     ng-model="claim.capacity"
+                     required min="0"
+                     pattern="\d+(\.\d+)?"
+                     select-on-focus class="form-control"
+                     aria-describedby="claim-capacity-help">
+            </div>
+            <div class="resource-unit">
+                <label class="sr-only">Unit</label>
+                <ui-select search-enabled="false"
+                           ng-model="claim.unit"
+                           input-id="claim-capacity-unit">
+                    <ui-select-match>{{$select.selected.label}}</ui-select-match>
+                    <ui-select-choices repeat="option.value as option in units" group-by="groupUnits">
+                        {{option.label}}
+                    </ui-select-choices>
+                </ui-select>
+            </div>
+          </div>
+          <div id="claim-capacity-help" class="help-block">
+            Desired storage capacity.
+          </div>
+          <div ng-if="expandPersistentVolumeClaimForm.capacity.$invalid">
+            <div class="has-error" ng-show="expandPersistentVolumeClaimForm.capacity.$error.required">
+              <span class="help-block">Size is required.</span>
+            </div>
+            <div class="has-error" ng-show="expandPersistentVolumeClaimForm.capacity.$error.number">
+              <span class="help-block">Must be a number.</span>
+            </div>
+            <div class="has-error" ng-show="expandPersistentVolumeClaimForm .capacity.$error.min">
+              <span class="help-block">Must be a positive number.</span>
+            </div>
+            <div class="has-error" ng-show="expandPersistentVolumeClaimForm.capacity.$error.checkCurrentCapacity">
+              <span class="help-block">The requested capacity may not be less than the current capacity.</span>
+            </div>
+            <div ng-if="expandPersistentVolumeClaimForm.capacity.$error.limitRangeMin" class="has-error">
+              <span class="help-block">
+                Can't be less than {{limits.min | usageWithUnits : 'storage'}}.
+              </span>
+            </div>
+            <div ng-if="expandPersistentVolumeClaimForm.capacity.$error.limitRangeMax" class="has-error">
+              <span class="help-block">
+                Can't be greater than {{limits.max | usageWithUnits : 'storage'}}.
+              </span>
+            </div>
+            <div ng-if="expandPersistentVolumeClaimForm.capacity.$error.willExceedStorage" class="has-error">
+              <span class="help-block">
+                Storage quota will be exceeded. <a ng-href="project/{{projectName}}/quota" target="_blank">View Quota&nbsp;</a>
+              </span>
+            </div>
+          </div>
+
+        </fieldset>
+      </div>
+
+      <div class="modal-footer" style="padding-right: 0;">
+        <button class="btn btn-default" type="button" ng-click="cancel()">Cancel</button>
+        <button class="btn btn-primary" type="submit" ng-disabled="expandPersistentVolumeClaimForm.$invalid || disableButton" ng-click="expand()"> Expand
+        </button>
+      </div>
+    </ng-form>
+  </div>
+</div>
--- a/app/index.html
+++ b/app/index.html
@@ -296,6 +296,7 @@
         <script src="scripts/controllers/modals/confirmScale.js"></script>
         <script src="scripts/controllers/modals/confirmSaveLog.js"></script>
         <script src="scripts/controllers/modals/deleteModal.js"></script>
+        <script src="scripts/controllers/modals/editPvcModal.js"></script>
         <script src="scripts/controllers/modals/debugTerminal.js"></script>
         <script src="scripts/controllers/modals/confirmReplaceModal.js"></script>
         <script src="scripts/controllers/modals/processOrSaveTemplateModal.js"></script>
@@ -312,6 +313,7 @@
         <script src="scripts/directives/date.js"></script>
         <script src="scripts/directives/deleteLink.js"></script>
         <script src="scripts/directives/editConfigMapOrSecret.js"></script>
+        <script src="scripts/directives/editPvc.js"></script>
         <script src="scripts/directives/editEnvironmentFrom.js"></script>
         <script src="scripts/directives/events.js"></script>
         <script src="scripts/directives/eventsSidebar.js"></script>
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -7042,9 +7042,14 @@ title: "Storage",
 link: "project/" + n.project + "/browse/storage"
 }, {
 title: n.pvc
-} ], t.pvcVersion = r.getPreferredVersion("persistentvolumeclaims"), t.eventsVersion = r.getPreferredVersion("events");
-var i = [], s = function(e, n) {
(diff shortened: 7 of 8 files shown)


app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,66 @@
+<div class="modal-resource-action">
+    <!-- Use a form so that the enter key submits when typing a project name to confirm. -->
```
spadgett: comment looks wrong

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,66 @@
...
+      <span class="pficon pficon-close"></span>
+    </button>
+    <h1 class="modal-title">Expand Persistent Volume Claim</h1>
+  </div>
+
+  <div class="modal-body">
+    <h3>Increase the Capacity of claim
+      <strong>{{displayName ? displayName : resourceName}}</strong>.
+      <a href="{{'persistent_volumes' | helpLink}}" target="_blank">
+        <span class="learn-more-inline">Learn More&nbsp;
```
spadgett: remove `&nbsp;`

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,66 @@
...
+    <p><span>This can be a time-consuming process.</span></p>
+
+    <ng-form name="expandPersistentVolumeClaimForm">
+      <fieldset class="form-inline compute-resource">
+        <label>Capacity</label>
+
+        <div class="resource-size" ng-class="{ 'has-error': expandPersistentVolumeClaimForm.capacity.$invalid && expandPersistentVolumeClaimForm.capacity.$touched}">
+          <div class="resource-amount">
+            <label for="claim-amount" class="sr-only">Amount</label>
+            <input type="number" name="capacity" id="claim-amount" ng-model="claim.capacity" required min="0" pattern="\d+(\.\d+)?" select-on-focus class="form-control" aria-describedby="claim-capacity-help" ng-change="validateLimitRange(expandPersistentVolumeClaimForm)">
```
spadgett: aria-describedby is wrong

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,66 @@
...
+            <span class="help-block">Size is required.</span>
+          </div>
+          <div class="has-error" ng-show="expandPersistentVolumeClaimForm.capacity.$error.number">
+            <span class="help-block">Must be a number.</span>
+          </div>
+          <div class="has-error" ng-show="expandPersistentVolumeClaimForm .capacity.$error.min">
+            <span class="help-block">Must be a positive number.</span>
+          </div>
+          <div class="has-error" ng-show="expandPersistentVolumeClaimForm.capacity.$error.checkCurrentCapacity">
+            <span class="help-block">The requested capacity may be less than the current capacity.</span>
```
spadgett: The requested capacity must be larger than the current capacity.

spadgett:
/hold

Hold for 3.11

app/views/browse/persistent-volume-claim.html
```diff
@@ -18,6 +18,17 @@ <h1 class="contains-actions">
               <li ng-if="!pvc.spec.volumeName">
                 <a ng-href="{{pvc | editYamlURL}}" role="button">Edit YAML</a>
               </li>
+              <li>
+                <edit-pvc
+                  ng-if="(pvcVersion | canI : 'update') && isExpansionAllowed"
```
spadgett: nit: I'd flip the order of this check so it can short circuit more efficiently since `isExpansionAllowed` is already calculated

```
ng-if="isExpansionAllowed && (pvcVersion | canI : 'update')"
```

app/scripts/controllers/modals/editModal.js
```diff
@@ -0,0 +1,72 @@
...
+
+        $scope.claim = {}
+        $scope.claim.capacity = parseInt(allocatedAmountAndUnit[0]);
+        $scope.claim.unit = allocatedAmountAndUnit[1];
+
+        $scope.currentCapacityUnits = angular.copy($scope.claim);
+
+        $scope.claim.capacity += 1;
+
+        $scope.units = [{
```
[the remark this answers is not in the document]
spadgett: OK. Even so, do we know that it will always come back from the API with that unit? If I request 1.25GiB, can the API server return the values in MiB to avoid the decimal, and what happens in this editor?

(I have no idea which is why I'm asking :)

spadgett:
> Is there any validation we'd want to do against the actual quota? It would be good to show it in some way, either directly and/or limiting the input field.

Good question. On PVC creation, we do check limit ranges and quota. It seems the same would apply here:

https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/oscPersistentVolumeClaim.js#L100-L127

cloudbehl:
@spadgett @sspeiche will update the PR with limit ranges and quota checks.

cloudbehl:
@spadgett Updated the PR can you please review.

benjaminapetersen:
@openshift/team-ux-review

app/scripts/controllers/modals/editModal.js
```diff
@@ -0,0 +1,123 @@
...
+    var resourceQuotasVersion = APIService.getPreferredVersion('resourcequotas');
+    var appliedClusterResourceQuotasVersion = APIService.getPreferredVersion('appliedclusterresourcequotas');
+
+    var amountAndUnit = $filter('amountAndUnit');
+    var usageWithUnits = $filter('usageWithUnits');
+    var usageValue = $filter('usageValue');
+
+    var allocatedAmountAndUnit = amountAndUnit($scope.currentCapacity);
+
+    $scope.projectName = JSON.parse($scope.pvc).metadata.namespace;
```
spadgett: Why do we need to call `JSON.parse` here?

app/scripts/controllers/modals/editModal.js
```diff
@@ -0,0 +1,123 @@
...
+        case 'M':
+        case 'G':
+        case 'T':
+          return 'Decimal Units';
+      }
+      return '';
+    };
+
+    var validateQuota = function() {
+      var willExceedStorage = QuotaService.willRequestExceedQuota($scope.quotas, $scope.clusterQuotas, 'requests.storage', (($scope.claim.capacity -$scope.currentCapacityUnits.capacity) + $scope.claim.unit));
```
spadgett: I'm not sure the math here is right. Couldn't the current value and requested value use different units?

app/scripts/controllers/modals/editModal.js
```diff
@@ -0,0 +1,123 @@
...
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('willExceedStorage', !willExceedStorage);
+    };
+
+    var validateLimitRange = function() {
```
spadgett: This looks like it's copied from oscPersistentVolumeClaim.js. Is there a way we can reuse the code without the copy/paste?

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+                      DataService,
+                      ProjectsService,
+                      NotificationsService,
+                      Logger) {
+
+    return {
+      restrict: "E",
+      scope: {
+        // pvc object
+        pvc: "@",
```
spadgett: Why `@`? Shouldn't this be `<` (or possibly `=` if you need two-way binding)?

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+        // Alerts object for success and error alerts.
+        alerts: "=",
+        // Optional link label. Defaults to "Expand".
+        label: "@?",
+        // Optional redirect URL when the delete succeeds
+        redirectUrl: "@?",
+        // currentCapacity of the pvc
+        currentCapacity: "@"
```
spadgett: Why does this need to be passed in as an attribute? Isn't it already in the PVC?

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+                      NotificationsService,
+                      Logger) {
+
+    return {
+      restrict: "E",
+      scope: {
+        // pvc object
+        pvc: "@",
+        // Alerts object for success and error alerts.
+        alerts: "=",
```
spadgett: I don't see where this is used.

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+        // Optional link label. Defaults to "Expand".
+        label: "@?",
```
spadgett: Can we remove this if it's not needed?

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+        // Optional redirect URL when the delete succeeds
+        redirectUrl: "@?",
```
spadgett: Can we remove this if it's not needed?

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+
+          modalInstance.result.then(function(updatedSize) {
+            // upon clicking edit button, edit resource and send notification
+            ProjectsService
+            .get($routeParams.project)
+            .then(_.spread(function(project, context) {
+              scope.project = project;
+              scope.projectContext = context;
+              DataService
+                .get(scope.pvcVersion, $routeParams.pvc, context)
```
spadgett: Why are we getting the PVC again if it's passed in?

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+                .then(function(pvc) {
+                  var updatedPvc = angular.copy(pvc);
+                  _.set(updatedPvc, ["spec", "resources", "requests", "storage"], updatedSize);
+                  var kind = pvc.kind;
+                  var resourceName = pvc.metadata.name;
+                  var typeDisplayName = scope.typeDisplayName || $filter('humanizeKind')(kind);
```
spadgett: I don't see where scope.typeDisplayName is ever set

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+                    NotificationsService.addNotification({
+                      type: "success",
+                      message: formattedResource + " expand request has been submitted."
+                    });
+                  })
+                  .catch(function(err) {
+                   // called if failure to edit
+                    NotificationsService.addNotification({
+                      type: "error",
+                      message: _.capitalize(formattedResource),
```
spadgett: The message is the resource name?

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,84 @@
...
+                  }, resourceName, updatedPvc, context)
+                  .then(function() {
+                    NotificationsService.addNotification({
+                      type: "success",
+                      message: formattedResource + " expand request has been submitted."
+                    });
+                  })
+                  .catch(function(err) {
+                   // called if failure to edit
+                    NotificationsService.addNotification({
```
spadgett: The notification should have an ID so it can be cleared when the form is resubmitted. See

https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/createPersistentVolumeClaim.js#L74

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,101 @@
+<div class="modal-resource-action">
+  <div class="modal-header">
+    <button type="button" class="close" data-dismiss="modal" aria-hidden="true" aria-label="Close" ng-click="cancel()">
+      <span class="pficon pficon-close"></span>
```
spadgett:
```html
 <span class="pficon pficon-close" aria-hidden="true"></span>
```

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,101 @@
...
+    <h3>Increase the Capacity of claim
+      <strong>{{displayName ? displayName : resourceName}}</strong>.
+      <a href="{{'persistent_volumes' | helpLink}}" target="_blank">
+        <span class="learn-more-inline">Learn More
+          <i class="fa fa-external-link" aria-hidden="true"></i>
+        </span>
+      </a>
+    </h3>
+
+    <p><span>This can be a time-consuming process.</span></p>
```
spadgett: Remove span here

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,101 @@
...
+
+    <ng-form name="expandPersistentVolumeClaimForm">
```
spadgett: Why `ng-form` instead of `<form>`?

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,101 @@
...
+    <button type="button" class="close" data-dismiss="modal" aria-hidden="true" aria-label="Close" ng-click="cancel()">
+      <span class="pficon pficon-close"></span>
+    </button>
+    <h1 class="modal-title">Expand Persistent Volume Claim</h1>
+  </div>
+
+  <div class="modal-body">
+    <h3>Increase the Capacity of claim
+      <strong>{{displayName ? displayName : resourceName}}</strong>.
+      <a href="{{'persistent_volumes' | helpLink}}" target="_blank">
```
spadgett: ng-href

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,101 @@
...
+            <span class="help-block">
+              Storage quota will be exceeded. <a ng-href="project/{{projectName}}/quota" target="_blank">View Quota&nbsp;</a>
+            </span>
+          </div>
+        </div>
+
+      </fieldset>
+
+      <div class="modal-footer">
+          <button class="btn btn-default" type="button" ng-click="cancel();">Cancel</button>
```
spadgett: unnecessary `;` in ng-click in both buttons

app/scripts/controllers/modals/editModal.js
benjaminapetersen: Also, this is a really long line & rather hard to understand.

app/views/modals/edit-resource.html
benjaminapetersen: Prob should clarify why we would suggest not using `ng-form` followed by a suggestion to use`ng-href`?

app/scripts/controllers/modals/editModal.js
```diff
@@ -0,0 +1,123 @@
...
+      $scope.updatedCapacity = $scope.claim.capacity + $scope.claim.unit;
+      $uibModalInstance.close($scope.updatedCapacity);
+    };
+
+    $scope.cancel = function() {
+      $uibModalInstance.dismiss('cancel');
+    };
+
+    $scope.groupUnits = function(unit) {
+      switch (unit.value) {
```
benjaminapetersen: A little subjective, but using a `switch` with passthrough + no `default` is a code-smell.  I wonder if this can be done in a way that is less brittle.  

Two possible solutions:

```JavaScript 
// not everyone likes nested ternaries, but this is fairly straightforward,
// though pardon the es2015 syntax:
const groupUnits = (unit) => {
  const binary = ['Mi','Gi','Ti'];
  const decimal = ['M','G','T']; 
  return _.includes(binary, unit)
    ? 'Binary Units'
    : _.includes(decimal, unit)
      ? 'Decimal Units'
      : null;
}

// can also use an object literal approach:
// again, pardon the es2015 syntax:
// this impl is probably quite a bit more performant, though 
// I doubt performance is super important here:

const groupUnits = (unit) => {
  const binary = 'Binary Units';
  const decimal = 'Decimal Units';
  return ({
    'Mi': binary,
    'Gi': binary,
    'Ti': binary,
    'M': decimal,
    'G': decimal,
    'T': decimal
  })[unit];
}
```

beanh66:
@benjaminapetersen Can you add screenshots or GIF so we can review? I didn't see anything linked above but I may have missed it!

app/scripts/controllers/modals/editModal.js
cloudbehl: ACK! 

I think I missed considering the different units.

app/scripts/directives/editPvc.js
cloudbehl: a message is something like "Persistent volume claim '{pvc_name}'"

example: Persistent volume claim 'demo1'

spadgett: Right, I guess I'm saying we should say something like, "Could not save persistent volume claim my-pvc."

serenamarie125:
@cloudbehl can we get screenshots added here please so we can review the screens?

app/scripts/directives/editPvc.js
cloudbehl: I updated the PR with the message you provided.

cloudbehl:
@serenamarie125 Attached the screenshots below. Please let me know if anything else is required.

screenshot from 2018-06-25 15-13-03
screenshot from 2018-06-25 15-13-13
screenshot from 2018-06-25 15-13-37
screenshot from 2018-06-25 15-14-00
screenshot from 2018-06-25 15-14-11
screenshot from 2018-06-25 15-14-23
screenshot from 2018-06-25 15-17-49
screenshot from 2018-06-25 15-14-36
screenshot from 2018-06-25 15-15-01

app/scripts/controllers/modals/editModal.js
cloudbehl: Ack! Done

app/scripts/directives/editPvc.js
cloudbehl: ACK!

the two way binding was not required so I used <.

app/scripts/directives/editPvc.js
cloudbehl: ACK

app/scripts/controllers/modals/editModal.js
cloudbehl: I used the first approach, please review!

cshinn:
@cloudbehl There's an extra space before the period in "Increase the Capacity of claim" Additionally, I don't think "capacity" should be capitalized in this instance, and it would be great if "claim" could be replaced by the name of the current claim (demo in this case).

[bootstrap touchspin](http://www.patternfly.org/pattern-library/widgets/#bootstrap-touchspin) could also be used for the capacity field

beanh66:
@cloudbehl In the first modal screen, is there a reason the field does not have the red error border around it? The modals in all of the following error cases do seem to have it. (https://user-images.githubusercontent.com/6695156/41843554-7346f30a-788b-11e8-9237-cbe66cd4ece6.png)

Also wondering if we can right align the GiB dropdown and the Expand button.

cloudbehl:
@spadgett @benjaminapetersen can you help me with [bootstrap touchspin](http://www.patternfly.org/pattern-library/widgets/#bootstrap-touchspin).

As I can see bootstrap-touchspin is apart of bower-component but how to use it is what I am facing a problem. Do I need to put a package of [angular-touchspin](https://www.npmjs.com/package/angular-touchspin) in package.json

Also is it really required to put  [bootstrap touchspin](http://www.patternfly.org/pattern-library/widgets/#bootstrap-touchspin) ? As I haven't seen any other component using this component.

Ask from UX team:
https://github.com/openshift/origin-web-console/pull/3004#issuecomment-400029976

spadgett:
> can you help me with bootstrap touchspin.

Let's skip that for the moment. I'd like to avoid adding a new dependency in this PR. Also it would affect other similar forms that don't use it.

cshinn:
@cloudbehl I don't have a strong opinion about it, especially if it's very inconvenient. Just thought it might be a useful option

app/scripts/controllers/modals/editModal.js
```diff
@@ -0,0 +1,121 @@
...
+    };
+
+    $scope.groupUnits = function(unit) {
+      var binary = ['Mi','Gi','Ti'];
+      var decimal = ['M','G','T'];
+      return _.includes(binary, unit.value)
+      ? 'Binary Units'
+      : _.includes(decimal, unit.value)
+      ? 'Decimal Units'
+      : null;
```
spadgett: Avoid nested ternaries, which are hard to understand.

app/scripts/controllers/persistentVolumeClaim.js
```diff
@@ -54,6 +57,15 @@ angular.module('openshiftConsole')
       DataService
         .get($scope.pvcVersion, $routeParams.pvc, context, { errorNotification: false })
         .then(function(pvc) {
+          var storageClassName = storageClass(pvc);
+          if(storageClassName) {
+            DataService.get($scope.storageClassesVersion, storageClassName, {namespace: pvc.metadata.namespace}).then(function(storageClassData) {
```
spadgett: Storage classes are cluster scoped, correct? You should be able to leave off namespace here

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,75 @@
...
+                      DataService,
+                      ProjectsService,
+                      NotificationsService,
+                      Logger) {
+
+    return {
+      restrict: "E",
+      scope: {
+        // pvc object
+        pvc: "<",
```
spadgett: 👍  Thanks, this looks much cleaner

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,75 @@
...
+          var hideErrorNotifications = function() {
+            NotificationsService.hideNotification("expand-pvc-error");
+          };
+
+          scope.$on('$destroy', hideErrorNotifications);
+
+          modalInstance.result.then(function(updatedSize) {
+            // upon clicking edit button, edit resource and send notification
+            ProjectsService
+            .get($routeParams.project)
```
spadgett: I don't think you need to get the project. I'd just use `{ namespace: scope.pvc.metadata.namespace }` as the context.

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,75 @@
...
+                  type: "success",
+                  message: formattedResource + " expand request has been submitted."
+                });
+              })
+              .catch(function(err) {
+               // called if failure to edit
+                NotificationsService.addNotification({
+                  id: "expand-pvc-error",
+                  type: "error",
+                  message: "Could not save " + _.capitalize(formattedResource),
```
spadgett: remove _.capitalize

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,101 @@
+<div class="modal-resource-action">
```
spadgett: Give a more specific filename than edit-resource.html since this is specific to PVCs.

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,75 @@
...
+            ProjectsService
+            .get($routeParams.project)
+            .then(_.spread(function(project, context) {
+              hideErrorNotifications();
+              var updatedPvc = angular.copy(scope.pvc);
+              _.set(updatedPvc, ["spec", "resources", "requests", "storage"], updatedSize);
+              var kind = scope.pvc.kind;
+              var resourceName = scope.pvc.metadata.name;
+              var typeDisplayName = $filter('humanizeKind')(kind);
+              var formattedResource = _.capitalize(typeDisplayName) + ' ' + "\'"  + (scope.displayName ? scope.displayName : resourceName) + "\'";
```
spadgett: I don't see where scope.displayName is ever set

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,101 @@
...
+  <div class="modal-header">
+    <button type="button" class="close" data-dismiss="modal" aria-hidden="true" aria-label="Close" ng-click="cancel()">
+      <span class="pficon pficon-close" aria-hidden="true"></span>
+    </button>
+    <h1 class="modal-title">Expand Persistent Volume Claim</h1>
+  </div>
+
+  <div class="modal-body">
+    <h3>Increase the Capacity of claim
+      <strong>{{displayName ? displayName : resourceName}}</strong>.
```
spadgett: I don't see where scope.displayName is ever set

app/views/modals/edit-resource.html
```diff
@@ -0,0 +1,101 @@
+<div class="modal-resource-action">
+  <div class="modal-header">
+    <button type="button" class="close" data-dismiss="modal" aria-hidden="true" aria-label="Close" ng-click="cancel()">
+      <span class="pficon pficon-close" aria-hidden="true"></span>
+    </button>
+    <h1 class="modal-title">Expand Persistent Volume Claim</h1>
+  </div>
+
+  <div class="modal-body">
+    <h3>Increase the Capacity of claim
```
spadgett: Consistent case in the title

app/scripts/controllers/persistentVolumeClaim.js
cloudbehl: @spadgett I don't think it's cluster scoped.

I removed namespace and it is also giving an error.

spadgett: What is the error?

cloudbehl: Ahh! ok I wasn't passing {}.

cloudbehl: now it worked, thanks!

cloudbehl:
@spadgett I updated the PR.

> Also wondering if we can right align the GiB dropdown and the Expand button.
@beanh66 even while creating a pvc the dropdown is left aligned. I think we should maintain that consistancy. Thoughts?

beanh66:
@cloudbehl Sorry I don't think I articulated well. I agree it is left aligned, I guess I would also expect the items to align on the right. Seems like the dropdown extends further than the button on the right and curious why?

app/scripts/controllers/persistentVolumeClaim.js
```diff
@@ -54,6 +57,15 @@ angular.module('openshiftConsole')
       DataService
         .get($scope.pvcVersion, $routeParams.pvc, context, { errorNotification: false })
         .then(function(pvc) {
+          var storageClassName = storageClass(pvc);
+          if(storageClassName) {
```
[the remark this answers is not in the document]
cloudbehl: @spadgett can you tell which logic are you talking about?

by looking pvcResolved method it only accepts the PVC and the action as a parameter.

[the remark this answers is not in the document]
spadgett: If the storage class changes for the PVC changes after the page loads (assuming it can be changed), that changed won't be picked up.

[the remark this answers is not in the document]
cloudbehl: so should I put `pvcResolved(pvc) ` and `watches.push(DataService.watchObject($scope.pvcVersion, $routeParams.pvc, context, pvcResolved));` after the then statement?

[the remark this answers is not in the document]
spadgett: No, you should move the get storage class logic into `pvcResolved`

OK as a follow on. It's an edge case.

cloudbehl:
> Seems like the dropdown extends further than the button on the right and curious why?

@beanh66 I corrected the spacing issue, now the dropdown and button are both alligned properly, 
Also have added the red border to the capacity text box. 

please check the attachment.

screenshot from 2018-07-02 15-18-40

beanh66:
@cloudbehl thanks! LGTM

app/scripts/directives/editPvc.js
```diff
@@ -0,0 +1,73 @@
...
+          });
+
+          var hideErrorNotifications = function() {
+            NotificationsService.hideNotification("expand-pvc-error");
+          };
+
+          scope.$on('$destroy', hideErrorNotifications);
+
+          modalInstance.result.then(function(updatedSize) {
+            // upon clicking edit button, edit resource and send notification
```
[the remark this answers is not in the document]
cloudbehl: I think we can remove the comment

app/scripts/controllers/modals/editPvcModal.js
```diff
@@ -0,0 +1,125 @@
...
+
+    DataService.list(limitRangesVersion, { namespace: $scope.projectName }, function(limitRangeData) {
+      var limitRanges = limitRangeData.by('metadata.name');
+      $scope.disableButton = false;
+      if (_.isEmpty(limitRanges)) {
+        return;
+      }
+
+      $scope.limits = LimitRangesService.getEffectiveLimitRange(limitRanges, 'storage', 'PersistentVolumeClaim');
+      // If min === max, set the capacity to the required value and make the field readonly.
```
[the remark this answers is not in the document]
cloudbehl: @benjaminapetersen ack!

app/scripts/controllers/modals/editPvcModal.js
```diff
@@ -0,0 +1,125 @@
...
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('willExceedStorage', !willExceedStorage);
+    };
+
+    var validateLimitRange = function() {
+      // Use usageValue filter to normalize units for comparison.
+      var value = $scope.claim.capacity && usageValue($scope.claim.capacity + $scope.claim.unit);
+      var currentValue = $scope.currentCapacityUnits.capacity && usageValue($scope.currentCapacityUnits.capacity + $scope.currentCapacityUnits.unit);
+      var min = _.has($scope, 'limits.min') && usageValue($scope.limits.min);
+      var max = _.has($scope, 'limits.max') && usageValue($scope.limits.max);
+      var minValid = true;
```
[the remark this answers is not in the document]
cloudbehl: divided the bigger function into three smaller functions.

app/scripts/controllers/modals/editPvcModal.js
```diff
@@ -0,0 +1,125 @@
...
+      $scope.updatedCapacity = $scope.claim.capacity + $scope.claim.unit;
+      $uibModalInstance.close($scope.updatedCapacity);
+    };
+
+    $scope.cancel = function() {
+      $uibModalInstance.dismiss('cancel');
+    };
+
+    $scope.groupUnits = function(unit) {
+      var binary = ['Mi','Gi','Ti'];
```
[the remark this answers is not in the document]
cloudbehl: ack!

app/scripts/controllers/modals/editPvcModal.js
```diff
@@ -0,0 +1,141 @@
...
+    $scope.expand = function() {
+      $scope.updatedCapacity = $scope.claim.capacity + $scope.claim.unit;
+      $uibModalInstance.close($scope.updatedCapacity);
+    };
+
+    $scope.cancel = function() {
+      $uibModalInstance.dismiss('cancel');
+    };
+
+    $scope.groupUnits = function(unit) {
```
benjaminapetersen: nit: but this function doesn't actually group, I would assume "grouping" would mean returning new grouped arrays.  Thoughts?

cloudbehl: I agree it's not returning a grouped array but on other side it's responsible for showing the GroupWise distribution of units in the view. So that's why the name is groupUnits name I have used.

If you have another suggestion I am ok changing the name.

benjaminapetersen: Right, I think the html clarifies this:

```html
<ui-select-choices 
   repeat="option.value as option in units" 
   group-by="groupUnits">
```
The `group-by` already tells me it is grouping.  So the function you pass can be named something more specifc, like `unitTextLabel()` for example.   `group-by="unitTextLabel"` gives me a better idea of what the grouping actually is.

cloudbehl: ack!

spadgett:
I haven’t looked closely at the last review, but given that this has gone
through several rounds of comments and is close to merging, I’d like to
avoid too much rework at this point. I don’t want to put this in danger of
missing feature freeze.

cloudbehl:
@spadgett ok, so when are are we planning to merge this PR?

Is there anything else required from my side, do let me know?

spadgett:
I’m on PTO, so I’ll need to look when I get back Friday

app/scripts/controllers/persistentVolumeClaim.js
```diff
@@ -30,14 +30,25 @@ angular.module('openshiftConsole')
...
 
+    var storageClass = $filter('storageClass');
     var watches = [];
 
     var pvcResolved = function(pvc, action) {
       $scope.pvc = pvc;
       $scope.loaded = true;
+      var storageClassName = storageClass($scope.pvc);
+
+      if(storageClassName) {
```
spadgett: Only request the storage class again if the value has changed.

Make sure we handle the case when there's no storage class. Right now it looks like isExpansionAllowed is never set when there is no storage class on the PVC.

```jsx
var storageClass;
var updateStorageClass = function(pvc) {
  var storageClassName = storageClass($scope.pvc);
  if (storageClassName !== _.get(storageClass, 'metadata.name')) {
    DataService.get($scope.storageClassesVersion, storageClassName, {}).then(function(sc) {
      storageClass = sc;
    });
  }
};

var pvcResolved = function(pvc, action) {
  $scope.pvc = pvc;
  $scope.loaded = true;
  updateStorageClass(pvc);
  $scope.isExpansionAllowed = (!storageClass || storageClass.allowVolumeExpansion) && pvc.status.phase === 'Bound';
};
```

app/scripts/controllers/modals/editPvcModal.js
```diff
@@ -0,0 +1,141 @@
...
+    var validateMaxValid = function (value) {
+      var max = _.has($scope, 'limits.max') && usageValue($scope.limits.max);
+      var maxValid = value <= max;
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('limitRangeMax', maxValid);
+    };
+
+    var validateCapacityValid = function (value) {
+      var currentValue = $scope.currentCapacityUnits.capacity && usageValue($scope.currentCapacityUnits.capacity + $scope.currentCapacityUnits.unit);
+      var requestValid = value > currentValue;
+      $scope.expandPersistentVolumeClaimForm.capacity.$setValidity('checkCurrentCapacity', requestValid);
```
spadgett: Something like 'greaterThanCurrent' would be more descriptive.

app/views/modals/edit-pvc-resource.html
```diff
@@ -0,0 +1,110 @@
...
+                Storage quota will be exceeded. <a ng-href="project/{{projectName}}/quota" target="_blank">View Quota&nbsp;</a>
+              </span>
+            </div>
+          </div>
+
+        </fieldset>
+      </div>
+
+      <div class="modal-footer" style="padding-right: 0;">
+          <button class="btn btn-default" type="button" ng-click="cancel()">Cancel</button>
```
spadgett: indentation

app/scripts/controllers/modals/editPvcModal.js
```diff
@@ -0,0 +1,141 @@
...
+    var checkMinEqualsMax = function() {
+      // If min === max, set the capacity to the required value and make the field readonly.
+      if ($scope.limits.min && $scope.limits.max) {
+        var minUsage = usageValue($scope.limits.min);
+        var maxUsage = usageValue($scope.limits.max);
+        if (minUsage === maxUsage) {
+          var requiredAmountAndUnit = amountAndUnit($scope.limits.max);
+          $scope.claim.capacity = Number(requiredAmountAndUnit[0]);
+          $scope.claim.unit = requiredAmountAndUnit[1];
+          $scope.capacityReadOnly = true;
```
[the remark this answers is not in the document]
cloudbehl: @spadgett I copied this from createPvc function. I agree this is a confusion, so if you ack I will change the code.

benjaminapetersen:
@cloudbehl agree, sorry about the conflicting messages.  I would def encourage you to go the path suggested by @spadgett.  I didn't realize there was some code duplication here, but do agree then that the extract-to-service approach is best.

spadgett:
> but do agree then that the extract-to-service approach is best.

Yeah, agreed. But to be clear: I wouldn't suggest that at this stage in the release, and I don't want to miss the cutoff.

@cloudbehl I think the best path forward is to revert to the very last changes, but make sure we fix the following:

* The display issue that @beanh66 noted
* The file rename (since that's easy)
* Any bugs around resize being disabled when the pvc doesn't have a storage class

Let me know if that's problematic or if you've lost the commit history. (It's good in large reviews like this to add new commits instead of always amending. It's hard to follow what's changed otherwise.)

app/scripts/controllers/persistentVolumeClaim.js
```diff
@@ -38,16 +38,20 @@ angular.module('openshiftConsole')
     var storageClass = $filter('storageClass');
     var watches = [];
 
+    var updateStorageClass = function(pvc) {
+      var storageClassName = storageClass($scope.pvc);
+      if (storageClassName !== _.get(storageClass, 'metadata.name')) {
+        DataService.get($scope.storageClassesVersion, storageClassName, {}).then(function(sc) {
+          $scope.isExpansionAllowed = (!sc || sc.allowVolumeExpansion) && pvc.status.phase === 'Bound';
```
spadgett: If there's no storage class name, isExpansionAllowed is never set to true. It never reaches this block. (I'm assuming a storage class is not required for resize.)

`sc` var itself should never be null here.

cloudbehl:
@spadgett I have squashed the commits, please have a look.

PR #3013, opened 2018-06-07
[WIP] Update grunt-contrib-jshint & dependencies

### LINKED ISSUE #3012: Integrity check fails for ariya/esprima when running npm install
Best I can tell, jshint references a fork of esprima (master branch?) that can cause integrity checks to periodically fail when running `npm install`. Looks like we have an old version of jshint.

From package-lock.json:

https://github.com/openshift/origin-web-console/blob/master/package-lock.json#L4847-L4864

```json
        "esprima": {
          "version": "https://github.com/ariya/esprima/tarball/master",
          "integrity": "sha512-W5+ntA5gmzYpBFDpLLw7yAJPohkq5yKF8JnP9UOaXSfnKcibZQcIM9anoxztmB2NwRWLLidEseBbuCdwgQsnig==",
          "dev": true
        },
        "jshint": {
          "version": "1.1.0",
          "resolved": "https://registry.npmjs.org/jshint/-/jshint-1.1.0.tgz",
          "integrity": "sha1-mYe5C4YFVsvH84DVdVoj1QQhNRM=",
          "dev": true,
          "requires": {
            "cli": "0.4.5",
            "esprima": "https://github.com/ariya/esprima/tarball/master",
            "minimatch": "0.4.0",
            "peakle": "0.0.1",
            "shelljs": "0.1.4",
            "underscore": "1.4.4"
          }
        },
```

Ben, can you take a look?
/assign @benjaminapetersen 

@smarterclayton fyi

benjaminapetersen:
Fix #3012


dist/scripts/vendor.js
```diff
@@ -74682,9 +74682,6 @@ kind: "HorizontalPodAutoscaler"
 }, {
 group: "extensions",
 kind: "NetworkPolicy"
-}, {
```
benjaminapetersen: Odd.

spadgett:
Try bower install

package-lock.json
```diff
@@ -4846,7 +4525,7 @@
         },
         "esprima": {
           "version": "https://github.com/ariya/esprima/tarball/master",
-          "integrity": "sha512-W5+ntA5gmzYpBFDpLLw7yAJPohkq5yKF8JnP9UOaXSfnKcibZQcIM9anoxztmB2NwRWLLidEseBbuCdwgQsnig==",
+          "integrity": "sha512-MMxO9L3pF24K8IjZSUFlbLeVmS5bAsXY2sQr9jbV8yw+RseYh3G9yKU1hBHJNb0zfIp/qLIFKV6ZhhqOY4s4VA==",
```
spadgett: Looks like we'll still have the problem even bumping grunt-contrib-jshint since it still pulls in jshint 1.1.0, which has this dependency on esprima master

benjaminapetersen:
@spadgett bower, good grief, already so used to `yarn` for everything in the new console... 😑

benjaminapetersen:
Bumping `grunt-htmlhint` didn't fix, still tracking down.

spadgett:
jshint 1.x still has the dependency, so i know at least grunt-contrib-jshint is a problem. Unfortunately there's not a newer version. We might even consider getting rid of package-lock.json since they're only development dependencies.

I wonder if this is a problem with the changes for yarn.

cc @TheRealJon

benjaminapetersen:
I'm open to eliminating `package-lock.json`.

benjaminapetersen:
Though, I think eliminating `package-lock.json` would not be considered the correct approach.  Doing some research on that. Not sure yet.

TheRealJon:
Since we don't enforce using NPM >= v5 in this repo, it is hit or miss whether anyone is actually using `package-lock.json`. That means it is really not serving it's purpose and is probably safe to remove.

benjaminapetersen:
It's probably a bad assumption that each of us (and CI) are reasonably current, but CI is prob our validator with that.  If anyone checks in a bad `/dist` it tends to imply something is funky with their build tools. 

But agree, we aren't enforcing it, so value is pretty questionable.

spadgett:
Thinking out loud: how hard is it to change from jshint to eslint? It would be better to use eslint anyway.

Ideally we'd use yarn with yarn.lock and fix the underlying issue. Then we'd be locking the dependencies we ship. package-lock.json is only locking our dev dependencies, which doesn't seem as valuable.

benjaminapetersen:
I can look into `eslint`.

I think I had a PR to swap us to `yarn` a long time ago, pretty minimal.  It might even still be open, I'll fish around.

spadgett:
> I think I had a PR to swap us to yarn a long time ago, pretty minimal. It might even still be open, I'll fish around.

@TheRealJon has a PR open that swaps everything over to yarn.

TheRealJon:
@benjaminapetersen https://github.com/openshift/origin-web-console/pull/2863

benjaminapetersen:
Ah thats right, you have the mega-overhaul :D

PR #3016, opened 2018-06-11
[enterprise-3.9] Bump origin-web-common to 3.9.4

spadgett:
https://github.com/openshift/origin-web-common/releases/tag/v3.9.4
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1589838

Removes package-lock.json since install fails on integrity check for
espirma dependency (tracked by #3012).

/assign @rhamilto


spadgett:
See also discussion in #3013 for why I've removed package-lock.json, at least in this branch. It is not locking the dependencies we ship anyway, so provides minimal value. We should look at the correct solution going forward when we fix #3012.

PR #3020, opened 2018-06-12
List Mobile client builds

sedroche:
Adding components to display the buildconfigs created in https://github.com/openshift/origin-web-console/pull/3019.

client-build-configs

This will list all buildconfigs created for a client. Displaying:
- Buildconfig details
- latest build status and ability to download latest build artifact.
- list of previous builds with status of builds and ability to download build artifact

Downloads have 2 options, a URL, or a QR code which resolves to the download URL.
latest-qrcode
history-qrcode

### DIFF
--- a/app/scripts/directives/jqQrCode.js
+++ b/app/scripts/directives/jqQrCode.js
@@ -1,0 +1,31 @@
+'use strict';
+
+/**
+ * @ngdoc component
+ * @name jq-qr-code
+ */
+angular.module('openshiftConsole').directive('jqQrCode', function () {
+  return {
+    template: "<div class='qr-code-container'></div>",
+    scope: {
+      content: '<?'
+    },
+    link: function ($scope, element) {
+      var qrCodeContainer = $('.qr-code-container', element);
+
+      qrCodeContainer.qrcode({
+        text: $scope.content,
+        size: 250
+      });
+
+      $scope.$watch('content', function () {
+        qrCodeContainer.find('canvas').remove();
+
+        qrCodeContainer.qrcode({
+          text: $scope.content,
+          size: 250
+        });
+      });
+    }
+  };
+});
--- a/app/scripts/directives/jqQrCode.js
+++ b/app/scripts/directives/jqQrCode.js
@@ -1,9 +1,5 @@
 'use strict';
 
-/**
- * @ngdoc component
- * @name jq-qr-code
- */
 angular.module('openshiftConsole').directive('jqQrCode', function () {
   return {
     template: "<div class='qr-code-container'></div>",
@@ -12,19 +8,18 @@ angular.module('openshiftConsole').directive('jqQrCode', function () {
     },
     link: function ($scope, element) {
       var qrCodeContainer = $('.qr-code-container', element);
+      var createQrCode = function(container, content) {
+        container.qrcode({
+          text: content,
+          size: 250
+        });
+      };
 
-      qrCodeContainer.qrcode({
-        text: $scope.content,
-        size: 250
-      });
+      createQrCode(qrCodeContainer, $scope.content);
 
       $scope.$watch('content', function () {
         qrCodeContainer.find('canvas').remove();
-
-        qrCodeContainer.qrcode({
-          text: $scope.content,
-          size: 250
-        });
+        createQrCode(qrCodeContainer, $scope.content);
       });
     }
   };
--- a/app/scripts/directives/mobileClientBuildsList.js
+++ b/app/scripts/directives/mobileClientBuildsList.js
@@ -27,9 +27,10 @@
     var buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
     var watches = [];
     var anchor;
+    var PAGE_HEADER_HEIGHT = 100;
 
     ctrl.$onChanges = function(changes) {
-      var mobileClientChanged = _.get(changes, "mobileClient.currentValue", false);
+      var mobileClientChanged = _.get(changes, 'mobileClient.currentValue', false);
       if (mobileClientChanged && !ctrl.context) {
         ctrl.context = {namespace: _.get(ctrl, 'mobileClient.metadata.namespace')};
       }
@@ -38,7 +39,8 @@
         ctrl.buildsWatchSet = true;
         watches.push(DataService.watch(buildConfigsVersion, ctrl.context, function(buildConfigs) {    
           ctrl.mobileBuildConfigs = _.filter(buildConfigs.by('metadata.name'), function(buildConfig) {
-            return _.get(buildConfig, 'metadata.labels.mobile-client-build') === 'true' && _.get(buildConfig, 'metadata.labels.mobile-client-id') === _.get(ctrl, 'mobileClient.metadata.name');
+            return _.get(buildConfig, 'metadata.labels.mobile-client-build') === 'true' &&
+              _.get(buildConfig, 'metadata.labels.mobile-client-id') === _.get(ctrl, 'mobileClient.metadata.name');
           });
           ctrl.loaded = true;
           ctrl.goToBuildConfig();
@@ -51,7 +53,7 @@
         anchor = $location.hash();
 
         $timeout(function() {
-          $anchorScroll.yOffset = 100;
+          $anchorScroll.yOffset = PAGE_HEADER_HEIGHT;
           $anchorScroll();
         }, 400);
       }
--- a/app/scripts/directives/mobileClientBuildsList.js
+++ b/app/scripts/directives/mobileClientBuildsList.js
@@ -1,0 +1,68 @@
+'use strict';
+
+(function() {
+  angular.module('openshiftConsole').component('mobileClientBuildsList', {
+    controller: [
+      '$timeout',
+      '$anchorScroll',
+      '$location',
+      'APIService',
+      'DataService',
+      MobileClientBuildsList
+    ],
+    bindings: {
+      mobileClient: '<'
+    },
+    templateUrl: 'views/mobile-client-builds-list.html'
+  });
+
+  function MobileClientBuildsList(
+    $timeout,
+    $anchorScroll,
+    $location,
+    APIService,
+    DataService
+  ) {
+    var ctrl = this;
+    var buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
+    var watches = [];
+    var anchor;
+
+    ctrl.$onChanges = function(changes) {
+      var mobileClientChanged = _.get(changes, "mobileClient.currentValue", false);
+      if (mobileClientChanged && !ctrl.context) {
+        ctrl.context = {namespace: _.get(ctrl, 'mobileClient.metadata.namespace')};
+      }
+
+      if (mobileClientChanged && !ctrl.buildsWatchSet) {
+        ctrl.buildsWatchSet = true;
+        watches.push(DataService.watch(buildConfigsVersion, ctrl.context, function(buildConfigs) {    
+          ctrl.mobileBuildConfigs = _.filter(buildConfigs.by('metadata.name'), function(buildConfig) {
+            return _.get(buildConfig, 'metadata.labels.mobile-client-build') === 'true' && _.get(buildConfig, 'metadata.labels.mobile-client-id') === _.get(ctrl, 'mobileClient.metadata.name');
+          });
+          ctrl.loaded = true;
+          ctrl.goToBuildConfig();
+        }));
+      }
+    };
+
+    ctrl.goToBuildConfig = function() {
+      if ($location.hash() && anchor !== $location.hash()) {
+        anchor = $location.hash();
+
+        $timeout(function() {
+          $anchorScroll.yOffset = 100;
+          $anchorScroll();
+        }, 400);
+      }
+    };
+
+    ctrl.goToCreateClientBuild = function() {
+      $location.url('project/' + ctrl.mobileClient.metadata.namespace + '/create-client-build/' + ctrl.mobileClient.metadata.name);
+    };
+
+    ctrl.$onDestroy = function(){
+      DataService.unwatchAll(watches);
+    };
+  }
+})();
--- a/app/scripts/directives/mobileClientBuildsRow.js
+++ b/app/scripts/directives/mobileClientBuildsRow.js
@@ -30,7 +30,7 @@
     row.state = {};
     row.buildConfigsInstantiateVersion = APIService.getPreferredVersion('buildconfigs/instantiate');
     row.buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
-    row.buildsVersion = APIService.getPreferredVersion('builds');
+    var buildsVersion = APIService.getPreferredVersion('builds');
     var watches = [];
     var anchor;
 
@@ -45,22 +45,25 @@
 
       if (buildConfigChanged && row.context && !row.watchSet) {
         row.watchSet = true;
-        watches.push(DataService.watch(row.buildsVersion, row.context, function(builds) {
-          var _builds = _.filter(builds.by('metadata.name'), function(build) {
+        watches.push(DataService.watch(buildsVersion, row.context, function(buildData) {
+          var builds = _.filter(buildData.by('metadata.name'), function(build) {
             return _.get(build, 'metadata.labels.buildconfig') === _.get(row, 'apiObject.metadata.name');
           });
-          var sortedBuilds = BuildsService.sortBuilds(_builds, true);
-          row.latestBuild = sortedBuilds[0];
+          var sortedBuilds = BuildsService.sortBuilds(builds, true);
+          row.latestBuild = _.head(sortedBuilds);
           row.historyBuilds = _.tail(sortedBuilds);
           row.toggleExpandForAnchor();
         }));
       }
     };
 
-    row.toggleExpandForAnchor = function() {
+    row.isMatchingAnchor = function() {
       var buildId = _.get(row, 'apiObject.metadata.name');
+      return $location.hash() && $location.hash() !== anchor && buildId === $location.hash();
+    };
 
-      if ($location.hash() && $location.hash() !== anchor && buildId === $location.hash()) {
+    row.toggleExpandForAnchor = function() {
+      if (row.isMatchingAnchor()) {
         anchor = $location.hash();
         row.expanded = 'true';
         var key = 'overview/expand/' + _.get(row, 'apiObject.metadata.uid');
--- a/app/scripts/directives/mobileClientBuildsRow.js
+++ b/app/scripts/directives/mobileClientBuildsRow.js
@@ -1,0 +1,98 @@
+'use strict';
+
+(function() {
+  angular.module('openshiftConsole').component('mobileClientBuildsRow', {
+    controller: [
+      '$location',
+      'APIService',
+      'BuildsService',
+      'DataService',
+      'ListRowUtils',
+      'Navigate',
+      MobileClientBuildsRow
+    ],
+    controllerAs: 'row',
+    bindings: {
+      apiObject: '<'
+    },
+    templateUrl: 'views/mobile-client-builds-row.html'
+  });
+
+  function MobileClientBuildsRow(
+    $location,
+    APIService,
+    BuildsService,
+    DataService,
+    ListRowUtils,
+    Navigate
+  ) {
+    var row = this;
+    row.state = {};
+    row.buildConfigsInstantiateVersion = APIService.getPreferredVersion('buildconfigs/instantiate');
+    row.buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
+    row.buildsVersion = APIService.getPreferredVersion('builds');
+    var watches = [];
+    var anchor;
+
+    _.extend(row, ListRowUtils.ui);
+
+    row.$onChanges = function(changes) {
+      var buildConfigChanged = _.get(changes, 'apiObject.currentValue', false);
+
+      if (buildConfigChanged && !row.context) {
+        row.context = {namespace:_.get(row, 'apiObject.metadata.namespace')};
+      }
+
+      if (buildConfigChanged && row.context && !row.watchSet) {
+        row.watchSet = true;
+        watches.push(DataService.watch(row.buildsVersion, row.context, function(builds) {
+          var _builds = _.filter(builds.by('metadata.name'), function(build) {
+            return _.get(build, 'metadata.labels.buildconfig') === _.get(row, 'apiObject.metadata.name');
+          });
+          var sortedBuilds = BuildsService.sortBuilds(_builds, true);
+          row.latestBuild = sortedBuilds[0];
+          row.historyBuilds = _.tail(sortedBuilds);
+          row.toggleExpandForAnchor();
+        }));
+      }
+    };
+
+    row.toggleExpandForAnchor = function() {
+      var buildId = _.get(row, 'apiObject.metadata.name');
+
+      if ($location.hash() && $location.hash() !== anchor && buildId === $location.hash()) {
+        anchor = $location.hash();
+        row.expanded = 'true';
+        var key = 'overview/expand/' + _.get(row, 'apiObject.metadata.uid');
+        sessionStorage.setItem(key, row.expanded);
+      }
+    };
+
+    row.startBuild = function() {
+      BuildsService.startBuild(row.apiObject);
+    };
+
+    row.deleteBuildConfig = function() {
+      BuildsService.deleteBuild(row.apiObject);
+    };
+
+    row.navigateToBuildConfig = function() {
+      var resource = _.get(row, 'apiObject.metadata.name');
+      var kind = _.get(row, 'apiObject.kind');
+      var namespace = _.get(row, 'apiObject.metadata.namespace');      
+      return Navigate.resourceURL(resource, kind, namespace);
+    };
+
+    row.toggleBuildHistory = function() {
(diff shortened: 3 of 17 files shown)


benjaminapetersen:
@sedroche Just taking a quick look here before a deep dive, have a couple observations:
- Can you provide some basic instructions for how to run / test this code?  That would be helpful.  Perhaps you have a template you can share.
- Visually you will prob need some UXD feedback. Off the cuff I see a few things:
    - Columns need to line up with existing columns.  The "Download" button is clearly out of place.
        - Actually, I don't think columns is necessarily the right display for this data.  The above pipelines are in a row to visually indicate a progression between states.  The information "Build #1", "Complete", "View log", "3 hours ago", "Duration" does not indicate progression. I would stack this as you see we do with other resource details.
    - Button color consistency... the second "Download", should it also be blue?    
    - The placement of the QR code may need some adjusting.  Visually it seems out of place.

Do you have a screenshot handy with your new `build-row` next to some of our other `overview` resources?  I think that will help verify visual consistency.

app/scripts/directives/mobileClientBuildsRow.js
```diff
@@ -0,0 +1,98 @@
...
+
+    row.$onChanges = function(changes) {
+      var buildConfigChanged = _.get(changes, 'apiObject.currentValue', false);
+
+      if (buildConfigChanged && !row.context) {
+        row.context = {namespace:_.get(row, 'apiObject.metadata.namespace')};
+      }
+
+      if (buildConfigChanged && row.context && !row.watchSet) {
+        row.watchSet = true;
```
benjaminapetersen: `watchSet` can be local.

app/scripts/directives/mobileClientBuildsList.js
```diff
@@ -0,0 +1,68 @@
...
+        }));
+      }
+    };
+
+    ctrl.goToBuildConfig = function() {
+      if ($location.hash() && anchor !== $location.hash()) {
+        anchor = $location.hash();
+
+        $timeout(function() {
+          $anchorScroll.yOffset = 100;
```
benjaminapetersen: What is the significance of `100`?  I assume `px` value, and that this could be thrown off by CSS changes.  Could this be a named variable giving us a better clue what its relationship is to the UI?

app/scripts/directives/mobileClientBuildsList.js
```diff
@@ -0,0 +1,68 @@
...
+          $anchorScroll();
+        }, 400);
```
benjaminapetersen: Is the `400` arbitrary?

app/scripts/directives/mobileClientBuildsList.js
```diff
@@ -0,0 +1,68 @@
...
+    APIService,
+    DataService
+  ) {
+    var ctrl = this;
+    var buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
+    var watches = [];
+    var anchor;
+
+    ctrl.$onChanges = function(changes) {
+      var mobileClientChanged = _.get(changes, "mobileClient.currentValue", false);
```
benjaminapetersen: Change `"mobileClient.currentValue"` to `'mobileClient.currentValue'` to keep quotes consistent.

app/scripts/directives/mobileClientBuildsList.js
```diff
@@ -0,0 +1,68 @@
...
+    $location,
+    APIService,
+    DataService
+  ) {
+    var ctrl = this;
+    var buildConfigsVersion = APIService.getPreferredVersion('buildconfigs');
+    var watches = [];
+    var anchor;
+
+    ctrl.$onChanges = function(changes) {
```
benjaminapetersen: You have a lot of similarities between the `$onChanges` functions, aside from different variable names.  Possible to factor any of it out into shared utility functions via a service?

app/scripts/directives/mobileClientBuildsList.js
```diff
@@ -0,0 +1,68 @@
...
+      var mobileClientChanged = _.get(changes, "mobileClient.currentValue", false);
+      if (mobileClientChanged && !ctrl.context) {
+        ctrl.context = {namespace: _.get(ctrl, 'mobileClient.metadata.namespace')};
+      }
+
+      if (mobileClientChanged && !ctrl.buildsWatchSet) {
+        ctrl.buildsWatchSet = true;
+        watches.push(DataService.watch(buildConfigsVersion, ctrl.context, function(buildConfigs) {    
+          ctrl.mobileBuildConfigs = _.filter(buildConfigs.by('metadata.name'), function(buildConfig) {
+            return _.get(buildConfig, 'metadata.labels.mobile-client-build') === 'true' && _.get(buildConfig, 'metadata.labels.mobile-client-id') === _.get(ctrl, 'mobileClient.metadata.name');
```
benjaminapetersen: Mind breaking the line on the `&&`?

app/scripts/directives/jqQrCode.js
```diff
@@ -0,0 +1,31 @@
...
+
+      qrCodeContainer.qrcode({
+        text: $scope.content,
+        size: 250
+      });
+
+      $scope.$watch('content', function () {
+        qrCodeContainer.find('canvas').remove();
+
+        qrCodeContainer.qrcode({
```
benjaminapetersen: duplication of the same 4 lines.

app/scripts/directives/jqQrCode.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+/**
```
benjaminapetersen: We no longer use the `@ngdoc` annotations, feel free to remove.

bower.json
```diff
@@ -20,6 +20,7 @@
     "moment-timezone": "0.5.3",
     "patternfly": "3.43.0",
     "jquery": "3.2.1",
+    "jquery-qrcode": "*",
```
benjaminapetersen: Since this will affect the shippable code, pin it to an exact version.  We are more flexible with `package.json` since it is our build tools, but anything in `bower.json` should be explicit.  We verify our build via CI, so any difference in version will fail tests.

sedroche:
@benjaminapetersen Thanks for the review.

- There are some prerequisites needed to run this feature such as a CRD and a [set of services](https://github.com/aerogearcatalog) so the easiest way to get setup is to use [this installer](https://github.com/aerogear/mobile-core/blob/master/docs/walkthroughs/local-setup.adoc) which sets up a cluster with the CRD and broker correctly.

That cluster should have an image containing all the mobile work but if you want to work from source just point your local UI at the cluster and checkout https://github.com/aerogear/origin-web-console `aerogear-mcp` branch.

When that is installed go to the catalog and provision an Android App:

android-app

And this particular piece of work is connected to a specific service so provision a Mobile CICD as well.

mobile-cicd

When those are provisioned you should have a Mobile Client on the overview screen:

mobileclientoverview

If you follow the link (something like `console/project/test/browse/mobile-clients/myapp-android?tab=configuration`) into the Mobile Client view, from there you need to configure the build. See these [docs](https://github.com/aerogear/mobile-core/blob/master/docs/walkthroughs/create-mobile-client-build.adoc#creating-a-build) on how to create one.

A configuration that I know works is:
- Git Repository URL = https://github.com/feedhenry-templates/helloworld-android-gradle
- Git Reference: 5xjenkinsfile

create-build-2

Once that is done you should see the build component on the builds tab and you should be able to create builds like in the description images

sedroche:
@benjaminapetersen 
We have been collaborating with Chris Shinn on UXD folowing these[ wireframes](https://redhat.invisionapp.com/share/GRJJGUHNSET#/screens/295058247).
@cshinn  what do you think? -> https://github.com/openshift/origin-web-console/pull/3020#issuecomment-397329858

sedroche:
@benjaminapetersen I will go through the comments now but about this comment:

> Actually, I don't think columns is necessarily the right display for this data. The above pipelines are in a row to visually indicate a progression between states. The information "Build #1", "Complete", "View log", "3 hours ago", "Duration" does not indicate progression. I would stack this as you see we do with other resource details

Do you mean to use this table?
table-resource

cshinn:
@sedroche I agree with @benjaminapetersen that the spacing seems a bit off in a couple places, but I don't necessarily think the info in the build history should be stacked.

Grouping pieces of information together rather than distributing them evenly would help with organization. For instance, moving the build status and log actions closer together in the build history rows would better mimic the organization of the current build while taking up less vertical space. Reducing the row height of the history rows would also help to show that they're subordinate to the current build. Duration information can be spread out a bit more because it shows similar information to the pipeline illustration and will help a wide row feel less "empty". Download buttons should be aligned to the right side of their container. 

mobile control panel 17

benjaminapetersen:
@openshift/team-ux-review

sedroche:
@cshinn @benjaminapetersen 
Styling updates.

updated-build-row

cshinn:
@sedroche I think that looks clearer. It might be good to use pf-button-small for the grey download buttons and give a bit more vertical padding, maybe ~13px (or closest pf spacer) top and bottom of the button and 20 on the left and right sides of the box to make it look a bit less cramped. Matching the above text color on "3 hours ago" would be good as well

benjaminapetersen:
@sedroche ok working to see if I can get this installed so I can look at it locally.

app/scripts/directives/mobileClientDownloadUrl.js
```diff
@@ -0,0 +1,49 @@
...
+      showPanel: '<'
+    },
+    templateUrl: 'views/mobile-client-download-url.html'
+  });
+
+  function MobileClientDownloadUrl(
+    APIService,
+    DataService
+  ) {
+    var ctrl = this;
```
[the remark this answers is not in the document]
sedroche: I have created a PR for this here -> https://github.com/openshift/origin-web-common/pull/340

app/scripts/directives/mobileClientBuildsRow.js
```diff
@@ -0,0 +1,98 @@
...
+        }));
+      }
+    };
+
+    row.toggleExpandForAnchor = function() {
+      var buildId = _.get(row, 'apiObject.metadata.name');
+
+      if ($location.hash() && $location.hash() !== anchor && buildId === $location.hash()) {
+        anchor = $location.hash();
+        row.expanded = 'true';
```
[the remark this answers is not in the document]
sedroche: We are using `ListRowUtils.ui` functionality here. This value gets stored into `sessionStorage` so it is expected to be a string by `ListRowUtils.ui`

app/scripts/directives/mobileClientBuildsList.js
sedroche: This was the height of the header on the page. We need this offset to scroll our components correctly. I have renamed the variable to suit.

app/scripts/directives/mobileClientBuildsList.js
sedroche: It is arbitrary. This handles the case where the DOM is still being generated and the anchor doesn't exist yet. I am not sure this is the best way to handle this, have you any examples of handling this elsewhere?

app/scripts/directives/mobileClientBuildsList.js
sedroche: I left this as is as it just attaching event handlers. I am not sure there is too much benefit from having this within a service.

sedroche:
@cshinn Updated with the last spacing changes

2ndchanges

cshinn:
I think that looks better. @jennyhaines what are your thoughts?

sedroche:
@benjaminapetersen I have applied some feedback if you can take another look please.

jennyhaines:
@sedroche @cshinn - awesome visual improvements! Just some points/questions:

- Is the repo url supposed to function as a link? If so, I'd use the link style that Chris showed in his mockup (blue coloring plus icon): link

- I think that the title/value pairs (like "Repo URL" and "https: //github...") should be spaced much closer together. (140px closer, in my opinion) like this: titlevaluepairs rather than this: titlevaluepairs-before

- The "View Log" links might work better spaced 8px closer to the "Complete"/"Failed" statuses

- I agree with Chris that the "27 minutes ago" information underneath the build name, and the "3 days ago" information within the build history rows should be lighter. In our PatternFly color palette, the lightest we should go for text is pf-black-600 (#72767b) to stay accessible. 

- Right now these carots carot should actually stay black like this: carot2

sedroche:
@jennyhaines I have updated with your suggestions.
The repo url is for cloning a repo so it is not supposed to be linkable. There wasn't a fixed distance on the title/value pairs but I have tried to match to your pic. Let me know if anything needs updating.

This is how the component looks now:

jenny-changes

app/views/mobile-client-builds-row.html
```diff
@@ -0,0 +1,125 @@
...
+      </div>
+    </div>
+  </div>
+  <div class="list-pf-expansion collapse" ng-if="row.expanded" ng-class="{ in: row.expanded }">
+    <div class="list-pf-container">
+      <div class="expanded-section no-margin">
+        <div class="component-label section-label">BUILD CONFIG</div>
+        <div class="build-config-details row">
+          <div class="col-sm-12 col-md-12 col-lg-6">
+            <b class="title">Repo Url</b>
```
serenamarie125: Please change URL to be in all caps

app/views/mobile-client-builds-row.html
```diff
@@ -0,0 +1,125 @@
...
+            <span>{{row.apiObject.spec.source.git.uri}}</span>
+          </div>
+          <div class="col-sm-12 col-md-12 col-lg-6">
+            <b class="title">Branch</b>
```
serenamarie125: please add a colon "Branch:"

app/views/mobile-client-builds-row.html
```diff
@@ -0,0 +1,125 @@
...
+            <span>{{row.apiObject.spec.source.git.ref}}</span>
+          </div>
+        </div>
+        <div class="build-config-details row">
+          <div class="col-sm-12 col-md-12 col-lg-6">
+            <b class="title">Build Platform</b>
```
serenamarie125: please add a colon "Build Platform:"

app/views/mobile-client-builds-row.html
```diff
@@ -0,0 +1,125 @@
...
+            <span>{{row.apiObject | label:'mobile-client-build-platform'}}</span>
+          </div>
+          <div class="col-sm-12 col-md-12 col-lg-6">
+            <b class="title">Jenkinsfile Path</b>
```
serenamarie125: please add a colon "Jenkinsfile Path:"

dist/scripts/templates.js
```diff
@@ -10839,6 +10839,219 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
...
+    "<div class=\"list-pf-container\">\n" +
+    "<div class=\"expanded-section no-margin\">\n" +
+    "<div class=\"component-label section-label\">BUILD CONFIG</div>\n" +
+    "<div class=\"build-config-details row\">\n" +
+    "<div class=\"col-sm-12 col-md-12 col-lg-6\">\n" +
+    "<b class=\"title\">Repo Url</b>\n" +
+    "<span>{{row.apiObject.spec.source.git.uri}}</span>\n" +
+    "</div>\n" +
+    "<div class=\"col-sm-12 col-md-12 col-lg-6\">\n" +
+    "<b class=\"title\">Branch</b>\n" +
```
serenamarie125: please add a colon "Branch:"

dist/scripts/templates.js
```diff
@@ -10839,6 +10839,219 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
...
+    "<span>{{row.apiObject.spec.source.git.ref}}</span>\n" +
+    "</div>\n" +
+    "</div>\n" +
+    "<div class=\"build-config-details row\">\n" +
+    "<div class=\"col-sm-12 col-md-12 col-lg-6\">\n" +
+    "<b class=\"title\">Build Platform</b>\n" +
```
serenamarie125: please add a colon "Build Platform:"

dist/scripts/templates.js
```diff
@@ -10839,6 +10839,219 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
...
+    "<span>{{row.apiObject | label:'mobile-client-build-platform'}}</span>\n" +
+    "</div>\n" +
+    "<div class=\"col-sm-12 col-md-12 col-lg-6\">\n" +
+    "<b class=\"title\">Jenkinsfile Path</b>\n" +
```
serenamarie125: please add a colon "Jenkinsfile Path:"

dist/scripts/templates.js
```diff
@@ -10839,6 +10839,219 @@ angular.module('openshiftConsoleTemplates', []).run(['$templateCache', function(
...
+    "</div>\n" +
+    "</div>\n" +
+    "</div>\n" +
+    "<div class=\"list-pf-expansion collapse\" ng-if=\"row.expanded\" ng-class=\"{ in: row.expanded }\">\n" +
+    "<div class=\"list-pf-container\">\n" +
+    "<div class=\"expanded-section no-margin\">\n" +
+    "<div class=\"component-label section-label\">BUILD CONFIG</div>\n" +
+    "<div class=\"build-config-details row\">\n" +
+    "<div class=\"col-sm-12 col-md-12 col-lg-6\">\n" +
+    "<b class=\"title\">Repo Url</b>\n" +
```
serenamarie125: please add a colon & change to "Repository URL:"

app/scripts/directives/mobileClientBuildsRow.js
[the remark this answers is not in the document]
benjaminapetersen: Isn't it evaluated back out into a real `boolean`?  Fine that data is stringified into `sessionStorage`, but it should be evaluated before use in JS.

benjaminapetersen:
@sedroche with so many comments, I may suggest you post a reply as you address things, esp if you are squash/rebasing as you go.  It will make it easier for us to follow-up with your changes.  

You could also wait to squash until we are ready for final review.  That might help keep the process moving more quickly.

app/scripts/directives/mobileClientBuildsList.js
benjaminapetersen: But you are setting `anchor` to `$location.hash()`, do you mean an html anchor? 

This is a bit more confusing in that the function is `goToBuildConfig()` but all of the contents are in a single `if`. Is there a potential that this will be clicked but nothing will happen?

Ah, hold on, this isn't a `view` callback. Perhaps just do `var goToBuildConfig` to make it clear this isn't a callback.

app/scripts/directives/mobileClientBuildsList.js
```diff
@@ -0,0 +1,70 @@
...
+    var anchor;
+    var PAGE_HEADER_HEIGHT = 100;
+
+    ctrl.$onChanges = function(changes) {
+      var mobileClientChanged = _.get(changes, 'mobileClient.currentValue', false);
+      if (mobileClientChanged && !ctrl.context) {
+        ctrl.context = {namespace: _.get(ctrl, 'mobileClient.metadata.namespace')};
+      }
+
+      if (mobileClientChanged && !ctrl.buildsWatchSet) {
```
benjaminapetersen: It may be helpful to introduce some named functions here to self-document.  I want to know why `$onChanges` could trigger `goToBuildConfig()`.

app/scripts/directives/mobileClientBuildsList.js
benjaminapetersen: Do we do this kind of scrolling elsewhere?

Perhaps also comment on the specific CSS class name that it matches.  This kind of thing can very easily break when there are CSS changes. 

Possible to use `element` within the `directive` and get the actual `y` coords of the component you want to navigate to && do a direct scroll?   Not sure how viable this is within your context.

app/views/mobile-client-builds-row.html
```diff
@@ -0,0 +1,125 @@
...
+        </ul>
+      </div>
+    </div>
+  </div>
+  <div class="list-pf-expansion collapse" ng-if="row.expanded" ng-class="{ in: row.expanded }">
+    <div class="list-pf-container">
+      <div class="expanded-section no-margin">
+        <div class="component-label section-label">BUILD CONFIG</div>
+        <div class="build-config-details row">
+          <div class="col-sm-12 col-md-12 col-lg-6">
```
benjaminapetersen: This is a tad heavy-handed.  `<div class="col-sm-12 col-md-12 col-lg-6">` can just be `<div class="col-sm-12  col-lg-6">`.  You should only have to add the alternate class when it changes, `-md` is currently the same as `-sm`.

app/views/mobile-client-builds-row.html
```diff
@@ -0,0 +1,125 @@
...
+            <b class="title">Repo Url</b>
+            <span>{{row.apiObject.spec.source.git.uri}}</span>
+          </div>
+          <div class="col-sm-12 col-md-12 col-lg-6">
```
benjaminapetersen: drop `col-md-12`?

app/views/mobile-client-builds-row.html
```diff
@@ -0,0 +1,125 @@
...
+            <b class="title">Branch</b>
+            <span>{{row.apiObject.spec.source.git.ref}}</span>
+          </div>
+        </div>
+        <div class="build-config-details row">
+          <div class="col-sm-12 col-md-12 col-lg-6">
```
benjaminapetersen: drop `col-md-12`?

app/views/mobile-client-builds-row.html
```diff
@@ -0,0 +1,125 @@
...
+            <b class="title">Build Platform</b>
+            <span>{{row.apiObject | label:'mobile-client-build-platform'}}</span>
+          </div>
+          <div class="col-sm-12 col-md-12 col-lg-6">
```
benjaminapetersen: drop `col-md-12`?

app/views/mobile-client-download-url.html
```diff
@@ -0,0 +1,13 @@
+<div ng-if="$ctrl.showPanel" class="mobile-client-download-url col-md-12">
+  <div ng-class="{'spinner spinner-lg': !$ctrl.url}" aria-hidden="true"></div>
+  <div ng-if="$ctrl.url">
+    <div class="col-md-6">
+      <p>Download from URL:</p>
+      <a ng-href="{{$ctrl.url}}" target="_blank">{{$ctrl.url}}</a>
+    </div>
+    <div class="col-md-6">
+      <span> Or scan QR code: </span>
```
benjaminapetersen: `<span>` or should be `<p>` to match?  Also, spacing.

app/views/mobile-client-build-history-item.html
```diff
@@ -0,0 +1,57 @@
...
+          </span>
+        </span>
+      </div>
+    </div>
+  </div>
+  <div class="download-trigger col-xs-3 col-sm-2 col-md-2">
+    <button ng-if="$ctrl.build.status.phase === 'Complete'" ng-click="$ctrl.toggleBuildDownloadUrl()" class="btn btn-default btn-sm">Download</button>
+  </div>
+  <div class="col-md-12">
+    <mobile-client-download-url show-panel="$ctrl.buildDownloadPanelExpanded" mobile-build="$ctrl.build"></mobile-client-download-url>
```
benjaminapetersen: This is also more like `<mobile-client-download-links>`.

app/scripts/directives/mobileClientDownloadUrl.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+(function() {
```
benjaminapetersen: This might be better named `mobileClientDownloadLinks` as it is going to print out a number of links, not just a url (string).  

See [this comment](https://github.com/openshift/origin-web-console/pull/3020/files#r196550463) as well.

app/scripts/directives/mobileClientDownloadUrl.js
```diff
@@ -0,0 +1,53 @@
...
+    var ctrl = this;
+    var annotationName = $filter('annotationName');
+    var BUILD_DOWNLOAD_ANNOTATION = annotationName('mobileGetBuildDownload');
+    var SET_BUILD_DOWNLOAD_ANNOTATION = annotationName('mobileSetBuildDownload');
+    var buildsVersion = APIService.getPreferredVersion('builds');
+
+    ctrl.$onChanges = function(changes) {
+      var mobileBuildChanges = changes.mobileBuild && changes.mobileBuild.currentValue;
+      var context = {namespace:_.get(ctrl, 'mobileBuild.metadata.namespace')};
+
```
benjaminapetersen: I suggest some named functions here again, to be more self-documenting.

app/styles/_mobile-client-builds-list.less
```diff
@@ -0,0 +1,174 @@
...
+
+            .col-sm-12 {
+              padding-left: 10px;
+            }
+
+            .info {
+              display: inline;
+              padding-right: 12px;
+
+              &.logs {
```
benjaminapetersen: Does the `&.logs` here tightly couple the html?

app/styles/_mobile-client-builds-list.less
```diff
@@ -0,0 +1,174 @@
...
+  .build-config-details {
+    .title {
+      display: inline-block;
+      margin-right: 40px;
+      width: 100px;
+    }
+  }
+
+  .build-timestamp {
+    color: #72767b;
```
benjaminapetersen: Is there a `$pf-color-var` that can be used here?

sedroche:
This PR depends on https://github.com/openshift/origin-web-common/pull/340

PR #3021, opened 2018-06-12
List mobile service to mobile client connections

sedroche:
Displaying a list of mobile services that can be bound to a mobile client.

servicestobind

A binding can be created from within the component.

empty-connected
pending-binding

Each service will have configuration created during a binding which will be displayed to the user in this component.
binding-config

**NOTE:** This PR makes use of https://github.com/openshift/origin-web-common/pull/334 to identify specific bindings between client <-> service
and https://github.com/openshift/origin-web-common/pull/320
and https://github.com/openshift/origin-web-common/pull/341

### DIFF
--- a/app/scripts/directives/bindService.js
+++ b/app/scripts/directives/bindService.js
@@ -47,6 +47,8 @@
     var serviceInstancesVersion = APIService.getPreferredVersion('serviceinstances');
     var serviceClassesVersion = APIService.getPreferredVersion('clusterserviceclasses');
     var servicePlansVersion = APIService.getPreferredVersion('clusterserviceplans');
+    var mobileBindingProviderAnnotation = $filter('annotationName')('mobileBindingProviderId');
+    var mobileBindingConsumerAnnotation = $filter('annotationName')('mobileBindingConsumerId');
 
     var preselectService = function(){
       var newestReady;
@@ -200,13 +202,12 @@
 
     var getMobileBindingMetadata = function(svcToBind, serviceClass) {
       var consumerId = _.get(ctrl, 'parameterData.CLIENT_ID');
-
+      var annotations = {};
+      annotations[mobileBindingConsumerAnnotation] = consumerId;
+      annotations[mobileBindingProviderAnnotation] = _.get(svcToBind, 'metadata.name');
       return {
         generateName: consumerId.toLowerCase() + '-' + _.get(serviceClass, 'spec.externalMetadata.serviceName').toLowerCase() + '-',
-        annotations: {
-          'binding.aerogear.org/consumer':  consumerId,
-          'binding.aerogear.org/provider': _.get(svcToBind, 'metadata.name')
-        }
+        annotations: annotations
       };      
     };
 
--- a/app/scripts/directives/bindService.js
+++ b/app/scripts/directives/bindService.js
@@ -19,6 +19,7 @@
       target: '<',
       project: '<',
       onClose: '<',
+      onBindCreated: '<',
       parameterData: '<'
     },
     templateUrl: 'views/directives/bind-service.html'
@@ -197,7 +198,7 @@
       });
     };
 
-    var getBindingMetadata = function(svcToBind, serviceClass) {
+    var getMobileBindingMetadata = function(svcToBind, serviceClass) {
       var consumerId = _.get(ctrl, 'parameterData.CLIENT_ID');
 
       return {
@@ -259,24 +260,15 @@
       }
     };
 
-    ctrl.bindService = function() {
-      var svcToBind = ctrl.target.kind === 'ServiceInstance' ? ctrl.target : ctrl.serviceToBind;
-      var application = ctrl.bindType === 'application' ? ctrl.appToBind : undefined;
-
+    var createBinding = _.bind(function(svcToBind, application, serviceClass, parameterData, bindingMeta) {
       var context = {
         namespace: _.get(svcToBind, 'metadata.namespace')
       };
-
-      var serviceClass = BindingService.getServiceClassForInstance(svcToBind, ctrl.serviceClasses);
-
-      var bindingMeta = {};
-      if (ctrl.isMobileEnabled && isMobileService(serviceClass)) {
-        bindingMeta = getBindingMetadata(svcToBind, serviceClass);
-      }
-
-      BindingService.bindService(svcToBind, application, serviceClass, ctrl.parameterData, bindingMeta).then(function(binding){
+      BindingService.bindService(svcToBind, application, serviceClass, parameterData, bindingMeta).then(function(binding){
         ctrl.binding = binding;
         ctrl.error = null;
+
+        ctrl.bindCreated(ctrl.binding);
 
         bindingWatch = DataService.watchObject(BindingService.bindingResource, _.get(ctrl.binding, 'metadata.name'), context, function(binding) {
           ctrl.binding = binding;
@@ -284,11 +276,37 @@
       }, function(e) {
         ctrl.error = e;
       });
+    }, ctrl);
+
+    var bindMobileService = _.bind(function(svcToBind, serviceClass) {
+      var bindingMeta = getMobileBindingMetadata(svcToBind, serviceClass);
+      createBinding(svcToBind, undefined, serviceClass, ctrl.parameterData, bindingMeta);
+    }, ctrl);
+
+    var bindService = _.bind(function(svcToBind, serviceClass) {
+      var application = ctrl.bindType === 'application' ? ctrl.appToBind : undefined;
+      createBinding(svcToBind, application, serviceClass, ctrl.parameterData);
+    }, ctrl);
+
+    ctrl.bindService = function() {
+      var svcToBind = ctrl.target.kind === 'ServiceInstance' ? ctrl.target : ctrl.serviceToBind;
+      var serviceClass = BindingService.getServiceClassForInstance(svcToBind, ctrl.serviceClasses);
+      if (ctrl.isMobileEnabled && isMobileService(serviceClass)) {
+        bindMobileService(svcToBind, serviceClass);  
+      } else {
+        bindService(svcToBind, serviceClass);
+      }
     };
 
     ctrl.closeWizard = function() {
       if (_.isFunction(ctrl.onClose)) {
         ctrl.onClose();
+      }
+    };
+
+    ctrl.bindCreated = function(binding) {
+      if (_.isFunction(ctrl.onBindCreated)) {
+        ctrl.onBindCreated(binding);
       }
     };
   }
--- a/app/scripts/directives/bindService.js
+++ b/app/scripts/directives/bindService.js
@@ -3,6 +3,7 @@
 (function() {
   angular.module('openshiftConsole').component('bindService', {
     controller: [
+      '$rootScope',
       '$scope',
       '$filter',
       'APIService',
@@ -17,12 +18,14 @@
     bindings: {
       target: '<',
       project: '<',
-      onClose: '<'
+      onClose: '<',
+      parameterData: '<'
     },
     templateUrl: 'views/directives/bind-service.html'
   });
 
-  function BindService($scope,
+  function BindService($rootScope,
+                       $scope,
                        $filter,
                        APIService,
                        ApplicationsService,
@@ -39,6 +42,7 @@
     var bindingWatch;
     var statusCondition = $filter('statusCondition');
     var enableTechPreviewFeature = $filter('enableTechPreviewFeature');
+    var isMobileService = $filter('isMobileService');
     var serviceInstancesVersion = APIService.getPreferredVersion('serviceinstances');
     var serviceClassesVersion = APIService.getPreferredVersion('clusterserviceclasses');
     var servicePlansVersion = APIService.getPreferredVersion('clusterserviceplans');
@@ -187,13 +191,32 @@
       });
     };
 
+    var loadServiceClasses = function() {
+      DataService.list(serviceClassesVersion, {}).then(function (serviceClasses) {
+        ctrl.serviceClasses = serviceClasses.by('metadata.name');
+      });
+    };
+
+    var getBindingMetadata = function(svcToBind, serviceClass) {
+      var consumerId = _.get(ctrl, 'parameterData.CLIENT_ID');
+
+      return {
+        generateName: consumerId.toLowerCase() + '-' + _.get(serviceClass, 'spec.externalMetadata.serviceName').toLowerCase() + '-',
+        annotations: {
+          'binding.aerogear.org/consumer':  consumerId,
+          'binding.aerogear.org/provider': _.get(svcToBind, 'metadata.name')
+        }
+      };      
+    };
+
     $scope.$watch("ctrl.serviceToBind", updateServiceInstance);
 
     ctrl.$onInit = function() {
+      ctrl.isMobileEnabled = $rootScope.AEROGEAR_MOBILE_ENABLED;
       ctrl.serviceSelection = {};
       ctrl.projectDisplayName = $filter('displayName')(ctrl.project);
       ctrl.podPresets = enableTechPreviewFeature('pod_presets');
-      ctrl.parameterData = {};
+      ctrl.parameterData = ctrl.parameterData || {};
 
       ctrl.steps = [ bindFormStep, bindParametersStep, resultsStep ];
       ctrl.hideBack = bindParametersStep.hidden;
@@ -204,6 +227,9 @@
         ctrl.serviceToBind = ctrl.target;
         if (ctrl.podPresets) {
           loadApplications();
+        }
+        if (ctrl.isMobileEnabled) {
+          loadServiceClasses();
         }
       }
       else {
@@ -242,7 +268,13 @@
       };
 
       var serviceClass = BindingService.getServiceClassForInstance(svcToBind, ctrl.serviceClasses);
-      BindingService.bindService(svcToBind, application, serviceClass, ctrl.parameterData).then(function(binding){
+
+      var bindingMeta = {};
+      if (ctrl.isMobileEnabled && isMobileService(serviceClass)) {
+        bindingMeta = getBindingMetadata(svcToBind, serviceClass);
+      }
+
+      BindingService.bindService(svcToBind, application, serviceClass, ctrl.parameterData, bindingMeta).then(function(binding){
         ctrl.binding = binding;
         ctrl.error = null;
 
--- a/app/scripts/directives/mobileServiceInstanceList.js
+++ b/app/scripts/directives/mobileServiceInstanceList.js
@@ -1,0 +1,55 @@
+'use strict';
+
+(function() {
+  angular.module('openshiftConsole').component('mobileServiceInstanceList', {
+    controller: [
+      '$filter',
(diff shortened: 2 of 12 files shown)


benjaminapetersen:
@openshift/team-ux-review

benjaminapetersen:
@cshinn I think similar items related to button placement, etc.

cshinn:
Separating the unbound services more strongly would help get them out of the way for people who don't necessarily want to bind all of their services to a given mobile client. They wouldn't need to expand, but a short message like the PR screenshot could be useful if it's easier for them to expand. In that case, I think it would look better for the button to be fixed width relative to the text inside rather than sized based on the grid.

mobile control panel 13

sedroche:
@cshinn Thanks chris. I'll start on these changes now.

app/scripts/directives/bindService.js
```diff
@@ -187,13 +191,32 @@
...
+        ctrl.serviceClasses = serviceClasses.by('metadata.name');
+      });
+    };
+
+    var getBindingMetadata = function(svcToBind, serviceClass) {
+      var consumerId = _.get(ctrl, 'parameterData.CLIENT_ID');
+
+      return {
+        generateName: consumerId.toLowerCase() + '-' + _.get(serviceClass, 'spec.externalMetadata.serviceName').toLowerCase() + '-',
+        annotations: {
```
benjaminapetersen: I believe the `annotations` `$filter` can be used to get & set annotations.

app/scripts/directives/bindService.js
```diff
@@ -242,7 +268,13 @@
       };
 
       var serviceClass = BindingService.getServiceClassForInstance(svcToBind, ctrl.serviceClasses);
-      BindingService.bindService(svcToBind, application, serviceClass, ctrl.parameterData).then(function(binding){
+
```
benjaminapetersen: I don't know if other kinds of services would cause this function to bloat over time, but it may be wise to branch early:

```JavaScript 
// call separate functions early, which can share another function if necessary to deduplicate.
ctrl.bindService = function() {
   if (ctrl.isMobileEnabled && isMobileService(serviceClass)) {
      bindMobileService();  
   } else {
      bindService();
   }
}
```

app/scripts/directives/mobileServiceInstanceList.js
```diff
@@ -0,0 +1,55 @@
...
+
+    $scope.$on('$destroy', function(){
+        DataService.unwatchAll(watches);
+    });
+
+    ctrl.goToCatalog = function() {
+      Navigate.toProjectCatalog($routeParams.project, {category: 'mobile', subcategory: 'services'});
+    };  
+
+    watches.push(DataService.watch(APIService.getPreferredVersion("clusterserviceclasses"), {namespace: $routeParams.project}, function(serviceClasses){
```
benjaminapetersen: Should this be within `$onInit`?

```JavaScript 
var watchClusterServiceClasses = _.once(function() { ... });
ctrl.$onInit = function() {
   watchClusterServiceClasses();
   // etc.
}
```

app/scripts/directives/mobileServiceInstanceList.js
```diff
@@ -0,0 +1,55 @@
...
+      ctrl.serviceClasses = serviceClasses.by("metadata.name");
+      watches.push(DataService.watch(APIService.getPreferredVersion("serviceinstances"), {namespace: $routeParams.project}, function(serviceInstances){
```
benjaminapetersen: I would probably suggest putting the `var serviceInstanceVersion = APIService.getPreferredVersion('serviceinstances')` at the top.  For the most part we do that everywhere.

app/scripts/directives/mobileServiceInstanceList.js
```diff
@@ -0,0 +1,55 @@
...
+        ctrl.serviceInstances = _.filter(serviceInstances.by('metadata.name'), function(serviceInstance){
```
benjaminapetersen: try running `grunt jshint`, just to ensure spacing & what not won't fail CI.  I think there may be a rule about `function() {`.

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+                              ListRowUtils) {
+    var row = this;
+    var serviceInstanceDisplayName = $filter('serviceInstanceDisplayName');
+    row.serviceBindingsVersion = APIService.getPreferredVersion('servicebindings');
+    row.state = row.state || {};
+    var annotationSpace = "org.aerogear.binding." + this.apiObject.metadata.name;
+    var extendedAnnotationSpace = "org.aerogear.binding-ext." + this.apiObject.metadata.name;
+    var watches = [];
+    var bindingsWatch = false;
+    row.bindingsLimit = _.get(row.serviceClass, "spec.externalMetadata.bindingsLimit", 1);
```
benjaminapetersen: `row.$onInit`?  Not sure if there is a risk on a race condition here, for things like `row.bindingsLimit`.

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+    _.extend(row, ListRowUtils.ui);
+
+    $scope.$on('$destroy', function(){
```
benjaminapetersen: By convention we put `$scope.$on('$destroy'` at the end of the file.

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+        DataService.unwatchAll(watches);
+        DataService.unwatch(bindingsWatch);
+    });
+
+    watches.push(DataService.watchObject(APIService.getPreferredVersion('mobileclients'), $routeParams.mobileclient, {namespace: $routeParams.project}, function(mobileClient) {
```
benjaminapetersen: I think this should all really be called form `$onInit`, rather than be loosely scattered in the `MobileServiceInstanceRow` function body.  Preferably broken down into named functions.

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+        row.mobileClient = mobileClient;
+
+        row.annotations = _.filter(row.mobileClient.metadata.annotations, function(annotation, name){
+          return name.split("/")[0] === annotationSpace;
+        })
+        .map(JSON.parse)
+        .sort(function(a, b) {
```
benjaminapetersen: There are quite a few `lodash` sort functions that might be a bit more expressive than this.

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+          if (a.label < b.label) {
+            return -1;
+          }
+          if (a.label > b.label) {
+            return 1;
+          }
+          return 0;
+        });
+
+        row.serviceType = _.get(row, "serviceClass.spec.externalMetadata.serviceName"); // safe get
```
benjaminapetersen: Feel free to eliminate `// safe get` for final pass.

app/scripts/directives/mobileServiceInstanceRow.js
benjaminapetersen: Possibly applies to anything that deals with `this.apiObject` as well.  

Even if not a race condition issue, its probably a better convention.

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+          .pickBy(function(annotation, name){
+            return name.split("/")[0] === extendedAnnotationSpace;
+          })
+          .transform(function(result, strVal, name) {
+            var key = name.split("/")[1];
+            result[key] = JSON.parse(strVal);
+          })
+          .value();
+
+        row.servicePlanPromise = DataService.list(APIService.getPreferredVersion('clusterserviceplans'), {}, function(servicePlans) {
```
benjaminapetersen: You aren't passing to the view, so `row.servicePlanPromise` isn't needed.

benjaminapetersen: Do you need a fail case? If a user isn't allowed to see cluster scoped resources?

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+          row.providerServicePlan = _.find(servicePlanData, function(plan) {
+            return row.serviceClass.metadata.name === plan.spec.clusterServiceClassRef.name;
+          });
+          var kind = row.mobileClient.kind.toLowerCase();
+          var bindDataPath = 'providerServicePlan.spec.externalMetadata.' + kind + '_bind_parameters_data';
+          var bindData = _.get(row, bindDataPath, []);
+          row.bindData = _(bindData).map(JSON.parse).value();
+        });
+
+        if(!bindingsWatch){
```
benjaminapetersen: `_.once()` to eliminate `bindingsWatch ` check.

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+          bindingsWatch = DataService.watch(APIService.getPreferredVersion("servicebindings"), {namespace: $routeParams.project}, function(bindings){
+            row.bindings = _.filter(bindings.by("metadata.name"), function(binding){
+              return _.get(binding.metadata.annotations, "binding.aerogear.org/consumer") === row.mobileClient.metadata.name &&
```
benjaminapetersen: Perhaps break this line down, at least on the `&&` into two lines.

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+                    _.get(binding.metadata.annotations, "binding.aerogear.org/provider") === row.apiObject.metadata.name;
+            });
+
+            row.bindingInProgress = _.sumBy(row.bindings, function(binding){
```
benjaminapetersen: Assuming if this returns `0` at least one `binding` is in progress?

app/scripts/directives/mobileServiceInstanceRow.js
```diff
@@ -0,0 +1,125 @@
...
+    row.$onChanges = function() {
+      row.displayName = serviceInstanceDisplayName(row.apiObject, row.serviceClass);
+    };
+
+    row.closeOverlayPanel = function() {
+      _.set(row, 'overlay.panelVisible', false);
+    };
+
+    row.showOverlayPanel = function(panelName, state) {
+      row.parameterData = _.reduce(row.bindData, function(acc, current) {
```
benjaminapetersen: `parameterData` is passed in as an attribute, just want to check & verify you do need 2-way data binding here.

benjaminapetersen: Nit: i think we typically use `result` for the accumulator var in a `_.reduce`.  That, or a name that describes the object being built, in this case something like `paramData`.

app/styles/_mobile-service-instance-list.less
```diff
@@ -0,0 +1,32 @@
+.mobile-service-instance-list {
+
+  .note {
+    text-align: center;
+  }
+
+  .mobile-service-instance-row {
+    .icon {
+      clear: both;
+      color: #888;
```
benjaminapetersen: Any `$pf-` var for the color?

app/styles/_mobile-service-instance-list.less
```diff
@@ -0,0 +1,32 @@
...
+    }
+    .list-pf-name {
+      h3 {
+        .logo {
+          width: 30px;
+          height: 30px;
+          margin: 0 10px;
+          display: block;
+          float: left;
+          font-size: 1.6em;
```
benjaminapetersen: `$pf-` var for font-size?

app/views/_mobile-service-instance-row.html
```diff
@@ -0,0 +1,90 @@
+<div class="list-pf-item provisioned-service mobile-service-instance-row" ng-class="{ active: row.expanded }">
+  <div class="list-pf-container" ng-click="row.toggleExpand($event)">
+    <div class="list-pf-chevron">
+      <div ng-include src=" 'views/overview/_list-row-chevron.html' " class="list-pf-content"></div>
+    </div>
+    <div class="list-pf-content">
+      <div class="list-pf-name">
+        <h3>
+          <span class="logo" ng-if="row.serviceClass.spec.externalMetadata.imageUrl">
+            <img src="{{row.serviceClass.spec.externalMetadata.imageUrl}}">
```
benjaminapetersen: `title` or `alt` on the `img` for accessibility purposes?

app/scripts/directives/bindService.js
sedroche: Yes, I will update this.

app/scripts/directives/bindService.js
sedroche: :+1:

app/scripts/directives/bindService.js
sedroche: Annotations are already ready to go -> https://github.com/openshift/origin-web-common/pull/341

app/scripts/directives/mobileServiceInstanceList.js
sedroche: :+1:

app/scripts/directives/mobileServiceInstanceList.js
sedroche: Yep, I will update these

app/scripts/directives/mobileServiceInstanceList.js
sedroche: There doesn't seem to be a rule but i have updated this style anyway

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: yes, makes sense. Will update.

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: :+1:

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: :+1:

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: Cheers. `sortBy` did the job.

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: The reason this was not in `$onInit` was that we extend `ListRowUtils.ui` which has it's own `$onInit`.

```.js
 _.extend(row, ListRowUtils.ui);
```

I have enclosed this in `$onChanges` guarded by `isFirstChange`
```.js
    row.$onChanges = function(changes) {
      row.displayName = serviceInstanceDisplayName(row.apiObject, row.serviceClass);

      if (changes.apiObject && changes.apiObject.isFirstChange()) {
        row.getClient()
          .then(function(client) {
            row.mobileClient = client;
            row.serviceType = _.get(row, 'serviceClass.spec.externalMetadata.serviceName');

            row.sortAnnotations();
            row.getBindingParametersData();
            row.watchServiceBindings();
            row.watchClient();
          });       
      }
    };
```

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: :+1:

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: `row.servicePlanPromise` will be removed.

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: This watch is removed in favour of a get on the mobile client instead of `watchObject`

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: :+1:

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: Yes that is correct. I have slightly updated this to:

```.js
        row.bindingInProgress = _.some(clientBindings, function(binding) {
          return !isBindingReady(binding);
        });
```

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: I am not sure about the call to `APIService.getPreferredVersion('clusterserviceplans')`. The bindService uses this APi to get the binding parameters so we did the same -> https://github.com/openshift/origin-web-console/blob/master/app/scripts/directives/bindService.js#L177

Is there an authorization for creating bindings I need to apply?

app/scripts/directives/mobileServiceInstanceRow.js
sedroche: No, I believe one way is correct. The child component only uses `parameterData`

sedroche: :+1:

app/styles/_mobile-service-instance-list.less
sedroche: Could you point to where the `$pf-` vars are defined please?

app/views/_mobile-service-instance-row.html
sedroche: :+1:

sedroche:
@cshinn I have updated this view with your suggested changes. Hopefully everything looks good, let me know if you require any more changes. 

### Unbound Services
unbound-chris

### Bound and unbound
1and1-chris

### Both Bound
bothbound

### Creating binding flow
binding

### Deleting binding flow
unbinding

sedroche:
@benjaminapetersen I have applied some feedback here if you can take a second look. Cheers.

app/styles/_mobile-service-instance-list.less
benjaminapetersen: The console's own vars are in `styles/_variables.less`. The paths to the PF files we are importing can be found in `styles/main.less`.  That will be in `bower_components/patternfly/`.  A good start would be `patternfly/src/less/variables.less`.

cshinn:
@sedroche looks good! I'm not sure if this is the correct PR to address it, but it'd probably be good to get mobile service icons with transparent backgrounds so that they don't look off when hovered/selected

beanh66:
@cshinn We removed the section header called "DETAILS" for the CNV flows. I believe @rhamilto pointed out that this style is reserved for specific items such as BUILDS, NETWORKING, etc. Wondering if the same applies here.
See CNV design: https://redhat.invisionapp.com/static-signed/live-embed/123952085/299406327/6/latest/4DyVgQEyeHerl4kOzzDLZelEuiXvA9iCCak31MlEf0WJO1amQl7RAlxbueoF4lP4FnGow0a1KooiHbjTlEj0xi08QlE/no-migration.png

cshinn:
@beanh66 that's a good point. So you're saying it only goes on object detail views?

PR #3030, opened 2018-06-18
[enterprise-3.9] Lock datatables dependency

spadgett:
Backport of #3028

/assign @benjaminapetersen

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -47,7 +47,8 @@
     "angular-utf8-base64": "0.0.5",
     "file-saver": "1.3.3",
     "origin-web-common": "3.9.4",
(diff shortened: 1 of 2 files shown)


spadgett:
> This is a dependency of a dependency, right? Definitely one of those situations where I wish package.json supported comments :/

Yeah, it's brought in by Patternfly. We're not using it directly.

benjaminapetersen:
Pondering if we should have put it in `resolutions`, to kind-of document this. Annoying, but `bower` isn't super robust.

Prob no matter, hopefully we will have few changes here anyway!

PR #3036, opened 2018-07-06
Prefer bower from node_modules directory

spadgett:
@adammhaile Is there an easy way to test if this fixes the build problem you're seeing?


adammhaile:
Not really... we never do that part manually, just as full OCP build.

adammhaile:
@spadgett 3.11 is totally hosed right now anyways.... it's not like this could break it MORE.

PR #3041, opened 2018-07-23
Bug 1607150 - UI Timeout causing IE 11 to automatically log out

jhadvig:
The bug was cause by the fact that events on the `storage`:
```js
$(window).on('storage', function (event) {
...
}
```
are in IE11 also triggered in the same tab, whereas in other browsers the event is triggered only by duplicate tabs.
For that case we need to check if the `inactivityLogoutKey` was set to `true` before logging out.
The event itself is triggered by [onUserChanged call](https://github.com/openshift/origin-web-console/blob/master/app/scripts/app.js#L595)

@spadgett PTAL

### DIFF
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -585,7 +585,7 @@ angular
     $(window).on('storage', function (event) {
       if (event.originalEvent.key === lastInteractionKey) {
         restartCheckInteractionInterval();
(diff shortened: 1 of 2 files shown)


spadgett:
OK we got past the previous failure. This looks like a flake.

/retest

PR #3050, opened 2018-08-01
Update context switch to use new config value

spadgett:
Uses config option set in https://github.com/openshift/origin-web-console-server/pull/41

/assign @rhamilto


rhamilto:
I assume the DataTable patch version bumps are intentional?

spadgett:
> I assume the DataTable patch version bumps are intentional?

Yeah, it's intentional.

Issue #3051, opened 2018-08-02
When editing DeploymentConfig Editor. Editor show nothing about the  invalidation input fields


linzhaoming:
Version: Openshift Console 3.9

## Problem

I have a Java springboot app, and need to set the special format Env which input editor does not support now, for example `SPRING_KAFKA_BOOTSTRAP-SERVERS`.

1. I can set the env in the YAML Editor in the **Console** successfully.
```
      containers:
        - env:
            - name: SPRING_KAFKA_BOOTSTRAP-SERVERS
              value: test
```
2. After previous setting, I edit the *DeploymentConfig* again in the Edit UI. But whenever I change any input text int the editor, the **Save** Button will  be always **disabled**.  Which essentially is because the input editor will consider the value of  `SPRING_KAFKA_BOOTSTRAP-SERVERS` is invalid. **But the UI Editor displays nothing about this error**.

![image](https://user-images.githubusercontent.com/1192635/43572347-e26c576c-9671-11e8-9258-ae6f61ee7af5.png)

3.  When I click the input of env `SPRING_KAFKA_BOOTSTRAP-SERVERS`, and move away, the **Editor** will show an error message.

![image](https://user-images.githubusercontent.com/1192635/43572499-4888e402-9672-11e8-9545-be36a4f862b6.png)

4. The same problem exists in the *DeploymentConfig*'s Environment Editor
![image](https://user-images.githubusercontent.com/1192635/43572897-2267b702-9673-11e8-8099-8dcc4e99ff7c.png)

## Analyze

1.  The codes is in [app/scripts/controllers/edit/deploymentConfig.js](https://github.com/openshift/origin-web-console/blob/master/app/scripts/controllers/deploymentConfig.js). Line 371-374
```
                 <div class="buttons gutter-top-bottom">
                  <button ng-click="save()" class="btn btn-primary btn-lg" ng-disabled="form.$invalid || form.$pristine || disableInputs">
                    Save
                  </button>
```
 The **Save** button status replys the state of  *form.$invalid*. 

## Expected
- Choice 1: Show the error message when the YAML of the editor is invalid to make it clear
- Choice 2: Many Java SpringBoot apps need the environment setting format for such as *SPRING_ZIPKIN_KAFKA_BOOTSTRAP-SERVERS*.  It would be better to allow the format in the input text.

spadgett:
[...]

PR #3052, opened 2018-08-02
Update key-validator regexp for key-value-editor

jhadvig:
Fixes: https://github.com/openshift/origin-web-console/issues/3051 

@spadgett PTAL


jhadvig:
Since the change was introduced in 1.8, I suppose we want to backport this change all the way to that version ?

spadgett:
@jhadvig I'd probably backport this to enterprise-3.9 and enterprise-3.10

PR #3057, opened 2018-08-03
Update key-validator regexp for key-value-editor

jhadvig:
Backport of https://github.com/openshift/origin-web-console/pull/3052

@spadgett PTAL

### DIFF
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -474,9 +474,9 @@
                     key-placeholder="Name"
                     value-placeholder="Value"
                     value-from-selector-options="valueFromObjects"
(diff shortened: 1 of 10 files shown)


spadgett:
@jhadvig Looks like Jenkins doesn't like the dist files.

jhadvig:
About to re-generate them

jhadvig:
@spadgett PR updated.PTAL

spadgett:
```
[INFO] Verifying that checked in built files under dist match the source...
[FATAL] Built /dist does not match what is committed, run 'grunt build' and include the results in your commit.
```

Interestingly, travis didn't complain

jhadvig:
Will update ASAP

PR #3060, opened 2018-08-14
 Fix wrong 'archiveLink' for operations user

josefkarasek:
Opening this PR mostly to get feedback. I'll open a new PR against 3.9 and 3.10 ones the fix is ready, with proper `grunt build` and `test`.

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1587807

### DIFF
--- a/app/scripts/directives/logViewer.js
+++ b/app/scripts/directives/logViewer.js
@@ -8,9 +8,10 @@ angular.module('openshiftConsole')
     '$window',
     '$filter',
     '$q',
-    'AuthService',
+    'AggregatedLoggingService',
     'APIService',
     'APIDiscovery',
+    'AuthService',
     'DataService',
     'HTMLService',
     'ModalsService',
@@ -21,9 +22,10 @@ angular.module('openshiftConsole')
              $window,
              $filter,
              $q,
-             AuthService,
+             AggregatedLoggingService,
              APIService,
              APIDiscovery,
+             AuthService,
              DataService,
              HTMLService,
              ModalsService,
@@ -434,16 +436,19 @@ angular.module('openshiftConsole')
                 if(!(projectName && containerName && name && url)) {
                   return;
                 }
-                $scope.$watchGroup(['context.project.metadata.name', 'options.container', 'name'], function() {
-                  angular.extend($scope, {
-                    kibanaArchiveUrl: $sce.trustAsResourceUrl(logLinks.archiveUri({
-                                        baseURL: url,
-                                        namespace: $scope.context.project.metadata.name,
-                                        namespaceUid: $scope.context.project.metadata.uid,
-                                        podname: name,
-                                        containername: $scope.options.container,
-                                        backlink: URI.encode($window.location.href)
-                                      }, $filter('annotation')($scope.context.project,'loggingDataPrefix')))
+
+                AggregatedLoggingService.isOperationsUser().then(function(canViewOperationsLogs) {
+                  $scope.$watchGroup(['context.project.metadata.name', 'options.container', 'name'], function() {
+                    angular.extend($scope, {
+                      kibanaArchiveUrl: logLinks.archiveUri({
+                        baseURL: url,
+                        namespace: $scope.context.project.metadata.name,
+                        namespaceUid: $scope.context.project.metadata.uid,
+                        podname: name,
+                        containername: $scope.options.container,
+                        backlink: URI.encode($window.location.href)
+                      }, $filter('annotation')($scope.context.project,'loggingDataPrefix'), canViewOperationsLogs)
+                    });
                   });
                 });
               });
--- a/app/scripts/services/aggregatedLoggingService.js
+++ b/app/scripts/services/aggregatedLoggingService.js
@@ -1,0 +1,43 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .factory("AggregatedLoggingService", function($q, Logger, DataService) {
+
+    // cache previous request
+    var userAllowed;
+
+    // Create SelfSubjectAccessReview request againt authorization API to check whether
+    // current user can view pods/log from 'default' project.
+    // Users who can view such logs are 'operations' users
+    var isOperationsUser = function() {
+      if (userAllowed !== undefined) {
+        // Using cached data.
+        Logger.log("AggregatedLoggingService, using cached user");
+        return $q.when(userAllowed);
+      }
+      Logger.log("AggregatedLoggingService, loading whether user is Operations user");
+      var ssar = {
+        apiVersion: 'authorization.k8s.io/v1',
+        kind: 'SelfSubjectAccessReview',
+        spec: {
+          resourceAttributes: {
+            resource: 'pods/log',
+            namespace: 'default',
+            verb: 'view'
+          }
+        }
+      };
+      return DataService.create({ group: 'authorization.k8s.io', version: 'v1', resource: 'selfsubjectaccessreviews'},
+                        null, ssar, {namespace: 'default'}).then(
+        function(data) {
+          userAllowed = data.status.allowed;
+          return userAllowed;
+        }, function() {
+          return false;
+      });
+    };
+
+    return {
+      isOperationsUser: isOperationsUser
+    };
+  });
--- a/app/scripts/services/logLinks.js
+++ b/app/scripts/services/logLinks.js
@@ -73,7 +73,10 @@ angular.module('openshiftConsole')
       ].join(''));
 
 
-      var archiveUri = function(opts, prefix) {
+      var archiveUri = function(opts, prefix, canViewOperationsLogs) {
+        if(canViewOperationsLogs && (!prefix || prefix.startsWith('project.'))) {
+            prefix = 'project';
+        }
         prefix = prefix || 'project.' + opts.namespace + '.' + opts.namespaceUid;
         opts.index = prefix + '.*';
 
--- a/app/views/directives/logs/_log-viewer.html
+++ b/app/views/directives/logs/_log-viewer.html
@@ -47,16 +47,8 @@
     {{emptyStateMessage}}
   </p>
 
-  <div ng-if="kibanaAuthUrl">
-    <form
-      action="{{kibanaAuthUrl}}"
-      method="POST">
-      <input type="hidden" name="redirect" value="{{kibanaArchiveUrl}}">
-      <input type="hidden" name="access_token" value="{{access_token}}">
-      <button class="btn btn-primary btn-lg">
-        View Archive
-      </button>
-    </form>
+  <div ng-if="kibanaArchiveUrl">
+    <a href="{{kibanaArchiveUrl}}">View Archive</a>
   </div>
 </div>
 
--- a/app/index.html
+++ b/app/index.html
@@ -185,6 +185,7 @@
         <script src="scripts/services/shims/hawtioExtension.js"></script>
         <script src="scripts/services/browserStore.js"></script>
         <script src="scripts/services/discovery.js"></script>
+        <script src="scripts/services/aggregatedLoggingService.js"></script>
         <script src="scripts/services/applicationGenerator.js"></script>
         <script src="scripts/services/navigate.js"></script>
         <script src="scripts/services/nameGenerator.js"></script>
--- a/dist/scripts/scripts.js
+++ b/dist/scripts/scripts.js
@@ -1,20 +1,20 @@
 "use strict";
 
 function OverviewController(e, t, n, r, a, o, i, s, c, l, u, d, m, p, f, g, v, h, y, b, S, C, w, P, j, k, I, R, E, T, N) {
-function D(e) {
+function A(e) {
 var t = e.metadata.ownerReferences;
 return t ? _(t).filter({
 kind: "OfflineVirtualMachine"
 }).map("uid").first() : null;
 }
-function A() {
+function D() {
 if (B.offlineVirtualMachines && B.virtualMachines && B.pods) {
 var e = _(B.pods).values().filter(function(e) {
 return !!_.get(e, 'metadata.labels["kubevirt.io/vmUID"]');
 }).keyBy('metadata.labels["kubevirt.io/vmUID"]').value(), t = _(B.virtualMachines).values().filter(function(e) {
-return !!D(e);
+return !!A(e);
 }).keyBy(function(e) {
-return D(e);
+return A(e);
 }).value(), n = [];
 _.each(B.offlineVirtualMachines, function(r) {
 var a = r.metadata.uid, o = t[a];
@@ -130,14 +130,14 @@ return _.isEmpty(be.pipelinesByDeploymentConfig[t]);
 B.deploymentConfigsNoPipeline = _.sortBy(e, "metadata.name"), B.pipelineViewHasOtherResources = !(_.isEmpty(B.deploymentConfigsNoPipeline) && _.isEmpty(B.vanillaReplicationControllers) && _.isEmpty(B.deployments) && _.isEmpty(B.vanillaReplicaSets) && _.isEmpty(B.statefulSets) && _.isEmpty(B.monopods));
 }, Ne = function() {
 B.disableFilter = "pipeline" === B.viewBy && _.isEmpty(B.pipelineBuildConfigs);
-}, De = function(e) {
+}, Ae = function(e) {
 return b.getLabelSelector().select(e);
-}, Ae = [ "metadata.name", "spec.clusterServiceClassExternalName" ], $e = function(e) {
-return y.filterForKeywords(e, Ae, be.filterKeywords);
+}, De = [ "metadata.name", "spec.clusterServiceClassExternalName" ], $e = function(e) {
+return y.filterForKeywords(e, De, be.filterKeywords);
 }, Be = function(e) {
 switch (B.filterBy) {
 case "label":
-return De(e);
+return Ae(e);
 
 case "name":
 return $e(e);
@@ -157,7 +157,7 @@ B.filteredDeploymentConfigs = Be(B.deploymentConfigs), B.filteredReplicationCont
 B.viewBy = localStorage.getItem(Oe) || "app", e.$watch(function() {
 return B.viewBy;
 }, function(e) {
-localStorage.setItem(Oe, e), Ne(), Ae = "app" === B.viewBy ? [ "metadata.name", "metadata.labels.app" ] : [ "metadata.name" ], Le(), "pipeline" === B.viewBy ? b.setLabelSuggestions(he) : b.setLabelSuggestions(ve);
+localStorage.setItem(Oe, e), Ne(), De = "app" === B.viewBy ? [ "metadata.name", "metadata.labels.app" ] : [ "metadata.name" ], Le(), "pipeline" === B.viewBy ? b.setLabelSuggestions(he) : b.setLabelSuggestions(ve);
 }), d.DISABLE_OVERVIEW_METRICS || (C.isAvailable(!0).then(function(e) {
 be.showMetrics = e;
 }), e.$on("metrics-connection-failed", function(e, t) {
@@ -240,7 +240,7 @@ _.isEmpty(e) || (b.addLabelSuggestionsFromResources(e, he), "pipeline" === B.vie
 }, nt = function(e) {
 return "Succeeded" !== e.status.phase && "Failed" !== e.status.phase && (!H(e, "openshift.io/deployer-pod-for.name") && (!O(e, "openshift.io/build.name") && "slave" !== H(e, "jenkins")));
 }, rt = function() {
-be.podsByOwnerUID = j.groupByOwnerUID(B.pods), B.monopods = _.filter(be.podsByOwnerUID[""], nt), A();
+be.podsByOwnerUID = j.groupByOwnerUID(B.pods), B.monopods = _.filter(be.podsByOwnerUID[""], nt), D();
 }, at = function(e) {
 return !!_.get(e, "status.replicas") || (!O(e, "deploymentConfig") || x(e));
 }, ot = function(e) {
@@ -495,13 +495,13 @@ poll: V,
 pollInterval: 6e4
 })), e.KUBEVIRT_ENABLED) {
 Rt.push(m.watch(N.offlineVirtualMachine, r, function(e) {
-B.offlineVirtualMachines = e.by("metadata.name"), A(), Le();
+B.offlineVirtualMachines = e.by("metadata.name"), D(), Le();
 }, {
 poll: V,
 pollInterval: 6e4
 }));
 Rt.push(m.watch(N.virtualMachine, r, function(e) {
-B.virtualMachines = e.by("metadata.name"), A(), Le();
+B.virtualMachines = e.by("metadata.name"), D(), Le();
 }, {
 poll: V,
 pollInterval: 6e4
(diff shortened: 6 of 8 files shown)


app/scripts/directives/logViewer.js
```diff
@@ -446,6 +446,10 @@ angular.module('openshiftConsole')
                   access_token: AuthService.UserStore().getToken()
                 });
 
+                var canI = $filter('canI');
+                var podsLogVersion = APIService.getPreferredVersion('pods/log');
+                var canViewOperationsLogs = canI(podsLogVersion, 'view');
```
spadgett: This is checking that you can view the log you're already viewing... @jcantrill Is this really what we want?

app/scripts/services/logLinks.js
```diff
@@ -73,7 +73,12 @@ angular.module('openshiftConsole')
       ].join(''));
 
 
-      var archiveUri = function(opts, prefix) {
+      var archiveUri = function(opts, prefix, canViewOperationsLogs) {
+        if(canViewOperationsLogs) {
+          if(!prefix || (prefix && prefix.startsWith('project.'))) {
```
spadgett: These two lines can be collapsed into a single `if` condition

app/scripts/directives/logViewer.js
jcantrill: This specific issue is that we changed the available index patterns if you are an operations user.  You used to get a list of every project in the cluster which was exceedingly long and we modified the seeding to provide only one which matched them all with 'project.*'.   The complementary change was not captured in the web console.  Operations users are being redirected to Kibana to use an index-pattern which does not exist.

Here we are trying to determine if the use is an operations user by using the same SAR we use in the multi-tenant plugin.

josefkarasek: Only operations users should be able to view pods/log from the logging project, so:
```
oc auth can-i view pods/log -n openshift-logging
```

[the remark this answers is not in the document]
josefkarasek: That's what the `canI` filter does under the hood, isn't it?
I'm trying to use its `projectName` argument, I assume that 
```
AuthorizationService.canI(resource, verb, projectName)
```
will issue a request against the Authorization API and SSAR is used on the server, or not?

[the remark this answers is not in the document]
spadgett: No, that's not what it does... You should run a separate SelfSubjectAccessReview request to check access in another project.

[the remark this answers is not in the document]
josefkarasek: Is there some support for it in the code base?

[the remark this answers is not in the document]
spadgett: No, but it should be pretty easy to call DataService.create with a SelfSubjectAccessReview object:

https://kubernetes.io/docs/reference/access-authn-authz/authorization/#checking-api-access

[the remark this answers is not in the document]
josefkarasek: That's the hint I was hoping for, thanks a lot!

app/scripts/directives/logViewer.js
```diff
@@ -446,17 +446,38 @@ angular.module('openshiftConsole')
...
-                    // The archive URL violates angular's built in same origin policy.
-                    // Need to explicitly tell it to trust this location or it will throw errors.
-                    kibanaArchiveUrl: $sce.trustAsResourceUrl(logLinks.archiveUri({
-                                        namespace: $scope.context.project.metadata.name,
-                                        namespaceUid: $scope.context.project.metadata.uid,
-                                        podname: name,
-                                        containername: $scope.options.container,
-                                        backlink: URI.encode($window.location.href)
-                                      }, $filter('annotation')($scope.context.project,'loggingDataPrefix')))
+                // var ssarVersion = APIService.getPreferredVersion('selfsubjectaccessreviews');
```
josefkarasek: I tried to add SSAR to preferredVersions but somehow this change didn't take effect, even after grunt build.
@spadgett can you advice how to add it?

app/scripts/directives/logViewer.js
```diff
@@ -446,17 +446,38 @@ angular.module('openshiftConsole')
...
+                    }
+                  }
+                };
+                var promise = DataService.create('selfsubjectaccessreviews', null, ssar, {namespace: 'openshift-logging'});
+                // var promise = DataService.create(ssarVersion, null, ssar, {namespace: 'openshift-logging'});
+                promise.then(function(data) {
+                  return data.status.allowed;
+                }, function() {
+                  return false;
+                }).then(function(canViewOperationsLogs) {
```
josefkarasek: @spadgett @jcantrill I'm struggling with the async nature of `promise`s. As this function is not called externally (click on a button etc), at this point log view is blocked. Can you help me how to make sure this promise gets evaluated?

app/scripts/services/logLinks.js
```diff
@@ -73,9 +73,14 @@ angular.module('openshiftConsole')
...
 
-      var archiveUri = function(opts, prefix) {
+      var archiveUri = function(opts, prefix, canViewOperationsLogs) {
+        console.log('admin: ' + canViewOperationsLogs);
+        if((canViewOperationsLogs && !prefix) || (canViewOperationsLogs && prefix && prefix.startsWith('project.'))) {
+            prefix = 'project';
+        }
         prefix = prefix || 'project.' + opts.namespace + '.' + opts.namespaceUid;
         opts.index = prefix + '.*';
+        console.log('link: ' + opts.index);
```
josefkarasek: Will delete once the fix is ready.

app/scripts/directives/logViewer.js
josefkarasek: I had an issue with my VM, all log files disappeared from the file system. That's why I was seeing weird behavior in the console. The code should be fine.

app/scripts/directives/logViewer.js
spadgett: Don't worry about adding a preferred version. You'd need to update the origin-web-common repo and cut a release. I don't think it's worth it here.

app/scripts/directives/logViewer.js
```diff
@@ -446,17 +446,38 @@ angular.module('openshiftConsole')
...
+                  kind: 'SelfSubjectAccessReview',
+                  spec: {
+                    resourceAttributes: {
+                      resource: 'pods/log',
+                      namespace: 'openshift-logging',
+                      verb: 'view'
+                    }
+                  }
+                };
+                var promise = DataService.create('selfsubjectaccessreviews', null, ssar, {namespace: 'openshift-logging'});
```
spadgett: Make sure the API version in the request is the same as the endpoint you're calling. I believe it would be

```js
DataService.create({ group: 'authorization.k8s.io', version: 'v1', resource: 'selfsubjectaccessreviews'}, ...
```

spadgett: We should be caching this response somewhere so we only check it once per session. You probably want to move this into an Angular service.

spadgett:
/cc @jcantrill 

Jeff should review as well

app/scripts/directives/logViewer.js
josefkarasek: What if I remove role from a user? That could be a possible danger it that user is logged in a session and has dirty cash.

josefkarasek: Disregard, we'll do another round of auth on our side.

josefkarasek:
Depends on openshift/origin-web-common#342

josefkarasek:
@jcantrill there are still a few issues that need to be resolved

1) The 'Archive link' button sends a kibana search query[1] in the request's body,
which is discarded during the oauth process (http 302 redirects remove the body).
As a result the user lands at default kibana index pattern

2) The token sent in the request body[1] is ignored by the proxy.
Oauth proxy can be configured to read `Authorization: Bearer: token` header.
Problem in our case is that the web console uses tokens with scope `user:info user:check-access`,
but we need `user:info user:check-access user:list-projects` scope so that Elasticsearch plugin can use this token to check whether the user is operations user. 

[1] https://github.com/openshift/origin-web-console/blob/master/app/views/directives/logs/_log-viewer.html#L10-L16

spadgett:
If you guys have switched to the OAuth proxy, I don't think we should be sending the token from the console at all anymore, correct?

cc @jwforres

jwforres:
the original discussion was the switch to using the oauth proxy was supposed to be transparent to console users and that the existing token mechanism was going to behave the same

the discussion about changing the behavior to stop passing token had been a converged console discussion

jcantrill:
@jwforres @spadgett I incorrectly was under the impression the oauth-proxy was able to function in the same manor as the original proxy.  Pending investigation from @josefkarasek about if it its possible to function in the previous capacity, we may need to revert the proxy being used in order to not have a functional regression.  We discussed within the logging team regarding the 4.0 release not having a link from console to kibana which means this work  would be uneeded.  It means however we need to continue to release the old proxy which I was hoping to avoid

app/views/directives/logs/_log-viewer.html
```diff
@@ -53,16 +47,8 @@ <h2>Logs are not available.</h2>
...
-    <form
-      action="{{kibanaAuthUrl}}"
-      method="POST">
-      <input type="hidden" name="redirect" value="{{kibanaArchiveUrl}}">
-      <input type="hidden" name="access_token" value="{{access_token}}">
-      <button class="btn btn-primary btn-lg">
-        View Archive
-      </button>
-    </form>
+  <div ng-if="kibanaArchiveUrl">
```
josefkarasek: @spadgett why is this form here again?

josefkarasek:
@jcantrill @spadgett please let me know what you think. I will then update the PR so we can finish this. Thank you

app/views/directives/logs/_log-viewer.html
spadgett: It was there to POST the user's API token to kibana to avoid another login

jcantrill:
@josefkarasek @spadgett this https://github.com/openshift/origin-web-console/pull/3067 will resolve the form post/get issue

app/scripts/directives/logViewer.js
```diff
@@ -27,6 +28,7 @@ angular.module('openshiftConsole')
              DataService,
              HTMLService,
              ModalsService,
+             AggregatedLoggingService,
```
jcantrill: This is not defined in this repo.  Please move to this repo per @spadgett request

app/scripts/directives/logViewer.js
```diff
@@ -434,16 +436,22 @@ angular.module('openshiftConsole')
...
-                                        baseURL: url,
-                                        namespace: $scope.context.project.metadata.name,
-                                        namespaceUid: $scope.context.project.metadata.uid,
-                                        podname: name,
-                                        containername: $scope.options.container,
-                                        backlink: URI.encode($window.location.href)
-                                      }, $filter('annotation')($scope.context.project,'loggingDataPrefix')))
+
+                var currentUser = AuthService.UserStore().getUser().metadata.name;
+                AggregatedLoggingService.isOperationsUser(currentUser).then(function(canViewOperationsLogs) {
```
jcantrill: I don't think this needs to be encapsulated this way but I defer to @spadgett.  You should be able to simply call:
`AggregatedLoggingService.isOperationsUser(currentUser)` and assign it to a var. and then pass it in as a param to `archiveURI`

app/scripts/directives/logViewer.js
josefkarasek: It's in vendor.js

jcantrill: You are missing the [point](https://github.com/openshift/origin-web-common/pull/342#issuecomment-415014888).  The intention is not do any other release of origin-web-common.  All the files in 'dist' are generated files that are vendored in through an automated process.  The recommendation is o close #342 and move the [file](https://github.com/openshift/origin-web-common/pull/342/files#diff-de4757cd302ff949e0ec3e237a13d05d) to origin-web-console

app/scripts/directives/logViewer.js
spadgett: It has to be a promise since it makes a selfsubjectaccessreview request to the server, which is asynchronous

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+angular.module("openshiftConsole")
```
josefkarasek: @spadgett angulars DI engine cannot find this service - do I need to manually register it somewhere?

I did run `hack/clean-deps.sh && hack/install-deps.sh && grunt build`

josefkarasek: I see that other services from app/scripts/services are in dist/scripts/scripts.js, but aggregatedLoggingService is not among them.

app/scripts/directives/logViewer.js
```diff
@@ -14,6 +14,7 @@ angular.module('openshiftConsole')
     'DataService',
     'HTMLService',
     'ModalsService',
+    'AggregatedLoggingService',
```
spadgett: We try to keep these alphabetized

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
...
+      number: 10
+    });
+
+    var inFlightPermissionsRequest = null;
+    var authPromise = null;
+
+    // Create SelfSubjectAccessReview request againt authorization API to check whether
+    // current user can view pods/log from 'openshift-logging' project.
+    // Users who can view such logs are 'operations' users
+    var isOperationsUser = function(userName) {
```
spadgett: This should not need to take in a user name since it uses self subject access review. (The cached data will be cleared on logout because the page is reloaded, so that's not a problem here.)

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+angular.module("openshiftConsole")
+  .factory("AggregatedLoggingService", function($q, $cacheFactory, Logger, $interval, APIService, DataService) {
+
+    var cachedUserPermissions = $cacheFactory('operationsUsersCache', {
+      number: 10
+    });
```
spadgett: This seems like overkill since we are only ever storing one value (if the current user is an operations user)

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
...
+
+    var inFlightPermissionsRequest = null;
```
spadgett: I don't think this is needed since you should never have multiple in flight requests for this specific permission

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
...
+            return Promise.resolve(false);
+        }).finally(function() {
+          inFlightPermissionsRequest = null;
+        });
+        inFlightPermissionsRequest = authPromise;
+      } else {
+        // Using cached data.
+        Logger.log("AggregatedLoggingService, using cached user " + userName);
+        if ((_.now() - userAllowed.cacheTimestamp) >= 600000) {
+          userAllowed.forceRefresh = true;
```
spadgett: I wouldn't worry about forcing refresh... This permission does not change often. Let's just always use the cached value.

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+        return $q.when(!!_.get(userAllowed, ['allowed']));
```
spadgett: Why not just...

```jsx
return $q.when(userAllowed.allowed);
```

app/views/directives/logs/_log-viewer.html
```diff
@@ -53,16 +47,8 @@ <h2>Logs are not available.</h2>
...
+    <a href="{{kibanaArchiveUrl}}" class="btn btn-primary btn-lg">View Archive</a>
```
spadgett: `ng-href`

Why does this have button styling? It should just be a normal link.

app/scripts/services/logLinks.js
```diff
@@ -73,9 +73,17 @@ angular.module('openshiftConsole')
       ].join(''));
 
 
-      var archiveUri = function(opts, prefix) {
+      var archiveUri = function(opts, prefix, canViewOperationsLogs) {
+        if((canViewOperationsLogs && !prefix) || (canViewOperationsLogs && prefix && prefix.startsWith('project.'))) {
```
spadgett:
```js
if (canViewOperationsLogs && (!prefix || prefix.startsWith('project.')) {
```

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        Logger.log("AggregatedLoggingService, loading whether user " + userName + " is Operations user");
+        var ssar = {
+          apiVersion: 'authorization.k8s.io/v1',
+          kind: 'SelfSubjectAccessReview',
+          spec: {
+            resourceAttributes: {
+              resource: 'pods/log',
+              namespace: 'openshift-logging',
```
jcantrill: This should be 'default' namespace since this is how we determine similar in the multi-tenant plugin

jcantrill: You will also need to set scopes to an empty array like:

* https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/util/RequestUtils.java#L96
* https://github.com/fabric8io/openshift-elasticsearch-plugin/blob/master/src/main/java/io/fabric8/elasticsearch/plugin/OpenshiftAPIService.java#L116

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
...
+              verb: 'view'
+            }
+          }
+        };
+        authPromise = DataService.create({ group: 'authorization.k8s.io', version: 'v1', resource: 'selfsubjectaccessreviews'},
+                          null, ssar, {namespace: 'openshift-logging'}).then(
```
jcantrill: namespace 'default'

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,66 @@
...
+          function(data) {
+            var isAllowed = data.status.allowed;
+            cachedUserPermissions.put(userName, {'allowed': isAllowed,
```
jcantrill: This will i presume will be removed based on @spadgett comments regarding caching

app/scripts/services/aggregatedLoggingService.js
josefkarasek: @jcantrill  I didnt find scope in resourceAttributes definition. I assume it's used by the above mentioned java library for something else than the auth api.

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/authorization/types.go#L80-L98

josefkarasek:
@spadgett are you happy with the latest changes?
If yes, what do we need to do to have this patch merged? This PR is against master, is that ok?
I ran tests against okd 3.10 with latest logging images and they pass.

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,47 @@
...
+              namespace: 'default',
+              verb: 'view'
+            }
+          }
+        };
+        authPromise = DataService.create({ group: 'authorization.k8s.io', version: 'v1', resource: 'selfsubjectaccessreviews'},
+                          null, ssar, {namespace: 'default'}).then(
+          function(data) {
+            userAllowed = data.status.allowed;
+            return Promise.resolve(userAllowed);
```
spadgett:
```js
return userAllowed;
```

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,47 @@
...
+          }, function() {
+            return Promise.resolve(false);
```
spadgett:
```js
return false;
```

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,47 @@
...
+          kind: 'SelfSubjectAccessReview',
+          spec: {
+            resourceAttributes: {
+              resource: 'pods/log',
+              namespace: 'default',
+              verb: 'view'
+            }
+          }
+        };
+        authPromise = DataService.create({ group: 'authorization.k8s.io', version: 'v1', resource: 'selfsubjectaccessreviews'},
```
spadgett: You don't need to create a Promise yourself. Just return the Promise from DataService.create

```js
return DataService.create(...).then(...);
```

app/scripts/services/logLinks.js
```diff
@@ -73,7 +73,10 @@ angular.module('openshiftConsole')
       ].join(''));
 
 
-      var archiveUri = function(opts, prefix) {
+      var archiveUri = function(opts, prefix, canViewOperationsLogs) {
+        if((canViewOperationsLogs) && (!prefix || prefix.startsWith('project.'))) {
```
spadgett: style nit: i think the extra parens around canViewOperationsLogs actually make it harder to read

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,47 @@
...
+    // cache previous request
+    var userAllowed;
+
+    // Create SelfSubjectAccessReview request againt authorization API to check whether
+    // current user can view pods/log from 'default' project.
+    // Users who can view such logs are 'operations' users
+    var isOperationsUser = function() {
+      var authPromise = null;
+
+      if (userAllowed === undefined) {
```
spadgett: I'd switch this to an early return if we have a cached value. It's easier to reason about.

```js
if (userAllowed !== undefined) {
  return $q.when(userAllowed);
}

var ssar = { ... };
return DataService.create(...).then(...);
```

spadgett:
> [FATAL] Built /dist does not match what is committed, run 'grunt build' and include the results in your commit.

app/scripts/directives/logViewer.js
```diff
@@ -434,16 +436,21 @@ angular.module('openshiftConsole')
...
-                                        podname: name,
-                                        containername: $scope.options.container,
-                                        backlink: URI.encode($window.location.href)
-                                      }, $filter('annotation')($scope.context.project,'loggingDataPrefix')))
+
+                AggregatedLoggingService.isOperationsUser().then(function(canViewOperationsLogs) {
+                  $scope.$watchGroup(['context.project.metadata.name', 'options.container', 'name'], function() {
+                    angular.extend($scope, {
+                      // The archive URL violates angular's built in same origin policy.
+                      // Need to explicitly tell it to trust this location or it will throw errors.
```
spadgett: I'm not sure if this is still needed now that it's a link instead of a form. You can probably get rid of the `trustAsResourceUrl` call

app/scripts/services/aggregatedLoggingService.js
```diff
@@ -0,0 +1,43 @@
...
+        }
+      };
+      return DataService.create({ group: 'authorization.k8s.io', version: 'v1', resource: 'selfsubjectaccessreviews'},
+                        null, ssar, {namespace: 'default'}).then(
+        function(data) {
+          userAllowed = data.status.allowed;
+          return userAllowed;
+        }, function() {
+          return false;
+      });
```
spadgett: 👍 thanks, this looks much cleaner

josefkarasek:
@spadgett Thank you for your help and feedback. If there's still something to be done please let me know.

app/scripts/directives/logViewer.js
spadgett: Comment is out of date with the code now

app/scripts/directives/logViewer.js
```diff
@@ -434,16 +436,21 @@ angular.module('openshiftConsole')
...
+                      kibanaArchiveUrl: logLinks.archiveUri({
+                                          baseURL: url,
```
spadgett: nit: indentation

josefkarasek: not sure what's wrong with indentation here - a new block of code - indented by 2 spaces against its parent block

jcantrill:
@josefkarasek please backport to 3.10 and 3.9 to fix relevent issues

spadgett:
Note that this will need some rework if we aren't using oauth proxy in 3.10 and 3.9

PR #3062, opened 2018-08-17
[Bug 1608289] Expand button should be enabled as long as the size don…

cloudbehl:
…'t exceed limit

We don't need to check if we are out of claims or not if we are expanding the PVC because the PVC is already created.

Signed-off-by: Ankush Behl <cloudbehl@gmail.com>


spadgett:
Thanks @cloudbehl. Needs dist files

cloudbehl:
@spadgett aah, sorry! Added dist files now.

spadgett:
@cloudbehl thanks, i think you'll need to run `bower install` and `grunt build` again as well

cloudbehl:
@spadgett is it good now? Let me know anything else is required.

PR #3064, opened 2018-08-21
Docker server address should default to dockerhub

jhadvig:
Aligning to CLI where the server address is optional

Screen:

@spadgett PTAL


jhadvig:
Closing this one per conversation with @spadgett

PR #3067, opened 2018-08-28
No longer post token for kibana

jcantrill:
This PR fixes https://bugzilla.redhat.com/show_bug.cgi?id=1614255

by no longer posting the access token for Kibana since the proxy was migrated to oauth-proxy.  User's may need to re-auth which is a change from 3.10 and prior

### DIFF
--- a/app/scripts/services/logLinks.js
+++ b/app/scripts/services/logLinks.js
@@ -46,7 +46,7 @@ angular.module('openshiftConsole')
 
       // broken up for readability:
       var template = _.template([
-        "/#/discover?",
+        "<%= baseURL %>#/discover?",
         "_g=(",
           "time:(",
             "from:now-1w,",
@@ -76,6 +76,11 @@ angular.module('openshiftConsole')
       var archiveUri = function(opts, prefix) {
         prefix = prefix || 'project.' + opts.namespace + '.' + opts.namespaceUid;
         opts.index = prefix + '.*';
+
+        if (opts.baseURL.substr(-1) !== '/') {
+            opts.baseURL += '/';
+        }
+
         return template(opts);
       };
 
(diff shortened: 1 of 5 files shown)


app/views/directives/logs/_log-viewer.html
```diff
@@ -9,9 +9,8 @@
     <span ng-if="kibanaAuthUrl">
       <form
         action="{{kibanaAuthUrl}}"
-        method="POST">
+        method="GET">
```
spadgett: This should no longer be a form. It can be something like

```html
<a ng-href="{{kibanaAuthUrl}}>View Archive</a>
```

We should remove "auth" from the name if it no longer handles auth.

app/scripts/services/logLinks.js
```diff
@@ -46,7 +46,7 @@ angular.module('openshiftConsole')
 
       // broken up for readability:
       var template = _.template([
-        "/#/discover?",
+        "<%= url %>/#/discover?",
```
spadgett: Do we know if url can ever end in a `/` causing a double forward slash? We should give it a name like `baseURL` if it's not the full URL.

I'm a little confused why we need to do this now... How did the base URL get prepended before?

app/views/directives/logs/_log-viewer.html
spadgett: Oh this should be using `kibanaArchiveURL` and we can remove `kibanaAuthURL`. I see now why you need to prefix the URL.. It used to be passed as a redirect.

app/scripts/services/logLinks.js
spadgett: OK, well if either is a valid config value, we will want to make sure there is no double slash in the URL. Do you'd want some JS to handle that.

jcantrill:
@spadgett Updated with suggestions.  Please review

Issue #2694, opened 2018-01-22
how to build bindata.go 


pennpeng:
TODO - run script to build bindata.go from the dist in this repo 

hi , i can't find the script to build bindata.go

spadgett:
If this is for 3.9, you need to clone the origin-web-console-server repo and run the vendor console script:

https://github.com/openshift/origin-web-console-server/blob/master/hack/vendor-console.sh

then rebuild the image

```
$ OS_BUILD_ENV_PRESERVE=_output/local/bin hack/env make build build-images 
```

For 3.8 and earlier, you run the same script, but in the origin repo:

https://github.com/openshift/origin/blob/master/hack/vendor-console.sh

/assign
/kind question
/close

pennpeng:
@spadgett  hi，when i  run https://github.com/openshift/origin/blob/master/hack/vendor-console.sh，there is an error：
```
[root@oc-3 origin]# ./hack/vendor-console.sh 
Making sure go-bindata binary has been built...
go install: no install location for directory /home/origin/vendor/github.com/jteeuwen/go-bindata outside GOPATH
	For more details see: 'go help gopath'
go install: no install location for directory /home/origin/vendor/github.com/jteeuwen/go-bindata/go-bindata outside GOPATH
	For more details see: 'go help gopath'
[ERROR] PID 21383: hack/vendor-console.sh:23: `go install ./...` exited with status 1.
[INFO] 		Stack Trace: 
[INFO] 		  1: hack/vendor-console.sh:23: `go install ./...`
[INFO]   Exiting with code 1.

```
i set gopath:
```
[root@oc-3 origin]# env GOPATH=/home/go ./hack/vendor-console.sh 
Making sure go-bindata binary has been built...
go install: no install location for directory /home/origin/vendor/github.com/jteeuwen/go-bindata outside GOPATH
	For more details see: 'go help gopath'
go install: no install location for directory /home/origin/vendor/github.com/jteeuwen/go-bindata/go-bindata outside GOPATH
	For more details see: 'go help gopath'
[ERROR] PID 20799: hack/vendor-console.sh:23: `go install ./...` exited with status 1.
[INFO] 		Stack Trace: 
[INFO] 		  1: hack/vendor-console.sh:23: `go install ./...`
[INFO]   Exiting with code 1.

```

spadgett:
@pennpeng Your `origin` directory needs be inside your `GOPATH`, for instance

```
/home/go/src/github.com/openshift/origin
[...]

PR #3068, opened 2018-08-30
complete/correct README for Production builds

zhanghaohao:
origin-web-console repo use go-bindata tool to generate bindata.go from dist folder , but the README has not completed this part , give out approaches for different branches .
Also correct the rebuild images and binaries part according to different branches .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 node_js:
   - "6"
 before_script:
-  - make build 
+  - make build
   # FIXME:
   # This needs to be enabled again when we eliminate PhantomJS for Firefox
   # Even the unit tests require a browser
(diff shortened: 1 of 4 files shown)


zhanghaohao:
Fix https://github.com/openshift/origin-web-console/issues/2694

spadgett:
I think it would be better to link to the instructions in https://github.com/openshift/origin-web-console-server so we only have them in one place. We should correct any errors if those instructions aren't accurate.

zhanghaohao:
sure, i have changed REAMDE.md according to your comment

zhanghaohao:
could we ignore the Travis CI build failed message ? because i successfully built on my mac

zhanghaohao:
run "grunt build" , then commit due to Travic CI build failed 
now Travis CI build succeed

PR #3075, opened 2018-09-12
Bug 1607150 - Additional fix for: UI Timeout causing IE 11 to automatically log out

jhadvig:
Not really sure why this haven't been backported. But it's fixing the issue.

@spadgett PTAL


spadgett (approved):
/lgtm

PR #3080, opened 2018-10-01
[enterprise-3.9] Bug 1633285 - No secrets dropdown in buildconfig edit page if any of the secrets has no data

jhadvig:
@spadgett PTAL


spadgett (approved):
/lgtm

PR #3083, opened 2018-10-01
Bug 1633285 - No secrets dropdown in buildconfig edit page if any of the secrets has no data

jhadvig:
@spadgett PTAL


jhadvig:
@spadgett do we need still to hold this one ?

spadgett (approved):
/lgtm

PR #3097, opened 2018-11-27
Bug 1652645 - Webhook urls in configuration page are invalid with undefined included

jhadvig:
From the bug description there are two issues:
1. When a freshly created webhook secret is not available right away so the secret value of the trigger URL is undefined: *.../buildconfigs/webhooks/**undefined**/generic*  
2. When the referenced secrets is deleted, but still used in the buildConfig trigger
Just met the issue one time after trying at least ten times. This requests the copy action taking place right after the secret is created and added as webhook.

Haven't found any reasonable precedence for showing warnings in the description list terms in the codebase(hopefully I haven't overlooked anything), so created a proposal that would cover both issues:

 Marked the PR as WIP to open the design discussion, code is not ready for review...

cc'ing @openshift/team-ux-review 

@spadgett PTAL

### DIFF
--- a/app/scripts/filters/resources.js
+++ b/app/scripts/filters/resources.js
@@ -182,6 +182,9 @@ angular.module('openshiftConsole')
       var secretsVersion = APIService.getPreferredVersion('secrets');
       if (canIFilter(secretsVersion, 'list')) {
         secret = SecretsService.getWebhookSecretValue(secret, webhookSecrets);
+        if (!secret) {
+          return '';
+        }
         return DataService.url({
           // arbitrarily many subresources can be included
           // url encoding of the segments is handled by the url() function
--- a/app/scripts/services/secrets.js
+++ b/app/scripts/services/secrets.js
@@ -152,7 +152,7 @@ angular.module("openshiftConsole")
       // to list Secrets
       if (_.get(secret, 'secretReference.name') && webhookSecrets) {
         var matchingSecret = _.find(webhookSecrets, {metadata:{name: secret.secretReference.name}});
-        return decodeSecretData(matchingSecret.data).WebHookSecretKey;
+        return _.has(matchingSecret, 'data') ? decodeSecretData(matchingSecret.data).WebHookSecretKey : '';
       } else {
         return _.get(secret, 'secret');
       }
--- a/app/views/browse/build-config.html
+++ b/app/views/browse/build-config.html
@@ -386,29 +386,28 @@
                               <dt>Bitbucket Webhook URL:
                               </dt>
                               <dd>
-                                <copy-to-clipboard clipboard-text="buildConfig.metadata.name | webhookURL : trigger.type : trigger.bitbucket : project.metadata.name : webhookSecrets"></copy-to-clipboard>
-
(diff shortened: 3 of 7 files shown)


app/views/browse/build-config.html
```diff
@@ -408,7 +408,17 @@ <h3>Triggers <span class="pficon pficon-warning-triangle-o" ng-if="!(secretsVers
                               <dt>Generic Webhook URL:
                               </dt>
                               <dd>
-                                <copy-to-clipboard clipboard-text="buildConfig.metadata.name | webhookURL : trigger.type : trigger.generic : project.metadata.name : webhookSecrets"></copy-to-clipboard>
+                                <div ng-if="trigger.generic | isWebhookAvailable : webhookSecrets">
```
spadgett: Isn't this potentially a problem for all webhook types, not only generic webhooks?

app/scripts/filters/resources.js
```diff
@@ -177,6 +177,19 @@ angular.module('openshiftConsole')
       return null;
     };
   })
+  .filter('isWebhookAvailable', function(SecretsService) {
```
spadgett: I'd suggest having the `webhookURL` filter return an empty string instead of creating a separate filter.

app/views/browse/build-config.html
jhadvig: yeah its definitely problem of all the webhook triggers... this is just a WIP

jhadvig:
@spadgett I've updated the PR and removed the WIP flag.
Found out that the cleanest and easiest way is to create a wrapper `copyWebhookTrigger` directive, check for the `webhook-url` inside and render the "Not Available" together with the warning icon or the `copyToClipboard` directive.

PTAL

app/scripts/filters/resources.js
```diff
@@ -182,6 +182,9 @@ angular.module('openshiftConsole')
       var secretsVersion = APIService.getPreferredVersion('secrets');
       if (canIFilter(secretsVersion, 'list')) {
         secret = SecretsService.getWebhookSecretValue(secret, webhookSecrets);
+        if (!secret) {
+          return '';
```
jhadvig, line 186: We could return `secret`, but was rather explicit with the return value here.

app/scripts/services/secrets.js
```diff
@@ -152,7 +152,7 @@ angular.module("openshiftConsole")
       // to list Secrets
       if (_.get(secret, 'secretReference.name') && webhookSecrets) {
         var matchingSecret = _.find(webhookSecrets, {metadata:{name: secret.secretReference.name}});
-        return decodeSecretData(matchingSecret.data).WebHookSecretKey;
+        return _.has(matchingSecret, 'data') ? decodeSecretData(matchingSecret.data).WebHookSecretKey : '';
```
spadgett, line 155: Prefer `_.get` so we never try to decode `null` values. (`_.has` only checks for `undefined`.)

```suggestion
        return _.get(matchingSecret, 'data') ? decodeSecretData(matchingSecret.data).WebHookSecretKey : '';
```

spadgett (approved):
One small comment, but I don't think it's an issue in practice. Tagging for merge. Thanks 👍 

/lgtm

spadgett:
We should fix in 3.10 and 3.11 as well.

PR #3098, opened 2018-11-28
[enterprise-3.10] Bug 1652645 - Webhook urls in configuration page are invalid with undefined included

jhadvig:
Port of https://github.com/openshift/origin-web-console/pull/3097

@spadgett PTAL


spadgett (approved):
/lgtm

PR #3101, opened 2018-11-28
Bug 1652645 - Webhook urls in configuration page are invalid with undefined included

jhadvig:
Port of https://github.com/openshift/origin-web-console/pull/3097

@spadgett PTAL


spadgett:
@jhadvig looks like we need to fix this for master, too.

```
+ oc cluster up --tag=latest --public-hostname=localhost --loglevel=5 --server-loglevel=5
Error: unknown command "cluster" for "oc"
```

PR #3108, opened 2019-01-22
Master deps

jhadvig:
Cherry-picked last changes to the 3.11. branch that havent made it do master.

/assign @spadgett


spadgett:
I don't know why we have both ci/prow/unit and ci/openshift-jenkins. They're running the same tests, but unit appears to be broken. (It's also probably misnamed since it's not just unit tests.) Can we simply remove ci/prow/unit?

cc @stevekuznetsov

stevekuznetsov:
Do you actually have two jobs doing the same thing reporting two different ways? I'd expect Jenkins based stuff to be long dead...

spadgett:
> Do you actually have two jobs doing the same thing reporting two different ways? I'd expect Jenkins based stuff to be long dead...

It looks like it...

We have been focused on improving CI for openshift/console, since that is where new development is happening. This repo is now only for fixes to previous 3.x releases.

I don't know the history, but somewhere along the way test unit was added, and I'm not sure it has ever worked. It runs `make test` which runs our entire e2e suite. That is also run by jenkins.

Note that master is not part of any shipped image anymore, only the `enterprise-3.x` branches.

stevekuznetsov:
Alright, then you'll want to remove it from the configuration in `openshift/release` if it's not useful.

PR #3109, opened 2019-01-23
Fix bug where polling quotas causes annoying toast notifications

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1651090


spadgett (approved):
/lgtm

spadgett:
this will need dist files

PR #3116, opened 2019-02-06
[enterprise-3.11] Fix bug where cannot create new role and new rolebinding in single operation using template 

rhamilto:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1673109

Big assist from @spadgett on this one.  Thank you!

### DIFF
--- a/app/scripts/directives/processTemplate.js
+++ b/app/scripts/directives/processTemplate.js
@@ -106,38 +106,35 @@
       var helpLinks = getHelpLinks(ctrl.template);
       TaskList.clear();
       TaskList.add(titles, helpLinks, ctrl.selectedProject.metadata.name, function() {
-        var d = $q.defer();
-        DataService.batch(processedResources, context).then(
-          function(result) {
-            var alerts = [];
-            var hasErrors = false;
-            if (result.failure.length > 0) {
-              hasErrors = true;
-              result.failure.forEach(
-                function(failure) {
-                  alerts.push({
-                    type: "error",
-                    message: "Cannot create " + humanize(failure.object.kind).toLowerCase() + " \"" + failure.object.metadata.name + "\". ",
-                    details: failure.data.message
-                  });
-                }
-              );
-              result.success.forEach(
-                function(success) {
-                  alerts.push({
-                    type: "success",
-                    message: "Created " + humanize(success.kind).toLowerCase() + " \"" + success.metadata.name + "\" successfully. "
-                  });
-                }
-              );
-            } else {
-              alerts.push({ type: "success", message: "All items in template " + ctrl.templateDisplayName +
-                " were created successfully."});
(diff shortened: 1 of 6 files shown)


app/scripts/directives/processTemplate.js
```diff
@@ -106,38 +106,35 @@
...
+                hasErrors = true;
+                alerts.push({
+                  type: "error",
+                  message: "Cannot create " + humanize(obj.kind).toLowerCase() + " \"" + obj.metadata.name + "\". ",
+                  details: failure.data.message
+                });
+              });
+          });
+        });
+        return chain.then(function(alerts) {
```
spadgett:
```suggestion
        return chain.then(function() {
```

app/scripts/directives/processTemplate.js
```diff
@@ -106,38 +106,35 @@
...
+                  type: "success",
+                  message: "Created " + humanize(obj.kind).toLowerCase() + " \"" + obj.metadata.name + "\" successfully. "
+                });
+              })
+              .catch(function(failure) {
+                hasErrors = true;
+                alerts.push({
+                  type: "error",
+                  message: "Cannot create " + humanize(obj.kind).toLowerCase() + " \"" + obj.metadata.name + "\". ",
+                  details: failure.data.message
```
spadgett: I'm not sure, but this might need to be

```suggestion
                  details: failure.message
```

Can you check?

spadgett (approved):
/lgtm

test/integration/features/user_adds_template_to_project.spec.js
```diff
@@ -34,20 +34,22 @@ describe('User adds a template to a project', () => {
...
-            // verify we have the two services in the template
-            let servicesPage = new ServicesPage(project);
-            servicesPage.visit();
-            expect(element(by.cssContainingText('td', 'mongodb')).isPresent()).toBe(true); // TODO: use fixture
-            expect(element(by.cssContainingText('td', 'nodejs-mongodb-example')).isPresent()).toBe(true); // TODO: use fixture
-            // verify we have one route for the mongo app
-            let routesPage = new RoutesPage(project);
-            routesPage.visit();
-            expect(element(by.cssContainingText('td', 'nodejs-mongodb-example')).isPresent()).toBe(true); // TODO: use fixture
+            return browser.sleep(3000).then(() => {
```
rhamilto, line 37: @spadgett, I took a guess this would resolve the failing test issue.  Mind taking a look?

spadgett: This seems reasonable. Ideally we would wait for the page to update instead of an arbitrary sleep, but I think it's OK if it unblocks CI.

spadgett (approved):
/lgtm

PR #3127, opened 2019-05-23
Update bower to 1.8.8

spadgett:
/assign @rhamilto


spadgett:
This has no impact on the built dist, only updating a dev dependency.

PR #3128, opened 2019-06-03
Bug 1713211: Correctly update overview donut when stateful set is scaled

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1713211

/assign @rhamilto

### DIFF
--- a/app/scripts/directives/overview/listRow.js
+++ b/app/scripts/directives/overview/listRow.js
@@ -66,7 +66,6 @@
 
     var updateCurrent = function(apiObject) {
       if (!apiObject ||
-          row.current ||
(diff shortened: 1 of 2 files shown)


app/scripts/directives/overview/listRow.js
```diff
@@ -66,7 +66,6 @@
 
     var updateCurrent = function(apiObject) {
       if (!apiObject ||
-          row.current ||
```
spadgett, line 69: This check prevented `row.current` from being updated to the new value for `apiObject`, which meant the donut was showing the old desired replicas value forever.

rhamilto (approved):
/lgtm

PR #3131, opened 2019-06-05
Bug 1717028: Fix edge case evaluating services for deployment config

spadgett:
Include labels automatically added to deployment config pods since the service might select them.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1717028

/assign @rhamilto


rhamilto (approved):
/lgtm

PR #3132, opened 2019-06-18
Bug 1721428: Remove misleading "Metrics might not be available" warning

spadgett:
A metrics URL is not required to use HPAs. Since we can't reliably
detect whether metrics server is configured, remove the misleading
warning that might show up when HPAs will otherwise work.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1721428

/assign @rhamilto


spadgett:
Passed CI... @rhamilto can you tag again?

rhamilto (approved):
/lgtm

PR #3140, opened 2019-07-28
Bug 1733806: Avoid runtime errors checking quota when processing template

### LINKED ISSUE #2136: Unable to provision an OpenShift template from UI, when the CustomResource object in the OpenShift template has  "spec.template" as a property defined
Unable to provision an OpenShift template from UI, when the CustomResource object in the OpenShift template has  "spec.template" as a property defined

**Describe the bug**
Provisioning a cluster from OpenShift console is not proceeding after you click "create" button when you have "template" property(eg: Kafkaconnects2i.spec.template.pod.labels) defined in Kafkamirrormaker, Kafkaconnect or Kafkaconnects2i OpenShift templates. Provisioning works fine via oc process command.

**To Reproduce**

Create a Kafkaconnect/kafkamirrormaker/kafkaconnects2i template with custom pod/deployment/statefulset option which has template parameter configured in the CR object to add a label

Refer https://github.com/strimzi/strimzi-kafka-operator/issues/1831 for an example template

Provision a kafkaconnects2i instance using above created template from UI. After clicking create button in the template provisioning window, it wont proceed to the creation of custom resource. Even after clicking "create" button multiple times, the provisioning window will still remain in the same stage.
See error
**Expected behavior**
It was expected to create a kafkaconnects2i custom resource and start a kafka connect pod

**Environment** 

Strimzi version: 0.11.1-AMQ Streams 1.1.1
Kubernetes cluster: OpenShift 3.11
Infrastructure: GCP
Additional context

Provisioning succeeds when i use oc process command using the same template

spadgett:
Fixes https://github.com/openshift/console/issues/2136

### DIFF
--- a/app/scripts/services/quota.js
+++ b/app/scripts/services/quota.js
@@ -71,7 +71,7 @@ angular.module("openshiftConsole")
       if (resource.kind === 'Pod') {
         return filterQuotasForPodTemplate(resource, quotas);
       }
-      else if (_.has(resource, "spec.template")) {
+      else if (_.has(resource, "spec.template.spec")) {
         return filterQuotasForPodTemplate(resource.spec.template, quotas);
       }
       // We plan on having other resources that scopes will affect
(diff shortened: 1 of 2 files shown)


rhamilto (commented):
/lgtm

PR #3150, opened 2019-10-02
Bug 1753601: Correctly map kind to resource

spadgett:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1753601

/assign @rhamilto

### DIFF
--- a/app/scripts/directives/fromFile.js
+++ b/app/scripts/directives/fromFile.js
@@ -5,6 +5,7 @@ angular.module("openshiftConsole")
                                   $location,
                                   $q,
                                   $uibModal,
+                                  APIDiscovery,
                                   APIService,
                                   CachedTemplateService,
                                   DataService,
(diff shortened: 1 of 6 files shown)


rhamilto (approved):
/lgtm

rhamilto (commented):
/lgtm

dobbymoodge:
/retest 

note that this will probably fail at the `/bin/yum install ./google-chrome-stable_current_x86_64.rpm -y` step. As of right now, the Chrome RPM at https://dl.google.com/linux/direct/google-chrome-stable_current_x86_64.rpm appears bugged - I found the `yum` process hung while the underlying `rpm` was running one of the install scripts; it was in this block:

```
  if [ "$?" -eq "0" ]; then
    # Note, if this is run during the package install, it will fail because rpm
    # can't recursively run rpm, but it should work when run later as part of
    # the installed cron job (and probably nothing needs the new keys before
    # then).
    rpm -e --allmatches ${KEY_PACKAGE} >/dev/null 2>&1 || return
  fi
```

I don't know if there's a way to pull down an older, known-working RPM. I'll see if I can figure out how to file bugs against Chrome packaging.

dobbymoodge:
That test and `rpm` call exist in older versions of the RPM as well, as far as I can tell, so maybe it's triggering a new bug in RPM?

PR #3152, opened 2019-10-14
Bug 1756230: Only load config map metadata on config maps page

spadgett:
Only the metadata is needed to display the config map list. This greatly
improves performance on slow network connections since not all config
map data is loaded.

/assign @TheRealJon


TheRealJon (approved):
/lgtm

PR #3160, opened 2019-12-02
Bug 1777758: Checking for wrong default storageclass annotation

mbarnes:
Annotation for default storageclass is now `storageclass.kubernetes.io/is-default-class`.

As of https://github.com/openshift/openshift-ansible/commit/18b42ee

### DIFF
--- a/app/scripts/directives/oscPersistentVolumeClaim.js
+++ b/app/scripts/directives/oscPersistentVolumeClaim.js
@@ -135,7 +135,7 @@ angular.module("openshiftConsole")
            scope.storageClasses = _.sortBy(storageClasses, 'metadata.name');
            var annotation = $filter('annotation');
            scope.defaultStorageClass = _.find(scope.storageClasses, function(storageClass) {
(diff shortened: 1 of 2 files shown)


app/scripts/directives/oscPersistentVolumeClaim.js
```diff
@@ -135,7 +135,7 @@ angular.module("openshiftConsole")
            scope.storageClasses = _.sortBy(storageClasses, 'metadata.name');
            var annotation = $filter('annotation');
            scope.defaultStorageClass = _.find(scope.storageClasses, function(storageClass) {
-             return annotation(storageClass, 'storageclass.beta.kubernetes.io/is-default-class') === 'true';
+             return annotation(storageClass, 'storageclass.kubernetes.io/is-default-class') === 'true';
```
benjaminapetersen, line 138: If I recall correctly, we can't assume the old `annotation` goes away, however (it will linger in use).  Perhaps we could update to check both...

spadgett: Agree, we'd ideally check for either annotation.

PR #3161, opened 2019-12-17
Bug 1782695: Supply the navigateResourceURL filter with kind and name space in the ConfigMap list page [enterprise-3.11]

jhadvig:
Since we started to retrieving the `PartialObjectMetadateList` for ConfigMap list page in https://github.com/openshift/origin-web-console/pull/3152 we need to tell to the `navigateResourceURL` filter explicitly that we need URL for configmap resource, since the `resourceURL` service function is aware about [first-class resource](https://github.com/openshift/origin-web-console/blob/enterprise-3.11/app/scripts/services/navigate.js#L201-L260), not the `PartialObjectMetadate` that is used. 
Due to the fact that are are supplying the filter with `PartialObjectMetadata` object, `null` return value is [returned](https://github.com/openshift/origin-web-console/blob/enterprise-3.11/app/scripts/services/navigate.js#L282) from the `navigateResourceURL` filter.

/assign @spadgett

### DIFF
--- a/app/views/browse/config-maps.html
+++ b/app/views/browse/config-maps.html
@@ -56,7 +56,7 @@
             <tbody>
               <tr ng-repeat="configMap in configMaps">
                 <td data-title="Name">
-                  <a href="{{configMap | navigateResourceURL}}">{{configMap.metadata.name}}</a>
+                  <a href="{{configMap.metadata.name | navigateResourceURL : 'ConfigMap' : configMap.metadata.namespace}}">{{configMap.metadata.name}}</a>
                 </td>
                 <td data-title="Created">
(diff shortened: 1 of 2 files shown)


spadgett (approved):
/lgtm

benjaminapetersen:
/bugzilla refresh

We won't have bugs in more recent releases.

spadgett:
This bug only applies to 3.11.z since the 4.x console has a different code base. It was also introduced by #3152, which is only in 3.11.z. Manually adding the valid-bug label.

Additionally the Jenkins tests are broken due to the Chrome version in use. I've manually verified the change.

/override ci/openshift-jenkins

PR #3166, opened 2020-02-06
Bug 1760767:  fix bug where container image name does not display cor…

rhamilto:
…rectly

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1760767

Discussed this with @spadgett and determined the safest and best fix is to simply display the image name as is and don't try to do clever display of the data.

Screen Shot 2020-02-06 at 2 39 49 PM

### DIFF
--- a/app/views/_pod-template-container.html
+++ b/app/views/_pod-template-container.html
@@ -14,7 +14,7 @@
     </div>
     <div class="pod-template-detail word-break">
       <span class="pod-template-key">Image:</span>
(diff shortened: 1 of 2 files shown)


rhamilto:
@benjaminapetersen, not sure how to get around the bugzilla dependency requirement given this bug doesn't exist in 4.x.

spadgett:
/override ci/openshift-jenkins 

Manually adding the valid bug label. The 4.x console is a different code base that doesn't have this bug.

spadgett (approved):
/lgtm

